PR #25, opened 2018-07-20
Specify license is GPLv3+

TrashMacNugget:
The GPL itself recommends having this notice so that if a newer version of the GPL is released, your code is still compatible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,6 @@ If you want to explore the functionality and inner workings of the app, [head ov
 
 ## License
 
-This software is licenced via the GNU GPL v3-License.
+This software is licenced via the GNU GPL v3 License, or (at your option) any later version.
 
 The brand (including name, icons and everything Zettlr can be identified with) is exluded and all rights reserved. If you want to fork Zettlr to develop another app, feel free but please change name and icons.


nathanlesage:
Hey,

sorry for the late response. I've had a hard time thinking about your PR and after a week of thinking whether or not I should do this, I have come to the conclusion to not do this.

The reason is as follow: Of course I will always release my source code under the newest GPL license possible, but I want to make sure beforehand that the license indeed does cover what I want it to. It may in the future be that there is a GPL license that is not as restrictive concerning commercial use which I then wouldn't adopt. Also, the phrase "or (at your option) any later version" leaves open a lot of juridical loop-holes that may or may not be exploited by patent trolls or other people I don't want to get in touch with.

In short, I won't merge this PR because I want to retain unambiguous licensing, a little bit of control over the Open Sourceness of Zettlr and prevent any person potentially exploiting loop holes.

But thank you a lot for pointing this out to me!

TrashMacNugget:
I respect that.

The main issue with that though is that once people contribute, you no longer can just choose to release it under the newest GPL version. You have to get permission from all of the contributors, which is a difficult (sometimes impossible) process.

nathanlesage:
I see your point here. Hm, I guess I'll have to find something then. But you're absolutely right concerning the mess of getting everyone to agree!

PR #83, opened 2019-01-03
Simplified Chinese Supported

AroKz:
The statistics is wrong with Chinese because the characters in English are like words in Chinese.So It's no need to count the number of "words",just "characters" enough.

And there may be a few mistakes.


nathanlesage:
Hey, thanks for your translation!

I'm looking forward to integrating it into Zettlr.

Just three small points:

1. I would be thankful if you could add the metadata to your translation in any way you see fit. I never did that because until now there was no need to differentiate the authors for translations. Please refer to the current develop commit, in which I added the author metadata for my own translations and adapt them to match yours!

2. Concerning the characters: I never thought of that, but you're right. Yet, I need to build a preferences toggle for this, because there may be users having their app in Chinese but who write in English, or vice versa — therefore the users need to explicitly tell the app how to count words!

3. Just a curious note: I think I remember some people on the CodeMirror forum discussing multibyte chinese characters (i.e. words that are made up of two distinct characters instead of just one) — just how frequent is this? Or is this a complete false flag…? If there are such things, do you know if there is a specific Unicode range in which we know for sure there are multibyte characters so that I could adapt the word counter even more?

Thank you again and have a nice day!

AroKz:
1. Like line 423 in my newest translation file?I don't really know about that.
2. The person who do this may be a...weirdo:),I think.So just changing to only counting characters when the app language is Chinese is a better choice?At least Microsoft Word did that way.Eastern languages usually are the same,like Japanese , Korean etc.
3. Words that are made up of two distinct characters instead of just one?I don't know what that mean.Let me put it this way,"我" is a character in Chinese,but it means "I" or "me" which are words in English.In some way,character in English(alphabet) is just like stroke in Chinese(like "一","丨","丿","丶"),which are both meaningless. Then if you want to count words of an article in Chinese,it's too complex to achieve the goal.For example,"兄弟"(brother) sometimes is a word,but sometimes either “兄”(elder brother) or “弟”(younger brother) is a word,too.Even we Chinese don't count the number of words,so just characters enough(I think this may work with Japanese,Korean too.),lol.

nathanlesage:
Thanks for the quick reply!

1. Yeah, nearly. I meant more like the first lines in the current dev-build translation files — I've just added some "header" or metadata field to put into some authoring details, so that this becomes common! This also makes it possible to generate a list of all contributors for the about dialog :)
2. Alright, so if Word does it and nobody has a problem with that, than this might be a good thing. Maybe sometime in the future I'll implement even something that tries to detect the language and switches the mechanism automatically based upon the file's language (but this is still futuristic, I think).
3. Ah okay, fascinating. I must confess, neither am I a linguist nor someone particularly good at languages (and I built an app with the goal of being translated in many languages, lol — I already fear the day I have to face rtl-languages in terms of UX design). But I'm always fascinated at how languages work. So thank you very much for your explanation! :)

For how the metadata should look like: For now I've just added a license and author-field. Maybe some other will be added in the future, but this depends on the necessities Zettlr may face some day, so for now author and license are enough I think (and for license: I think GNU GPL v3 should suffice for most cases)

https://github.com/Zettlr/Zettlr/blob/36ded66affe90378a0a4f213720864f958704f24/source/common/lang/en_US.json#L1-L7

AroKz:
OK,It's all up to you and thanks for your app!

nathanlesage:
Alright, cool! To merge this PR, I still need you to do two small things: First, please target the develop branch, I don't push onto the master branch before releasing a new version. And second, please add the metadata header as you wish (i.e. with a name that you would like to have put there, etc).

Then I'll directly merge it and integrate it into the app!

nathanlesage:
Update! The translation files (that you can find in the develop-branch) have had some overhauls, so I'd like to suggest to implement them in the new pull request :) This way there are no nasty english strings within a Chinese environment!

nathanlesage:
I‘ll be closing this pull request now but I‘m waiting for it to occur in the develop-branch!

Issue #102, opened 2019-02-13
frac displays incorrectly


Wieke:
## Description

For some reason the \frac{}{} command in latex doesn't render properly, the horizontal line separating the numerator and denominator is missing.

## Reproducing

Just use the frac in a math environment like this for example:

```
$$
a = \frac{b}{c}
$$
```

## Expected behaviour

I was expecting a horizontal line to appear.

**Screenshots**

![image](https://user-images.githubusercontent.com/6737186/52738702-82c20580-2fcf-11e9-9656-d2df3ec5316b.png)

## Platform
 - OS: Archlinux 
 - OS Version N/A rolling distro (kernel version 4.20.6-arch1-1-ARCH)
 - Zettlr Version 1.1.0
 - Screen Resolution 1920x1080

## Additional information

Found another math thing that renders incorrectly:

![image](https://user-images.githubusercontent.com/6737186/52741459-abe59480-2fd5-11e9-9015-0bd411826a21.png)

nathanlesage:
Thanks for the hint! I have to be honest — I didn't pay very much attention to the formula rendering process, because I'm not using it myself, but I'll see into it!

Wieke:
Heh, I suspected as much. There are multiple bugs, inconveniences and a missing feature or two related to doing math/rendering equations. Which is a shame because I quite like the editor otherwise but I tend to do a lot of math. How open are you to contributions? Cause I have half a mind to try fix these issues myself.

nathanlesage:
PRs are always welcome! You can find all code pertaining to the KaTeX-rendering in [this file](https://github.com/Zettlr/Zettlr/blob/master/source/renderer/assets/codemirror/zettlr-plugin-render-math.js). As you can see there's not much usage of the KaTeX-library at the moment:

https://github.com/Zettlr/Zettlr/blob/1cd9b6a4ba137eaae9c50449f095307395799d42/source/renderer/assets/codemirror/zettlr-plugin-render-math.js#L110-L112

nathanlesage:
So the frac is showing up perfectly now, but strangely, the unequal-bug persists:

![image](https://user-images.githubusercontent.com/17251683/52855680-6da9bb80-3122-11e9-8f78-85b505ae231b.png)

Wieke:
I'm pretty sure that's [this known katex bug](https://github.com/KaTeX/KaTeX/issues/1842) since electron is based on chromium.

nathanlesage:
[...]

PR #103, opened 2019-02-14
Update katex.css, fixes \frac issue.

Wieke:
Here's a fix for issue #102 (frac displays incorrectly). I'm not sure how the bug came to be but the css appeared to be incorrect. (I assume `source/common/assets/katex.css` was out of date.) So I just copied the content of katex.css of the katex package to the one in the assets folder and corrected the paths to the fonts.

This seems to have solved a couple of issues, the missing horizontal line in \frac is now rendered correctly, the same goes for the missing line in \overline, and now variables in an equation are correctly rendered in italic.

The problem with \ne being rendered incorrectly appears to be a [known issue](https://github.com/KaTeX/KaTeX/issues/1842) so hopefully a future update of katex will solve that  particular issue.

### DIFF
--- a/source/common/assets/css/katex.css
+++ b/source/common/assets/css/katex.css
@@ -1,3 +1,4 @@
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
 @font-face {
   font-family: 'KaTeX_AMS';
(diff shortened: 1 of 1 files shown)


nathanlesage:
Oh my god, did I really make THIS mistake:

image

😳

Well, thank you very much for this commit! I'll gladly merge it, but please do the pull request to the `develop`-branch, where it belongs! :)

Issue #168, opened 2019-04-21
Long filenames truncated in folder view


ngirard:
Using Zettlr v1.2.3 on Ubuntu 18.10.

Depending on users' scenarios, the name of files may need to be lengthy.

In this case, long file names appear truncated in the folder view, leading to a usability issue.

I'm about to submit a pull request which solves the problem, but whose consequences when using such-called "snippets" are unknown to me, since the modified code has a comment mentioning them but I simply don't know what they are.

nathanlesage:
Okay, here's the potential workaround:

1. You have a lengthy filename, which you need to see in full length
2. You hover over it with your mouse
3. Without delay (b/c hovering over such a confined element) an element is display exactly where the truncated filename is, only that this is not truncated.

Something like this (obviously with better CSS, b/c this is just a mockup):

![image](https://user-images.githubusercontent.com/17251683/56487028-0cafb280-64da-11e9-8603-e4eb4c92b3b6.png)

ngirard:
> Okay, here's the potential workaround (...)
For sure, this would be better than nothing !

Allow me to paste here your [explanations](https://github.com/Zettlr/Zettlr/pull/169#issuecomment-485339922) for the technical reason behind this defect, in case someone eventually comes up with a better solution:

> To display even long file lists, Zettlr makes use of Clusterize.js, which does not display all files, but only the ones you see.

> It basically creates a very high container (the height equals the amount of files), and only display approx. 100 (?) files at once. If you scroll, older files will be removed from the DOM, and new files are added where you scroll to. Removing the no-wrap from the file titles results in different height in the containers, and this can have severe consequences such that you may even have to re-start the GUI by reloading the thing.

It seems a bit odd to me, that a single component (Clusterize.js) takes both a model-related decision (what files to display) and a view-related decision (how tall the files shall be displayed).

But hey, you had the courage I never really had, to embrace the Javascript stack while not being fond of it, so I cannot but trust your choice to go with Clusterize.js !

nathanlesage:
> But hey, you had the courage I never really had (yet!), to embrace the Javascript stack while not being fond of it

[...]

PR #169, opened 2019-04-21
Allow directory view to display long filenames

ngirard:
Solves #168, but the consequences when using such-called "snippets" are unknown to me, since I simply don't know what they are.

## Tested On
 - OS: Ubuntu
 - OS Version: 18.10
 - Zettlr Version: 1.2.3

### DIFF
--- a/resources/less/geometry/preview.less
+++ b/resources/less/geometry/preview.less
@@ -27,7 +27,6 @@
 
             // Bigger size when snippets are on.
             &.snippets {
-                height:@tile-height-full - (2*@pane-padding); // Remove top/bottom padding
                 padding:@pane-padding;
                 line-height:inherit;
             }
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hey, thanks for the PR! I can tell you what the modifications you've done do to the list. The snippets will still work fine, but there's another problem that will completely destroy the file list in some cases: To display even long file lists, Zettlr makes use of `Clusterize.js`, which _does not display all files_, but only the ones you see.

It basically creates a very high container (the height equals the amount of files), and only display approx. 100 (?) files at once. If you scroll, older files will be removed from the DOM, and new files are added where you scroll to. Removing the no-wrap from the file titles results in different height in the containers, and this can have severe consequences such that you may even have to re-start the GUI by reloading the thing.

But if you want to see even the longest filenames at one glance, an option would be to make use of an old trick that many apps implemented earlier: As soon as your mouse hovers over a file title, an element will be displayed immediately that contains the full filename, visible at once. This is the only workable workaround I can offer you, unfortunately :/

PR #402, opened 2019-11-21
In fenced code block, add syntax highlighting for every language known.

BAKFR:
## Description

Use the CodeMirror mode list to catch tag of fenced code blocks for each languages known. The tag are generated from the name and aliases CodeMirror modes.
Every tag is lowercase.

It support of a ton of languages, and tag aliases (example: I can use either `js` or `javascript`). 

## Changes

There is no more `objectivec` tag. It's been replaced by `objective-c` or `objc`.

I didn't see any performance changes, but I invite you to test it before.

## Tested On
 - OS: Linux
 - Zettlr Version: last develop

### DIFF
--- a/source/common/data.json
+++ b/source/common/data.json
@@ -149,5 +149,95 @@
       "executivepaper": "Executive",
       "legalpaper": "Legal"
     },
-    "events": []
+    "events": [],
+    "highlightingModes": {
+      "text/javascript": {
+        "mode": "javascript",
+        "selectors": [ "javascript", "js", "node" ]
+      },
+      "application/json": {
+        "mode": "javascript",
+        "selectors": [ "json" ]
+      },
+      "text/x-csrc": {
+        "mode": "clike",
+        "selectors": [ "c" ]
+      },
+      "text/x-c++src": {
+        "mode": "clike",
+        "selectors": [ "c\\+\\+", "cpp" ]
+      },
+      "text/x-csharp": {
+        "mode": "clike",
+        "selectors": [ "c\\#", "csharp", "cs" ]
+      },
+      "text/x-java": {
+        "mode": "clike",
+        "selectors": [ "java" ]
+      },
+      "text/x-kotlin": {
+        "mode": "clike",
+        "selectors": [ "kotlin", "kt" ]
+      },
+      "text/x-objectivec": {
+        "mode": "clike",
+        "selectors": [ "objective-c", "objectivec", "objc" ]
+      },
+      "text/x-scala": {
+        "mode": "clike",
+        "selectors": [ "scala" ]
+      },
+      "text/css": {
+        "mode": "css",
+        "selectors": [ "css" ]
+      },
+      "text/x-scss": {
+        "mode": "css",
+        "selectors": [ "scss" ]
+      },
+      "text/x-less": {
+        "mode": "css",
+        "selectors": [ "less" ]
+      },
+      "text/x-php": {
+        "mode": "php",
+        "selectors": [ "php" ]
+      },
+      "text/x-python": {
+        "mode": "python",
+        "selectors": [ "python", "py" ]
+      },
+      "text/x-rsrc": {
+        "mode": "r",
+        "selectors": [ "r" ]
+      },
+      "text/x-ruby": {
+        "mode": "ruby",
+        "selectors": [ "ruby", "rb" ]
+      },
+      "text/x-sql": {
+        "mode": "sql",
+        "selectors": [ "sql" ]
+      },
+      "text/x-swift": {
+        "mode": "swift",
+        "selectors": [ "swift" ]
+      },
+      "text/x-sh": {
+        "mode": "shell",
+        "selectors": [ "shell", "sh", "bash" ]
+      },
+      "text/x-yaml": {
+        "mode": "yaml",
+        "selectors": [ "yaml", "yml" ]
+      },
+      "text/x-go": {
+        "mode": "go",
+        "selectors": [ "go" ]
+      },
+      "text/x-rustsrc": {
+        "mode": "rust",
+        "selectors": [ "rust", "rs" ]
+      }
+    }
 }
--- a/source/renderer/assets/codemirror/autoload.js
+++ b/source/renderer/assets/codemirror/autoload.js
@@ -5,6 +5,8 @@
  * in the main class.
  */
 
+const highlightingModes = require('../../../common/data').highlightingModes
+
 // 1. Mode addons
 require('codemirror/addon/mode/overlay')
 require('codemirror/addon/mode/multiplex') // Multiplex needed for syntax highlighting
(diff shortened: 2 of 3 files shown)


nathanlesage:
Thanks for this optimization, it looks good so far, and I will test it after the weekend (currently I'm attending a conference and don't have time).

Only two suggestions/ideas:

1.  I believe the regex might not work the way you expect it to: The end-delimiter (`$`) will not be recognised by the multiplex mode; I tried it with YAML (see lines 136 and 137, commented out). Therefore, the mode-list needs to be sorted by length descending (so that, e.g., `cplusplus` will be checked for before `c`) -- otherwise only `c` will be detected, never `c++`. But I think this is a minor change!
2. Maybe we should also take a look at the autoload-script ([here](https://github.com/Zettlr/Zettlr/blob/master/source/renderer/assets/codemirror/autoload.js)), because all modes need to be pulled in for CodeMirror to recognise them, and maybe we could even automate the process of enabling all modes …? Suggestions or ideas to this? Might this be a problem concerning performance?
3. For instance, `shell` and `bash` are both possible, but as I indicated in a comment in my original file, it needs to be distinguished for things such as `highlight.js` work correctly -- does your change still enable both `shell` and `bash`…?

Thanks again and I'm looking forward to your comments!

All the best
Hendrik

BAKFR:
Thank for your quick reply.

1. From my tests, the end delimiter works perfectly fine. By example, `c_` is not recognized as a language, but would be if the `$` wasn't working.
2. You're right, there is something missing here.
I through it was automatic, since I've seen color highlight for some languages (ex: `java`) not listed in the autolaod. Testing with a less common language (ex: `erlang`) doesn't work without adding it to the autoload.
I don't know how we can load every code mode (and I'm not sure it's a good idea either). I'll look into it.
3. Both `shell` and `bash` are correctly detected. CodeMirror has a lot of aliases, such as `node`, `js` and `javascript`, or `c++` and `cpp`.
   
   > it needs to be distinguished for things such as `highlight.js` work correctly

   I'm not sure to understand. You need to apply different behavior if the block is a `shell` block instead of a `bash` block ?

nathanlesage:
> I'm not sure to understand. You need to apply different behavior if the block is a shell block instead of a bash block ?

Yes, because I originally used bash (I think), but highlight.js only detects shell — or vice versa. This is why I added both.

Concerning autoloading all languages, this might be true. I'm not sure, maybe I'll try loading all at once and see how performance does :D

> From my tests, the end delimiter works perfectly fine. By example, c_ is not recognized as a language, but would be if the $ wasn't working.

Curious, but if it works, this is good!

BAKFR:
I have added the load of every CodeMirror mode. I didn't see any performance change, but I've got a good computer, so feel free to test on your side.

nathanlesage:
Perfect, thanks! And sorry that I still didn't manage to review the changes (albeit it seems not much changes will be required) :see_no_evil:

nathanlesage:
Soooo. Sorry again for the delay, but I'm just testing your amendments, and here's what some benchmarks concluded:

Loading only the modes that we have currently took `38ms`. Loading *all* modes, instead pushed up the loading time to `270ms`, so `0.2s`, which will considerably increase on older and slower machines. So that's one argument against it. Additionally, I've logged all modes that were being loaded, and I noticed something odd: Some modes were loaded multiple times; `clike` for instance was loaded 4 times, of which were two times two in a row. No, even more: Just discovered another clike. Just have a look:

image

So I think we shouldn't do this. I have no idea why this happens, but it isn't good style to load modes several times. I mean, node will capture these and not attempt to load them again, but it indicates that the `CodeMirror` modeInfo array is not unique for some reason.

Therefore, let's just keep the autoloading to explicitly stating which modes we want to load — also gives more control (e.g., I think we don't need `brainfuck` like ever). Nevertheless, was interesting testing that out!

Now on to the other addition: The extraction of the modes works fine, and almost all modes are detected (except objectivec, the Swift predecessor). But after all, the approach also suffers from the inconsistencies in the `modeInfo` object. Nevertheless, the idea is great, and with some minor additions, it'll work:

1. Let's return to the "old fashioned" approach of explicitly requiring the modes we support (albeit I would like to add `rust` while we're at it)
2. Let's keep your approach to the multiplex, but with one small change: Create an array or object at the top of the mode that holds all currently supported modes, but only the part that changes, hence: the opening, and the mime type. The opening could then make use of your regexp-approach, which indeed seemed to work well! Only, I'd suggest moving the escape-regex out of the loop (if we need it at all), so that it doesn't recompile each iteration.

Then I can merge the PR finally, and the mode should be much smaller! :)

BAKFR:
I suggest we add a single file containing the list of modes we want to load. We could use this file in both `autoload.js` and `zettlr-mode-multiplex.js`. Something like ` source/renderer/assets/codemirror/list-highlighting-modes.js`.
What do you think about this ?

nathanlesage:
Good point! There is already a file for that: `./source/common/data.json` — we could just add it there and pull it in using ´require().modes` (or whatever key we choose!)

source/common/data.json
```diff
@@ -149,5 +149,20 @@
       "executivepaper": "Executive",
       "legalpaper": "Legal"
     },
-    "events": []
+    "events": [],
+    "highlightingModes": {
+      "javascript": [ "javascript", "js", "node", "json", "typescript", "ts"],
+      "clike": [ "c", "c\\+\\+", "cpp", "c\\#", "csharp", "java", "kotlin", "objective-c", "objectivec", "scala" ],
```
BAKFR: special characters such a `+` or `#` must be escaped here.

nathanlesage: … because this way we can immediately stuff them into a regexp, yes! I think it's fine to have it there because having to compile an escaping regex to parse the trigger words in the mode loader would add nothing but computational overhead, I think it's clear enough that these are regex-escapers!

nathanlesage:
Hey, so I finally had the time to test out your commit, and there seems to be something quite odd happening after all. Just have a look at some results from my syntax highlighting file — most of the highlighting is off, and not as it's supposed to be:

C Old:

c old

C# Old: 
csharp old

C# New:
csharp new

Java old:

java old


Java new:
java new

PHP Old:
php old

PHP New:
php old

It seems as if all the clike-languages are off completely; and I would assume this is due to missing mimetypes — It seems as if we need to retain them at all costs. Here's the mimetypes in the original branch:

image

For your reference, here's the test file I'm using: https://gist.github.com/nathanlesage/af91046476f484aaf9ddd4033475c083

BAKFR:
Okay, I'll fix it tomorrow :)

Le sam. 7 déc. 2019 à 16:36, Hendrik Erz <notifications@github.com> a
écrit :

BAKFR:
Sorry for the delay.

Third time is the one. I tested most of the languages, and it seems to work as expected.

nathanlesage:
Looks good! :) I'll have a test-run in the evening and hopefully be able to merge it immediately!

source/common/data.json
```diff
@@ -149,5 +149,95 @@
       "executivepaper": "Executive",
       "legalpaper": "Legal"
     },
-    "events": []
+    "events": [],
+    "highlightingModes": {
+      "text/javascript": {
+        "mode": "javascript",
+        "selectors": ["javascript", "js", ",node"]
```
nathanlesage: That comma in front of "node" seems to be a spelling mistake, right?

nathanlesage:
So, I just tested it out, and it's working like a charm! :)

There's only just one spelling mistake (I've added a comment on that; it's with the node selector), and one last demand: Could you increase the spacing for these arrays (ESLint is set to require one space after the opening and before the closing bracket, e.g. `[ "one", "two" ]` instead of `["one", "two"]`)

Then I'll be able to merge! (Sorry that I \*again\* have something to complain, I already feel bad :( )

BAKFR:
> So, I just tested it out, and it's working like a charm! :)
> 
> Could you increase the spacing for these arrays (ESLint is set to require one space after the opening and before the closing bracket, e.g. `[ "one", "two" ]` instead of `["one", "two"]`)

Done!

> Then I'll be able to merge! (Sorry that I *again* have something to complain, I already feel bad :( )

Don't be :) You're the only guardian of the code quality.
besides, I'm the one who hasn't checked eslint and missed the comma.

nathanlesage:
Awesome! All set now! Thank you very much for your initiative, a place in the Changelog is yours ❤️

PR #419, opened 2019-11-30
Added AppImage support

yashha:
## Description

Adds AppImage building feature for easy installing on every system.

## Changes

- When building linux it will build an AppImage for x64 and ia32.
- Changed artifact name to include the achitecture. Zettlr-linux-${version}-${arch}.${ext}
 I thought version should be before arch but I can change this back if you want.
- Changed file associations array. It was necessary for Appimage building.
  This could maybe have not intendend effects?

## Tested On
 - OS: Debian Ubuntu
 - OS Version: 18.04.03
 - Zettlr Version: v1.5.0-beta.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
         "release:mac": "node scripts/build-app.js --mac",
         "release:win": "node scripts/build-app.js --win",
         "release:linux": "node scripts/build-app.js --linux",
+        "release:app-image": "node scripts/build-app.js --app-image",
         "less": "node scripts/less.js",
         "less:extract": "node scripts/extract-css.js",
         "handlebars": "node scripts/handlebars.js",
--- a/scripts/build-app.js
+++ b/scripts/build-app.js
@@ -18,6 +18,7 @@ const path = require('path')
 let flags = process.argv // Contains the CLI flags
 let buildTargets // Contains all build targets
 let onlyDir = false // Is set to true, if --dir flag is given on command line
+const arifactFilenameFormat = 'Zettlr-${version}-${platform}-${arch}.${ext}'
 
 // Set the current working directory to the app's root.
 process.chdir(path.resolve(__dirname, '../'))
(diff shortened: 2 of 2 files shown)


nathanlesage:
Oh, and P.S.: PRs only to the develop-branch, please, I won't merge this into master! Thaaaankss

yashha:
To 2. It seems thats now the standard way to write it. And the longform of markdown is included.
To 3. I put the AppImageTarget into a contant, that looked cleaner.

Should I leave the arch before the version? I think it looks better. Also should I align the other packages then also?

nathanlesage:
Alright, if the long form is also included, why not. Concerning the filenames: I'm not sure whether or not changing the pattern would cause the API to crash, there's a lot of possibility of regression bugs :S I'll have to check its code to make sure! If the API doesn't care then we can certainly switch — is there any "unwritten standard" as to how the artifact names of released files should be? Something like semver for versioning? If so, I'd like to stick to the most widely adopted naming convention. The more common something is, the less other stuff will break when interacting :)

We'd definitely have to tell these guys: https://github.com/Homebrew/homebrew-cask/blob/master/Casks/zettlr.rb …

P.S.: Just checked the API, and it does not care at all, it'll happily eat whatever I feed it! So we can change the artifact name to the most common way of naming them!

In the meantime, I'll quickly test building AppImages on my machine!

yashha:
I just know it from other packages like visual studio code. The naming the other way around seemed odd to me.

https://code.visualstudio.com/Download
https://github.com/ramboxapp/community-edition/releases

Atom has it different:
https://github.com/atom/atom/releases

yashha:
Here not one standard file format, I guess.
https://askubuntu.com/questions/330018/what-is-the-standard-for-naming-deb-file-name

nathanlesage:
:D Well, well. Then we should use some reasonable format. What I've seen with a lot of programs, and which is pretty obvious, is to put the version number behind the name. And concerning sorting, we should start with the artifact parts that are less specific and end with those that are most specific. So this would mean:

`Zettlr-${version}-${platform}-${arch}.${ext}`; so compared to currently just move forward the version.

Besides, I think some variable like `platform` is already supported, I think? If so, we could make the artifact name much more modular. That I left the literal platforms in there was, if I remember correctly, because it wasn't possible back when I first wrote the script. But I'm not sure. Your opinion?

yashha:
Sounds good to me, I'll make a commit.

nathanlesage:
Looks wonderful! I'll be testing it out hopefully today, just to see how it goes, and then I'll merge this thing :)

nathanlesage:
Alright, it seems to work, although I have no chance to test this out due to missing hardware and disk space for virtual images. Do you happen to have a macOS to build and then test on a linux distribution?

yashha:
I have only a linux. It worked there.

nathanlesage:
Mh … I have an idea: Would you be willing of supplying the AppImage for the time being? I'm asking because I can't fix AppImage related problems obviously … In the meantime, I'll just merge that thing because it obviously works and does not interfere with the rest of the build process. :)

PR #422, opened 2019-12-02
Create a keyboard shortcut for Task Lists

jeffgeorge:
## Description
Adds a new keyboard shortcut to create a Task List (`Cmd / Ctrl + T`).

I'm looking to use Zettlr primarily as a todo-list tracker so I make frequent use of task lists. Toggling them without leaving the keyboard is helpful and the Markdown syntax is _so many characters_, so a shortcut seemed like a good solution.

## Changes
I just updated the Markdown Shortcuts CodeMirror plugin and Editor's Context Menu. As best I can tell these are the only places in the application that reference Markdown shortcuts.

## Tested On
 - OS: macOS
 - OS Version: 10.15.1 (Catalina)
 - Zettlr Version: Latest Develop

## Additional information
 - I chose `Cmd+T` / `Ctrl+T` because it both didn't appear to already be in use and because it's common across other note apps (Bear, for instance, uses `Cmd+T`). If we'd prefer to save T for other purposes (Insert Table, maybe?) we could consider using the somewhat unruly `Shift+Cmd+L` shortcut as used by Apple Notes.
 - I've only tested on macOS at the moment, but since the changes directly copy how other shortcuts are implemented I'm not *super* concerned.
 - Should this PR be accepted, the docs would need to be updated. Specifically, the [shortcuts reference](https://github.com/Zettlr/zettlr-docs/blob/master/docs/en/reference/shortcuts.md) and probably also the [Zettlr as a note-taking app](https://github.com/Zettlr/zettlr-docs/blob/master/docs/en/guides/guide-notes.md) guide.
 - Also, should this PR be accepted, I may consider implementing further enhancements to the Task List functionality, specifically adding another shortcut to toggle a given line's task list item as checked/unchecked, and updating the CSS to strike-through a checked list-item (or by applying actual strike-through Markdown -- `~` -- to the item)

### DIFF
--- a/source/renderer/assets/codemirror/zettlr-plugin-markdown-shortcuts.js
+++ b/source/renderer/assets/codemirror/zettlr-plugin-markdown-shortcuts.js
@@ -596,4 +596,6 @@ const { clipboard } = require('electron');
   CodeMirror.keyMap['default']['Shift-Ctrl-I'] = 'markdownImage'
   CodeMirror.keyMap['default']['Shift-Cmd-C'] = 'markdownComment'
   CodeMirror.keyMap['default']['Shift-Ctrl-C'] = 'markdownComment'
+  CodeMirror.keyMap['default']['Cmd-T'] = 'markdownMakeTaskList'
+  CodeMirror.keyMap['default']['Ctrl-T'] = 'markdownMakeTaskList'
 })
--- a/source/renderer/assets/context/editor.json
+++ b/source/renderer/assets/context/editor.json
@@ -32,6 +32,7 @@
   },
   {
     "label": "menu.insert_tasklist",
+    "accelerator": "CmdOrCtrl+T",
     "command": "cm-command",
     "content": "markdownMakeTaskList"
   },


nathanlesage:
You're pretty lucky, with `Cmd/Ctrl+T` you hit the only free key on the upper row on the keyboards 😂

image

> If we'd prefer to save T for other purposes (Insert Table, maybe?)

That would indeed be a question we'd need to answer. But I think there's no way we could possibly foresee how many table rows and cols the user would want, so by hitting the shortcut, we could only pop up the dialog to select rows and cols, which, in my opinion, does not benefit so much from having a shortcut assigned, because you'll be needing the mouse one way or the other.

Maybe, to make it easier to find out whether `Cmd/Ctrl+T` works: All shortcuts should follow the rules outlined here: https://docs.zettlr.com/en/reference/shortcuts/

What I should possibly add is, that, as Zettlr is meant as a transitory solution for people who only know Word, we should always have a look at the shortcuts Word and LibreOffice are using.

But I just checked: `Cmd/Ctrl+T` is used to "Manage styles" in LibreOffice and "Hanging Indent" in Word, so both functions where Zettlr explicitly departs from the Word processor way, so it seems harmless.

Concerning the files changed: At first I feared you may have forgotten the command, but then I realised it's already in there, so it seems good to go — the only question left is: Are there any potential problems concerning the `Cmd/Ctrl+T`-shortcut, or are we good to go?

P.S.: If you want to take a second look at possibilities with shortcuts, have a look here: https://zettlr.com/shortcuts

jeffgeorge:
> Maybe, to make it easier to find out whether `Cmd/Ctrl+T` works: All shortcuts should follow the rules outlined here: https://docs.zettlr.com/en/reference/shortcuts/

Yep, I checked that reference page before picking `Cmd/Ctrl+T`. I think it clearly fits all those guidelines, with -- as you said -- the added bonus of not conceptually colliding with traditional word processors.

Regarding follow-up work, based on the guidelines, does `Cmd/Ctrl + Option/Alt + T` sound reasonable for the command to toggle a TaskList Item as checked / unchecked? That key combo is a *little* uncomfortable, but it's open across all platforms and would fit the conceptual guidelines of being an alternative action on the same "target". Open to suggestions!

> P.S.: If you want to take a second look at possibilities with shortcuts, have a look here: https://zettlr.com/shortcuts

This is easily the best shortcut visualizer I've ever seen. Should I go file a PR vs [zettlr.config.yml](https://github.com/nathanlesage/cheatsheet-generator/blob/master/examples/zettlr.config.yml) to coincide with this update?

I can't think of any other issues with this, if it all looks good to you I'll prep a PR for the shortcut visualizer and at least the English version of the docs regarding this update.

nathanlesage:
> This is easily the best shortcut visualizer I've ever seen. Should I go file a PR vs zettlr.config.yml to coincide with this update?

Please! This would make my life easier because I can rest assured that this one doesn't miss any shortcuts :)

Concerning Ctrl+Alt+T — do you think we really need this? 🧐 I mean nothing is more satisfying than taking the time to actually click a checkbox and forget it!

I'll be merging the PR right away!

jeffgeorge:
Fair point re: bothering with a shortcut for checking the checkbox. I'm still interested in updating the checkbox logic from just marking an x in the `- [x]` markdown syntax to additionally adding the `~` strike-through syntax around the list item. I'll figure that out and then worry about a shortcut later.

Look for the other PR's to land in a couple hours. I'll definitely need help on the German portion of the update to the docs!

nathanlesage:
P.S.: Would you mind also updating the shortcuts docs page with the new shortcut? Then we have everything in place!

PR #424, opened 2019-12-02
add "y" two digit year

asteres:
Added a 2 digit year. ie: "19" instead of "2019".

### DIFF
--- a/source/common/util/replace-string-variables.js
+++ b/source/common/util/replace-string-variables.js
@@ -22,6 +22,8 @@ const uuid4 = require('uuid/v4')
 module.exports = function (string) {
   let date = new Date()
   let yyyy = date.getFullYear()
+  let yy = date.getFullYear().toString().substr(-2)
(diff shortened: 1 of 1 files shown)


source/common/util/replace-string-variables.js
```diff
@@ -22,6 +22,8 @@ const uuid4 = require('uuid/v4')
 module.exports = function (string) {
   let date = new Date()
   let yyyy = date.getFullYear()
+  let yy = date.getFullYear().toString().substr(-2)
```
nathanlesage: To make line 26 work, here we'd need to wrap the expression into `parseInt()` (**[and also specify the radix 10](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)!!**) to re-convert it into a two-digit number. Otherwise `yy <= 9` will behave counterintuitive.

nathanlesage:
Closing due to inactivity

Issue #391, opened 2019-11-11
[ENHANCEMENT] More Settings for File Linking


nathanlesage:
## Description

When linking files, the current behaviour of Zettlr is to put the ID in the brackets and add the file name (excluding the extension) afterwards in all cases, even if the file has no ID and will therefore be linked using only the filename. This is undesirable. Nevertheless, there may be certain preferences by different users regarding that policy, so we need to make it accessible using a settings option.

## Proposed Enhancements

1. If the link contents equal the filename, do not add the filename.
2. If desired, do not add the filename at all.

This can be achieved by adding a new radio option in the Zettelkasten-tab with three states:

- Never add the filename when choosing files to link to.
- Add the filename when choosing files to link to, if these files have an ID.
- Always add the filename when choosing files to link to, even if they do not have an ID.

## Wording

The proposed wording for the English translations is:

```
Add the filename after linking to a file:

( ) Always
( ) Only if linking with ID
( ) Never
```

raoul-van-oosten:
Yes, I would love this! Currently, mine is linked as "[[ID]] ID <title>". Is that normal behavior?

raoul-van-oosten:
Of course, that makes sense. Maybe I should discard ID from the title, then, because I always (have to) remove it manually now.

nathanlesage:
Tadamm 🎶

PR #440, opened 2019-12-20
De-duplicate ID when expanding ID links

patmaddox:
## Description

When expanding ID links, I would like the ID to be included in brackets, and the remainder of the file name to follow the brackets.

### Current behavior

Zettlr currently shows the file ID twice – once in the brackets, and once after the brackets:

zettlr-id-autocomplete-current

### Proposed new behavior

Display the ID in brackets, and remove it from the trailing text:

zettlr-id-autocomplete

(note: I do think it makes sense to keep the ID in the hints list)

## Changes

I made a small change to the `CodeMirror.on(completionObject, 'pick')` call in `zettlr-editor.js`.

## Tested On
 - OS and version: macOS 10.13.6
 - Zettlr version: 1.5

### DIFF
--- a/source/renderer/zettlr-editor.js
+++ b/source/renderer/zettlr-editor.js
@@ -148,15 +148,15 @@ class ZettlrEditor {
           CodeMirror.on(completionObject, 'pick', (completion) => {
             // When a user inserts an ID, neither the citeproc IDs nor the
             // tagDB will be loaded, but a generated library containing the
-            // displayText-property.
+            // expandedText-property.
             if (this._currentDatabase !== this._tagDB &&
               this._currentDatabase !== this._citeprocIDs &&
-              completion.displayText) {
-              // We need to add completion.displayText after the completed thing.
+              completion.expandedText) {
+              // We need to add completion.expandedText after the completed thing.
               let cur = JSON.parse(JSON.stringify(cm.getCursor()))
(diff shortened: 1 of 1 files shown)


patmaddox:
I do have two questions about this PR:

1. It changes the current behavior. Should I add a "De-duplicate ID when expanding ID links" toggle in the Editor options?

2. How should it handle files that don't have an ID, or are just an ID? e.g. `README` and `20191219164220`? I feel like it could or should de-duplicate those as well, so you would end up with just `[[README]]` and `[[20191219164220]]` rather than `[[README]] README` and `[[20191219164220]] 20191219164220`

nathanlesage:
Current behaviour is to use the ID, if any, in the Link and the full filename (even if it also contains the ID) afterwards.

Your questions, if I understand them correctly, are answered by this proposal: https://github.com/Zettlr/Zettlr/issues/391

This would require an additional preferences toggle. I’m currently on vacations, but please feel free to direct any questions to me after the 1st January. Then I’ll also have a look at your PR!

nathanlesage:
Hey, I'll be closing this PR now, as I just implemented the file linking (but with the preference setting). But thank you very much for reminding me that it's also possible that the user puts the ID in the filename, I would've forgotten this otherwise! So thank you very much! :)

PR #449, opened 2019-12-29
i18n: Fix typo in french translation ("% s" instead of "%s")

BAKFR:


nathanlesage:
Thank you for spotting that! Nevertheless, the translations are managed over at [Zettlr Translate](https://translate.zettlr.com). The four default language files are in source control here so that anybody cloning the repository has them available! If you'd be able to apply the fixes there, I would be greatful!

BAKFR:
My bad, I didn't see that.
I'm on my way to fix it the proper way.

BAKFR:
Done, but the fixes will not be effective until someone else has approved the translations I've added.

nathanlesage:
I've approved them, should work now! Thanks again :)

PR #450, opened 2019-12-29
Use raw string to transfer i18n data from main process to renderer process

BAKFR:
Previously, i18n data were transferred using a JSON object copied once at start.
It led to numerous IPC calls with heavy performance cost.
This commit add a loadI18n() method using raw string instead object, thus
avoiding these calls.

On my computer, i18n loading time went from ~150ms to less than 5ms

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fixed wrong error logging in the Citeproc provider.
 - Added the necessary `cslenvironment` to Zettlr's default TeX template so that Pandoc >2.8 does not throw errors. Thanks to @frederik-elwert for implementing!
 - Cleaned up the keymap for CodeMirror. It's now centralised within `generate-keymap.js` and not scattered in all the plugins anymore.
+- Rewrote the i18n loading logic, resulting in huge performance gains on startup. Thanks to @BAKFR for implementing!
 
 # 1.5.0
 
--- a/source/common/lang/i18n.js
+++ b/source/common/lang/i18n.js
@@ -50,15 +50,17 @@ function i18n (lang = 'en-US') {
 
   // Cannot do this asynchronously, because it HAS to be loaded directly
   // after the config and written into the global object
-  global.i18n = JSON.parse(fs.readFileSync(file.path, 'utf8'))
+  global.i18nRawData = fs.readFileSync(file.path, 'utf8')
+  global.i18n = JSON.parse(global.i18nRawData)
 
   // Also load the en-US fallback as we can be sure this WILL both stay
   // up to date and will be understood by most people.
   let fallback = getLanguageFile('en-US') // Will return either the shipped or updated file
-  global.i18nFallback = JSON.parse(fs.readFileSync(fallback.path, 'utf8'))
+  global.i18nFallbackRawData = fs.readFileSync(fallback.path, 'utf8')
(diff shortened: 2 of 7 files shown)


nathanlesage:
What ...? So basically the JSON-string-conversion built into Electron is actually **slower** than the JSON-string-conversion using `JSON.stringify`? Really? :D 

I've had a look at the code, and it looks good so far, but I need to take another one, somewhen this week. I like to rely upon builtin functionality as much as possible, but if the difference is **that** huge, I'd also consider manually stringifying the object.

BAKFR:
The problem is, the object returned by `remote.getGlobal()` is not a JSON object, but a proxy object.
The framework intent is to provides fully shared object (ie you have a reference over the object on the other side, not a copy).

When you do a `JSON.stringify()` on this proxy object, it will read each property one by one, with a round trip to the main process each time. The problem is not stringify() itself; attempting to do it manually will result in the same performance loss.

The rule is: shared JSON objects are a performance nightmare no matter what.

nathanlesage:
Ah, yes, you're right! I wasn't fully awake when I looked at the code and forgot that I do the detaching in the renderer, originally prompted because of low speed while accessing the object. This explains the performance increase.

source/renderer/util/load-i18n.js
```diff
@@ -0,0 +1,15 @@
+
```
nathanlesage: First, I would suggest to keep this file in the language-folder, to keep all language-related things together. Besides, this file is not only used by the renderer process (that is: the main window), but by others as well.  Therefore, the `common/lang`-directory is better for this function.

Then, I'm not sure whether or not `load-i18n` is a good name, because `i18n()` and `loadI18n()` in principle do exactly the same thing, only for different process types. Maybe keep `i18n()`, or rename it to `i18nMain()`, and name the other one `i18nRenderer()`, indicating the different processes this is being used in.

Last but not least: If we keep it in a separate file (which I would suggest), please copy over a file header from one of the other files, adapting it w/ your name/user-name/other nickname, optionally a mail, to indicate this was your work and to make sure we have the GNU GPL v3-license in the file!

Besides that, I'm pretty happy with it, and as it works for you, I'd happily merge it!

BAKFR: Here it is!
I've fixed all yours points, although I think we could have kept the new function into i18n.js as well.

Small question for the next time: should I update the CHANGELOG.md for each pull request , or you do it once per version update ?

nathanlesage: Please feel free to also update the Changelog!

In general, I tend to put visible improvements in user experience as "new features", everything else that users might notice into "GUI and Functionality" and everything users might not be able to see into "Under the Hood", so this issue would go to the latter section!

Plus, as you're a contributor, please feel free to add your GitHub handle (I've done this in the 1.5 Changelog, just have a look how I mentioned contributors there) :)

BAKFR: Also, when changes are required, I always rebase my commits out of habit, instead of adding more commits to fix the previous ones. Would you prefer I keep the history ?

nathanlesage: Eeeeeeeh … I don't know — I mean having more history means that in the worst case we can track down issues, right? I've never done a rebase, I tend to only glue all commits to the HEAD of a branch without merging them together, so … I don't know. For this particular instance, it should not be a problem, but in general I would say let's try to keep as much history as possible!

PR #463, opened 2020-01-07
Modify TeX template for pandoc 2.8

### LINKED ISSUE #455: [BUG] CSL Environment error on PDF exports for Pandoc > 2.8
## Description

[See the forum discussion](https://forum.zettlr.com/discussion/77/problem-pdf-export-with-citations-or-images).

## Reproducing

Apparently, having Pandoc 2.8 installed causes the error, as some additional TeX commands to the default template are missing which Pandoc expects.

## Expected behaviour

Produce a valid PDF, no matter the Pandoc version (although Pandoc >2.0 is required for Zettlr)

**Screenshots**

## Platform

 - OS and version: Not applicable
 - Zettlr Version: Not applicable (depends on Pandoc version)

## Additional information

Should you experience this issue, mitigation is possible by downloading the official template, adding the required code and pointing Zettlr to use that template instead of the built-in. The necessary steps are described also in the forum thread.

frederik-elwert:
Add the custom environment for CSL citations that pandoc 2.8 requires. Fixes #455.

## Description

Pandoc 2.8 uses a dedicated custom latex environment for typesetting indented bibliographies. If this is missing from the latex template, compilation will fail. This pull request adds the corresponding snippet.

## Changes

Just added the corresponding snippet from the original pandoc template.

## Tested On
 - OS and version: Ubuntu 18.04
 - Zettlr version: 1.5.0
 - pandoc version: 2.7.3, 2.8, 2.9.1.1 (works fine also under pre-2.8!)

### DIFF
--- a/source/main/assets/export.tex
+++ b/source/main/assets/export.tex
@@ -166,6 +166,8 @@ pdfborder={0 0 0}
 \providecommand{\tightlist}{%
   \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
 
+% The following commands are used by Pandoc if a CSL bibliography is present
+% to create proper indentation, especially for styles with hanging indentation.
 $if(csl-refs)$
 \newlength{\cslhangindent}
 \setlength{\cslhangindent}{1.5em}
--- a/source/main/assets/export.tex
+++ b/source/main/assets/export.tex
@@ -166,6 +166,15 @@ pdfborder={0 0 0}
 \providecommand{\tightlist}{%
   \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
 
+$if(csl-refs)$
+\newlength{\cslhangindent}
+\setlength{\cslhangindent}{1.5em}
+\newenvironment{cslreferences}%
+{$if(csl-hanging-indent)$\setlength{\parindent}{0pt}%
+\everypar{\setlength{\hangindent}{\cslhangindent}}\ignorespaces$endif$}%
+{\par}
+$endif$
+
 \title{$PDF_TITLE$}
 \author{$PDF_AUTHOR$}
 


source/main/assets/export.tex
```diff
@@ -166,6 +166,15 @@
 \providecommand{\tightlist}{%
   \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
 
+$if(csl-refs)$
```
nathanlesage: We need a short comment explaining why this csl-refs-clause is here, just so we don't wonder in the future!

source/main/assets/export.tex
```diff
@@ -166,6 +166,15 @@
...
+\newlength{\cslhangindent}
+\setlength{\cslhangindent}{1.5em}
+\newenvironment{cslreferences}%
+{$if(csl-hanging-indent)$\setlength{\parindent}{0pt}%
+\everypar{\setlength{\hangindent}{\cslhangindent}}\ignorespaces$endif$}%
+{\par}
+$endif$
+
```
nathanlesage: Did you check whether this code is also backwards-compatible to pre-2.8-versions of Pandoc?

alerque: See the main PR message, the OP says it was tested on 2.7.3. From experience in another context I'm pretty sure this looks right.

source/main/assets/export.tex
alerque: Half the stuff in the Pandoc template is pretty opaque if you don't know the internals of what Pandoc / LaTeX. I suggest just tracking the upstream changes to the data template as closely as possible.

source/main/assets/export.tex
nathanlesage: Oh damn it, I'm not completely awake today. Sorry, my mistake and thanks @alerque for pointing me to it! :)

source/main/assets/export.tex
nathanlesage: Absolutely, but we know for sure that this environment was introduced in 2.8, so I just want this comment here so that it's easier to spot.

Reading a short comment is always simpler than having to dig up the full version history of Pandoc's template.

frederik-elwert: I added a comment. Is this what you had in mind?

nathanlesage: Yes, thank you very much! :)

nathanlesage:
Aaaaand here we go! Thank you very much!

PR #516, opened 2020-02-09
Update build.yml

stephenbawks:
## Tested On
 - OS and version:
 - Zettlr version:

## Additional information

I read the article and commented in reddit, now I am making one small change to your New-Item.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
       # attempts literally the same thing.
       - name: Manually create the handlebars-directory (Windows only)
         if: matrix.os == 'windows-latest'
-        run: New-Item -Path "./source/common/assets" -Name "handlebars" -ItemType "directory" # Hello, PowerShell ._.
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hehe, were you the one not fond of my aversion to Windows? =D

What exactly does your proposed change do differently, and why is it better? Seems like … different Syntax, right? But I'm not a PowerShell power user so I have no idea

stephenbawks:
That was me.

Growing up doing Windows support for years  and having a special place in
my heart for Powershell I had to make one 'correction' to your code!

Primarily work on Linux of course but still love Powershell.

nathanlesage:
Haha okay :D But I'm still curious as to what the difference is between the two commands?! :see_no_evil:

stephenbawks:
Technically there is no difference. Mine is just easier to read for most
people coming in with bash experience.

nathanlesage:
Ahh, alright! Yeah, I mean why not, if it works … but I can't merge this into master because this would trigger the build workflow and result in, like, ten emails being sent to me :roll_eyes:

stephenbawks:
Yes I agree emails are annoying and I try to avoid them at all costs at
work.

PR #517, opened 2020-02-09
Paste current id if it exists

Zverik:
When you press Ctrl+L / Cmd+L, Zettlr inserts a unique id in the document. The id is always freshly generated: you can insert 100 different ids. It's okay I guess. But.

I think the idea for this key is to generate a new id when there isn't one. But where there's a registered id, it makes more sense to keep it, and not generate new ids.

This PR changes the behaviour of Ctrl+L, so that when a file has an id, it inserts it instead of generating a new one. That would help me in moving ids from file names to file contents (I made this mistake of keeping ids in file names), and I guess it will be helpful in not creating multiple ids for a single file.

One thing it's missing is that in case of a new file, it should paste just one id, instead of generating many. Not sure how to do that withoug breaking anything. Do I just `this.getCurrentFile().id = ...`?

Tested on Fedora Linux 31.

### DIFF
--- a/source/renderer/zettlr-renderer.js
+++ b/source/renderer/zettlr-renderer.js
@@ -181,8 +181,11 @@ class ZettlrRenderer {
     let image = clipboard.readImage()
     let rtf = clipboard.readRTF()
 
+    // If the file has an id, paste it instead of generating a new one
+    let f = this.getCurrentFile()
+    let id = (f && f.id) ? f.id : generateId(global.config.get('zkn.idGen'))
     // Write an ID to the clipboard
-    clipboard.writeText(generateId(global.config.get('zkn.idGen')))
+    clipboard.writeText(id)
     // Paste the ID
     remote.getCurrentWebContents().paste()
 


nathanlesage:
I see where you are going with this. However, what about cases where a part of a URL is recognised as an ID and you want to give it a "correct" ID? This would make this impossible. We already have the "copy ID" feature, so I'd like to keep it more verbose. I don't like it when apps do some magic under the hood as this might result in unepected behaviour.

Zverik:
Thanks Nathan. How do i activate the "Copy Id" feature? Could not find it in docs on [shortcuts](https://docs.zettlr.com/en/reference/shortcuts/), [editor](https://docs.zettlr.com/en/core/editor/), or on any of the Zettelkasten pages.

nathanlesage:
If a note has an ID, you can right-click it to copy it:

image

Apparently, the docs need updating … if you happen to see a place where it would fit, please feel free to add it! I'll gladly merge this one to the docs repository: https://github.com/Zettlr/zettlr-docs

PR #544, opened 2020-02-15
Fix IME position at all zoom levels

ryota-abe:
## Description
We have already applied [a workaround for the IME issue](https://github.com/Zettlr/Zettlr/commit/c64173d4947ad910f7560c400ca306511dea494f), but it was not complete.

## Changes
The source of the problem is the textarea, whose size and position must exactly match the actual displayed text.

### Current state
- The dimensions of the textarea do not match the actual displayed text.
- The vertical position of the textarea is in the middle of the parent element. (This is probably a Chromium bug.)
image

### Amendment
- Adjust the dimensions of the textarea.
- Change the shape of the parent div to work around a positioning bug in Chromium.
image

## Tested On
 - OS and version:
    1. Windows 10 Home 1903 (+ Japanese Microsoft IME)
    2. Windows 10 Home 1903 (+ Microsoft Pinyin)
    3. Ubuntu Desktop 18.04 LTS (+ Mozc)

### DIFF
--- a/resources/less/geometry/codemirror.less
+++ b/resources/less/geometry/codemirror.less
@@ -9,9 +9,18 @@
   // hidden. This way, pop-ups for non-western
   // fonts such as Hiragana/Katakana are correctly
   // aligned. Should also help accessibility-plugins.
-  &.CodeMirror-wrap > div > textarea {
-    font-size: 1.6em;
-    line-height: inherit;
+  &.CodeMirror-wrap > div:first-child {
+    height: 1.4375em !important;
+    width: 0 !important;
+    textarea {
+      font-family: inherit;
+      font-size: inherit;
+      line-height: inherit;
+      height: 1.4375em !important;
+      top: 0 !important;
+      bottom: 0 !important;
+      margin: auto;
+    }
   }
 
   line-height:1.4375;


nathanlesage:
Oh boy, I have to get back to that linter workflow 🙄

But crazy how much code is necessary to alleviate the problem, holy! Might well consider checking this on one of the[ CodeMirror demos](https://codemirror.net/mode/gfm/index.html), and, if it's reproducible, ping Marijn that he fixes this in the core…?

Does the textarea now work fully with all zoom levels?

ryota-abe:
The problem is partially reproducible on the CodeMirror demo page.
image
- The size of the text area is not correct.
- But its position looks correct. 🤔 

And [a related issue is alreadly open](https://github.com/codemirror/CodeMirror/issues/4257).

---
> Does the textarea now work fully with all zoom levels?

Absolutely.
fix_ime_position2

nathanlesage:
> And a related issue is alreadly open.

Perfect, thank you very much for looking it up!

> > Does the textarea now work fully with all zoom levels?
>
> Absolutely.

Perfect! :) Then I'll merge! ありがとうごさいます

PR #564, opened 2020-02-22
Remove stray copy file

Zverik:
## Description

There's one file that's not referenced from anywhere, and looks like a stray macOS copy of an older version. Introduced in 0cc3de469fb99f808d2aa3f9cf39472a09b0fff3. Mind if we removed it?

## Changes

Nothing changes for the UI.

## Tested On
 - OS and version: Fedora Linux 30
 - Zettlr version: 1.6.0 beta5


nathanlesage:
That's the backup of the Markdown mode I made during refactor. I wanted to keep it until 1.6 is released, but I think that more or less it can be removed by now …

Zverik:
Well, it's git: you can extract any version of any file :)

PR #565, opened 2020-02-22
 Fix to show search result in sidebar even for a single file

ryota-abe:
## Description
If the result of global search contains only a single file, nothing appears in the sidebar.

## Changes
Fix to show single file in sidebar.

## Tested On
 - OS and version: Windows 10 Home 1903

## Additional information
Not reproduced in v1.5.0


nathanlesage:
Thanks for spotting!

PR #569, opened 2020-02-25
Fix table editor to recognize first line of file

ryota-abe:
## Description
When inserting a table at the very beginning of the file, the first row is not recognized by the table editor.
image

## Changes
At the first line of the file, the index is 0 and evaluates to falsy. So I changed it to compare with undefined.

## Tested On
 - OS and version: Windows 10 Home 1903
 - Zettlr version:  Latest develop

### DIFF
--- a/source/renderer/assets/codemirror/zettlr-plugin-render-tables.js
+++ b/source/renderer/assets/codemirror/zettlr-plugin-render-tables.js
@@ -49,7 +49,7 @@ const Table = require('../../util/table-helper.js');
       // First get the line and test if the contents resemble a table
       let line = cm.getLine(i)
       if (tableRE.test(line)) {
-        if (!firstLine) {
+        if (firstLine === undefined) {
           firstLine = lastLine = i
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hey, thanks for the PR! I'll have a look in the next week (currently my time resources are segfaulting completely), but just a short info for you: I had to partially revert the changes you implemented for fixing the IME editor, because somehow the first child div introduced an invisible "barrier" at the top of files (if you scrolled down a file, you'd see that the text does not disappear under the toolbar, but actually some pixels before that) :S

ryota-abe:
Okay. I patiently wait 😄

> I had to partially revert the changes you implemented for fixing the IME editor

Thanks for letting me know about it. It seems better not to tweak the invisible div. I'll take another look at this problem.

nathanlesage:
So, just had time to verify, and I did verify, and it did work, as expected, so here we go! どうもありがとうごさいます！

PR #576, opened 2020-02-29
Use Joplin-Turndown for parsing HTML

### LINKED ISSUE #575: [ENHANCEMENT] Replace Showdown with Turndown
## Description

Zettlr uses Showdown.js for converting pasted HTML text into Markdown. That leaves some issues (e.g. #573) due to this conversion being experimental and unmaintained. I propose to replace it with turndown — more specifically, the [joplin-turndown](https://github.com/laurent22/joplin-turndown) branch, which looks better suited for this conversion.

## Proposed Changes

Replace showdown with joplin-turndown. There is also [joplin-turndown-gfm](https://github.com/laurent22/joplin-turndown-plugin-gfm) plugin that might be useful.

## Caveats

Cannot see any: turndown is better maintaned and more configurable.

Zverik:
## Description

Adds `joplin-turndown` with its GFM plugin to convert HTML to Markdown, instead of Showdown. Keeps Showdown for md→html conversion.

Fixes #575.

## Changes

* Added `joplin-turndown` and `joplin-turndown-plugin-gfm` to dependencies.
* Updated `yarn.lock` files.
* Switched showdown with turndown in the editor's `this._cm.on('beforeChange')`.

## Tested On
 - OS and version: Fedora Linux 30
 - Zettlr version: 1.6.0 beta5

I absolutely love how it works now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,8 @@
         "codemirror": "^5.51.0",
         "command-exists": "^1.2.7",
         "got": "^10.5.5",
+        "joplin-turndown": "^4.0.22",
+        "joplin-turndown-plugin-gfm": "^1.0.12",
         "jquery": "^3.4.1",
         "katex": "^0.11.1",
         "md5": "^2.2.1",
--- a/source/package.json
+++ b/source/package.json
@@ -25,6 +25,8 @@
         "codemirror": "^5.51.0",
         "command-exists": "^1.2.7",
         "got": "^10.5.5",
+        "joplin-turndown": "^4.0.22",
+        "joplin-turndown-plugin-gfm": "^1.0.12",
         "jquery": "^3.4.1",
         "katex": "^0.11.1",
         "md5": "^2.2.1",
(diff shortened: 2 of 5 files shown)


nathanlesage:
Holy hell, it's indeed much better than showdown's very strict not-throwing-away-information approach. Didn't know they made the joplin-library standalone! Again another big dependency, but it does indeed work very good!

Did you double check that there's not another HTML->MD stuff going on where there's still Showdown …?

Zverik:
Yes, that was the only invocation of Showdown's `makeMarkdown` function. The library primary purpose was to convert md→html, and the reverse conversion was an experiment and an afterthought, not very advanced because that wasn't the primary function. The dependency is indeed big, but pasting is among the most used functions, so I believe it's worth adding.

nathanlesage:
> and the reverse conversion was an experiment and an afterthought, not very advanced

… this was the feeling I got myself.

> The dependency is indeed big, but pasting is among the most used functions, so I believe it's worth adding.

Yes, I don't like two different libraries, but for what it's worth, 2 MB won't really matter with the already, like 260 Megabytes, Zettlr is punching on the disks 😄

PR #591, opened 2020-03-08
clarification new block math render CHANGELOG.md

mkoliba:
Added clarification for significant change in the required format for math block equation to render.

## Description
Added clarification for significant change in the required format for math block equation to render. Before double dollar `$$` tags could be the same line with the equation. In this version have to be separate lines. I consider important to document this change in behaviour. 

## Changes
Added clarification in CHANGELOG.md


nathanlesage:
Ah, alright. I see; I removed that behaviour because I thought it's not as often used …?

In case double-dollar-signs should work again, please just open up a new issue! However, retrospectively fixing the Changelog will not be seen by anybody, plus it's a little bit like "rewriting history". However, I'll immediately dispatch a tweet informing everyone of the change!

PR #616, opened 2020-03-15
Fix IME pop-up position again

ryota-abe:
## Description
This is a continuation of [this PR](https://github.com/Zettlr/Zettlr/pull/544) and [this commit](https://github.com/Zettlr/Zettlr/commit/8e27fc760d99a72ac965d24a2856e4c1cbe82637).
Fix the position of the IME popup so that it does not overlap with the typed text.

## Changes
In version 1.6.0, it looks like:
image

with this PR:
image

## Tested On
 - OS and version: Windows 10 1903

## Additional information
The key point is `transform: scaleY(1.1);`. This is a workaround to render the vertical position correctly. Without this, `bottom:` would be ignored 🤷‍♂

### DIFF
--- a/resources/less/geometry/codemirror.less
+++ b/resources/less/geometry/codemirror.less
@@ -15,9 +15,8 @@
       font-size: inherit;
       line-height: inherit;
       height: 1.4375em !important;
-      top: 0 !important;
-      bottom: 0 !important;
-      margin: auto;
+      bottom: -1.4375em !important;
+      transform: scaleY(1.1);
     }
   }
 


nathanlesage:
> The key point is `transform: scaleY(1.1);`.

And this then also works with none of the previously described artefacts and misbehaviours? Especially the invisible margin at the top of the editor?

ryota-abe:
There is no longer invisble margin at the top of the editor (see screenshot above). The appearance and behaviour this time looks good to me 👀

nathanlesage:
Alright, thanks a lot! ☺️

PR #625, opened 2020-03-23
Add configuration file for VS Code

tobiasdiez:
## Description
Add run configuration file for VS Code.

## Changes
- Add run configuration for VS Code so that one simply needs to press F5 to lunch and debug Zettlr.  This is taken from https://www.electronjs.org/docs/tutorial/debugging-main-process-vscode.
- Add gitignore entries for VS Code according to http://www.gitignore.io/api/visualstudiocode

(One could argue that IDE-specific files shouldn't be checked in. I believe that the general consensus is that run configurations and linter / formatting rules are fine, while user-specific settings should be ignored. This is also reflected in the above gitignore.) 

## Tested On
 - OS and version: Win 10
 - Zettlr version: develop


.gitignore
```diff
@@ -13,3 +13,16 @@ yarn.log
 
 # Don't commit ESLint error reports
 eslint_report.htm
+
+# Created by https://www.gitignore.io/api/visualstudiocode
```
nathanlesage: I would change that simply to "Taken from" because semantically it makes more sense without removing the attribution to gitignore.io!

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+      "cwd": "${workspaceFolder}",
+      "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron",
+      "windows": {
+        "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron.cmd"
+      },
+      "args" : ["."],
+      "outputCapture": "std"
+    }
+  ]
+}
```
nathanlesage: Ending linebreak? :)

nathanlesage:
Cool! Thanks for implementing! Just had a quick rundown and it surely helps developing a lot!

Please feel free to add this change to the "Under the hood"-section of the changelog including a mention to your GitHub name ("thanks @tobiasdiez for implementing" for example)! Then I'll merge!

tobiasdiez:
Included your feedback.

nathanlesage:
Perfect! Thanks a lot!

Issue #578, opened 2020-03-02
Indentation for sub-lists is way too high


Zverik:
## Description

Consider this list:

```
1. item
2. item
    1. sub-item
3. last item
```

When a sub-item is spanning multiple lines, its indentation is very big, almost half a page. Sometimes the editor fails to see sub-item indexing, and bullets or numbers are not highlighted. See screenshot:

![zettlr-sub-indentation](https://user-images.githubusercontent.com/766031/75715661-d4265b00-5cde-11ea-90ab-77fa04c8e679.png)

(I have 4 spaces before "1." and "2.")

## Reproducing

For a simple case, copy the list above to Zettlr and try adding more words to the sub-item.

## Expected behaviour

Indentation should be smaller — if not ideal (see #511 for the general list indentation issue).

## Platform
 - OS and version: Fedora Linux 30
 - Zettlr Version: 1.6.0 beta5

## Additional information

I understand this is hard to fix, and it's nothing too important. It's okay if this waits until 1.8.

nathanlesage:
> For a simple case, copy the list above to Zettlr and try adding more words to the sub-item.

Have a look at this:

https://github.com/Zettlr/Zettlr/blob/f89e537fcd73335643dcfc5ccb3708fd871ef900/source/renderer/zettlr-editor.js#L446-L458

The "MINOR BUG"-part is what's causing this. Have to go to this again at some point.

Zverik:
Still being worked on in #628.

tobiasdiez:
This should be fixed in the latest development version, or at least not as bad as on the screenshot.

nathanlesage:
Yes, tested it out earlier, seems to work much, much better, even with indentation up to three levels:

![image](https://user-images.githubusercontent.com/17251683/84077806-2ced4a00-a9d8-11ea-82d7-312cc3542d5b.png)

(Ignore the additional indentation caused by the numbers, this is not how one should use lists, but I was too lazy :D)

Issue #511, opened 2020-02-07
List indentation is broken


Zverik:
## Description

When writing lists, both numbered and bulleted, second and further lines for each item are indented incorrectly.

## Reproducing

Write a list with items spanning two or more lines.

## Expected behaviour

![Screenshot 2020-02-07 at 11 47 48](https://user-images.githubusercontent.com/766031/74014639-273b1580-49a0-11ea-94fb-957d0b2607a7.png)

I expected for second lines to be aligned with first, that is, their first letters forming a nice row. Example:

1. Today is the second birthday of Zettlr. Exactly two years, or 730 days, ago, on December 26th, 2017, version 0.5 of Zettlr was publicly released on GitHub.
2. Last year we bumped the version to 1.0 as a birthday present, but as I have reiterated multiple times already, the “real” version 1.0 was reached when we released Zettlr 1.3, so this year Zettlr will not receive a major version update.
1. Back then, I promised to be releasing the results of the data that has amassed over the past years to show you how an app can proliferate globally.

## Platform
 - OS and version: macOS 10.14.5, same for Fedora 31
 - Zettlr Version: 1.5.0
 - Screen Resolution: 2560×1440

nathanlesage:
This has to do with the font. It does work with monospaced fonts (that is, currently, the Bielefeld theme). It has to do with the character widths. It is both known and (probably?) more of a feature than a bug. I'm not sure I can fix this, but well, I'll keep this open, maybe something works out.

Zverik:
Thing is, for all four themes the indentation is wrong, even for the monospaced one:

![Screenshot 2020-02-07 at 12 38 58](https://user-images.githubusercontent.com/766031/74018400-ea265180-49a6-11ea-8506-e30650f6c948.png)

![Screenshot 2020-02-07 at 12 38 48](https://user-images.githubusercontent.com/766031/74018403-eabee800-49a6-11ea-92b2-bccf94929b93.png)

![Screenshot 2020-02-07 at 12 38 24](https://user-images.githubusercontent.com/766031/74018404-eabee800-49a6-11ea-93a4-b0dd31e44214.png)

nathanlesage:
> I understand you're using not built-in lists, but are approximating the formatting using spaces or calculated indents, but let's keep this one as a possible enhancement

In fact, it's one of these instances where I stackoverflowed back in the day:

[...]

Issue #470, opened 2020-01-10
Nested lists: tabulations cannot be used


Beurt:
Hi,

Thank you for Zettlr, it seems a very interesting tool!

I found a strange behaviour with nested list in my own Mardown files (I have already thousands!): when the list levels are marked with tabs they are not recognized by Zettlr. For instance (`\t` is for a tabulation):

```
- level 1
\t- level 2
- level 1
```
Renders as:

> - level 1
> - level 2
> - level 1

But it should render as:

> - level 1
>     - level 2
> - level 1

When tabs are converted into (4) spaces, the rendering is correct.

nathanlesage:
Sorry for only answering now, I have forgotten to do so earlier!

To your question: Tab doesn't insert a tab character in Zettlr, but this is by design. Why?

- Tabs can be replaced with four spaces (most commands do that)
- Tabs are sometimes hard to distinguish from multiple spaces, and depending on implementation, you may accidentally end up exporting tabs instead of spaces, so this can be seen as a failsafe.
- Concerning list indentation: If the cursor is between the list sign and the text, you can use `Tab` and `Shift-Tab` to (un)indent the list automatically.

Beurt:
I'm sorry too for the late answer.

The problem was not with new files created with Zettlr, but that if we use in Zettlr some old Markdown files including nested lists using tabulations instead of (4) spaces, their render will be messed up.

For instance, I already have hundreds of such files (my previous Markdown notes) and it is a no go for Zettlr since all my nested lists (that I use a lot) are broken (no more nesting).

nathanlesage:
Ah! Mh … well, after all, maybe it should display either way. Let me have a look

felipehw:
@nathanlesage 
I have a lot of files with markdown indentation marked consistently by TABs. I tried to write with Zettlr but he opts for spaces instead of TABs for indentation.
For what I've read here .... this is by design.
So Zettlr never will support TAB indentation?

nathanlesage:
Well, you can use tabs in Markdown files, but spaces are recommended in cross-device-situations, so that is what's the default, yes. But Zettlr does support tab indentation

felipehw:
But is there a way to configure the nested items in a list, for example, to be marked using TAB instead x whitespaces?
Because I perceive that my TABs got mixed with the whitespaces of the "automatic" indentation ...

nathanlesage:
Umm, might be config option …?

felipehw:
[...]

PR #628, opened 2020-03-23
Fix indent of lists

tobiasdiez:
## Description
Fix the indention of lists (hanging indent).

## Changes
The origin of the problem was the assumption that the characters have width given by `cm.defaultCharWidth()`. Instead, now we really measure the width of the characters and use this value for the indention - this works as expected!

image

## Tested On
 - OS and version: Win 10
 - Zettlr version: develop (isn't this always the case?)

## Additional information
Do you have at test file where I can add the above example (to make testing easier in the future)?


source/renderer/zettlr-editor.js
```diff
@@ -503,6 +521,38 @@ class ZettlrEditor {
...
+  * Computes and returns the width of a character in the monospace font in pixels.
+  * @see https://stackoverflow.com/a/118251/873661
+  */
+  computeSpaceWidth () {
+    if (this._spaceWidth > 0) return this._spaceWidth
+
+    // We use 50 characters in order to get an approximately correct width (clientWidth is an integer)
+    var container = document.createElement('div')
+    container.id = 'measureWidth'
+    container.innerHTML = '<span>&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp&nbsp</span>'
```
Zverik: I suggest `'<span>' + '&nbsp'.repeat(50) + '</span>'`

Zverik: (and the same for dashes above)

source/renderer/zettlr-editor.js
```diff
@@ -503,6 +521,38 @@ class ZettlrEditor {
...
+    this._cm.getWrapperElement().appendChild(container)
```
Zverik: Don't we need to remove the container after the calculation? (here and above)

source/renderer/zettlr-editor.js
```diff
@@ -503,6 +521,38 @@ class ZettlrEditor {
...
+    this._monospaceWidth = (container.clientWidth + 1) / 50
+    return this._monospaceWidth
+  }
+
+  /**
+  * Computes and returns the width of a character in the monospace font in pixels.
+  * @see https://stackoverflow.com/a/118251/873661
+  */
+  computeSpaceWidth () {
```
Zverik: This and the above method look very similar. Would it be possible to extract common code to a function?

source/renderer/zettlr-editor.js
```diff
@@ -124,6 +124,9 @@ class ZettlrEditor {
     // Caches the "left" style property during distraction free
     this._leftBeforeDistractionFree = ''
 
+    this._spaceWidth = 0
```
Zverik: Since we're caching the values, what would happen when a user changes the theme with the font?

nathanlesage:
Alright, just had a look at the screenshots — if this really works this great out of the box this is one hell of a relief for me! Thank you so much!

It's rather late here now, so I'll get to it tomorrow (sorry for not getting back to everything today, privately there's a lot happening today, but I'll give it a thorough view tomorrow! Btw also thanks to @Zverik for the first code review, I'll look through all of it :)

source/renderer/zettlr-editor.js
tobiasdiez: They are indeed similar, but have enough difference that I couldn't extract a meaningful common ancestor. Do you have a concert idea of how to achieve this?

source/renderer/zettlr-editor.js
tobiasdiez: Good point. Didn't know that there were different themes. I now added code that listens to theme changes and resets the width. This seems to work somewhat.

For some reason, some of the value of the text-indent in the other themes are off (the padding seems to be ok). I tried to investigate but couldn't find the problem. Maybe this is something for another PR or do you have a suggestion?

tobiasdiez:
Thanks for the review @Zverik. I've now mostly included your suggestions and, moreover, also fixed the display for numbered lists. With this, the following issues should be fixed now https://github.com/Zettlr/Zettlr/issues/511 https://github.com/Zettlr/Zettlr/issues/470 https://github.com/Zettlr/Zettlr/issues/578 (as least for the default theme).

No worries @nathanlesage! There is no hurry.

source/renderer/zettlr-editor.js
```diff
@@ -126,6 +127,11 @@ class ZettlrEditor {
 
     this._spaceWidth = 0
     this._monospaceWidth = 0
+    ipc.on('toggle-theme', (e) => this.resetSpaceCalculation())
+    ipc.on('switch-theme-berlin', (e) => this.resetSpaceCalculation())
+    ipc.on('switch-theme-bielefeld', (e) => this.resetSpaceCalculation())
+    ipc.on('switch-theme-frankfurt', (e) => this.resetSpaceCalculation())
+    ipc.on('switch-theme-karl-marx-stadt', (e) => this.resetSpaceCalculation())
```
nathanlesage: Better approach to this: Only listen to the config-change event (I don't know exactly what it's called, but the `config-provider.js` emits an event as soon as the config has changed — then check if the value of the theme-config-value is different and handle this instead. Keeps the boilerplate a little bit in check! :)

source/renderer/zettlr-editor.js
```diff
@@ -447,30 +453,35 @@ class ZettlrEditor {
     this._cm.on('renderLine', (cm, line, elt) => {
       let monospaceWidth = this.computeMonospaceWidth()
       let spaceWidth = this.computeSpaceWidth()
+      let tabWidth = spaceWidth * 5
       let basePadding = 2.5 * monospaceWidth
 
       // Show continued list/quote lines aligned to start of text rather than first non-space char (hanging indent)
       // MINOR BUG: also does this inside literal blocks.
```
nathanlesage: I think we might drop that comment, as I think I fixed that with my last PR (I don't think that your changes re-introduced that, but we might double-check!)

source/renderer/zettlr-editor.js
```diff
@@ -531,9 +542,9 @@ class ZettlrEditor {
     // We use 50 characters in order to get an approximately correct width (clientWidth is an integer)
     var container = document.createElement('div')
     container.id = 'measureMonoWidth'
-    container.innerHTML = '<span>--------------------------------------------------</span>'
+    container.innerHTML = '<span>' + '-'.repeat(100) + '</span>'
     this._cm.getWrapperElement().appendChild(container)
-    this._monospaceWidth = (container.clientWidth + 1) / 50
+    this._monospaceWidth = (container.clientWidth + 1) / 100
```
nathanlesage: If I understand your approach correct, you're trying to guess the character width by seeing how far a container can go — what does this do on 4k monitors? Is this not a problem, or does it work regardless of screen size? If the former one, we should try to implement as many edge-cases as possible to not have to deal with this in the future again!

source/renderer/zettlr-editor.js
Zverik: Could we make another event that's fired on theme switch and on customcss update? Smth like `update-style`.

source/renderer/zettlr-editor.js
```diff
@@ -447,30 +453,35 @@ class ZettlrEditor {
     this._cm.on('renderLine', (cm, line, elt) => {
       let monospaceWidth = this.computeMonospaceWidth()
       let spaceWidth = this.computeSpaceWidth()
+      let tabWidth = spaceWidth * 5
```
Zverik: Are you absolutely sure it's 5? Quite an unusual number.

source/renderer/zettlr-editor.js
Zverik: Maybe by introducing a function `measureCharWidth(divId, symbol)` — although I'm not sure why you use `-` for measuring spaces for the monospace font, when spaces should have the same width. When using `&nbsp;` for all, you'd have only one argument to the function.

source/renderer/zettlr-editor.js
nathanlesage: Puhhh… until I manage to refactor all the boilerplate in the IPC classes, I'm hesitant in adding new events.

After all, I'm currently planning to remove these `message` IPC calls altogether and replace them with appropriate channels (e.g. `config` for config-events, `fsal` or similar for file system changes, etc …). This would make it more manageable, modular and decentral. But I have to refactor that damned file system beforehand :D

source/renderer/zettlr-editor.js
```diff
@@ -27,6 +27,7 @@ const { trans } = require('../common/lang/i18n.js')
 const generateKeymap = require('./assets/codemirror/generate-keymap.js')
 const EditorSearch = require('./util/editor-search')
 const openMarkdownLink = require('./util/open-markdown-link')
+const ipc = require('electron').ipcRenderer
```
nathanlesage, line 30: I just saw this when looking over this again — please don't do this as of now! Currently all IPC calls are centrally handled in the IPC wrapper classes. It's not good design, I'm that fare, but I'd like to refactor and decentralise the IPCs in one block, and not piece by piece. See also my other comment on the theme switching

source/renderer/zettlr-editor.js
```diff
@@ -406,7 +415,7 @@ class ZettlrEditor {
 
     // We need to update citations also on updates, as this is the moment when
     // new spans get added to the DOM which we might have to render.
-    this._cm.on('update', (cm) => {
+    this._cm.on('update', () => {
```
nathanlesage, line 418: … is this a general behaviour of your linter that it simply strips of all parameters if they aren't used? Because I'd like to revert that. The reason: They don't take up that much real estate and then you know what you can work with if you have to refactor any of the callback functions. It's much easier not to forget some thing this way.

nathanlesage:
Just had a curious look at the Abricotine repository, and look what I found:

https://github.com/brrd/abricotine/blob/develop/app/renderer/cm-init.js#L130

This is how Thomas is dealing with this and it looks much simpler, doesn't it …? Would this work here?

source/renderer/zettlr-editor.js
tobiasdiez: Ok, done!

source/renderer/zettlr-editor.js
tobiasdiez: Yes, this was done automatically. May I suggest you create format guidelines you like and reformat the code accordingly (using ESLint or whatever). In this way, other contributors can safely reformat according to these styles (If I hit reformat now in VS code, there are a lot of changes).

source/renderer/zettlr-editor.js
tobiasdiez: To be honest, I've no idea why it's 5. Was try and error

source/renderer/zettlr-editor.js
tobiasdiez: The character size is measured by creating a span with a lot of spaces in it, and then take the width of this span. This should work regardless of the system. The applied css prevents that this span is ever displayed, so there shouldn't be any side effects.

source/renderer/zettlr-editor.js
tobiasdiez: Good idea. I've now implemented it that way. 👍

tobiasdiez:
I updated the code according to your feedback. Sorry for taking that long to come back to this PR.
Sadly I couldn't try out the new version because of the problems described in #727.

The code in the Abricotine rep doesn't seem to be fundamentally different. They insert a character in the editor, measure it's width and then use this to calculate the proper indent/padding. This is exactly what the new code in this PR does as well. However, we have to deal with a few issues along the way that they don't face (spaces vs tabs, mixed font styles, ...). Any suggests what I should change in particular?

tobiasdiez:
So, I finally was able to test this. It works for me perfectly for the standard theme. For the other themes, it works for the first level but it's still a bit off for the second level. I would leave this to another PR.

So from my point this is now ready for review!

PR #634, opened 2020-03-26
Use gulp for the less > css conversion

tobiasdiez:
## Description
Use gulp for the less > css conversion instead of the custom js script.

## Changes
The usage of gulp has a few advantages:
- Using gulp's watch feature, after running `gulp less` all updates to a less file now automatically trigger a new conversion so that the css files in `source` are always up-to-date, which makes development way easier (just edit the less file, save, and refresh in Zettlr using F5)
- Easier integration with IDE (for example the `less` task can now be run in VS Code using `Ctrl` + `Shift` + `B`).
- Use sourcemaps so that the correct less file (where the relevant style is defined) is displayed in the Developer Console
image

- Shorter code (which in my opinion is easier to understand once you get a bit used to gulp)

## Tested On
 - OS and version:
 - Zettlr version:


nathanlesage:
Hello, thanks for this PR, and sorry for me taking so long to get back to you, maybe you've already seen the tweet that there's something up here …

While in general I like the idea, there are some questions:

First, I would like to retain the `yarn less`-command, but instead add a `yarn less:watch` (or similar) task to make it verbose what is happening. Removing the `yarn less`-command will break the build scripts, which may not happen (and simply changing it in the test workflow is not enough). Besides, I'd like the one-time-build and the watch-command — whose benefits I can clearly see — to be separate.

Furthermore, it would be really helpful to do short one-liner comments to each of the gulp pipes for other people (and me, who's so far not used gulp) to understand why the different pipes are necessary. Which brings me to the point: Are all new dev-dependencies necessary? Just out of curiosity! The sourcemap-dependency, however, is pretty much self-explanatory and you are right that sourcemaps are indeed a good idea. They will end up in the CSS-assets directory, next to the generated CSS files, correct?

Last but not least: There's an added ´.vscode/launch.json`, which has already been added in another commit, if I am not mistaken…? If so, then I can't merge this PR with the other one!

tobiasdiez:
Thanks for the review, and no worries - these are strange times and I know that it's not always easy to maintain an open source project. I hope you and your dears are fine.

I've updated the code according to your feedback:
- `yarn less` will no invoke `gulp less`
- split the less compilation `gulp less` from watching `gulp watch` (the latter can also be invoked by `yarn watch`). I went for a generic `watch` because I assume that the other build processes will be added there in the future. This would allow you to invoke `gulp watch` once and never worry about updated css files or handlebars etc.
- Added a bit of documentation about gulp and some of the pipes
- The dependencies are all coming from gulp. I was a bit surprised myself.
- There are minor additions to the vscode launch file (to debug also the render process which is more important actually then the main process). Github should be able to merge it (or I have to merge it myself).

nathanlesage:
Yeah … I wish I could say all of my dears were great, but this is for another time! Thanks for your consideration!

> split the less compilation gulp less from watching gulp watch (the latter can also be invoked by yarn watch). I went for a generic watch because I assume that the other build processes will be added there in the future. This would allow you to invoke gulp watch once and never worry about updated css files or handlebars etc.

This is a good thought, having one single command initiating all the watchers is better than having ten terminal sessions each with one watcher! So I like this idea. Thank you also for restoring the default behaviour, which will make migration to other contributors easier :)

> There are minor additions to the vscode launch file (to debug also the render process which is more important actually then the main process).

Could you explain how debugging the renderer works in this? (Because yes, you are right! The renderer process is some delicate piece of software where debugging is hard!) And could you port these changes to your other PR so that I can grab these .vscode-files before the FSAL-merger? Would make it easier for me to just keep the overview over the project :)

Concerning comments: What is `plumber()` supposed to mean? Sounds as if someone is cleaning a toilet 🧐

But after all, it looks good for merging now!

tobiasdiez:
After you merge, I'll rebase my other PRs so that the use the correct vsstudio files and you can then also use them for your developments. 

The debugging goes as follows. You set a breakpoint somewhere in the code (e.g. in the editor). Then go to the debug console in visual studio (the anti-bug icon on the right toolbar, 4th icon from the top). There you start the application (using the right profile depending on the type of code you want to debug). Visual studio does the heavy lifting, starting the app, attaching the debugger and stopping at the breakpoint, where you get access to the current state of variables, can run code etc. Works quite nicely I have to say.

Plumber is there just to improve error handling https://www.npmjs.com/package/gulp-plumber and is now documented.

Thanks for this nice project and the great communication. It's been a pleasure to contribute so far. 😃

nathanlesage:
Perfect, I just had a look at it, and there's only one issue — comments in the vscode json-file :D Linters don't like that so much, so I'd suggest we leave it out! Once you have done so, I'll finally merge this thing, and again apologies for being rather sloppy lately! I promise to do better :)

tobiasdiez:
Ok, removed the comments. So this should be ready to merge from my point of view.

nathanlesage:
Yes! Here we go :)

PR #642, opened 2020-04-03
Add support for custom latex commands and autocompletion

tobiasdiez:
## Description
This is still work in progress. The aim is to allow using custom latex command and provide an autocompletion for latex commands.

## Changes
- For the autocompletion, I added support to read a cwl file, which enables support of [LaTeX-cwl](https://github.com/LaTeXing/LaTeX-cwl) (that's the only reasonably big library of Latex commands I know). In the end, it's just a text file where each line contains the command definition (e.g. `\implies` or `\sideset{left}{right}{symbol}`). 
- KaTeX supports custom commands which need to be passed via the `macros` option. Right now this is a hard coded array. The better way would be to let the user specify a javascript file defining the commands.

I need a bit of help with the overall structure. In particular, there are three things were I need input:
- What's the best way to make the autocompletion cwl file as well as the macro javascript file configurable? Some hard-coded path relative to Zettlr's install location, some %AppData% path or a preference option?
- As of now, the `cwl` file is read and parsed in the editor. This is probably not a good thing to do. Where should be the parsing done?
- How do I load a custom-specified javascript file (the macros) and access the variables defined therein? I tried `require(absolute path to js)` path this lead to a complete freeze. 

## Tested On
 - OS and version: Win 10
 - Zettlr version: dev

### DIFF
--- a/source/renderer/assets/codemirror/zettlr-plugin-render-math.js
+++ b/source/renderer/assets/codemirror/zettlr-plugin-render-math.js
@@ -128,7 +128,396 @@
         // Enable on-click closing of rendered Math elements.
         elem.onclick = (e) => { textMarker.clear() }
 
-        require('katex').render(myMarker.eq, elem, { throwOnError: false })
+        // TODO: Read from file
+        // const fs = require('fs')
+        // var customMacrosFile = 'D:\\Research\\A_Central\\Latex\\tex\\generic\\myMathHeader.json'
+        // let customMacros = new Map(JSON.parse(fs.readFileSync(customMacrosFile, { encoding: 'utf8' })))
+        require('katex').render(myMarker.eq, elem, {
+          throwOnError: false,
+          macros: {
+            // # Math commands
+            /*
+            \begin{thm}
+            \end{thm}
+            \begin{prop}
+            \end{prop}
+            \begin{lemma}
+            \end{lemma}
+            \begin{coro}
+            \end{coro}
+            \begin{defn}
+            \end{defn}
+            \begin{fact}
+            \end{fact}
+            \begin{remark}
+            \end{remark}
+            \begin{remarks}
+            \end{remarks}
+            \begin{example}
+            \end{example}
+            \begin{counterexample}
+            \end{counterexample}
+            \begin{proof}
+            \end{proof}
+
+            \begin{thmenumerate}
+            \end{thmenumerate}
+            */
+
+            '\\abs': '',
+            '\\norm': '',
+            '\\nnorm': '',
+            '\\normDot': '',
+            '\\equivClass': '',
+            '\\bracket': '',
+            '\\scalarProd': '',
+            '\\scalarProdDot': '',
+            '\\scalarProdR': '',
+            '\\poisson{left}{right}': '',
+            '\\poissonDot': '',
+            '\\commutator': '',
+            '\\commutatorDot': '',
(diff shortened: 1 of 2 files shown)


nathanlesage:
Thanks for the initial work! Two things that come to mind right away:

1. Are you planning to integrate it using autocomplete? https://codemirror.net/doc/manual.html#addon_show-hint
2. I was planning something similar for the frontmatters (only with less vars, obviously), so in this case: Is your approach scalable to do this with multiple modes? (KaTeX/LaTeX and also YAML, maybe other languages later on?)

Cheerio, and thank you so much for the effort you put into this! :)

tobiasdiez:
Yes, I used the `showHint` method. The code for the auto completion is here
https://github.com/Zettlr/Zettlr/pull/642/files#diff-3000d5f5546b73135d3dce057785da12R303-R307.
It works but the logic right now is somewhat minimal. For example, I don't check if we are in a math environment (what's the easiest way to do this?). Depending on the structure of the commands in the frontmatter, the approach scales to this as well (maybe with a bit of refactoring).

So it's working for me right now, but I need your advice on the above points to have configurable version that also is usabale by others.

nathanlesage:
Alright, I just had another look at the code.

First, I think we should not implement code autocomplete using the editor hints, because out of two reasons: These autocompletes will only trigger in certain environments (KaTeX or YAML)\*, and second there's already autocomplete support built into CodeMirror. Why I did it this way is simply because the Markdown needs only these three databases (which were easier to built this way).

The CodeMirror show-hint plugin enables us to define the things to be autocompleted within the mode file itself, which would be much more convenient (hence, in the mode, call a function that adds autocomplete for certain environments). Or, we could even outsource them to their respective files. I currently don't remember in what way I hooked into the show-hint capabilities, maybe we can also port the current logic to the show-hint portion — in this case we could make the editor class a little bit leaner (I'm currently hoping to get that behemoth below 1.000 lines).

This might also enable us to define autocompletes for all supported code blocks, which would be somewhat mindblowing, especially given that afaik no other editor does this :D 

In case, you're very welcome to have a look at the current autocomplete logic and determine the best course of action — in any case, the more we save on code, the better.

***

\* If I'm not mistaken, one can do so using the codemirror hint object. I have yet to fully read the documentation on that.

tobiasdiez:
Macro file: add textfield In advanced preferences

As preparation for this PR, refactor IPC handler (similar to citeproc provider) and add broadcast (e.g. in order to send preference changes to all open windows).

nathanlesage:
@tobiasdiez Relevant issue: https://github.com/Zettlr/Zettlr/issues/859

There I also already outlined a preliminary API (I'm not sure that's _Der Weisheit letzter Schluss_, so we might need to alter it, but I'm not sure!)

tobiasdiez:
Call: Implement using asset manager.

tobiasdiez:
Sorry, I somehow managed to "merge" this PR by pushing my changes directly to the `develop` branch instead of my fork. Opened now https://github.com/Zettlr/Zettlr/pull/3099 as a follow-up.

nathanlesage:
You have removed the corresponding files again, right? I couldn't find them anywhere?

tobiasdiez:
Yes, I've reverted my accidental push.

nathanlesage:
Alright, thanks a bunch :)

PR #646, opened 2020-04-04
Decrease size of math

tobiasdiez:
## Description
Decrease math font size from 1.21em to 1.1em.

Before
image

After
image

## Tested On
 - OS and version: Win 10
 - Zettlr version: latest


nathanlesage:
Does this work with zooming in and out, and in all themes to a satisfying degree?

tobiasdiez:
Sorry for the late answer.

Yes, it looks good in all themes, and zoom works as expected as well. Math in the Bielefeld theme doesn't look that nice, but that's more an issue of the very different fonts instead of its size (but then again, what do you expect from a theme that is called Bielefeld ;-) )
image

nathanlesage:
> what do you expect from a theme that is called Bielefeld

😁

Aaaand merged!

PR #647, opened 2020-04-04
Update pull request template

tobiasdiez:
## Description
Not a functional change, but I since I always forget to add a changelog entry, I added it in the list of things to do before creating a PR.

## Further changes
- Removed ESlint since there is already a github action for this (and most editors should pick it up anyway)
- Removed Zettlr version since this should always be the current development branch. (It might make sense to make `develop` the default branch here on github so that new PRs target the correct branch by default) 

## Tested On
 - OS and version:
 - Zettlr version:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,8 +8,7 @@ following points:
   - [ ] I target the develop-branch, and *not* the master branch.
   - [ ] I have tested this extensively and paid extra attention to
         potential cross-platform issues (e.g. Cmd/Ctrl/Super-key bindings)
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hey, thanks! Makes total sense!

Regarding:

> Removed ESlint since there is already a github action for this (and most editors should pick it up anyway)

Unfortunately, I already had some guys who did not ESLint the code before opening the PR, so this can serve as a reminder for them — reduces the clutter and failing tests! So I'd like to keep that in. The rest makes absolute sense, and yes, the default branch is also a good thing. I'll do that!

tobiasdiez:
Ok, I readded the ES lint part.

nathanlesage:
Thanks a lot! :)

PR #648, opened 2020-04-04
Render equations with double-dollar in display mode

tobiasdiez:
## Description
Equations encapsulated with double dollars as in `$$ equation $$` render now in display mode, meaning they are centered in the line and have a bit of space around them.

image

## Tested On
 - OS and version: Win 10
 - Zettlr version: dev


nathanlesage:
What exactly does this alter at the element itself? Because CodeMirror cannot have block-style elements for rendering; is this still inline? Does KaTeX only add some margin within the element…?

tobiasdiez:
As far as I see, it adds the `katex-display` style (instead of `katex`). 
https://github.com/KaTeX/KaTeX/blob/a92e39d810c319c805d22026f4873a50409272e0/src/katex.less#L591-L611

According to the documentation, there is a bit more magic going on with how integrals and sums are displayed.
> If true the math will be rendered in display mode, which will put the math in display style (so \int and \sum are large, for example), and will center the math on the page on its own line. If false the math will be rendered in inline mode.

nathanlesage:
> Maybe because it's in a widget?

This might prevent it, the outer span of the widget should have an inline-setting, maybe this overrides it — but I've had some funky behaviour with block-display on such widgets, so we should make sure this doesn't interfere, because the CodeMirror-documentation is really point out that we should be careful with block widgets! (The tables, for instance, are also inline, and only appear so wide because I add a 100% width to them.)

tobiasdiez:
Is there something I should do concerning the block-display? I haven't noticed any problems in my testing so far.

nathanlesage:
> Is there something I should do concerning the block-display

Hmm … if you can set the block display to inline-block or inline without it compromising the display, I'd advise we do this just to be sure. If not, then … I'd say the best way is to simply edit a file with something like that displayed and have a look if something breaks with the display!

tobiasdiez:
I played around a bit, but couldn't find a satisfying solution using `inline` or `inline-block`. The space around the equation always turned out strange. On the other hand, I didn't experienced any problems using the current solution using `display`. 

However, I took the opportunity and improved the equation detection a bit. For example, there were problems with two directly following equations such as `$x$$a$`.  Moreover, the closing dollar signs of multiline equations were displayed if they had some white space in front of them.

nathanlesage:
> two directly following equations such as $x$$a$

The question is: Should this be that way? I'd argue that it's a — typographically — super ugly solution, and therefore would opt against allowing this. To a certain extent, I want to force people to write cleaner, and henceforth not allow too many exceptions from clean "code" (if you want it like that).

> Moreover, the closing dollar signs of multiline equations were displayed if they had some white space in front of them.

Alright, but this has nothing to do with the regex, but with how the characters are placed for where the element should end. So I'm curious if the crazy regular expression is really necessary, of if these things couldn't be achieved with the existing regex? Because the current expression looks daunting, and I fear that in two to three months if someone has to look up what it actually detects and what not it might become almost impossible … you get me here …?

Concerning the block: If it doesn't cause any problems, then I'd say let's throw it out in the beta 1.7 and see if people complain or if it works out of the box :D

source/renderer/assets/codemirror/zettlr-plugin-render-math.js
```diff
@@ -12,13 +12,12 @@
...
-  // Matches all inlines according to the Pandoc documentation
+  // Matches all equations contained in one line according to the Pandoc documentation
   // on its tex_math_dollars-extension.
   // More information: https://pandoc.org/MANUAL.html#math
-  // First alternative is only for single-character-equations
-  // such as $x$. All others are captured by the second alternative.
-  var inlineMathRE = /(?<!\\)\${1,2}([^\s\\])\${1,2}(?!\d)|(?<!\\)\${1,2}([^\s].*?[^\s\\])\${1,2}(?!\d)/g
-  var multilineMathRE = /^\s*\$\$\s*$/
+  // First alternative is only for single-character-equations such as $x$. All others are captured by the second alternative.
+  var inlineMathRE = /(?<!\\)(?<bracket>\${1,2})(?<eq>[^\s\\])\k<bracket>(?!\d)|(?<!\\)(?<bracketTwo>\${1,2})(?<eqTwo>[^\s].*?[^\s\\])\k<bracketTwo>(?!\d)/g
```
tobiasdiez, line 19: I didn't really change this regex (or make it more complicated per se), only required that the beginning and closing dollar signs have to be the same, i.e. `$x$` and `$$x$$` are fine but `$x$$` not. 

The side effect was that `$x$$y$` now works. I agree that it's ugly, but semantically correct. In my opinion it is more important to flag `$x$$` as wrong.

source/renderer/assets/codemirror/zettlr-plugin-render-math.js
```diff
@@ -56,33 +55,37 @@
...
         continue
-      } else if (isMultiline && multilineMathRE.test(line)) {
+      } else if (isMultiline && isMultilineStartOrEnd) {
         // We have left the multiline equation and can render it now.
         isMultiline = false
         newMarkers.push({
           'curFrom': { 'ch': 0, 'line': fromLine },
-          'curTo': { 'ch': 2, 'line': i },
-          'eq': eq
+          'curTo': { 'ch': multilineMathMatch[1].length, 'line': i },
```
tobiasdiez, line 73: That's the correction for the second point concerning the space in front of `$$`.

nathanlesage:
I just took some time to test the regex and immediately found a problem:

image

The second equation should not match due to the space in front of the matching bracket-group. P.S.: Shouldn't we rename "bracket" to "dollar" just to be clear that we don't mean `{{` but `$$`…?

tobiasdiez:
You are right, there were some problems with the regex. I've now improved it:
image
Moreover, I added tests so that future updates don't break this.
What do you think?

benrbray:
I was ready to submit my own PR for display math until I noticed this one waiting to be merged :)

My [solution](https://github.com/benrbray/Zettlr/commit/2d10f91b2258c87f182bf5a6b1699a87d55a90ad) makes the same change to `zettlr-plugin-render-math.js`, but I use the following in `editor.less`:

```css
.katex {
    font-size: 1.1em; // reduce font-size of math a bit
    display: inline-block; // needed for display math to behave properly
}
.katex-display, .katex-display>.katex>.katex-html {
  display: inline-block; // needed for display math to behave properly
}
```

This solution does not center the display math, but it avoids some weird issues I had with selection and cursor navigation using the default `display:block` applied by KaTeX.

nathanlesage:
> This solution does not center the display math, but it avoids some weird issues I had with selection and cursor navigation using the default display:block applied by KaTeX.

This is what I was afraid of!

BUT I think I have an idea, because the TableEditor actually makes use of a pretty bad, yet somehow working hack: It just takes an `inline-block` with the width of `100%`, so maybe this'll help with the display equations as well? I'm not sure how the usability of that would be, but then we'd have centered equations!

Anyone wants to check that?

cben:
FWIW, I've been using `inline-block` with width of 100% with MathJax: 
http://cben.github.io/CodeMirror-MathJax/demo.html and it worked pretty OK.

nathanlesage:
> Moreover, I added tests so that future updates don't break this.
What do you think?

Just saw this, and yes, tests are always good! Code coverage of the codebase is pretty bad, so anything helps!

Now I only need the time to test the PR out and resolve the conflicts … unless Tobias, you want to resolve the conflicts in the meantime?

gonike:
(Just chiming in to say that this is a real deal-breaker for me. I write a lot of math, and the non-centered, non-display mode equations just look horrible to my eyes. I'm really looking forward to this being implemented!)

christopherkang:
Is this still a WIP? Would love to see this added!

ThaddaeusWiedemer:
I would like to add that adding \displaystyle in a formula set with only single dollar signs can be used as a workaround, although it doesn't center the equation.

tobiasdiez:
Sorry that it took me so long, but it's now working and cleaned up. @nathanlesage please have a look!

@benrbray @cben Thanks for the pointers. Using `inline-block`  with `width=100%` indeed works well.

tobiasdiez:
@nathanlesage could you please have a look at this? Thanks!

(of course, will merge the latest develop branch later)

nathanlesage (commented):
So, I just had a look at your changes. First, it looks like it's doing what it is supposed to, and looks compatible in general with my envisioned final stage of Zettlr 2.0. However:

- Why did you create the class in the KaTex plug-in? I'm not on a crusade for more functional programming, but I really don't see the necessity for a class there.
- Then the PR seems to introduce quite some changes in the test suite. While it is certainly necessary, and I see where your going with this, I think that would be better handled in an additional PR, which could be an initiative to improve the code coverage more generally.
- lastly, I want to completely move all CSS rules into the Vue components (except the themes) and there introduce a more fine grained styling. This would mean to move these styles to the editor components (right now both in the main window and in the quickbooks, although I hope to make Quicklooks editable in due time and merge both files).

What do you think?

tobiasdiez:
Thanks for the feedback @nathanlesage. Concerning your comments:
- I agree, I was still in the Java mindset where everything needs to be in a class. I've now removed the `EquationFinder` class.
- I also agree that the changes to the tests are not directly related to the goal of this PR. However, without them the typescript tests are not executed, so the changes are necessary. If you want me to extract them to a new PR, that's no problem but has to wait at least two weeks as I'm now away from my main developer machine.
- The css is now in the editor.vue file, in accordance with your recent changes.

I couldn't try the latest version, because of https://github.com/Zettlr/Zettlr/issues/1869. But everything should work hopefully since my recent changes were only refactorings without touching the underlying functionality.

nathanlesage:
Ah, I just saw that you converted the plugin to typescript … I'm still a little bit uneasy with having TypeScript modules below a JavaScript module, although it works … hm. But I mean, it doesn't rush, since you apparently also uncovered a bug in CodeMirror; it'll still be part of 2.0!

tobiasdiez:
I didn't encountered any problems with using typescript here. In the end, ts is compiled to js anyway, so the "require" to a ts file is just importing the compiled js version.

What do you mean with "bug in codemirror"? The main changes for the implementation are to the regex and how they are used to analyze the test, so that's pretty zotero-specific.

nathanlesage:
> What do you mean with "bug in codemirror"?

The one you mentioned in one of your comments in the code!

tobiasdiez:
Sorry that it took me so long, but I finally managed to extract the test-related changes to https://github.com/Zettlr/Zettlr/pull/1974 (this branch this contains these changes, since otherwise the tests don't work).

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -0,0 +1,167 @@
+// This plugin renders MathJax parts in CodeMirror instances
+
+import { getBlockMathRE, getInlineMathRenderRE } from '../../../../common/regular-expressions'
+import * as CodeMirror from 'codemirror'
+import katex from 'katex'
+
+require('katex/dist/contrib/mhchem.js') // modify katex module
```
nathanlesage, line 7: Doesn't import work to modify it or does webpack/TS do gruesome things to it?

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -0,0 +1,167 @@
...
+      { 'ch': match.index + match[0].length, 'line': line },
+      match.groups.eq ?? '',
+      // Equations surrounded by two dollars should be displayed as centred equation
+      (match.groups.dollar ?? '').length === 2
+    ))
+  }
+  return newMarkers
+}
+
+export function findEquations (lines: LineInfo[]): EquationMarker[] {
```
nathanlesage, line 119: Do we need to export these functions …?

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -0,0 +1,167 @@
...
+  let eq: string[] = []
+  let fromLine = 0
+
+  // This array holds all markers to be inserted (either one in case of the
+  // final line of a multiline-equation or multiple in case of several
+  // inline equations).
+  let equations: EquationMarker[] = []
+
+  for (let line of lines) {
+    if (![ 'markdown', 'stex' ].includes(line.modeName)) continue
```
nathanlesage, line 130: While I personally did this myself for quite a long time, I have switched to always using curly brackets (just to make sure we don't overlook a continue, break, or something like this). Does make the code more verbose, but I have found it easier to read this way

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -0,0 +1,167 @@
...
+export class LineInfo {
+  constructor (
+    readonly lineNumber: number,
+    readonly text: string,
+    readonly modeName: string,
+    readonly tokenType?: string
+  ) {}
+}
+
+export class EquationMarker {
```
nathanlesage, line 81: Micro-classes again …? :D

test/parse-equations.spec.ts
```diff
@@ -0,0 +1,232 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        equation parser tester
+ * CVM-Role:        TESTING
+ * Maintainer:      Hendrik Erz
```
nathanlesage, line 7: Feel free to add your name here!

nathanlesage (approved):
Looks pretty good, I'll test this as well over the next week, but I don't think anything will prevent merging! (Except the remnants of the testing setup PR, which I've just merged)

source/common/modules/markdown-editor/plugins/render-math.ts
tobiasdiez: Yes, they are needed for the tests

source/common/modules/markdown-editor/plugins/render-math.ts
tobiasdiez: In my opinion, they really make sense here. What would you propose as an alternative?

tobiasdiez:
Merge latest develop branch and addressed feedback. Cannot test it right now because of https://github.com/Zettlr/Zettlr/issues/2033. However, the only non-trivial change was require > import, so that may be easy to revert if it doesn't work.

source/common/modules/markdown-editor/plugins/render-math.ts
tobiasdiez: In my opinion, the readability that comes from having a constructor really helps.

tobiasdiez:
Merged the most recent develop version, tests pass and everything is working for me in a few test documents. Thus, this is now finally ready for review from my side.

@nathanlesage ping 😉

nathanlesage:
Holy hell, your master branch was still on 1.6, no wonder node gave me loads of errors :D But works perfectly, so I'm going to merge it – although I would still argue you're too much into classes 😝

Issue #660, opened 2020-04-13
 Weird behavior in search field using Korean character (possibly CJK)


pizzakinggod:
## Description
When typing a Korean letter, letter is duplicated when there is a match for auto-completion.
See 

## Reproducing
 I type "코", then a matching note titled "코로나" appears. Then I try to use the found match, so presse enter or direction keys. Then a previous letter "코" is replicated and the search field is now filled with "코코" not "코로나"

## Expected behaviour
Auto-complete matching note names.

**Screenshots**
![gif](https://i.imgur.com/algxmE2.gif)

## Platform
 - OS and version: Windows 10 Home
 - Zettlr Version: 1.6.0
 - Screen Resolution: 1080p

## Additional information
Only tested with Korean characters, so not sure about other Asian inputs.

nathanlesage:
Might very well be related to these characters being double in width. Or maybe it has to do with the fact that non-latin scripts are written using an IME — @ryota-abe, do you have experienced something similar?

In either case, I'll just drop the corresponding code in here:

https://github.com/Zettlr/Zettlr/blob/b3d551d2b59625d6b332274369455a5ca867ba19/source/renderer/zettlr-toolbar.js#L71-L79

pizzakinggod:
So I spent time investigating and Japanese characters do reproduce this behavior, at least on my computer. And it does seem like it has to do with characters being fullwidth.

A quick hack would be to change line 76 to select from 1 length less if last character input is non-latin, but this might cause more problems in the future.

So I tried using [wcwidth](https://www.npmjs.com/package/wcwidth) to check if the last character matches the width defined by POSIX.1-2001. This worked okay but I am not sure it is okay performance-wise or will not cause other problems in the future development. Let me know what you think.

```javascript
if (e.setSelectionRange) { 
              let d = this._oldval.charAt(this._oldval.length-1);
              if (d.length === wcwidth(d))
                e.setSelectionRange(this._oldval.length, this._searchbar.val().length) 
              else
                e.setSelectionRange(this._oldval.length-1, this._searchbar.val().length) 
            } 
```

ryota-abe:
It is reproducible in my environment as well, but I hadn't noticed this until now because I was using only ASCII characters for file names.

pizzakinggod:
So I was able to cook up a solution that works well for a Korean character, but not with Japanese. This is due to the difference in how IME handles the characters.

[...]

PR #666, opened 2020-04-14
Add composition check CJK IME when using search bar

pizzakinggod:
## Description
Solves [#660](https://github.com/Zettlr/Zettlr/issues/660)
Check composing state(IME) during key event of search bar (zettlr-toolbar.js).
Append and select auto-complete only when a character is fully composed.

## Changes
Take event.isComposing from key event and use it to check for composing state.
Using an actual DOM element is necessary because jQuery keyevent does not provide composing state.

## Additional information
Let me know if it needs an additional changes.

Tested on: Windows 10 Home


nathanlesage:
Didn't know of that specific state check, but it does indeed look like a good way to check for stuff. And i wanted to migrate away from jQuery for some time now, so one reason more to do so.

Nevertheless, I'm more concerned that the `isComposing`-variable is being garbage-collected due to the `let`-statement which is only valid for a certain block …?

nathanlesage:
Btw concerning the jQuery/normal event thingy: Don't worry, I found [this statement here](https://api.jquery.com/category/events/event-object/):

> To access event properties not listed above, use the `event.originalEvent` object:

I'm as of now building this in, shouldn't take more than a minute!

PR #672, opened 2020-04-16
SPDX: clarify used license

marado:
"GPL-3.0" has been deprecated as a License identifier since the
release of version 3.0 of the SPDX specification[1].

This patch updates the package.json files with the now correct
"GPL-3.0-only", reflecting that this project is licensed with GPL
version 3, without an 'or later' clause.

[1] https://spdx.org/licenses/

## Additional information

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
         "url": "git://github.com/Zettlr/Zettlr.git"
     },
     "version": "1.6.0",
-    "license": "GPL-3.0",
+    "license": "GPL-3.0-only",
     "description": "A powerful Markdown editor for researchers.",
     "main": "./source/main.js",
     "scripts": {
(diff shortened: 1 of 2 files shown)


nathanlesage:
Hey, sorry for answering late, but especially concerning licenses, I tend to be wary and think about what I want to do. I am not sure, and I think I'll need to further think about whether I will be using the "only" or "or later" version of the GPLv3. Nevertheless, thank you for bringing this to my attention! I'll be getting to this once I had a chance to evaluate this!

marado:
Sure! Just FYI, I did not try to take a decision for you on this matter:
the current lack of an "or later" clause on the license currently in place
makes it (currently) be, de facto, an "3 only", even if you don't
explicitly declare it as such on the package.json file.

nathanlesage:
I know that you didn't want to take a decision, don't worry! And I see the reason for choosing "only" in this case! :)

PR #676, opened 2020-04-17
Fix #668 by using openItem instead of openExternal on files.

### LINKED ISSUE #668: Links to local notes do not work if the target file name contains an "é" (Windows)
## Description

If I create links to notes in Zettlr, following the link fails if the file name contains "é".

## Reproducing

1. Create two notes, one called "note" and one called "é".
2. Create a link in "note" to "é": `[link](é.md)`
3. Try to use `Ctrl+click` on the link

Zettlr displays "could not open file://C:/.../é.md", even though the path seems right.
 

## Expected behaviour

Zettlr should create a link to the file (in fact, creating the link with `[[é]]` works)

**Screenshots**

None.

## Platform
 - OS and version: Windows 10 (Doing this on Linux works)
 - Zettlr Version: 1.6.0
 - Screen Resolution: Irrelevant

## Additional information

Since `[[é]]` works, what path does this method use?

pizzakinggod:
## Description
Fixes #668 by changing path RegEx to work for unicode characters and using `openItem` instead of `openExternal` when opening files.

## Changes
Same as above

## Additional information
Let me know if it needs any other changes.

Tested on: Windows 10

### DIFF
--- a/source/common/util/make-valid-uri.js
+++ b/source/common/util/make-valid-uri.js
@@ -17,7 +17,7 @@ const path = require('path')
 
 const protocolRE = /^([a-z]{1,10}):\/\//i
 const linkRE = /^[a-z0-9]+\.[a-z0-9]+/i
-const mdFileRE = /[a-z0-9]+\.(?:md|markdown|txt)$/i
+const mdFileRE = /.+\.(?:md|markdown|txt|tex)$/iu
 
 module.exports = function (uri, base = '') {
   // Why do we need a helper function for this?
--- a/source/renderer/util/open-markdown-link.js
+++ b/source/renderer/util/open-markdown-link.js
@@ -24,11 +24,21 @@ module.exports = function (url, editorInstance) {
     // we need to convert it to absolute.
     let base = editorInstance._cm.getOption('markdownImageBasePath')
     let validURI = makeValidUri(url, base)
-    shell.openExternal(validURI).catch((err) => {
-      // Notify the user that we couldn't open the URL
-      if (err) {
-        global.notify(trans('system.error.open_url_error', validURI) + ': ' + err.message)
+    if (validURI.indexOf('file://') === 0) {
+      // Check if we are opening a file
+      if (!shell.openItem(validURI)) {
+        // If it failed to open, then show an error message
+        // Unlike openExternal, it returns a boolean value, so there is additional no error message
+        // If we wanted to append one, it will require us to write translation for all other locales...
+        global.notify(trans('system.error.open_url_error', validURI))
       }
-    })
+    } else {
+      shell.openExternal(validURI).catch((err) => {
+      // Notify the user that we couldn't open the URL
+        if (err) {
+          global.notify(trans('system.error.open_url_error', validURI) + ': ' + err.message)
+        }
+      })
+    }
   }
 }


nathanlesage:
Thanks for the PR, but some questions:

- Why did you switch to another opening function? I could not find any difference according to the docs, besides that `openItem` is less powerful. What is the benefit of using it?
- What do you mean with "If we wanted to append one" in one of your comments ...?

pizzakinggod:
> 
>     * Why did you switch to another opening function? I could not find any difference according to the docs, besides that `openItem` is less powerful. What is the benefit of using it?

Yes, you are right. I am not using it because it offers a benefit, but rather using `openExternal` does not open files with unicode. It is weird because in electron repository it says they fixed the [issue](https://github.com/electron/electron/pull/6436).
If `openExternal` worked as expected, we would only need to change the RegEx, but it didn't for me...
Maybe I will try more tests. 

> 
>     * What do you mean with "If we wanted to append one" in one of your comments ...?

Sorry if it wasn't clear, what I meant was, with `openExternal`, since it returns a promise, we can catch error and append error message that comes with it, but with openItem, since it returns a boolean value there is none. So If we wanted to append an error message like that it is cumbersome. I think I went to far assuming it, so you can just ignore that comment. I'll delete it.

nathanlesage:
Did you get any further with testing the openItem/openExternal unicode issues? I'll be dropping a new beta today, so maybe it'll work with the newer electron binary …? I'm not sure, but maybe test it out!

pizzakinggod:
Sorry for a delayed reply. I just tested it with the new beta and it still shows the error message.  
It seems Electron9 has replaced the shell function `openItem` with `openPath`. [Spec doc about openPath](https://github.com/electron/governance/blob/master/wg-api/spec-documents/shell-openitem.md). 

To quote your last reply from #668 

> 
> 
> > However, shell.openItem(validURI) successfully opens the markdown file(with unicode characters) with an external editor.
> 
> I'm extremely against monkeypatching something on our side simply because Electron doesn't do its job … which is to have openExternal to actually open these files, if I understand the discussion correctly, right?

So my observation at that time was that 
1. `make-valid-url.js` that is in use right now has a regex that only matches alpha-numeric characters.

2-1. `openExternal` does not handle non-ascii file names well.
2-2. `openItem` exists for properly opening a local file and works alright.

Anyhow, now that the return value of `openPath` changed, this code won't work...

nathanlesage:
> It seems Electron9 has replaced the shell function openItem with openPath

Yes, you are right, I had to migrate the attachment opening, that was why I wanted you to double-check that. The `make-valid-uri` should not affect the attachments, as it as far as I'm concerned only makes links work in the editor itself. The attachments are more or less simply displayed, their paths taken and that gotten to the operating system.

nathanlesage:
Hey, sorry that it took me so long to have a look, but the Electron 9 documentation does not include `openItem` anymore, it has become `openPath`: https://www.electronjs.org/docs/breaking-changes#api-changed-shellopenitem-is-now-shellopenpath

nathanlesage:
I just realized that you've been targeting the wrong branch (I double-checked all PRs just now because I overlooked that one issue targeted master and now I have three build pipelines vomiting code into the repository …). Please target `develop`!

PR #680, opened 2020-04-18
updated multiplex regex to correctly match the whole line

JorySchossau:
There was something nagging me about this - that the text after the language keyword wasn't being styled. In [discussion with Nathan](https://forum.zettlr.com/discussion/140/code-discussion-how-to-apply-styles-to-arbitrary-patterns#latest) I realized I was being an idiot and merely needed to update the regex again. So here's that fix.

Old Regex Example:
```
```\s*(python|sh)\b
```

New Regex Example:
```
```\s*(python|sh)\b.*$
```


nathanlesage:
Looks good!

Issue #200, opened 2019-05-22
Hope to support vim shortcut :)


Alopex4:
Totally, I think this application is great.   
However as a Linux user, I very miss the `vim` shortcut. Maybe someone has the same feeling too.  May I ask the developer would you plant to add `vim` support as a new feature ?   
Thanks again for your awesome job !

nathanlesage:
Support for a `vim` mode in CodeMirror seems to be not so good:

> Note: The CodeMirror vim bindings do not have an active maintainer. That means that if you report bugs in it, they are likely to go unanswered.

sndnyang:
I gave a simple try on the vim mode. And two small issues:

1. The 'Ctrl-f' is still 'find' instead of 'PageDown'.  Other shortcuts I have tried work fine.
2. The status/shape of the cursor keeps the same in normal(command) mode and insert mode.  In normal mode, the cursor usually shows a flickering block rather than the normal flickering line. It will help us to know the current mode we are using(but not very crucial).

It's an excellent software!

nathanlesage:
To point 1: It's because I override the keymap with some keys, refer to this file to know which ones!

To point 2: Yeah, I didn't implement the necessary CSS styles, as I don't officially support these (yet) — If you have an idea, please feel free to PR!

antoniy:
I would love to have option to enable vim mode in the settings. I've tested the suggested method and it actually makes the app usable to already broken Vim user like myself. There are just two small things I need personally to make this way better.

1. When in normal mode (vim) the caret should be a block style so you could tell you're in normal mode. When in insert mode - switch back to pipe style caret.
2. For a variety of vim motions the ability to see line numbers is actually useful to reformat whole blocks of text. So a support for line numbering, relative line numbering and hybrid line numbering will bring all that joy back.

Can we at least have the option in the settings as an experimental feature so we can actually use it without manually executing commands in the developer console?

PR #683, opened 2020-04-19
Feature editor mode preferences (Normal,Vim,Emacs)

JorySchossau:
This simply adds a UI element to the preferences (Editor) as a drop down list to change the codemirror editor mode, as pointed out by @nathanlesage in #200 . It works, but I need input on the things below.

Feedback especially needed on:
* **where to put it** in the window, since maybe putting it with the other boolean options is confusing. Should it be after the readability algorithm setting?
* **why label is not getting 'translated'**, I can't tell why this isn't working. I've provided translations for the 4 default language files (DE,GB,US,FR). I must be doing something stupid.

Will this new field get automatically picked up by translate.zettlr.com, or how do I have to add it?

To Do on a future PR?
* change visual cursor to block for vim in insert-mode. (I'm not too familiar with emacs, so don't know if it needs a cursor mode too).

Tested on Window 10:
I performed the following
* starting, setting mode=vim, saving, verifying mode, setting mode=normal, saving, verifying
* starting, setting mode=emacs, saving, verifying mode, setting mode=normal, saving, verifying
* starting, setting mode=vim, saving, closing, starting, verifying
* starting, setting mode=emacs, saving, closing, starting, verifying
* starting, setting mode=vim, saving, closing, starting, verifying, setting mode=normal, saving, verifying

### DIFF
--- a/source/main/providers/config-provider.js
+++ b/source/main/providers/config-provider.js
@@ -185,6 +185,7 @@ class ConfigProvider extends EventEmitter {
         'enableTableHelper': true, // Enable the table helper plugin
         'indentUnit': 4, // The number of spaces to be added
         'countChars': false, // Set to true to enable counting characters instead of words
+        'inputMode': 'default', // Can be default, vim, emacs
         'boldFormatting': '**', // Can be ** or __
         'italicFormatting': '_', // Can be * or _
         'readabilityAlgorithm': 'dale-chall', // The algorithm to use with readability mode.
--- a/resources/templates/dialog/preferences.handlebars
+++ b/resources/templates/dialog/preferences.handlebars
@@ -286,6 +286,16 @@
             </label>
             <label for="pref-editor-direction-rtl">{{i18n "dialog.preferences.editor_setting.direction_rtl"}}</label>
           </div>
+          
+          <hr>
+          <div class="cb-group">
+            {{i18n "dialog.preferences.input_mode"}}
+            <select name="editor.inputMode">
+              <option value="default" {{#ifCond editor.inputMode '=' 'default'}}selected="selected"{{/ifCond}}>Normal</option>
+              <option value="vim" {{#ifCond editor.inputMode '=' 'vim'}}selected="selected"{{/ifCond}}>Vim</option>
+              <option value="emacs" {{#ifCond editor.inputMode '=' 'emacs'}}selected="selected"{{/ifCond}}>Emacs</option>
+            </select>
+          </div>
         </div>
         <div class="clear"></div>
       </div>
--- a/source/renderer/assets/codemirror/autoload.js
+++ b/source/renderer/assets/codemirror/autoload.js
@@ -16,6 +16,8 @@ require('codemirror/addon/edit/closebrackets')
 require('./continuelist.js')
 require('./indentlist.js')
 require('codemirror/keymap/sublime') // This will load the extra commands from SublimeText
+require('codemirror/keymap/vim') // This will load the extra commands from Vim
+require('codemirror/keymap/emacs') // This will load the extra commands from Emacs
 
 // 3. Display addons
 require('codemirror/addon/display/fullscreen')
--- a/source/renderer/zettlr-editor.js
+++ b/source/renderer/zettlr-editor.js
@@ -773,6 +773,9 @@ class ZettlrEditor {
     this._cm.setOption('direction', global.config.get('editor.direction'))
     this._cm.setOption('rtlMoveVisually', global.config.get('editor.rtlMoveVisually'))
 
+    // Set input mode (vim, emacs, default (sublime))
+    this._cm.setOption('keyMap', global.config.get('editor.inputMode'))
+
     // Last but not least set the Zettelkasten options
     this._cm.setOption('zkn', global.config.get('zkn'))
 


nathanlesage:
After finalising the FSAL support today, I don't have much energy to dive deep into this PR, but I'll certainly review it next week! Thanks a lot :)

To your immediate questions:

> Will this new field get automatically picked up by translate.zettlr.com, or how do I have to add it?

Localisation-strings have to be added by me, so I'll see to it after the PR is merged!

> change visual cursor to block for vim in insert-mode.

This should be done by CodeMirror automatically, I already tested this and it works, there's even styling for that block style cursor in the LESS files, because I added it a while ago.

Concerning the code: Apparently you had an older version of the language files — I need to see to removing them from git soon. These should not be edited manually, but yes, it's my bad, because I should've done so far earlier :D Concerning the code, I'll be doing a review soon enough!

JorySchossau:
> I don't have much energy to dive deep into this PR, but I'll certainly review it next week! Thanks a lot :)

No worries! Take care. Thanks for the quick response.

nathanlesage:
Sooooooo, just had a look at your PR! The basic idea looks good, however, a few things:

1. The language files should not be changed. If you need translations (e.g. for the preferences settings), please choose reasonably namespaced localisation IDs ("reasonably" means simply copy the style from adjacent string translations :D) and tell me all that you've added so that I can add them to Zettlr Translate so they can be translated!
2. The swtch-construct seems a little bit verbose in setting the option, so maybe we could scale it down a little bit :)
3. At least concerning sublime mode I know that CodeMirror needs the corresponding keymap file (which is more like a plugin) to be included. What I already included is the `keymap/sublime` in the `renderer/assets/codemirror/autoload.js) and I think we need to add vim and emacs plugins similarly (but the CodeMirror manual should be more specific on that!)
4. Then, the final thing would be that we need to double-check if these things conflict with anything from the `generate-keymap.js` (also in the `assets/codemirror`-folder) or maybe other shortcuts … but I assume you already tested this thoroughly…?

What's your opinion on this?

Cheerio, and sorry for taking so long!

JorySchossau:
1. Lang Files
This makes sense, now that I know you are the black box behind language strings! I added "editorMode" and "dialog.preferences.editor_mode" but I'm still having the issue of strings not showing up correctly for the label. Maybe you can help with this? I'd love to know what the problem is; I don't have a lot of experience with vue and this templating framework. (`resources/templates/dialog/preferences.handlebars`)
2. Switch Stmt Verbosity
Haha that's an understatement. This was historical contingency from debugging :p I made it 1 line now.
3. Keymap Loading
Ah, yep, thank you. I moved the `require` commands for vim and emacs to sit next to the one for sublime.
4. Keybind Conflicts
Unfortunately, I can only speak about vim. I would expect that the `generate-keymap.js` bindings would all still work in insert mode (of vim mode), and indeed this is the case. The only conflict is in insert mode using `ctrl+o`, which appropriately is overridden by Zettlr with the Open File dialog. `ctrl+o` is a rare and easy to work around command in vim insert mode. So, all looks good here.

nathanlesage:
> you are the black box behind language strings

😈

But in fact this reminds me that we should find a better way to organize translations; Zettlr Translate basically was my shot at getting something GDPR-conforming translation service where no private information was needed while still keeping the bots out … but that's a whole different story that we might want to tackle with the users…!

I'll see to adding the localisation string soon! But it does show up "raw", right? Like, the identification string for the translation does show up the way you write it in the source code, right?

(Basically the algorithm goes — or SHOULD go — like this: 1.) Try to display the localisation in the user's language, 2.) if that doesn't work display the english default, and 3.) if that fails too, simply display the localisation string)

> The only conflict is in insert mode using ctrl+o, which appropriately is overridden by Zettlr with the Open File dialog.

This is what I meant with conflicts :D Can the `Ctrl+O` be mapped to a different key? Plus, we could even think about removing Ctrl+O for opening a new directory, as this is seldomly used …

I'll be testing this thing out later and have a look at it; I recently read some tutorials on vim, so I may even be able to exit insert mode without crashing the app! 🎉

source/main/providers/config-provider.js
```diff
@@ -185,6 +185,7 @@ class ConfigProvider extends EventEmitter {
         'enableTableHelper': true, // Enable the table helper plugin
         'indentUnit': 4, // The number of spaces to be added
         'countChars': false, // Set to true to enable counting characters instead of words
+        'editorMode': 'default', // Can be default, vim, emacs
```
nathanlesage, line 188: Maybe rename that to `insertMode`? Because as it is currently, in the codebase it would look like `global.config.get('editor.editorMode')` which sounds repetitively; and at least my gut feeling tells me that `global.config.get('editor.inputMode')` (or insert, something like that) tells more about what the config var is for?

JorySchossau:
1. variable name
I changed it to `inputMode` and `input_mode`, because "insert mode" is a different concept, at least in the vim world, and "editor mode" was a bit redundant and ambiguous as you said.
2. commits missing
Oops, I rebased my commits away trying to clean up the history! I put them back.
3. `Ctrl+O` conflict
I really don't think this is an issue worth thinking about. All other editors have the same problem and approach it the same way: "tough luck -- get used to it." 🙄  Sublime Text with vim mode has exactly the same behavior, and so does Visual Studio Code with vim mode. That is, `Ctrl+O` in insert mode opens the Open File Dialog instead of doing the vim command. It's just not a problem.
4. localised description in the preferences panel
Yes, the "raw" string shows up, so that could use some TLC from you when you get a chance. Thanks!

nathanlesage:
Perfect to all four points!

Note to myself: I should switch from master to the actual feature branch for testing if I don't want to end up running an old 1.6 base 🤦🏼‍♂️

Works fine, let's merge that baby! 🚀

Issue #536, opened 2020-02-12
[ROADMAP] Zettlr 1.7


nathanlesage:
Yes, I know, Zettlr 1.6 is still in beta and not yet released, but due to the crazy amount of enhancements and feature requests I feel it's necessary to outline the path to the future, both for you to know what will be happening and for me to have mental ease to actually ignore quite a bunch of these issues due to the following reasons:

In the past months, the functionality and codebase of Zettlr was extremely enlarged, with more and more features adding to the app. While this is incredibly cool, and I actually switched to using many of these new features in my own workflow (such as YAML frontmatters, it's such a great thing!), it has also destabilised some core modules to a frightening degree. What I'm especially referring to is two broader issues that have become quite unstable over the last few releases: the IPC (InterProcessCommunication) between server and client processes, which is, bluntly put, an utter mess. This needs refactoring fast. And the second thing is the file system. Currently, it's a monkey-patch recursive read-in of the root directories and files using only two classes. It's both ineffective, straining on computer's resources, and has become prone to errors. For example, virtual directories are, as far as I know, pretty much unusable due to a lot of mistakes and logical fallacies I accidentally introduced.

Therefore, here's what the roadmap of Zettlr for the next two releases will be:

1. Finish beta testing of 1.6, closing only bugs, and then releasing it.
2. Afterwards, I will be completely ignoring feature and enhancement requests and dedicate all my spare time to fixing the above-mentioned issues. Which means: First the file system abstraction layer, and secondly, fixing the IPC.
3. After this is done, 1.7 will be released without any major visible overhauls, but with most bugs being solved. Then, it'll be fully safe to use the sidebar without any more issues.
4. Then, in 1.8 I'll begin again to incorporate some of these issues.

You can track the status on the 1.7 Project: https://github.com/Zettlr/Zettlr/projects/2

If you really want a feature, what you can do in the meantime is discuss it on the forum with other users. The benefit? Once I'm back at integrating more features, the feature request by you will be fleshed out to such a degree that I can basically just read along and hack in the lines of code required to make it happen! Division of labor, or something :D 

Any more questions? I'm happy to answer them!

danieltomasz:
Great plans! I just registered to the forum :)

Zverik:
Happy to see a planned future for the editor! All is clear now, it bothered me a bit that with the constant stream of issues you might have had little time to keep looking at the bigger picture.

One question: what would you do with pull requests until releasing 1.7?

[...]

Issue #392, opened 2019-11-11
Rewrite File System Abstraction Layer w/ Caching


nathanlesage:
Just to keep you informed, here's something I've been planning for some time but will tackle in the near future: A full rewrite of the file system abstraction layer.

Rewriting the FSAL is a huge task, so a few issues will be tackled before the work on this is begun. This issue does not have a due date.

## But why?!

Zettlr uses a lightweight file system abstraction to mimic a file tree in its sidebar. The reason for this is that you do not have to use your operating system's file browser for each and every task but can perform most actions from within the app so that the experience is more immersive. Additionally, the internal FSAL includes some gadgets that are unnecessary or even counterproductive from the perspective of the all-purpose file browser, but come in handy for some certain tasks regarding Zettlr (such as pinning notes, also a proposed issue which will be resolved with a new FSAL). Furthermore, the internal FSAL performs an important filtering function by leaving out and not touching all files that are neither Markdown nor any of the specified attachment types.

Nevertheless, the current FSAL is actually comprised of a plethora of functionality scattered across the main process, which makes handling files an extremely volatile task. Not in the way of actually destroying data, because Zettlr will always bail before anything happens to the data, but this means that Zettlr will sometimes choke in doing this.

Some misbehaviour has been spotted with regard to handling files:

- Moving around a huge number of files (such as a full zettelkasten directory) sometimes results in an equal amount of notifications that "files have been changed externally" because the watchdog gave up freeing the file watchers before the actual move happened, resulting in poor experience until all notifications were gone.
- During startup, an undesirable user experience is caused, as the user is not informed about the current status of booting -- when the files have been loaded, suddenly the sidebar updates and moves around to its final state of, e.g., showing the file list of the current selected directory. Until then, a lot of actions are simply impossible without any visual indication. In its current state, the FSAL cannot provide such status updates.
- On every start, each file is read in and parsed completely, lowering performance especially on older machines. Even on moderately new Windows desktop machines, the startup duration is undesirably long (my test system: Intel i5-4590, 8 GB of RAM and an SSD)

## And how?

The proposed solution is to implement the new FSAL as a `service provider`, as it needs to provide its functionality from the boot of the app to the shutdown. Furthermore, the FSAL will actually be composed of three modules: (a) the file tree itself, (b) the file watcher (currently residing in: `watchdog-provider.js`), and (c) a caching system.

[...]

PR #685, opened 2020-04-20
Fsal refactor

nathanlesage:
## Description

Good morning y'all — it has been done. Quite some work, it turned out. However, here's why I do this via a Pull Request:

* I'd like other people to look over it. It's still a little bit rough, but more eyes can spot more bugs more easily, so please do some code reviews where I missed some things!
* Furthermore, it's still not fully complete, I have to do some chores such as removing the excess `console.log`s after all remaining bugs have been addressed

## Changes

* As mentioned in #392 and #536 I rewrote the complete file system abstraction layer to make it more scalable for many of the feature requests here on the issue tracker and to make it easier maintainable.

So far I can say the following things have changed:

- Added a file cache reducing file loading times up to 95 percent (especially for huge files)
- Went from object-oriented classes towards functional programming --> reduces memory overhead and prevents endless prototype chains stuffing up the RAM.
- The FSAL is now a self-contained module that works a lot via emits, which means that a lot of the caring for the internal file state is now a thing of the past — **but here I'm sure we can even further reduce the amount of file updates and directory updates that emanate from the commands, so please look out for these**
- The new FSAL supports multiple open files, and tabs have been added to the editor. There are still some quirks in the renderer process that have to be addressed (such as correctly replacing an open file if it has changed remotely or removing the selection from a file if none is open)

**Known Bugs**

Of course, something is still afoul:

1. Sometimes, moving directories around does not work on the first attempt, sometimes these have to be selected prior
2. Some of the remote changes are still being processed even if they emanate from within the app (e.g. are not "remote" in the sense of the word), because the changes are slower than the current internal timeout used to ignore these when they emanate from the app. I'm currently curious whether or not we should simply use the remote change watcher instead of manually adapting the state? This might be a shortcut and even further reduces the source of truth to only the file system itself, with no further ado from the app
3. Dragging and rearranging the open files still has some small nuisances
4. Is everything really happening to a degree where we can guarantee a safe state truth within the FSAL?

## Additional information

No additional information that I know of, except: Comment as if there was no morning, everything you spot now can be addressed before we push this monstrosity onto the develop branch :)

Tested on: macOS 10.15.4


source/common/util/generate-filename.js
```diff
@@ -23,5 +25,9 @@ module.exports = function () {
   let pattern = global.config.get('newFileNamePattern')
   pattern = replaceStringVariables(pattern)
   pattern = pattern.replace(/%id/g, generateId(global.config.get('zkn.idGen')))
+  // In case a funny guy has removed the pattern from config.
+  if (pattern.length === '') pattern = uuid()
```
wgslr, line 29:
```suggestion
  if (pattern.length == 0) pattern = uuid()
```

source/common/util/is-attachment.js
```diff
@@ -17,14 +17,11 @@ const isFile = require('./is-file')
...
-module.exports = function (p) {
+module.exports = function (p, skipExistenceCheck = false) {
   let ext = global.config.get('attachmentExtensions')
-  if (!ext) {
-    // Something went wrong on init. Hey ZettlrConfig, are you even there?
-    return false
-  }
-
-  return isFile(p) && ext.includes(path.extname(p).toLowerCase())
+  let fileExists = skipExistenceCheck ? true : isFile(p)
```
vorticalbox, line 25: ext and fileExists could be const

source/common/util/generate-filename.js
nathanlesage: I must've been asleep while coding that … thanks!

source/common/util/is-attachment.js
nathanlesage: Hey, I saw your comment over on reddit, thanks for reviewing this! I've been playing with the thought to `const` such variables myself, but I want to do this on the whole repository in one piece. But you are right!

source/common/util/flatten-directory-tree.js
```diff
@@ -13,28 +13,26 @@
...
-    return newarr.push(tree)
+    return [tree] // well ...
   } else if (tree.type === 'directory') {
     // Append directory (for easier overview)
     newarr.push(tree)
-    if (tree.children != null) {
+    if (tree.hasOwnProperty('children') && tree.children != null) {
       for (let c of tree.children) {
-        newarr.concat(flattenDirectoryTree(c, newarr))
+        newarr.concat(flattenDirectoryTree(c))
```
ryota-abe, line 33: `Array.concat()` never changes `this` object. Did you mean `newarr = newarr.concat(flattenDirectoryTree(c))` in this line?

nathanlesage: I just scrapped the full file, because I realised that `source/common/util/object-to-array.js` does the same thing but is also highly optimized. But yes, this is the file I meant over on Twitter!

nathanlesage:
Soooo, merging so that the develop branch is again the single source of truth here!

PR #688, opened 2020-04-20
Filename in link

Wieke:
## Description

If a `[[link]]` format is used the autocomplete now places the filename/title within the link like `[[link|filename]]`,rather than placing it after the link like `[[link]] filename`.

If the user changed the link start and end to something other than `[[` and `]]` the filename will be added behind the link instead.

## Changes

I changed the on autocomplete callback so the behaviour described above is achieved. 

I also changed the click event listener of the editor so it removes the filename from the autosearch term.

## Additional information

The reason I propose this change is because it would result in somewhat less ambigious links, making it clearer what text correctly describes the content of the file being linked. (Eg. `[[202004202053]] apple pie` vs `[[202004202053|apple]] pie` or `[[202004202053|apple pie]]`.) And because it would make it easier for someone exporting the notes to html (or some other format) to transform the internal links to proper links with anchor text. (Eg. `[[202004202053|link to notes]]` to `<a href="202004202053.html">link to notes</a>`.)

Due to the changes to the click event listener if the link contains multiple `|` (for example if the filename or id contains a `|`) then it will strip away part of the link and do a search on a part of the link rather than directly going to the file. I figured this was an acceptable trade off since using `|` in either filenames or ids is rather uncommon. (I tried to implement the use of an escape character like `\` but that trips up editor a bit.)

Tested on: Archlinux

### DIFF
--- a/source/renderer/zettlr-editor.js
+++ b/source/renderer/zettlr-editor.js
@@ -165,8 +165,8 @@ class ZettlrEditor {
               let linkPref = global.config.get('zkn.linkWithFilename')
               // Prepare the text to insert, removing the ID if found in the filename
               let text = completion.displayText
-              if (completion.id && text.indexOf(completion.id) >= 0) {
-                text = text.replace(completion.id, '').trim()
+              if (text.startsWith(completion.text)) {
+                text = text.slice(completion.text.length).trim()
               }
               // In case the whole filename consists of the ID, well.
               // Then, have your ID duplicated.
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hey, I refactored the full autocompletion out of the editor class into its respective utility module, so now this PR conflicts! But also, why should Zettlr support this more traditional wikilink-syntax …?

nathanlesage:
I was just going through the PRs, and I will close this PR, as I don't want to enable this functionality.

The reasoning for this is here: https://github.com/Zettlr/Zettlr/issues/992#issuecomment-653802307

Issue #719, opened 2020-05-05
Search for a new icon set


Wieke:
As discussed in #538 I'm going to be looking for a GNU GPL v3 compatible replacement icon set for Zettlr. I'm going to start by listing all the functions we need icons for and seeing if I can find appropriate icons in some icons sets I know of. These are the sets I've considered:

* [Feather](https://feathericons.com/) (MIT license)
* [Bootstrap](https://icons.getbootstrap.com/) (MIT license)
* [Material Design](https://material.io/resources/icons/) (Apache 2.0 license)
* [Clarity Design](https://clarity.design/icons) (MIT license) (Thanks to @mhyfritz)

According to [this graph](https://en.wikipedia.org/wiki/License_compatibility#/media/File:Floss-license-slide-image.png) they should all be compatible with GPLv3. Let me know if I missed any good sets.

Here's my attempt at matching icons to functions (I listed multiple if there wasn't a clear best option in my opinion). 

|  Function               | Clarity | Material | Bootstrap | Feather |
|-------------------------|---------|----------|-----------|---------|
| Menu                    | ![](https://raw.githubusercontent.com/vmware/clarity-assets/master/icons/core/bars-line.svg) | ![](https://fonts.gstatic.com/s/i/materialiconsoutlined/menu/v5/24px.svg?download=true) | | ![](https://raw.githubusercontent.com/feathericons/feather/master/icons/menu.svg) |
| Open directory          | ![](https://raw.githubusercontent.com/vmware/clarity-assets/master/icons/core/folder-open-line.svg) | ![](https://fonts.gstatic.com/s/i/materialiconsoutlined/create_new_folder/v5/24px.svg?download=true) | ![](https://raw.githubusercontent.com/twbs/icons/master/icons/folder-plus.svg)   | ![](https://raw.githubusercontent.com/feathericons/feather/master/icons/folder-plus.svg) |
| Create file             | ![](https://raw.githubusercontent.com/vmware/clarity-assets/master/icons/essential/file-line.svg) | ![](https://fonts.gstatic.com/s/i/materialiconsoutlined/note_add/v5/24px.svg?download=true) | ![](https://raw.githubusercontent.com/twbs/icons/master/icons/file-plus.svg) ![](https://raw.githubusercontent.com/twbs/icons/master/icons/file-earmark-plus.svg)   | ![](https://raw.githubusercontent.com/feathericons/feather/master/icons/file-plus.svg) |
[...]

PR #751, opened 2020-05-08
Clarity icons

Wieke:
## Description
Adoption of the clarity icon set. Changed existing icons to icons from this set where possible. See issue #719.

## Changes

Replaced these icons:

* All buttons in the toolbar except pomodoro timer button.
* All buttons, except headers, footnote and remove footnote, in the formatting menu.
* Open directory button in attachments panel.
* Request file, reset field, remove autocorrect, add autocorrect buttons in pdf preferences, project preferences and preferences screens.
* Add button in tags screen.
* Project icon and open/collapse carret in file tree.
* Music note in pomodor timer panel.
* Close button of tabs.

Added a icons utility that is called by the renderer, it loads the clarity module and adds custom icons. Various places where html was being created in javascript where changed (for toolbar, attachment, autocorrect, tabs, filetree icons).

Oh I also added a little hover effect to the open directory button in the attachment pane, so it's a bit more consistent with the other buttons in the application.

## Additional information

I tested by building and running the app-image. I don't have the means of testing it on windows or mac platforms. I wasn't sure where to place the call to `loadCustomIcons` I ended up putting it at the end of the renderer constructor.

Tested on: Archlinux

### DIFF
--- a/source/renderer/util/icons.js
+++ b/source/renderer/util/icons.js
@@ -1,0 +1,45 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        Clarity icons helper
+ * CVM-Role:        Utility
+ * Maintainer:      Wieke Kanters
+ * License:         GNU GPL v3
+ *
+ * Description:     This module loads the clarity library and adds custom icons.
+ *
+ * END HEADER
+ */
+
+const clarityIcons = require('@clr/icons').ClarityIcons
+require('@clr/icons/shapes/all-shapes')
+const fs = require('fs')
+const path = require('path')
+
+function loadCustomIcons (dir) {
+  fs.readdir(dir, { withFileTypes: true }, (err, list) => {
+    if (err) throw err
+    list.forEach(x => {
+      if (x.isDirectory()) {
+        loadCustomIcons(path.join(dir, x.name))
+      } else if (x.isFile()) {
+        let ext = path.extname(x.name)
+        if (ext && ext.toLowerCase() === '.svg') {
+          fs.readFile(path.join(dir, x.name), (err, data) => {
+            if (err) throw err
+            let icon = {}
+            icon[path.basename(x.name, ext)] = data.toString()
+            clarityIcons.add(icon)
+          })
+        }
+      }
+    })
+  })
+}
+
+module.exports = {
+  loadCustomIcons: function () {
+    loadCustomIcons(path.join(__dirname, '../assets/icons'))
+  }
+}
--- a/source/renderer/zettlr-renderer.js
+++ b/source/renderer/zettlr-renderer.js
@@ -31,7 +31,7 @@ const loadI18nRenderer = require('../common/lang/load-i18n-renderer')
 
 const reconstruct = require('./util/reconstruct-tree')
 
-const icons = require('./util/icons')
+const loadicons = require('./util/load-icons')
 
 const path = require('path')
 
@@ -81,7 +81,6 @@ class ZettlrRenderer {
         this.beginSearch(term)
       }
     }
-    setTimeout(() => icons.loadCustomIcons(), 0)
   }
 
   /**
@@ -127,6 +126,9 @@ class ZettlrRenderer {
 
     // Here we can init actions and stuff to be done after the startup has finished
     setTimeout(() => { this.poll() }, POLL_TIME) // Poll every POLL_TIME seconds
+
+    // Load the clarity icon modules and add custom icons
+    setTimeout(() => loadicons(), 0)
   }
 
   /**
--- a/source/renderer/zettlr-renderer.js
+++ b/source/renderer/zettlr-renderer.js
@@ -31,6 +31,8 @@ const loadI18nRenderer = require('../common/lang/load-i18n-renderer')
 
 const reconstruct = require('./util/reconstruct-tree')
 
+const icons = require('./util/icons')
+
 const path = require('path')
 
 // Pull the poll-time from the data
@@ -79,6 +81,7 @@ class ZettlrRenderer {
         this.beginSearch(term)
       }
     }
+    setTimeout(() => icons.loadCustomIcons(), 0)
   }
 
   /**
--- a/source/renderer/zettlr-renderer.js
+++ b/source/renderer/zettlr-renderer.js
@@ -127,8 +127,9 @@ class ZettlrRenderer {
     // Here we can init actions and stuff to be done after the startup has finished
     setTimeout(() => { this.poll() }, POLL_TIME) // Poll every POLL_TIME seconds
 
-    // Load the clarity icon modules and add custom icons
-    setTimeout(() => loadicons(), 0)
+    // Load the clarity icon modules, add custom icons and then refresh
+    // attachments (because it requires custom icons to be loaded).
+    setTimeout(() => loadicons().then(() => this._attachments.refresh()), 0)
   }
 
   /**
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Node modules
 node_modules
 
+# Package Lock
+package-lock.json
+
 # Compiled releases
 release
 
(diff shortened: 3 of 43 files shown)


source/renderer/util/icons.js
```diff
@@ -0,0 +1,45 @@
...
+ * END HEADER
+ */
+
+const clarityIcons = require('@clr/icons').ClarityIcons
+require('@clr/icons/shapes/all-shapes')
+const fs = require('fs')
+const path = require('path')
+
+function loadCustomIcons (dir) {
+  fs.readdir(dir, { withFileTypes: true }, (err, list) => {
```
nathanlesage, line 21: I'd recommend using `fs.promises` here, as it makes the text more legible. And, as we don't have to handle the promises generated by asynchronous functions, this should prove to be no problem (we can catch potential errors and log them, to be sure)

source/renderer/util/icons.js
```diff
@@ -0,0 +1,45 @@
...
+            icon[path.basename(x.name, ext)] = data.toString()
+            clarityIcons.add(icon)
+          })
+        }
+      }
+    })
+  })
+}
+
+module.exports = {
```
nathanlesage, line 41: This could be simplified by writing

```javascript
module.exports = function () {
    return loadCustomIcons( /* ... */ ) // Return for the promise
}
```

source/renderer/zettlr-renderer.js
```diff
@@ -79,6 +81,7 @@ class ZettlrRenderer {
         this.beginSearch(term)
       }
     }
+    setTimeout(() => icons.loadCustomIcons(), 0)
```
nathanlesage, line 84: This can be moved to the init-function where the initial configuration is pulled from main, because I thiiiink that this is only called once … but better check twice …

nathanlesage:
I'm a fan! I just cloned it and tested it out locally. I'm really aroused by the looks of the new toolbar, and I could only find super minor things!

image

First, both the file selection icons and the reset-icons are super slightly off to the bottom. Unfortunately, I have no quick fix :/

Then, there are two different clippy icons in the sidebar:

image

I'm not sure how to solve this (maybe with real file icons based on the attachment's extensions? But Clarity apparently does not include such icons :/)

And last, what about the root file/root directory indicators?

image

Wieke:
Yeah I'm not surprised I missed a couple of icons. For example I also forgot the file sort icons:

image

Clarity only has a single sort icon though, so I'd probably have to customize something.



As for the file selection icons, there is some styling on the toolbar buttons that keeps them centred, that might work there as well. Otherwise I suspect fiddling with margins might get it centred.

As for the missed attachment icon, I see a couple of options:

* Reuse the paperclip icon. 
* Just have no icon.
* Use a generic file icon. 
* Use an image icon for images , zipped file icon for zipped files  and generic file icons for everything else.

I'd probably go for the generic file icon for the time being. Having icons of specific mimetypes might be useful, though I would probably place that in a new issue cause figuring that out isn't going to be trivial. I did find some nice icons in the [elementary](https://github.com/elementary/icons/tree/master/mimes/symbolic) and [gnome](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/tree/master/Adwaita/scalable/mimetypes) themes. (Most of the other filetype icon sets I saw were really colourful and garish, not a good fit for the zettlr style.)

The file/root icons should be easy since there are file and tree-view icons.

 

I'll probably add the missing icons and process the other feedback within a day or two.

nathanlesage:
Perfect, thank you! :)

> Having icons of specific mimetypes might be useful, though I would probably place that in a new issue cause figuring that out isn't going to be trivial.

Concerning this, hot take: How about using the generic file icon and simply add the file extension in there? I mean, it's SVG anyway so we could make use of this. This might actually be pretty helpful as it shows people immediately what file type it is, includes also unknown file extensions and helps if the filename is too long so it's truncated.

Wieke:
Oh that's a nice idea I'll give it a try.

Wieke:
While fixing the icon not quite being in the center of the button issue I ended up consolidating the `file-select-group` and `form-inline-buttons` classes into a `input-button-group` class, allowing you to do something like this:

```html
<label>Foo</label>
<div class="input-button-group">
  <input type="text">
  <button><clr-icon shape="floppy"></clr-icon></button>
</div>
```
image

Which I then used to ensure the centeredness of the icons in the buttons that follow text fields. The one drawback though is that this changes the look of the reset and auto correct fields a bit.

Before
image
After
image

Before
image
After
image

Now personally I think this consolidation is worth the small change in appearance (which might even be a small improvement, especially in the autocorrect case). But I'd like to get your assent before I commit these changes.

nathanlesage:
> I ended up consolidating the file-select-group and form-inline-buttons classes into a input-button-group class

You have my blessings for every simplification of these things! And if they make stuff much more ordered and aligned, you go you! Looks pretty cool tbh.

Wieke:
Alright so I made the requested, and some extra, changes.

Loading of custom icons now uses promises.

As described above the `input-button-group` class was added and centeredness of the reset, file request and remove autocorrect buttons was fixed. The centeredness of the add tag/autocorrect buttons was fixed as well. (Presumably, the difference is minimal on my screen.)

The files and directories icons were also replaced. Though I couldn't figure out how to get the root file thing to show up, so no guarantees there (I gave it the same styling as the directories icon, so it shouldn't be too bad).

image (It's a bit small and hard to see on my screen.)

I also reworked the sorting component a bit. The sorting logic was deduplicated by moving it from the file tree/list to the sorting component itself. And the sorting component now emits a `sort-change` event to which the parent listens rather than using `this.$parent` (as far as i know this is a bit more vue-esque).

I also changed the outward behaviour because I thought the current behaviour was a bit vague and I had a hard time figuring out what to replace the icons with. (Honestly I thought the sort-by-time icon meant sort-by-file-size, what with it showing different sized bars.) There are still two toggle buttons/icons but now they ether toggle what is being sorted by (time or name) or what direction the sort goes (ascending or descending). This has the added benefit of the icons showing what way the directory is currently being sorted (something I don't think you could tell before) and it would be slightly easier to add additional sort-by options.

image (alphabetic ascending)
image (time descending)

The attachment list icon was also replaced. I used your idea of adding the extension to the generic icion. It seems to work pretty well.

image

(Only the first 3 letters of the extension are being shown.)

And finally I also replaced the times icons used in the tag cloud dialog and the user dictionary entries list.

Still haven't got them all. The regexes `font-family:.*(font-icon|WebHostingHub-Glyphs)` still has some hits but I'd have to see if I can track them down.

nathanlesage:
> There are still two toggle buttons/icons but now they ether toggle what is being sorted by (time or name) or what direction the sort goes (ascending or descending).

This is extremely good. We only need to tell people "Attention, it looks and works slightly different now", albeit I think your solution is even much more straight forward (but you never know :D)

> This has the added benefit of the icons showing what way the directory is currently being sorted (something I don't think you could tell before) and it would be slightly easier to add additional sort-by options.

Yes, agreed!

> (Only the first 3 letters of the extension are being shown.)

This is better than nothing, I think. We MAY want to think about a list of four-or-more-letter-endings that we could shorten down, but that's nothing for right-now!

> The regexes font-family:.*(font-icon|WebHostingHub-Glyphs) still has some hits but I'd have to see if I can track them down.

Yeah, the styling is an utter mess at the moment, we'd have to sort and clean it up at some point …

***

That being set: What about we just merge this today, and have it already in upstream, because then we can already integrate this in the beta phase (might uncover additional issues as well), and you have time to just go and check if you really found all icons!

Wieke:
> That being set: What about we just merge this today, and have it already in upstream, because then we can already integrate this in the beta phase (might uncover additional issues as well), and you have time to just go and check if you really found all icons!

Sounds good to me.

nathanlesage:
Almost forgot to merge this, but here we go!

Just one small annotation: Feel free to make use of `async`/`await` in the load utility. This'll bring that function more in sync with the rest of the codebase and also reduce that `))))}`-monstrosity at the end of the function :D

Plus we might want to write a test for that function (in case we decide to change something so that we will get a notification if we accidentally break something), that it loads the directory correctly. … or is it failsafe?

PR #758, opened 2020-05-09
Fix sidebar and include vue dev tools

tobiasdiez:
## Description
For me the sidebar was not loading properly. In the end, it was a problem in the `moment` library ("the function moment() is not defined"), which was fixed by downgrading it to 2.24. This seems to be the issue: https://github.com/moment/moment/issues/5494

## Changes
Along the way, I had problems debugging this. So I read a bit about vue, electron and developer tools. Resulting changes:
- Activate eslint for vue files and fixed the errors (there are a couple of warnings left)
- Install vue devtools which allow inspection of vue components in the developer console of the running application.
- Enable sourcemaps for vue files to help debugging
- Don't minimize webpack emits for development
- Add webpack watch to `yarn watch`.

## Additional information

Tested on: Win 10


nathanlesage:
> For me the sidebar was not loading properly. In the end, it was a problem in the moment library ("the function moment() is not defined"), which was fixed by downgrading it to 2.24. This seems to be the issue: moment/moment#5494

Actually, this is something I also had with Wieke's PR. I assume Yarn is not as accurate as I would like when it comes to precisely installing certain versions (as apparently it has never updated moment on my system at all, which I'm pretty pissed about right now). I'll see to it tomorrow, but thanks a lot for investigating ahead of time :) I was already wondering what was causing this!

tobiasdiez:
Not sure why you didn't experience this problem. `moment@^2.25.1` was clearly in the yarn.lock file.

nathanlesage:
> Not sure why you didn't experience this problem. `moment@^2.25.1` was clearly in the yarn.lock file.

This is my question as well. But it might be that the 2.24-dependency pulled in by chart.js has saved my ass here, if `yarn why` doesn't lie to me:

image

.eslintrc.json
```diff
@@ -5,7 +5,9 @@
         "node": true
     },
     "extends": [
-        "standard"
+        "standard",
+        "eslint:recommended",
```
nathanlesage, line 9: Why eslint-recommended…? Just asking because I thought it's already in there…?

package.json
```diff
@@ -26,12 +26,14 @@
...
     "csl:refresh": "node scripts/update-csl.js",
     "reveal:build": "node scripts/update-reveal.js",
-    "wp:dev": "cross-env NODE_ENV=development webpack",
-    "wp:prod": "cross-env NODE_ENV=production webpack",
-    "watch": "gulp watch",
+    "wp:dev": "cross-env NODE_ENV=development webpack --progress --colors",
+    "wp:prod": "cross-env NODE_ENV=production webpack --progress --colors",
+    "watch:gulp": "gulp watch",
+    "watch:wp": "cross-env NODE_ENV=development webpack --progress --colors --watch",
+    "watch": "npm-run-all --parallel watch:*",
```
nathanlesage, line 33: What exactly does this command do? I assume simply run all in parallel processes…? If so, can't we emulate that behaviour either by merging them in a single watch command in the gulp-file or by using Yarn? (Just to only have yarn in here, and not npm…?)

package.json
```diff
@@ -26,12 +26,14 @@
...
     "test": "mocha",
     "test-gui": "node scripts/test-gui.js",
-    "lint": "eslint \"./source/**/*.js\" -f html -o eslint_report.htm"
+    "lint": "eslint --ext js,vue -f html -o eslint_report.htm source resources"
```
nathanlesage, line 36: Can we put the additional parameters into the `.eslintrc`?

package.json
```diff
@@ -165,6 +167,7 @@
     "sanitize-filename": "^1.6.3",
     "semver": "^7.3.2",
     "showdown": "^1.9.1",
+    "source-map-support": "^0.5.19",
```
nathanlesage, line 170: For vue I guess?

package.json
```diff
@@ -179,13 +182,16 @@
     "csso": "^4.0.3",
     "electron": "^8.2.5",
     "electron-builder": "^22.6.0",
+    "electron-devtools-installer": "^3.0.0",
```
nathanlesage, line 185: Just so I understand this: This is a package that enables us to install any of the supported additional DevTools extensions, right? Does this download ALL available devtools or only the ones we explicitly specify? Don't know about the space impact though

package.json
```diff
@@ -179,13 +182,16 @@
...
     "electron-notarize": "^0.3.0",
-    "eslint": "^6.8.0",
+    "eslint": "^7.0.0",
```
nathanlesage, line 187: Have you cross-checked that nothing breaks? Especially not the custom rules in the `.eslintrc`?

package.json
```diff
@@ -179,13 +182,16 @@
...
     "eslint-config-standard": "^14.1.1",
+    "eslint-loader": "^4.0.2",
```
nathanlesage, line 189: I guess the same as above: While I understand adding a vue-plugin is totally necessary, the question is why do we need to explicitly need the eslint-loader?

package.json
```diff
@@ -179,13 +182,16 @@
...
     "eslint-plugin-import": "^2.20.2",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
+    "eslint-plugin-vue": "^7.0.0-alpha.1",
```
nathanlesage, line 194: I'm not really comfortable with adding alphas (or betas, for that matter) as dependencies, and tend to wait until the plugins become stable — is there a reason we need the alpha here?

package.json
```diff
@@ -198,6 +204,7 @@
     "lorem-ipsum": "^2.0.3",
     "mini-css-extract-plugin": "^0.9.0",
     "mocha": "^7.1.2",
+    "npm-run-all": "^4.1.5",
```
nathanlesage, line 207: I think my question with regard to this relates to the above mentioned comment :)

resources/vue/sidebar.vue
```diff
@@ -435,6 +435,7 @@ module.exports = {
 
       // Create instances for all elements without already existing
       // tippy-instances.
+      // eslint-disable-next-line no-undef
```
nathanlesage, line 438: Better: Try to use `window.tippy` — and if that works it's fine (I'm using this in other parts of the renderer as well and it works just fine).

If it doesn't work fine, recommended is to put it as a global in the first line (I've done so where I used jQuery, it's basically the first line consisting of a comment with `/* global $ jQuery define etc...*/`)

source/main.js
```diff
@@ -168,6 +171,15 @@ app.on('activate', function () {
   zettlr.openWindow()
 })
 
+/**
+ * Load Vue developer extension
+ */
+app.whenReady().then(() => {
```
nathanlesage, line 177: Can we put this in the `app.on('ready')`-event listener? This would ensure that this is called BEFORE the application is actually instantiated and keep the logic in one place

webpack.config.js
```diff
@@ -55,11 +59,23 @@ module.exports = {
...
   plugins: [
     new VueLoaderPlugin(),
-    new MiniCssExtractPlugin({
-      // The files will be placed next to the respective components
-      // i.e.: vue-sidebar.js will be in the same directory as sidebar.css
-      filename: '[name].css',
-      ignoreOrder: false // Maybe we need this if the plugin spits out warnings
-    })
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.NoEmitOnErrorsPlugin()
```
nathanlesage, line 63: Why is the MiniCssExtractPlugin gone? And if it's not needed anymore, did you remove it from the package.json …?

Furthermore, I don't know if I understand all the changes you've committed to these files, so any explanation would be warmly welcome! :)

.eslintrc.json
tobiasdiez: I saw it somewhere and thought "recommended sounds good". Couldn't really find a comparison of standard vs recommended but it seems that the standard style seems to be more strict in general. So I've removed recommended now again.

package.json
tobiasdiez: Yes, it runs them in parallel. Although it has npm in it's name, it's completely compatible with yarn (https://www.npmjs.com/package/npm-run-all#yarn-compatibility). Thus `yarn watch` is equivalent to running `yarn watch:wp` and `yarn watch:gulp` in parallel.

package.json
tobiasdiez: Target files have to be specified via command line arguments (the `eslintrc` is only to define the style): https://github.com/eslint/eslint/issues/11223

package.json
tobiasdiez: Good observation, that was a left over from me playing around. Removed.

package.json
tobiasdiez: Yes, it's only a loader and doesn't come with the tools themselves. Pretty lightweight at 16kb.

package.json
tobiasdiez: I didn't noticed any problems, and linting still works.

package.json
tobiasdiez: The eslint-loader makes sure that everything that passes through webpack is formatted nicely. This is not strictly necessary, since the `yarn lint` also applies to the vue files now. I can remove it if you prefer. 

(BTW: the vue files have a few style warnings. Should I fix them in a new PR afterwards?)

package.json
tobiasdiez: This is necessary for eslint 7 support. Should I downgrade both? (One could argue that the alpha is not a real problem as it is only a dev dependency anyway and there are no problems right now...but I share your sentiment to stick to stable releases.)

resources/vue/sidebar.vue
tobiasdiez: window.tippy worked!

source/main.js
tobiasdiez: I didn't really understood what I did there, just followed the official documentation ;-) I just googled a bit and apparently the `whenReady` is preferred over `on(ready)`, see https://github.com/electron/electron/pull/21972. I've now changed on(ready) to whenReady and merge both methods.

webpack.config.js
tobiasdiez: It didn't seem to do anything, especially since there are no css files but just a few vue files. Also the vue-loader had a `extractCSS` option. But I'm not sure about this...

tobiasdiez:
Thanks for the review. I've implemented your suggested changes now.

package.json
nathanlesage: Perfect, thanks!

package.json
nathanlesage: Alright. But: Why specify both source and resource directory? I'm pretty sure I did this in the eslintrc, … right? I have no overview right now

package.json
nathanlesage: Perfect!

package.json
nathanlesage: Alright, thanks!

package.json
nathanlesage: Mh … do you think they'll upgrade to stable quickly? Because ESLint 7 sounds good, so I'll trust your expectations regarding this.

package.json
nathanlesage: > BTW: the vue files have a few style warnings. Should I fix them in a new PR afterwards?

Style warnings? If it's about "not scoped styles", we can ignore them. I do not like to put styling into these files, and prefer to keep them concentrated in the LESS files, as I'm not a huge fan of this mingling of styles!

package.json
tobiasdiez: The eslintrc doesn't contain anything about the resource folder, which I added because it contains the vue files, some other js.

package.json
tobiasdiez: Sorry, I meant "code style" warnings. Like wrong indent or different recommended order of template vs source etc.

package.json
tobiasdiez: The plugin seems to be developed pretty actively. I would keep the alpha for now (with the option to downgrade in case of problems).

package.json
nathanlesage: Ah okay … but seriously, now I'm wondering why it did lint the source-directory correctly o_O

package.json
nathanlesage: Perfect! :)

package.json
nathanlesage: Ah alright. Yeah I wasn't able to adhere to the ESLint custom config for the rest of the app precisely because I didn't have time to integrate ESLint for the Vue files, so yeah, the style warnings are surely something to tackle!

webpack.config.js
nathanlesage: The MiniCssExtractPlugin was necessary (at least when I added the Vue components) because otherweise the vue compiler threw around with errors. If it works now without problems, then we could certainly drop it. Just make sure it doesn't break, then all is good! :) (+ then we can also remove it from the dependencies, obviously …)

source/main.js
nathanlesage: I did not realize they had this change, but then Electron sometimes fails at communication. However, if it's now merged, then perfect! :)

nathanlesage:
Super cool, thanks a lot! Now we're left with only two small things, which means it's more or less ready for merging!

webpack.config.js
tobiasdiez: It works (at least for me) without any problems. I've removed the dependency also from the package.json

package.json
tobiasdiez: The compiled files are ignored from linting using the `eslintignore` file.

nathanlesage: Ahh, alright. Thanks!

nathanlesage:
And here we go…!

PR #761, opened 2020-05-10
Create github workflow to update CSL and language files

tobiasdiez:
## Description
Create github workflow to update CSL and language files.

## Changes
I've noticed that I had changed language files and csl-styles after invoking `yarn install`. I think, it's preferably to update the files regularly, and don't need to worry when checking out a new branch etc. The added workflow runs twice a month, updates the language and csl files, commits everything (as two separate commits), and pushes the changes back to the develop branch. Thus, as a developer, you don't need to worry about this anymore (maybe update it manually before a release).

## Additional information

Tested on: couldn't really test it, that's a bit of a disadvantage with github flows...


nathanlesage:
> Tested on: couldn't really test it, that's a bit of a disadvantage with github flows...

I feel you there, see [this awesome commit message I had last week](https://github.com/Zettlr/Zettlr/commit/d005f01001b4c7dcc4ea63b8924238d2e070f7ed)

But I wonder what benefits this has? Because the `yarn csl:refresh`-command is run either way before each release, because it's in `postinstall`, so I thought I didn't have to think about this either way …?

tobiasdiez:
I think the usual workflow for working on a new feature/bug fix is: create branch from `debug`, run yarn install to make sure you have updated dependencies (I'm looking at you `moment`). Start developing and commit. 
However, `yarn install` takes quite some time now and also updates the language/style files. What should I do with the changes to these files? Commit as part of my PR, ignore? I think it's better to not worry update this when you checkout the repository.

nathanlesage:
> and also updates the language/style files. What should I do with the changes to these files? Commit as part of my PR, ignore?

Well, they are in the `.gitignore`, so the changes won't be commited either way …?

tobiasdiez:
Then there is something wrong with the gitignore file, as I have a few files under `source/common/lang` as well as under `source/main/assects/csl-style` that are marked as changed.

But I also think it's a good idea to commit these files to the repository. They are not versioned dependencies, so you don't have any guarantee which exact files are used. So better put them under version control and update them regularly in my opinion.

nathanlesage:
Yeah, I need to add gitignores to these as well! Wanted to do this already but somehow always forgot them …

> But I also think it's a good idea to commit these files to the repository. They are not versioned dependencies, so you don't have any guarantee which exact files are used. So better put them under version control and update them regularly in my opinion.

You are right that in that case I would have no control over which files they're actually using, BUT if we commit them to git and then have them update every fortnite or so, you'd still run into the problem that some users open PRs seconds before the action runs and then they might need to pull these changes in.

The good thing at least with these three types of files is that their version doesn't matter, so it's easier to leave them out and have people from time to time update them manually, I'd say. Or do I miss something here …?

tobiasdiez:
If there is really no possibility that these files lead to problems, then it's probably a matter of taste.
A few differences come to mind:
- `+` More robust: if it compiles under one environment, it compiles also with the same code in a different environment at a later point
- `+` More self-contained (still works if the cls server is down, or the user does not have internet)
- `+` Shorter execution time for `yarn install` 
- `-` Additional code to maintain (this gtihub workflow)
- `-` Need to run the update manually directly before a release

In my opinion, robustness is the biggest advantage - but as you say this is probably not a big gain for these kind of dependencies (although it's probably a good idea to adhere to the same principles regardless of the kind of dependency).

By the way, the problem with merges shouldn't be an issue at all, as devs never need to change these files manually so you never have merge conflicts.
None of these points seem to be

nathanlesage:
I guess you have a valid point. Soooo then … we would need a workflow that is easily extensible to also include the language files and the CSL style(s) (plural because I'm still tinkering with the idea of offering a variety of styles for the users for some more plurality, but this is absolutely not urgent, just an idea). This way all assets could be compiled at once, every week for instance, and automatically merged into the develop branch (I really don't want to manually merge these changes, just pull them). And to make sure these run always before a release, we can also require the workflow in the build action!

Would you mind seeing to adding the necessary changes? You have more overview over what is happening, right…? This would be awesome!

tobiasdiez:
Of course, I would update this PR accordingly. What exactly are you missing from the current implementation? It runs the `cls:refresh` and `lang:refresh` tasks, so whatever you put into these scripts is updated and committed to the develop branch.

nathanlesage:
> Of course, I would update this PR accordingly. What exactly are you missing from the current implementation? It runs the `cls:refresh` and `lang:refresh` tasks, so whatever you put into these scripts is updated and committed to the develop branch.

Yeah, just remove the .gitignore on the CSL Language directory again and update the build workflow (but I'm afraid I don't know from the top of my head how to trigger external workflows from within another workflow, but I remember there was some way to do this …)

Basically, the perfect state would be:

- No gitignores in the respective folders
- Remove the updaters from the postinstall-field
- Have this workflow run these things instead every x days/weeks
- Require this workflow from within the build to make sure it runs every time I trigger a new build!

… I think that's it, right…?

And thanks for doing this! :)

tobiasdiez:
Ok, everything should work now as wished. Only, instead of invoking the update workflow in the build workflow (which I couldn't figure out how to do), I reused the npm scripts.

nathanlesage:
> Only, instead of invoking the update workflow in the build workflow (which I couldn't figure out how to do), I reused the npm scripts.

Fair enough!

Just looked over this, so it looks perfect! Thank you very much again! :)

nathanlesage:
Hey @tobiasdiez 

I'm not sure; can you check out the workflow files? Something with the run seems to have failed, but I have absolutely no idea what it might be:

image

tobiasdiez:
Follow-up at https://github.com/Zettlr/Zettlr/pull/765

PR #765, opened 2020-05-10
Fix typo in update workflow

tobiasdiez:
I was probably still in "debug" mode 🐒

## Additional information


nathanlesage:
Hihihi, thanks for fixing! :)

nathanlesage:
So, had to add an `npm install` as well, but the languages are still crashing it … but today I'm too tired to get to the bottom of this … I've had enough workflows for one day :D

PR #766, opened 2020-05-10
Fix release workflow - chapter 4

tobiasdiez:
The commit failed as there were no changes. 

I also think `yarn install --frozen-lockfile` is better, because it doesn't update the lockfile which would otherwise be committed as well. Is there a reason not to use yarn?

## Additional information

### DIFF
--- a/.github/workflows/refresh.yml
+++ b/.github/workflows/refresh.yml
@@ -23,7 +23,7 @@ jobs:
           node-version: '12.x'
       - name: Install packages in main and source directories
         # the scripts require some modules that we need to install.
-        run: npm install && cd ./source && npm install && cd ..
+        run: yarn install --frozen-lockfile && cd ./source && yarn install --frozen-lockfile && cd ..
       - name: Config git
         run: |
           git config --local user.email "action@github.com"
@@ -31,18 +31,20 @@ jobs:
       - name: Update csl
         run: |
           yarn csl:refresh
-          git commit -m "Update csl" -a
+          git add -A
+          git diff-index --quiet HEAD || git commit -m "Update csl"
       - name: Update language files
         run: |
           yarn lang:refresh
-          git commit -m "Update language files" -a
+          git add -A
+          git diff-index --quiet HEAD || git commit -m "Update language files"
       - name: Test build # to make sure we don't break anything
         run: |
-          npm install
+          yarn install --frozen-lockfile
           cd ./source
-          npm install
+          yarn install --frozen-lockfile
           cd ..
-          npm test
+          yarn test
       - name: Push changes
         uses: ad-m/github-push-action@v0.5.0
         with:


nathanlesage:
> Is there a reason not to use yarn?

I just thought npm suffices, as yarn would require an additional step to install, and this way we don't even have to worry about the npm lock files, as they're in the gitignore!

Which means: The workflow as it is will fail as yarn is not pre-installed on those VMs (unless they changed something)

tobiasdiez:
According to https://help.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions#installing-dependencies, yarn is pre-installed. Thus, there shouldn't be a problem.

Moreover, it seems there is another problem with the gitignore as the previous workflow https://github.com/Zettlr/Zettlr/runs/661171025?check_suite_focus=true printed
> Untracked files:
  (use "git add <file>..." to include in what will be committed)
	package-lock.json
	source/package-lock.json

nathanlesage:
> yarn is pre-installed.

Whaaaaaat! Then they added that one later one, hence yes please: Let's switch everything to yarn! + Add those package-lock.jsons to the gitignores … uno momento!

nathanlesage:
And short question: `git add -A` equals `git add *` right?

And then I have a better idea than what you are doing right now:

We could revert the git actions to simply do "git commit -m", but add a next step that only runs if the previous failed and tells us "Hi, nothing changed, so all good!"

This should afaik be a much nicer solution to prevent failed builds (as explicit "failure()"-steps indicate to Github actions that this failure is not a problem)!

tobiasdiez:
I can convert it to the `if: failure()` version. 
> as explicit "failure()"-steps indicate to Github actions that this failure is not a problem

Are you sure that works indeed like this? The documentation is not totally clear about this
https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#failure but this reads more like "if one of the steps failed, run this task." But I think the whole job is still marked as failed, leading to a failed workflow.

The right `git add` argument depends a bit on what should happen to new files. The `git add -A` stages all changes to existing files and adds new files as well. `git add -a` doesn't add new files. Not sure what's better here (how do you add new languages for example?)

nathanlesage:
> But I think the whole job is still marked as failed, leading to a failed workflow.

This would be bad, but if the docs are not clear, I guess we have to torture the master branch yet again lol

> (how do you add new languages for example?)

I don't :D The current way is: Deliver the four "default" languages (totally subjective and arbitrary selection besides English) with the app so that it works out of the box, even in offline mode. New languages are then downloaded by the user through the preferences dialog. These four languages are only to make sure the app works out of the box, even if the user is offline.

tobiasdiez:
I had a look and one failing step indeed marks the whole workflow as falure. See for example, https://github.com/JabRef/jabref/runs/496617043?check_suite_focus=true where the "Report issue" step runs successful, is marked with "if: falirue()" https://github.com/JabRef/jabref/blob/master/.github/workflows/check-outdated-dependencies.yml#L21 but the workflow still fails.

Thus, I would stick with the "prevent commit if no changes" approach for now.

nathanlesage:
Alright, I also just read up on what the OR-operator means in a bash-context. Sounds perfectly reasonable!

PR #772, opened 2020-05-10
Use gulp for handlebar conversion

tobiasdiez:
## Description
Use gulp instead of custom script. 

## Changes
Has the advantage that we can use gulp's watch feature (and that it's way less code). I made sure that the compiled templates are exactly the same.
*Note:* I still had a `other/quicklook.handlebars.js` file in the compiled template folder. This file is no longer generated with the new code as I couldn't find the source file.

## Additional information
That's probably my last PR concerning the build script. Sorry for the overhead, I used this to familiarize myself with zettlrs architecture and hope I can contribute some more meaningful PRs in the near future.

Tested on: Win 10


source/common/zettlr-template.js
```diff
@@ -90,7 +90,7 @@ function makeTemplate (cat, tpl, data = {}) {
     let precompiled = require(`./assets/tpl/${cat}/${tpl}.handlebars.js`)
     return handlebars.template(precompiled)(data)
   } catch (e) {
-    global.log.error(e.message, e)
+    console.error(e.message, e)
```
nathanlesage, line 93: My dear, why did you replace the persisting global.log here? :(

(P.S.: Reminds me that I need to splice the stream of logs in the provider and also pipe one part to the STDOUT if the app is running from command line …)

gulpfile.js
```diff
@@ -37,7 +42,37 @@ function lessThemes () {
     .pipe(gulp.dest('source/common/assets/css/'))
 }
 
+/**
+ * Precompile handlebar templates
+ */
+function handlebarTemplates () {
+  return gulp.src(resourcesTemplatesDir + '**/*.handlebars')
```
nathanlesage, line 49: I would leave out resources and instead use only `tplDirectory` and `lessDirectory` (it's more consistent with the rest of the app's naming conventions, and "resources" is a little bit doubled, as gulp is basically the resources-to-assets-compiler file — this should also be something for us to keep in mind: Everything that needs to be compiled needs to be put in resources, and compiled to one of the assets directory, may help clarify things in future decisions!)

nathanlesage:
Looks wonderful, but something is afoul with the indentation in the package.json; I saw this myself the other day and I'm still unsure where it comes from …

P.S.: Just two things: You have replaced one `global.log` with `console.log`, which needs to be reverted prior to merging — but it does remind me that I need to pipe global.logs to STDOUT whenever the app is running from console.

Second: instead of `resourcesLessDir` and `resourcesTemplatesDir`, better: `tplDirectory` and `lessDirectory`, it's closer to the naming conventions of the app, and "resources" is somewhat doubled, as gulp will even in the future only be used to compile something from resources to source (it's basically the resources-compiler!


I'm not sure, GitHub somehow dropped my single comments on the files after navigating away -.-'

source/common/zettlr-template.js
tobiasdiez: It didn't worked. `global.log` was undefined here. Thus I had a look in the files where the make template method was used (I think it was zettlr dialog) and there `console` was used as well. What's the proper way to solve this?

gulpfile.js
tobiasdiez: Ok, changed.

tobiasdiez:
Thanks for the review. I've changed the code accordingly. For the `global.log` vs `console`, see above.

source/common/zettlr-template.js
nathanlesage: Oh wait … yes, I didn't implement the global logging in the renderer process -.-' My bad, just leave console for now! We can deal with that later on :)

nathanlesage:
Aaaand here we go!

Issue #178, opened 2019-05-05
Add new 'Zettel' by clicking on a newly generated ID


mobamus:
First of all, I really want to thank you for your effort in writing Zettlr. As many other people here, I was/am in a constant negotiation with software implementations of the Zettelkasten method and my own workflow. Years ago I started using [Daniel Lüdeckes great software](http://zettelkasten.danielluedecke.de/). After the recent reconsideration of my current approach, I came to realize, that: a) I am using the software all wrong and therefore will never benefit from it as I was planning to; and b) since I already suffered twice from using software, that was at some point not usable anymore, I came to agree with the guys from [zettelkasten.de](http://zettelkasten.de) and you,@nathanlesage, that the only sustainable way of dealing with this, is a plain-text approach. Nonetheless, I think Lüdeckes ZKN3 software is great, but – for me –, it obscures the fact that you have to do a lot of work yourself (I know, that is a user problem and not software-related). 

That’s why I came across Zettlr in the first place. Then I saw that the books of your library in the background of your videos and the 'Zettels' you are writing for your own Zettelkasten. Since they are pretty exactly the same as my own areas of interest/work, I thought: This *must* be the right software for me. :-) Additionally, I really like your political conviction to FLOSS. :+1: 

Now to my request:

## Description
After long theoretical considerations of how to implement the Zettelkasten method and the question, if the 'Folgezettel' is *the* essential characteristic of Luhmann or not (for former argument see [Daniel Lüdecke](https://strengejacke.wordpress.com/2015/09/08/luhmanns-arbeitsweise-im-elektronischen-zettelkasten/), for later see [Sascha Fast](https://zettelkasten.de/posts/luhmann-folgezettel-truth/)) I had the idea, that it would be nice to create a 'Folgezettel' by generating a new ID in Zettlr and then click on it. So you would write `[[new id]]`, then click on the link and find yourself in a new Zettel file, which already contains the generated ID and a Backlink to the former Zettel. I think this would be a good implementation of the ideas of Sascha Fast in the aforementioned blog post. The generated file would get the name `new id.md` and would be in the same directory as the original one.

I think this idea corresponds a little bit to the [comment in another issue](https://github.com/Zettlr/Zettlr/issues/170#issuecomment-485315920) by @annargrs. 

P.S.: @nathanlesage Thank you so much for writing this great software! Keep it up! :+1:

annargrs:
I got this idea from sumblimeless_zk.

[...]

Issue #368, opened 2019-10-18
Create new note from selected text with hotkey


raoul-van-oosten:
Hi!

Sublimeless-ZK has the option to create a new note from selected text by pressing Shift-Enter. In this case the selected text becomes the title of the new note.

It would be great if this could be implemented in Zettlr too

nathanlesage:
There's another open issue somewhere where something similar (concerning hotkeys) is demanded. Until I get to unifying the hotkey management within the app (which currently is, honestly, quite a mess), I can't implement this. But I'll keep this open for then.

raoul-van-oosten:
I guess it is related to [Issue 90](https://github.com/Zettlr/Zettlr/issues/90) and [Issue 178](https://github.com/Zettlr/Zettlr/issues/178)

nathanlesage:
To a certain extent, yes!

patmaddox:
This is a dream feature for me with zettelkasten. It's refactoring notes. Write a long note, run a shortcut to extract some text to a note. Also the reverse to inline notes, to build up a longer document from a list of zettels.

PR #775, opened 2020-05-11
[FEATURE] Add new 'Zettel' by clicking on a newly generated ID

halcyonquest:
## Disclaimer
This is my first attempt at javascript and also my first time using git/github, so you will not hurt my feelings at all if something needs to be changed. It's just been fun to dive into something new and I'm excited to keep learning! :)

## Description
This PR adds the ability to create and open files from ZKN links by keyboard shortcuts and mouse clicks. This is to resolve bug #178 

## Changes
Just to clarify behavior, there are 2 pieces
1. Searching for a tag (this is existing behavior that happens when doing Alt/Opt + Click on a tag.
2. Create/Open file - this is the new functionality added with this PR. This functionaly directs Zettlr to search for a file with the exact name of what is in the tag. If found it is opened, if not a new file with that name is created in the current directory and then opened.

#### Changes to zettlr-editor.js
Added an event handler on the CodeMirror keyup event to capture if a user has pressed Alt/Opt + Enter or Shift + Alt/Opt + Enter.
##### So behavior is now like this
**Alt/Opt + Enter**: Performs search like point 1 above
**Shift + Alt/Opt + Enter**: Creates/opens file like point 2 above
##### Also tweaked the current click action to be the following
**Alt/Opt + Click**: This was existing behavior which is unchanged, and performs search like point 1 above
**Shift + Alt/Opt + Click**: Creates/opens file like point 2 above.

#### Changes to zettlr-renderer.js
Added function **openOrCreate** which performs the operation of finding and opening or creating and opening the given file name.

## Additional information

Tested on: Windows 10, MacOS Catalina


nathanlesage:
Ah, I suspected that much. The openOrCreate-function in the renderer has to go, because the source of truth is the main processes' FSAL. However, I have to move some stuff in the main process to an appropriate command to further consolidate the new structure, so you can leave that for now, as I have to go to work on this again. The "openOrCreate"-function can be removed in favor of the already existing "autoSearch", which will with my modifications automatically create a new file if none has been found and force-open that one instead. This way we can reuse a good portion of the functionality already present and don't have to reinvent the wheel!

For the rest, I'll be adding some comments real quick! But for now, thank you very much for the PR, and I'm looking forward to merging it once I've done my chores :)

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ This release contains several breaking changes to 1.6 due to heavy internal refa
 - The About dialog of the application now also holds a tab with debug information about both the binary, the system, and the current environment.
 - Tags with diacritics are now also removed on export (with the respective setting turned on), so that the removed tags match the tags which are highlighted in the editor.
 - Fixed searches behaving irrationally if you search again while the previous search has not yet ended.
+- Added ability to open or create new files from a ZKN link [[]] by either _Shift + Alt/Opt + Click_ or _Shift + Alt/Opt + Enter_
```
nathanlesage, line 50: The correct structure would be to wrap the shortcuts into code-backticks. Just search the Changelog, so you'll find some examples!

Plus please add your name there, as this is your doing, not mine! :) Everyone deserves a shoutout for their work!

source/renderer/zettlr-editor.js
```diff
@@ -222,6 +222,39 @@ class ZettlrEditor {
 
     this._searcher.setInstance(this._cm)
 
+    this._cm.on('keyup', (cm, event) => {
```
nathanlesage, line 225: This whole logic can be done much easier with a CodeMirror shortcut. Have a look at the CodeMirror manual and at the file `generate-keymap.js` (resides in `source/renderer/assets/codemirror`) to get an idea of where to put this logic!

source/renderer/zettlr-editor.js
```diff
@@ -222,6 +222,39 @@ class ZettlrEditor {
...
+          linkEndPos !== -1 &&
+          linkStartPos < linkEndPos &&
+          (linkStartPos - 1) <= cur.ch &&
+          (linkEndPos + linkEnd.length) >= cur.ch) {
+          let linkName = line.substr(linkStartPos + linkStart.length, linkEndPos - linkStartPos - linkStart.length)
+          if (event.shiftKey) {
+            // Performa a save to avoid a problem where a file
+            // is saved while trying to create or open a link.
+            // That action cancels the file open/create
+            this._renderer.saveFile()
```
nathanlesage, line 248: This will not work as intended, as `saveFile` is a strictly asynchronous event. It sends the file to the main process in order to save, but immediately returns, which might result in unwanted behaviour if the write process of the file is not done while the main process is already instructed to create & open a new file. From the top of my head I don't really know of a way to handle this, so I'll have to come back to this 🧐

source/renderer/zettlr-editor.js
```diff
@@ -222,6 +222,39 @@ class ZettlrEditor {
...
+            this.markClean(this._currentHash)
+            this._renderer.openOrCreate(linkName)
```
nathanlesage, line 250: As stated, this function needs to go, as the function below (`autoSearch`) will implement that behaviour once I've done my part! The plus side of this is that we can get rid of this whole if/else-thingy.

Oh, plus I realised we don't even have to add new shortcuts in that case o_O

Oh god, please forgive me, I'm so utterly absent-minded at the moment … I'll just leave this comment here, and wait for your reply on that matter … 🙈

halcyonquest:
Thanks for the feedback!

I'm fine with the idea of using a setting to control creating files. Before I make any changes though, I was looking through the open issues and wanted to at least bring up #368 about creating a new note from selected text. Seems like a great feature and similar functionality to this issue. Maybe it makes sense to add that at the same time or at least consider how that would be implemented? Would something like that also be governed by this setting or should that be a feature that is always available since it's not specific to a tag? In that case the shortcuts might make more sense since they will be consistent for all file creation inside the editor.

Also in regards to the setting, I would vote for keeping the keyboard shortcut (maybe just alt-enter) for people (like me :)) who prefer the keyboard and want to avoid reaching for the mouse as much as possible.

In the meantime, I'm going to study CodeMirror and look at the key mapping file and see if I can wrap my head around how it works.

Thanks again!

halcyonquest:
I've been reviewing your comments and I'm not clear right now what direction is desired, so I'm hesitant to make any changes.

Here are the two paths that have been discussed and input from others with their ideas would be welcome. The more I think about it though, I really think we should include #368 (in regards to creating notes from selected text) in this as well. In what I have done already I was able to add just a few lines to make this _mostly_ work and I think this is something that would be useful to people. Please add new ideas or contribute your own pros and cons.

## Preference setting
This option involves adding a new preference to modify the behavior of the current `Alt-Click` so that when enabled it will not perform a search, rather it will either open or create a file. If we wanted to resolve #368, that same option could be used to either search or open/create depending on how the preference is set. If doing this I would like to keep the `Alt-Enter` shortcut so people don't need to leave their keyboard if they're so inclined.

| Pros                                                                                                                              | Cons                                                                                                              |
|-----------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|
| Simpler for the user. They make a choice and go with it from there. Assuming users would usually prefer one action over the other | The user has to choose either of the options and no easy way to provide search other than through the search bar. |
| Sounds like it would be easier coding once some additional FSAL changes are made.                                                                  | If we were to also do #368, it might be helpful to either search or create depending on the user preference       |

## Key combinations
This option involves adding key combinations that will either perform a search or search for and create a file with the given text. It will work like this:

* Pressing `Alt-Enter` or `Alt-Click` on a link will result in a search for that link text.
* Pressing `Alt-Shift-Enter` or `Alt-Shift-Click` on a link will result in either opening or creating a file with the name of the link text

**If we want to include #368 in this the above two combinations will also work for a piece of selected text which should resolve that issue as well**

| Pros                                                                                                | Cons                                                                        |
|-----------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------|
| Gives user flexibility to search or open/create depending on what they're doing at the moment       | It involves two additional shortcuts that users have to learn and remember. |
| The Alt-Shift-Enter could also be applied to the search bar to create the same functionality there. |                                                                             |

nathanlesage:
> I've been reviewing your comments and I'm not clear right now what direction is desired, so I'm hesitant to make any changes.

Completely understandable, I just realized mid-review that I myself had no idea what to do really …

> so that when enabled it will not perform a search, rather it will either open or create a file.

As I am currently envisioning, it would only enable the automatic file-creation, and do everything else as it's now!

> The user has to choose either of the options and no easy way to provide search other than through the search bar

Concerning this con see my notes above!

> If we were to also do #368, it might be helpful to either search or create depending on the user preference

#368 is both semantically and structurally independent from this issue, I'm afraid.

> will also work for a piece of selected text which should resolve that issue as well

… but with an auto-generated filename, right?

> It involves two additional shortcuts that users have to learn and remember.

Not really a con, as I observe that people who really want to do this are pretty inclined to learn these shortcuts! :)

halcyonquest:
> #368 is both semantically and structurally independent from this issue, I'm afraid.

>… but with an auto-generated filename, right?

They way I understand that issue is that they want to select a word or phrase then be able to press a shortcut and it would create a file with the selected text as the name. To me it seemed like the same action as generating from a link, but maybe there is a good reason to keep those separate from each other.

nathanlesage:
> with the selected text as the name

Okay, then I read it differently, because for me it was something like "Select some text and create a new note with the selection as content" — so then there was a misunderstanding on my part!

halcyonquest:
I was looking at the `generate-keymap.js` file that you suggested above. I see that it is very easy to add new shortcut combinations to it. What I could find was in that same folder were all the functions (files with `zettlr-plugin-*.js` that implement the behavior, but it all seems to deal only with CodeMirror. I couldn't find an example that called back to the renderer in order to make a new file or call the main process. What do you suggest is the best way to approach that?

This is more of a "me trying to learn" question and I see there is a lot of work to do on 1.7 so reply when you can and sorry for adding something else to your plate!

nathanlesage:
> I see that it is very easy to add new shortcut combinations to it.

Exactly!

> but it all seems to deal only with CodeMirror.

Yeah, this is basically why I implement methods that call the main process within the ZettlrEditor-class itself. It's all a little bit "hacky" because I piggy-back on a lot of functionality of CodeMirror but do have to take that one monkey patch from time to time. The whole thing has been growing for almost three years now and it certainly will improve with more people dealing with it!

Speaking of which; I'll be doing a short video implementing this feature tomorrow just so that I can simply point to that video, because the internal workings of Zettlr do be pretty difficult to understand!

> This is more of a "me trying to learn" question and I see there is a lot of work to do on 1.7 so reply when you can and sorry for adding something else to your plate!

Seriously, I super value your engagement, as this is what has been missing from this app for the past two and a half years. I'm very excited to welcoming everyone into the team but yeah, 1.7 was and still is indeed a super hard piece to work with, so it's rather me having to excuse not being more able to tend to these questions — whose answers would most certainly benefit a lot of people!

source/renderer/zettlr-editor.js
halcyonquest: I see that saveFile can be called synchronously if we added a parameter to saveFile in the renderer that could be used only for cases like this. I'm looking forward to seeing how this code is refactored with the new fsal changes.

halcyonquest:
If I'm adding a new user preference, is there a trick to getting the label to show? It is in the en-US.json file, but for some reason the app doesn't read the change. Not sure what I'm missing.

halcyonquest:
I submitted a new PR #827 that uses the option you discussed above. I would prefer to close this and use that one instead, but will wait for any feedback.

nathanlesage:
You got it!

PR #827, opened 2020-05-21
[FEATURE] Add new 'Zettel' by clicking on a newly generated ID - Round 2

halcyonquest:
## Description
This PR resolves Issue 178 related to creating or opening Zettels when clicking on the link.

## Changes
Current functionality is that when you press `Alt-Click` on a ZKN link the text of the link is put into the search bar and a search is performed.

With this PR, a new checkbox preference is added on the Zettelkasten tab of the preferences screen which reads:
> When clicking on a ZKN link, create and/or open file with link name instead of search

If that option is turned on, when a ZKN link is clicked, Zettlr will search for a file with exactly that name. If one is found, that file is opened. If it is not found, a new file will be created in the same directory and that new file will be opened.

## Additional information
I made an attempt at translating the text in the language files (I just pasted whatever Google Translate spit out), so that may need some revision.

**There is currently another PR #775 that I would like to close in favor of this one.**

Also wanted to point out that even though I added the text in the language file, I wasn't able to get them to render. Please let me know what I'm missing.

Tested on: Windows 10, MacOS Catalina

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ This release contains several breaking changes to 1.6 due to heavy internal refa
 - If there is a selection in the document, its contents fill up the global search field on focus, if the global search field does not have any contents.
 - Fixed wrong display of project property table of content evaluation level.
 - When linking files, Zettlr will now present you those files that match with at least one tag with the currently active file, making cross-linking of notes as easy as typing the link-start and hitting the arrow down-key. Bonus: It'll present you these options even if the files reside in a completely different root directory.
+- **New Feature** (By Halcyonquest) Adding preference under Zettelkasten section to allow users the option to create/open files instead of search/open when clicking on ZKN Links.
 
 ## Under the Hood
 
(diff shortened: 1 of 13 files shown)


nathanlesage:
Hey, #775 can certainly be closed now, yes!

And probably this one as well, because as promised I managed to implement this video and explain everything with code, in order to show how to easily implement that feature: https://youtu.be/sa_ehm0yDEo

> Also wanted to point out that even though I added the text in the language file, I wasn't able to get them to render. Please let me know what I'm missing.

This question is also answered in the video (at the end, I think after the 1:00 mark)

BUT I see regarding your PR that you almost exactly did what I did in the video, so really cool! :) I'll have a look especially since your solution in the renderer itself may be better than mine which resulted in some weird behaviour! I'll be committing my changes soon (after I figure them out) and then we can merge this as well because what I didn't do was add a validation rule for the setting :D

halcyonquest:
Great! Thank you for making the video. I'm fine if you decide to close this one as well or pull whatever might be useful. Just starting to get comfortable with this, so I will watch the video and we can talk about what I can work on next.

nathanlesage:
Okay, here we go! Just finalized the stuff, a little bit has actually changed since the video, but only things I also talk about (e.g. I'm now using for the first time multiple events in one command :D)

Furthermore, I added a mention to you in the Changelog for your contributions!

PR #893, opened 2020-06-08
Replace the make.sh script with a Makefile proper

herbygillot:
## Description

This change seeks to replace the `scripts/make.sh` shell script with a `Makefile`.

You can run `make help` to see available targets, but the workflow will usually be `make build`, followed by `make release-<platform>`.  Or one can simply run `make release-all`, which builds all deps and builds a release for every OS just as `make.sh` did.

This `Makefile` also supports disabling code signing, which can be useful in certain situations.  This is done via the `NO_SIGNING` variable; example: `make NO_SIGNING=1 release-mac`.  This is currently only supported by the `release-mac` target.

## Changes

* Adds `Makefile`
* Removes `scripts/make.sh`

## Additional information

Tested on: `macOS 10.15.5`


nathanlesage:
Hey, thanks for the PR! Unfortunately, the `make.sh`-script is not used anymore, it's a remnant from before we switched to GitHub actions. Right now, every build is done using simply the `yarn`-commands, so a Makefile is not really necessary!

Plus, I fear it would be one more place to keep updated whenever we change something in the build commands …

herbygillot:
Well OK, then in that case I'll close this.  Thanks.

PR #908, opened 2020-06-10
Add support for subIDs

kyaso:
## Description
Allow multiple IDs inside a file. The first ID is the "main" ID, while all other IDs are considered "subIDs". When a link to a subID is clicked, the corresponding file, in which that subID is defined, is opened.

## Changes
I added an array `subIDs` to the file descriptor. Then I extened the file parser to also read any subID into `subIDs`. When an internal link is clicked, in addition to `id` and `name`, `subIDs` is matched to find the target file. Further, I had to modify `findObject()` slightly to support `prop`s of type Array.

## Additional information
With this PR, we can have multiple IDs inside a file. This can be useful, for example when we want to link to a specific heading. Or any other part of a file.

In the current version of this PR, we have to first insert the ID at the location of our choice, then copy it, and paste it between `[[ ]]` in the file from where we want to link to that ID.

Tested on: Linux Mint 19.2

### DIFF
--- a/source/common/util/find-object.js
+++ b/source/common/util/find-object.js
@@ -23,7 +23,15 @@ module.exports = function findObject (tree, prop, val, traverse) {
   // Is the tree even defined?
   if (!tree) return undefined
   // First let's see if we can shortcut
-  if (!Array.isArray(tree) && tree.hasOwnProperty(prop) && tree[prop] === val) return tree
+  if (!Array.isArray(tree) && tree.hasOwnProperty(prop)) {
+    if(!Array.isArray(tree[prop]) && tree[prop] === val) {
+      return tree
+    }
+    if(Array.isArray(tree[prop]) && tree[prop].includes(val)) {
+      // The property is of type Array (e.g. subIDs)
+      return tree
+    }
+  }
 
   // Now search the tree
   let ret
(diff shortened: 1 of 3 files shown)


nathanlesage:
Hey, thank you for your PR! Interesting idea you had there, but unfortunately, the way you anticipate this feature to work will require much more. Plus, the concept of sub-IDs will make the overall workflow much more complicated, and hence I'm not really a fan and I am convinced that implementing this would counter the idea of a Zettelkasten. To link to sub-headings, we needed something like URI-fragments, but not ZKN-links. Please don't forget that these links are actually fuzzy-searches, and not *that much* "real" links. Introducing sub-IDs will unnecessarily increase the application workflow's complexity. Plus, it's not covered in any recommendations on how to work with Zettelkasten. So I'm afraid we can't add this to the code base …

kyaso:
Yeah, I was concerned too about the sub-IDs interferring with the idea of Zettels having only one unique ID. I thought of this more like a handy way to place bookmarks in larger (non-Zettel) files. Anyway, thanks for your feedback!

nathanlesage:
Yeah I see where you want to go, but bookmarks are something different than ideas — however, feel free to discuss how that might be achieved with other users or so, maybe some interesting ideas of how to solve that with the least amount of added functionality becomes possible!

PR #921, opened 2020-06-12
Added graph visualization of notes

JulienMir:
## Description
I really wanted the graphical view of the note network so I added it.

By selecting the button on the toolbar, a dialog pops up and displays a network of notes.
Each note is represented by a circle, links from one note to another are represented as edges between circles.
Each circle is labelled using the filename of the note (the extension and eventually the ID are removed first).
The more links (inbound and outbound) a note has, the bigger the radius of the circle representing it.

You can drag notes around to reorganize them.
Clicking on a note launches a search with the associated ID (therefore opening the note but also displaying connexe notes in the left panel).
All actions that can be done on the note (i.e. the circle) can be done on the associated label (its title)
Clicking and dragging the background lets you translate the whole network.
You can also zoom in and out with the mouse wheel.

As of now, the only information you can see besides notes and their relation is the centrality of a note.
This is indicated by the size and color of a node.
I hope this feature can grow into something more useful, displaying a variety of information.

## Changes
I added a new dialog.

I changed the toolbar to display a new button (next to "tag cloud").

I modified the fsal-file to create and update a "database" containing all informations about note's relationships. This is linked to a provider.

## Additional information
As it is the first time I ever did a PR, I did not want to overdo it by adding too much.

Most of the stuff I've done is based the "tag-cloud" functionality.
I copied and adapted almost all of it.
I might have added unnecessary code this way...

I did not manage to change textual informations (Like button labels, translated description, tool tips, etc.) despite adding all the necessary values in the different files (e.g. fr-FR.json, toolbar.json, etc.)

Tested on: 
Windows 10 using yarn


nathanlesage:
Hey, thanks a lot! That's quite a big thing, but, funny enough, almost implemented the way I would've done it, so for that, great!

I do not have time to review it currently, and it won't ship with 1.7, but in the meantime what I've already noticed:

- Please have a look whether or not we can make this happen without `d3`. I know d3 is kind of magic for a lot of stuff, but it's also because it's a huge dependency that would push Zettlr's binary beyond 300 megs, which we need to prevent. Henceforth, I'd like you to look into the possibility of using either ChartJS for that, or maybe even something different altogether.
- Second, your branch is not on the most recent state of the develop branch, so make sure to run a `git pull` on this PR so that, e.g. the Changelog is up to date again (+ please remember that once development of 1.8 starts you should insert a credit to you for the feature, in case I forget)
- Third, no modifications to the language files, please. They'll get overridden by our automation anyway and this keeps the git history clean
- Fourth, instead of saving all these links to file, please check whether or not we can add them to the cache instead. This saves disk-space and is more resilient against errors. The tags are saved because they are independent from the files (more or less), but with filelinks we need to be urgently aware of possible state-problems incurred due to outdated information in a separate source of truth
- Fifth, I had a great idea of how to supercharge the graph network, but for that it would make more sense to have a one-on-one-discussion, because this way we can even make sure that it's much more extensible in the future, and it should be super easy to add it to this PR!

So long, thanks again for the PR and I'm looking forward to working on this after 1.7+my vacation!

Aigeruth:
Hey @nathanlesage, I have taken a quick look on the application size and I noticed that the `app.asar` file is more than 100MB. I have compared Zettlr to other Electron-based application and they seem to be smaller. Please, see the data below.

```bash
➜  Zettlr git:(develop) du -sh /Applications/Zettlr.app/Contents/Resources/app.asar
107M	/Applications/Zettlr.app/Contents/Resources/app.asar
➜  Zettlr git:(develop) du -sh /Applications/Slack.app/Contents/Resources/app.asar
2.3M	/Applications/Slack.app/Contents/Resources/app.asar
➜  Zettlr git:(develop) du -h /Applications/Obsidian.app/Contents/Resources/*.asar
676K	/Applications/Obsidian.app/Contents/Resources/app.asar
8.1M	/Applications/Obsidian.app/Contents/Resources/obsidian.asar
```

If I understand correctly, the `.asar` file is just a tarball of the application source files. As far as I can see, Zettlr does not use ship a minified bundle as part of the package, and only uses Gulp for compiling LeSS and Handlebards, and webpack for transpiling Vue components. (Please, let me know if I misunderstand something as I have not fully grasped the code-base yet.) The Electron team recommends bundling for [performance](https://www.electronjs.org/docs/tutorial/performance#7-bundle-your-code). If you think, this would be beneficial for Zettlr and my time permits, I can look into setting up a pipeline for bunding Javascript. This would potentially create space for D3 as a dependency as well.

nathanlesage:
@Aigeruth Well. I see your point, but this is not Zettlr's fault.

Here's what's taking up so much space:

1. The `mermaid.js`-dependency sports over 40 megabytes, so without it it would be 60 MB.
2. The main process's assets are another 24 megabytes, which is to over 90 % due to the included dictionaries.

Removing both would according to simple math result in an `app.asar` of 40 megabytes, of which more than 20 megabytes are still dependencies.

> The Electron team recommends bundling for performance.

Don't you think I have thought about that myself? Fact is:

> ensure that the overhead included in calling require() is only paid once when your application loads.

This is already the case. Only a few scattered `require`s are evaluated later on, almost every require is evaluated right on application boot in the initial startup cascade. So there's no win in changing that.

TL;DR: Neither the application size nor the way it actually is technically composed are something we could in any way really optimize. However, what the real performance bottlenecks are is quite easy to spot: My own, shitty and un-optimized code. That needs to be updated. But until very recently I did not have any help, and maintaining such a large codebase is very difficult alone. It will get better, but it will take time. Please feel free to join in, if you spot something that's worth optimizing!

JulienMir:
1. d3's latest versions lets you import only only what you need. I have to admit I didn't look into it so far, but cherry picking only the useful modules could already reduce a lot the size. Especially since the `d3-force` might be the only really needed module here. ChartJS doesn't seem to have this kind of feature
2. Will do
3. Got it
4. I did not fully understand the code base, I will look into it.
5. I'd be happy to hear about your ideas.

Aigeruth:
Hey @nathanlesage, I'm sorry if it came across that way, but I have meant in no way to imply that this is Zettlr's fault or that you are not aware of the performance recommendations of the Electron team.

Thank you for explaining your point of view. Would you mind helping me with how you sized up mermaid.js, so I have the full picture here? I looked at the distributed versions of mermaid and it seems to be smaller to me (1.1M), but I may have missed something obvious:

```bash
$ du -h source/node_modules/mermaid/dist/*
 16K	source/node_modules/mermaid/dist/index.html
408K	source/node_modules/mermaid/dist/mermaid.core.js
688K	source/node_modules/mermaid/dist/mermaid.core.js.map
4.4M	source/node_modules/mermaid/dist/mermaid.js
4.8M	source/node_modules/mermaid/dist/mermaid.js.map
1.1M	source/node_modules/mermaid/dist/mermaid.min.js
5.1M	source/node_modules/mermaid/dist/mermaid.min.js.map
```

These bundled and minified versions already contain all necessary dependencies.

I agree with you and probably there is not much to shave off from dictionaries and other assets.

nathanlesage:
>  I'm sorry if it came across that way, but I have meant in no way to imply that this is Zettlr's fault or that you are not aware of the performance recommendations of the Electron team.

Don't worry, I was a little bit snappy myself, sorry for that!

> Thank you for explaining your point of view. Would you mind helping me with how you sized up mermaid.js, so I have the full picture here? I looked at the distributed versions of mermaid and it seems to be smaller to me (1.1M)

Huh, this is surprising — but it may very well be incurred by the peer-dependencies of mermaid itself. I remember walking the extra mile and I just downloaded mermaid into an empty NPM project to see what additional dependencies it would in itself pull, and in the end the `node_modules`-folder with everything from mermaid was 40 megs, and this held true when I then pulled in the dependency into Zettlr's code base (minus two or three packages that were already there). After all, it's not only about the modules themselves, but the ridiculous amount of other modules they're using.

This is basically why I'm always hesitant in adding new dependencies, because I fear that the additional peer dependencies bloat up the overall size :/

> These bundled and minified versions already contain all necessary dependencies.

Is there an additional package on NPM that I missed? Because I'd love having 1.1 meg over 40 megabytes, that's for sure. But I want to pull it in as an NPM package, and not having to create an additional CI run just to keep that dependency updated. It seems you've managed to only retrieve the mermaid dependency?

(But maybe we should now transfer this discussion someplace else, maybe you could open an additional PR if you managed to strip the codebase of these 39 unnecessary MB of packages)

nathanlesage:
And @JulienMir thanks for checking it out! I always thought d3 would be one of these monstrosities, but if you can keep it small, this might work out. How much additional overhead do the d3-modules incur? (+ we need to keep in mind to then maybe replace chartJS with d3 in the mid-term)

Concerning your fourth point: No worries, it's pretty recent and you couldn't know!

Keep your time, and after 1.7 is released, I'll come back to you (and if I forget, just ping me with a comment here) and we can have a look in-depth, as I'm really excited to add this functionality!

nathanlesage:
Quick update, I have to apologize: Mermaid pulls in d3 as a dependency, hence Zettlr already has this thing as a transient dependency, so I waive my reservations!

JulienMir:
> Quick update, I have to apologize: Mermaid pulls in d3 as a dependency, hence Zettlr already has this thing as a transient dependency, so I waive my reservations!

Good to hear!

I tried to look into how to use the cache instead of the file system.
The problem I see is that providers are instantiated before the FSAL (and the cache) and are therefore "harder to use" (in a clean way) together.

Should the FSAL be a provider as well or should I initialize the LinkProvider outside of the `_bootServiceProviders` function? (Or do you see another, better way to do this?)

nathanlesage:
Hmm … from the top of my head I'd argue that we might want to add a hook after the FSAL has loaded to initialize the link provider — I know that this trick has worked good in the past.

Oh, even better: Why not add a `global.linkProvider.sync()` (or something like that, I don't know what identifier the link provider has) in order to always call it when something in the underlying file structure has changed? Then we could hook that also into the callback in the main constructor for when the FSAL emits a change event!

JulienMir:
I did as you suggested: I added a sync function that gets called on 'fsal-state-changed' signals. It indeed seems like the best solution

ALien747:
### Hey Julien,

this is a pretty cool feature and I almost had started implementing the same thing if I didn't had found your pull request.

**Some remarks from side:**
- You are using `global.config.get('zkn.idRE')` as regex for finding the links. The default value is `(\d{14})` - a 14 digit number. So it would only find links that match this expression and I think we cannot expect that all links follow this format. At least my do not. For instance my link would look like `[[Journal - 18.07.2020]]`. Because I like a clear naming. So my sugestion is to simply match all _"links"_ that are in the `[[BRACKET FORMAT]]` / or whatever alternative is defined in the settings. You might compare this to [obsidian](https://obsidian.md/). _It could also be that I don't refference things in the correct zettler way but for me this is more intuitive_.

- When one defines an alternative `zkn.idRE` that matches everything it's quite fun to see that it also matches links to Youtube etc. which could be quite useful. So linking to websites, Zotero-citings etc. would be nice. Maybe one could add filters.

**Some ideas**
- Tags and links are cool. But when working with a graph network it's nice when one can have different types of objects that could be displayed in their own style. For instance: `@person/Julien`, `@location/germany`,  `@topic/computers`

Exr0n:
Hey @nathanlesage, I hope you had a good vacation :)
I think you asked Julien to ping you after 1.7. It looks like there's been no activity here for a while, and I am (and I'm sure many others are) very excited about this feature. In case @JulienMir isn't here, what can the rest of the community do to help get this up and running?

nathanlesage:
Hey, true! One of the reasons I didn't reply to this PR is that I have to really hard think again about whether this feature fits Zettlr, or not. The main reason being that I get a lot of comments along the lines of "Roam can do it, why not Zettlr?" or "Obsidian can do it, why not Zettlr?", and the thing is that I realize more and more that we're talking about two different types of application – one that is meant for heavy-lifting of writing text (Zettlr) and pure note-taking (such as Roam and Obsidian) apps.

And the more I talk with other people about the feature-set of Zettlr, and the more I think about this myself, the more I get the impression that instead of copying over functionality for pure note-taking from other apps we should focus more on getting Zettlr _en par_ with Word etc. Because, let's be honest: Note-taking apps are easy to build and it's much easier to get all of this running, but when you're dealing with longer texts such as papers or even whole book-chapters, this whole approach is prone to slow computing times, often breakages, and other stuff.

Besides, while graph visualizations do sound super awesome, it's one of these things where you can implement it and get a ton of likes, but there's actually no surplus value added with them. They are [bad for induction](https://reticular.hypotheses.org/1724), are really CPU-heavy and [do not boast productivity at all](https://www.reddit.com/r/RoamResearch/comments/ho7tmg/practicality_of_graph_overview/).

This is why I'm seriously considering to not implement this, which does hurt me super hard, because there's so much dedication and work in it, but on the other side I have real anxiety that this project becomes just that inch too much where it suddenly becomes useless to everyone :(

zifeo:
@nathanlesage In my opinion Zettlr would get a lot from this feature. There are plenty of reasons not to use Obsidian or Roam, and this enhancement bridge the missing gap I am missing for making Zettlr my main editor.

nathanlesage:
But … do we really win anything from this graph?

Exr0n:
@nathanlesage here's my humble opinion on the whole topic. 
## Process
Right off the bat, I will note that I am biased towards adding this feature. Before prior research, I think that graph visualization is highly helpful for learning new concepts and documenting the learning process for reflection. During the compilation of the following hypothesis, I will do my best to adhere to [waitbutwhy's thinking like a scientist](https://waitbutwhy.com/2019/09/thinking-ladder.html)(scroll down about a quarter of the way or search for to the heading "The Battle Over Our Beliefs") Please correct me if I misunderstand something you or one of your sources. 
## Responses
1. > Roam and Obsidian are note taking apps, while Zettlr is for heavy-lifting writing.
    I agree that writing large "publication style" text documents doesn't benefit much from backlinks and graph viewing. If Zettlr decides to poise itself as a heavy-lifting text editor, then I would agree that this feature shouldn't be added.
    However, I came to Zettlr as an open source alternative to Obsidian. If this feature doesn't get added, I would probably end up just using one of the alternatives. I believe that many others are in the same boat. Maybe this could become a spinoff of Zettlr or a "note-taking" version.
2. > Graph visualization is bad for induction.
    I will first note that [the VNA paper that this article references](http://www.tommasoventurini.it/wp/wp-content/uploads/2014/08/Venturini-Jacomy_Visual-Network-Analysis_WorkingPaper.pdf) uses an analysis of one example static graph created from web hyperlinks.
    Here is a recreation of my thought process while reading the linked article (some points are re-ordered):
    1. We don't know if or why it should or does works
        - I think this is a good point to take into consideration, but it does seem to work. If a feature is helpful for some people, I think its reasonable to add it to an application.
        - As it says, we use p-values too.
    1. Graph isn't reproducable because the algorithm is arbitrary.
        - This is true, but I don't see how this is a problem. Although the absolute positions of the nodes are not deterministic, we try to avoid reading into that anyways. At most, this boils down to "how do we ignore the absolute position?"
        - It's reasonable to run the algorithm multiple times to get a better overview of the structure of the graph and shy away from reading into absolute position.
    1. The knowledge is implicit so it's difficult to "teach and disseminate".
        - I don't see the impact of this point--the graph view would be a tool to pick out patterns in your own note taking. What matters is your own ability to read into the graph, and people seem to find this helpful for that. You wouldn't avoid a note taking strategy beacuse it makes it harder for other people to read...
    1. You might forget to read into absolute relative distance instead of absolute position
        - This seems to be the biggest problem people have with network vis, but there are reasonable work-arounds:
            - Run it multiple times--I don't think this is unreasonable. While this is not possible in static publications like papers, it is absolutely possible on websites and in Zettlr.
            - Inspect certain clusters with interactivity--Pull out one or some nodes, perhaps to different edges of the graph, to see what is affected. What gets pulled with it? What gets stretched between? This is how I would use the graph visualization, similar to how it works in Obsidian. 
    1. It's difficult to visually pick out the direction of an edge
        - I think this is absolutely a fair point, and something that we can address. For example, we might prefer that links point from right to left, perhaps by linking nodes to invisible nodes on either the right or left edge of the screen based on their in and out degrees.

That's all the analysis I'm able to do tonight. Ultamately, I think we can make use of a graph visualization by using it like a mind map and taking advantage of Zettlr interactive use case. If Zettlr doesn't want to be a note taking app then I guess people will have to find other alternatives.

nathanlesage:
Hey, thank you for this rich comment on the issue! In fact, we're now right at the center of a problematic that has been haunting Zettlr since the beginning: What does it want, and what not?

Judging purely based on the amount of reactions here on this PR and based off other indicators, I'd say that many of you do use Zettlr as a note-taking app, and it's merely me who's using it for full papers. But that – as you rightfully state – is not something that's hindering implementing of this feature. I thought about it, and upon reflection, I realized the biggest headache that's coming from this PR is: Who is going to maintain it? I won't be using it very much, and so bugs will very likely elude me, and given the size of this PR, there's quite something to maintain here. Maybe my whole aversion towards implementing this PR is that I only see all the work coming with it without any of the benefits. But this would then leave the area of "note taking vs. text editor" and enter a general problematic of Open Source.

I mean, the main point (besides some scientific points argued on in the post I shared above) speaking against this is really that I have fear not being able to keep the app together, because there are so few people contributing code, and many who make a PR and then leave never to be seen again :/ If that problem would somehow be solved by people committing to keeping tidy one single part of the code base this would make the decision easier for me to just merge this thing.

NorthernScott:
Like @tomkersten above, I saw this and was really excited and came back to check on progress. I am starting to use Zettlr for note taking and to organize information for projects. I'm definitely not looking for a Word replacement. That said, like others here, I empathize with your need for help maintaining the project. Unfortunately, I don't have the skills to assist. Looking forward to seeing the continued development on this project.

carregal:
Hi there, I'm quite late to the discussion but I believe I can relate both to why some people are attracted to a graph view of notes and what keeps @nathanlesage uncomfortable with the idea of adding such a feature.

If I may try to contribute, my 2c would go in the direction of looking for the "Jobs to be done" with a graph visualization.

As far as I've seen both on this thread and elsewhere, graph visualization of nodes usually tries to help answering questions like:
- Which are the more linked nodes?
- Which are the non linked nodes?
- Which nodes are "roots"?
- Which nodes are "leaves"?
- Which nodes link to this particular one?

Those questions are certainly useful for those going Zettelkasten style, but I'd say they could also be helpful for those trying to keep any other form of non linear writing, notes or not.

So maybe we could try thinking in terms of what minimal set of features Zettlr would have to offer to answer those questions but without needing a graph visualization.

Here are a few suggestions of what could be offered instead:
- A list of files linking to the current one
- A lis of files with linked-to and linked-from counts

If those 2 lists were available the previous questions could be answered like:

- Which are the more linked files?
   The top ones of the linked-to list
- Which are the non linked files?
   The ones where count(linked-to) and count(linked-from) is zero
- Which notes are "roots"?
   The one where count(linked-to) is zero but count (linked-from) is non zero
- Which notes are "leaves"?
   The one where count(linked-from) is zero but count (linked-to) is non zero
- Which files link to this particular one?
   The ones in te linking-to-current list

Another thing to keep in mind is that the need to check for files linking to the current one is probably a daily task, while checking the graph for "anomalies" is probably a weekly task. Thus the first scenario would deserve a GUI solution, while the second scenario could be used with a batch solution.

Thus, the first list (linking-to-current) would probably be more useful if accessible while using the editor, but the second list (with the counts) could be just an option on the File menu that generates a CSV file to be analyzed elsewhere (typically a spreadsheet).

This would probably lessen the overload both on size and maintenance for @nathanlesage but would still allow those of us checking on a per note basis for notes linking to the current one (in order to check backlinks) and hunting for orphan notes by the end of the week.

 If visualizing the graph end up still being a need after these two lists were available, we could consider adding a "graph-dump-as-csv" option too, but I believe it would be worth trying the simpler format first.

I apologize for the first comment being so long and I am looking forward for feedback on the suggestions, thanks in advance.

nathanlesage:
@carregal That's an interesting proposal I have not yet considered. Indeed, this would both be in line with the core philosophy of Zettlr and easier for me to maintain. The problem would then only be that a lot of minds are already very much hyped towards the graph view (as opposed to a simple list with information). This is one of the basic problem why many people specifically demand a graph _view_ (instead of just the graph information, as proposed by you). I guess this is related to the fact that such a _view_ is much more appealing, as it looks nice. But in fact you are very right in assuming that the real goals of such a graph network would only be to extract those information and it might even be more accessible (e.g. if presented in the sidebar).

carregal:
@nathanlesage, just to make it clear, I was not advocating against graph views per se. Being one of those who use and like them, I can relate to the attraction :)

My point was more in the separation of concerns direction. Zettlr would be responsible for keeping (and exporting) the graph itself, and other tools would be responsible for the analysis or visualization.

There are lots of graph visualization tools, both online line and for desktop use. I believe it wouldn't be that much of a trouble exporting graph data (nodes and edges) using a format compatible with them.

Examples of such tools, both free:
https://vue.tufts.edu
https://www.yworks.com/products/yed

I fully understand it's your call, but I strongly believe there could be common ground here for both point of views (maintainability and visualization).

tomkersten:
Building on @carregal's suggestion: A simple [graphiz](https://graphviz.org/) document may suffice as well (there are both local and web implementations that can render that file format).

nathanlesage:
We're definitely getting somewhere!

the-solipsist:
Do note that there's a lua filter for Pandoc ([diagram-generator](https://github.com/pandoc/lua-filters/tree/master/diagram-generator)) that helps Pandoc use external apps (e.g., GraphViz, PlantUML, matplotlib, and hopefully [Mermaid](https://github.com/pandoc/lua-filters/issues/93)) to convert the graph mark-up into an image/pdf. This allows for the graph text to be generated by Zettlr, the graph rendering to be done through an export function.

Qazzquimby:
Hi,
Forgive me if there's somewhere else to discuss this. I couldn't find one in issues. 

I realize a plug in system would be a great deal of work, but it would offer so much flexibility in terms of how to scale. 
I am one interested in this primarily as an opensource alternative to obsidian. Some of the features I would love (opt in automatic backlinking) others of you would hate, and fairly. 
Keeping such features in plugins would avoid bloating zettlr with ui and file size and maintenance costs. 

Could someone familiar with the code base summarize the challenges for me?

gabrielrudas:
Hi
I want to contribute with my humble opinion (as  a new user and Zettlr enthusiast) to second @carregal proposal.

-Instead of doing graphs because some other app does it, the reasoning should be: what problem does this graph-thing address? I think  the point is to help users to find relations between ideas by navigating and mapping networks of notes (or longer texts). 

- Offering a way to trace backlinks and understand networks of interlinked texts  should be pursued, but not to copy  Roam  or Obsidian. The point is to offer an alternative to ease the workflow between note-taking, concept-creation, draft-writing and paper-composition (which is, short, what we do in the humanities). If we can have such a tool as a clean, GIU feature (in the side bar, maybe), I don't mind not having graphs. 

 - If Zettlr is intended to replace Word, it has to be more than a less-bloated Word. It has to offer things you cannot do with it. Being zettlcasten-orientated is probably the most important feature Zettlr offers. I would use Zettlr instead of Word (or Scrivener for that matter), but I think that it should focus on helping users to trace and discover relations between ideas. It doesn't matter if that goal is achieved using graphs, lists, chirstmas threes or whatever. 

For the record. I am a researcher in Humanities with no coding experience. I love this app and I love that @nathanlesage understands our needs (and hates fascists).

gabrielrudas:
A suggestion of how the side bar porposed by @carregal option could work:

It shows a lists of nodes divided in:

- Backlinks 
- Links from this file to other files (especially useful for long texts)
- Files that are both backlinks and linked to.

The list allows three actions
-Normal click: opens selected file as new tab
-Ctrl-click: opens selected file as a QuickLook window
-Ctrl-alt-click: shows the nodes of the selected file

This last feature would be fundamental, since it would allow a "rabbit hole" experience, helping finding new connections between by "surfing" a zettlecasten. 

Exapmple: lets say I have a note with the title "Populism (for and against)". I click on the side bar. It shows:

Backlinks

* Chantal Mouffe 
* Ernesto Laclau

Links

* Zizeck-Lacalu Debate

Two-way links

* Crisis of Liberalism

Ctrl-alt-click on  "Zizeck-Lacalu Debate", so instead of taking me to the text, the side bar shows me that file links:

Backlinks

* Populism (for and against)

Links

* Pink Tide

Two-way links

* Zizeck does not get Adorno

Ctrl-alt-click on "Zizeck does not get Adorno". The side bar shows:

Backlinks

* Zizeck-Lacalu Debate
* Frankfurt School now

Ctrl-alt-click on "Frankfurt School now". The side bar then shows

Backlinks

* Zizeck does not get Adorno

Links

* Winfried Menninghaus

And so on. This way, I could find a distant connection between Laclau and Menninghaus using a tool that, I guess, is easier to code and maintain than a graph visualzitaion.

omartan:
> For the record. I am a researcher in Humanities with no coding experience. I'm starting learning the zettlekasten method thanks to Zettlr. I love this app and I love that @nathanlesage understands our needs (and hates fascists).

Excellent reply and I wholeheartedly agree, the graph is nice and fancy and @nathanlesage has stated his reservations which I think is justified. 

And I do think it's important to be able to visualise the links as that'll help into getting new insight, as someone who's getting into UX Design, I'm surprised with how much visuals help to give more insight. Perhaps there's a way to figure out how we can implement a UI that'll help the users visualise links. Thoughts?

pcuci:
Here's an alternative take on graphing concept relationships (for inspiration, for those who really need it in the meantime - works out of the box): https://marketplace.visualstudio.com/items?itemName=tchayen.markdown-links - you'd have to install vscode, a code editor, to use it; definitely not for non-developers...

The idea here would be to enable (existing) 3rd party plugins like Markdown Links (which are maintained by others) to work against Zettlr managed projects. For instance, we could download and install vscode in the background, and trigger it to open the directories to visualize our Zettelkästen.

This is just a bad idea 😂

nathanlesage:
> For instance, we could download and install vscode in the background, and trigger it to open the directories to visualize our Zettelkästen.

"Hello, thanks for downloading this app. Would you mind if we just downloaded a bunch of other apps in the background and use your disk space?" 🙃 I think that's the definition of malware hehehe

carregal:
Happy New Year! As this thread apparently keeps getting traction, I’d like to throw another 2c at it :)

What about using the built in Mermaid parser to display graphs?

Maybe in a separate tab or maybe as some form of macro expansion that would generate the necessary mermaid code and insert it on the current cursor position?

Maybe offer some alternatives like “Insert graph with notes linked from this one” or “Insert graph with notes linking to this one”?

This could end up being simpler than embedding another graph visualization module and could allow those interested in graphs to experiment with this protoviewer and check what would be the next step or if it would be good enough.

I have been experimenting with such scenarios by generating the mermaid code by hand and would certainly appreciate some automation :)

StoltHD:
This would be a great feature for my research project...
To be able to visualize all the links I have in my all my notes would be so helpful when I'm starting the job of combining them to one large writing project.

But I really understand your concern about maintenance problem.

StoltHD:
How about using cytoscape.js for this? 
https://js.cytoscape.org/
https://github.com/cytoscape/cytoscape.js

I'm not a developer, but out of what I can read and understand, it's not a really huge library, there is a Vue.js component for it, can be installed to node.js.

And most of all, it is very advanced, has many configuration options for those who need it, and it is actively developed...

Another benefit would most likely be that it would be easy to export data to any network graph or knowledge graph system if needed... For those that use zettlr for research writing and don't wish to learn R or Python...

Just a suggestion that might be in the scope of both usage and maintenance, since it's a stable library developed by the Cytoscape team...

cdaven:
I have been working on a tool that analyses notes and wikilinks between them, and while it doesn't give you the information within Zettlr itself, you can actually redirect the output of the tool to Markdown files, to read with Zettlr. Or just read on the command-line.

No visual output for now, sorry!

The tool is called [NoteExplorer](https://github.com/cdaven/noteexplorer), and among other things, you can list these kinds of notes:

- **Isolated** notes; notes with no incoming or outgoing links
- **Sinks**; notes with no outgoing links, but at least one incoming link
- **Sources**; notes with no incoming links, but at least one outgoing link

It also lets you find broken links and insert backlinks into your notes, if you wish.

Since this isn't at all a part of Zettlr, any questions or comments about NoteExplorer should be handled on its own GitHub pages, and not here.

StoltHD:
Emile, the developer of the Obsidian Advanced Graph told me if someone want to implement his project into any other project he would help where he could, and his project is active...

It's a really great graph project, with lot of really helpful features planned, including a inline block code function for graphs with cytoscape.js (already working...
He will also add a overlay for using neo4j as a backend graph engine for advanced users, the first version of that worked well, but he want to rewrite it after he has created the addon he is working on now, without database backend

If it is of any interest, check out his project: https://github.com/HEmile/obsidian-neo4j-graph-view

Only thing you will be responsible for is the plugin integration, not the plugin it self... just a tips for something that actually work...

tarahmarie:
One notable contribution to this debate (which is incredibly important to me as well, since if there was a graph feature in Zettlr I could opt entirely out of Obsidian), is that in the roadmap just published by @nathanlesage for V.2 (https://github.com/Zettlr/Zettlr/issues/1690) there is no mention of visualization or graphs. For me, that is the one missing piece of Zettlr to make backlinks functional.

nathanlesage:
@tarahmarie Right now it's not on the roadmap because there's just too much to do, and I want the new version to be stable before engaging in such in endeavour.

But nevertheless, seeing how many people *still* – after almost a year of me saying "No, no, no!" – still want such a feature, it's getting harder and harder to say "no".

So here's a small ray of light for y'all: I will definitely integrate links deeper into the file descriptors, and once that works, I'd be willing to give visualisations a shot and see how such a network would fit into Zettlr's user interface.

source/renderer/dialog/note-network.js
```diff
@@ -0,0 +1,220 @@
...
+        'title': a[1].name ? a[1].name.replace(a[0], '') : 'Undefined',
+        'inbound': a[1].inbound ? a[1].inbound.length : 0,
+        'outbound': a[1].outbound ? a[1].outbound.length : 0
+      })
+    }
+
+    this.graph = {
+      'nodes': tmpNodes,
+      'links': tmpLinks
+    }
```
djalilhebal, line 54: In the "Add nodes" part, you're calling `.attr('r', function (d) { return 1 + d.inbound + d.outbound })`.
When I first read it, I was like, "Aren't `inbound` and `outbound` arrays? Wouldn't calling this code return incorrect results like the following snippet?"
```js
(1 + [2, 3] + [4, 5]) === '12,34,5'
```

But it seems like `LinkProvider's Node/Link` objects are different from NoteNetwork's. In this case, I suggest using different names for them... by "them," I mean the types/objects, or at least rename the attributes from `inbound` to `inboundCount` and `outbound` to `outboundCount`.

source/main/providers/link-provider.js
```diff
@@ -0,0 +1,266 @@
...
+ */
+
+/**
+ * This class manages note's relations on the app. It reads the links on each
+ * start of the app and writes them after they have been changed.
+ */
+class LinkProvider {
+  /**
+   * Create the instance on program start and initially load the links.
+   * @param {FSALCache} cache a cache to store links
```
djalilhebal, line 22: The constructor accepts no `cache` param.

source/renderer/dialog/note-network.js
```diff
@@ -0,0 +1,220 @@
...
+    this._dialog = 'note-network'
+    this.graph = []
+  }
+
+  preInit (data) {
+    var tmpLinks = []
+    var tmpNodes = []
+
+    // Building link list
+    for (const a of Object.entries(data)) {
```
djalilhebal, line 31: Maybe consider renaming `a` to `idNodePair` and `i` to `targetNodeId` (if I got these correctly, that is).

source/main/providers/link-provider.js
```diff
@@ -0,0 +1,266 @@
...
+    this._links = []
+    // The global link database; it contains all links that are used in any of the
+    // files.
+    this._globalLinkDatabase = Object.create(null)
+
+    // Register a global helper for the link database
+    global.links = {
+      /**
+       * Adds an array of links to the database
+       * @param  {Array} linkArray An array containing the links to be added
```
djalilhebal, line 36: Aside: As a general suggestion about JSDoc comments, I believe it would be better if you could use generic types (e.g. `Array<String>`) instead of the non-generic version (e.g. `Array`). This would help others understand the code and (I guess) make the editor provide better hints and type-checking.

This is not really important as Zettlr is migrating to TypeScript anyways.

djalilhebal (changes requested):
So I've checked the last commit and reviewed/suggested some changes to some parts of the code.

***Disclaimer: I have not studied the whole codebase, and I have not tested your pull request (neither with nor without my suggestions).*** _Oh, and I'd like to mention that I've been using Zettlr for less than a week now (and liking it, TBH)._

djalilhebal:
Work on this pull request makes my comment mostly useless, but I'm post this comment anyways:

---

Before I knew of this project, I was considering forking the VSCode extension that @pcuci mentioned (to make it a standalone Node package).

These "factlets" may interest you:

- You're already watching the filesystem and declaring `md5` as a dependency, as does that extension.
- That extension relies on manipulating/traversing ASTs of Markdown files.  Not sure if `showdown` lets you do that.
- Their code is not really VSCode-dependent (save some helper methods provided by the editor to watch folders and communicating between windows). It should be easy to port to any Electron-based app.
- They use a minified version of D3 Version 4.13.0., which is 217 KB. (Just saying, their solution is not dissimilar to this pull requests.)
- If you are still considering using Mermaid (which you probably shouldn't) to output a (static, non-interactive) graph then you may want to know of their `getDot` function:

From [@tchayen/markdown-links/src/utils.ts](https://github.com/tchayen/markdown-links/blob/master/src/utils.ts#L110):  
(This function seems to be unused. It returns the graph in [the DOT language](https://graphviz.org/doc/info/lang.html).)
```js
export const getDot = (graph: Graph) => `digraph g {
  ${graph.nodes
    .map((node) => `  ${node.id} [label="${node.label}"];`)
    .join("\n")}
  ${graph.edges.map((edge) => `  ${edge.source} -> ${edge.target}`).join("\n")}
  }`;
```

You could do something similar to the following in Mermaid:  
(Although the result will get ugly as soon as many interconnected pages are created. This is because [it supports only a _Hierarchical_ and no _Organic_ layout algorithm](https://github.com/mermaid-js/mermaid/issues/867).)
```js
export const getMermaid = (graph: Graph) => `graph TB {
  ${graph.nodes
    .map((node) => `  ${node.id}(${node.label})`)
    .join("\n")}
  ${graph.edges.map((edge) => `  ${edge.source} --> ${edge.target}`).join("\n")}
  }`;

// TODO: Manually replace bidirectional arrows with one undirected arrow.
//       Like, if `A --> B` and `B --> A` exist, output only one `A --- B`.
/*
const getCorrectLink = (nodeAId, nodeBId) => {
    const existsEdgeBA = graph.edges.some(e => e.source === nodeBId && e.source.target === nodeAId)
    const shouldOutput = (!existsEdgeBA) || (existsEdgeBA && nodeAId > nodeBId)
    const arrowType    = existsEdgeBA ? '---' : '-->'
    return shouldOutput ? `  ${nodeAId} ${arrowType} ${nodeBId}` : ''
}
*/
```

mermaid-diagram-20210314200000

StoltHD:
@nathanlesage - Only thing I did was to ask Emile because his addon is something that will be active for a while, and it has a lot of customization options AND use Cytoscape.js that is under active development and is used by many projects...

What you or any others want to do with that information, that's entirely up to you.

UnoYakshi:
Any updates on this feature?

StoltHD:
> Work on this pull request makes my comment mostly useless, but imma post the following anyways:
> 
> Before I knew of this project, I was considering forking the VSCode extension that @pcuci mentioned (to make it a standalone Node package).

Maybe you should look at the work Emilie are doing for his addon Juggl for Obsidian, He is also thinking about making it more useful for other Editors by splitting the code of the graph from the Obsidian depended code...

I think He would be really happy for any real help he can get, I have linked to his github repo just above your last comment, and he can be reached on his discord channel.
He use cytoscape.js for the algorithms, and also have a way of saving the generated graph (really helpful to analyze in the cytoscape dektop version or in any other graph software with more feature and that is more optimized for rendering speed ...

If I don''t remember wrong, he also have a feature that makes it possible to create inline graphs using a code block.
Cytoscape.js also support a lot of network and layout algorithms through "addons", many of those is open source.

Personally I think it would be of great help for many users, even writers and maybe specially writers of technical documentation to have advanced features like this...

Just a tips

jchalifour:
> @djalilhebal
> Maybe you should look at the work Emilie are doing for his addon Juggl for Obsidian, He is also thinking about making it more useful for other Editors by splitting the code of the graph from the Obsidian depended code...
> 

I just came across Juggl and was quite impressed with what it's doing. I came here to mention it in case it was of use. Its developer wrote that although it's currently implemented for Obsidian, it doesn't rely on that and could be ported. I saw your comment also mentioning Juggl, so I'll just add a link https://juggl.io/Juggl -- who knows, it might be an interesting combination someday.

lduktus:
I am very late to the party, but I wanted to at least share my thougts that I had while reading through this issues.
First of all I don't want to advocate against this feature in general, but I think the question raised by @nathanlesage is crucial. Maybe as a Disclaimer I haven't used Zettlr for a large period of time now and I may have a very opionated (and sometimes ideological) view on software:

Zettlr is Open Source and Free, which makes it already better than tools à la Obsidian. Another aspect is that it respects sane Markdown defaults and integrating pandoc is a very sensible choice.

It also does a really good job in making a plaintext workflow more accessible, not only by it's design and features, but also in the way it seeks to provide infomation (the docs, youtube videos, the community) and encourage users to contribute.
Therefore it helps in democratizing plaintext workflows, which is the main reason I decided to write this comment. I have used a lot of tools in last years (Emacs, [Neo-]Vim, VSCode, Atom), but most of these tools are not thought for humanities or writing in first place. And while they do the job it requires some work and knowledge to get a nice workflow going. Especially if your needs tend more towards advanced citations, footnotes, etc. Until I discovered Zettlr there were no tools I would have suggested to colleagues and friends, as most people in humanities have no experience with editors.

I don't think that the crucial question is if Zettlr is either for notetaking or a replacement of Office software. Most people who regularly write may like the way to integrate there notes in their writing environment (btw I know a lot of people who use some kind of Office software for note taking). Imo the question is if Zettlr should have all features of a notetaking app, that aims to be this and nothing more. Imo Zettlr is a nice writing environment, that gives you some quality of life features, that are very valuable when you write longer texts in fields like humanities or even for journalists. As the basic idea of using Markdown and plaintext approches in general is independence from specific platforms and tools I don't think that Zettlr has to ship all features of apps that just do another job [check also this reddit pos](https://www.reddit.com/r/Zettlr/comments/mlmyjv/dendron_vs_zettlr_as_a_zettelkasten_system/).

Imo there is no problem in using various tools, scripts or maybe plugins (I read in antoher issue that there are no plans to add plugin support for now, however I think something like this may be an ideal example for something which could be implemented as a plugin and not as a core feature) to  provide some features. I know there is the tendency to add as much nice features as possible to a program (I am looking at you Emacs), however I am not sure if this is always a good idea. While there may always be a debate if something deserves to be a 'core' feature or not it could clearly help to have a clear definition of the long term goals.

My comment may be somehow off topic and maybe there is another plattform which is more suitable for this type of question.
Anyway my personal opionion is that graph visualization is not necessarily something Zettlr should provide as a central functionality.

jchalifour:
@lduktus well put. I would like to add (augment?) the following.

On the issue of note-taking application and writing tool, I believe that Zettlr's emphasis on a nice writing environment is precisely what makes its dual-purpose as a note-taking tool so valuable. The more it excels for taking, managing, and developing notes, the better I can convert those to my finalized writing. Maybe I'm just restating what's been said though. It's just that I think what is best about Zettlr is how it enables those processes to be wed together.

Those graphical/visualization tools are super interesting and might prove valuable in various ways. At the moment, I personally have absolutely no use for them--mostly a curious gimmick to me. Nevertheless, I like to follow what other people are doing with them because I suspect that there might eventually be some really good use that I just am not aware of yet.

tarahmarie:
I don't think this issue has lost any of its relevance; we're just waiting for the additions mentioned above.

tarahmarie:
Is it someone else's turn to note that this issue is not stale, just awaiting the fixes above? :-)

nathanlesage:
@tarahmarie It's now pinned, so stalebot should leave it alone.

nathanlesage:
Still not at the top of the list – prior to this are some nasty bugs, the transition to Vue3 and TypeScript, and a full clean up of the code base. Currently estimating that I can actually get to this issue sometime in Spring 2022, provided that I can take a week off from work to get to the app

almereyda:
Until then, you can try a manual visualisation of your Zettelkasten, e.g. with the help of https://github.com/mickael-menu/zk and the gist linked from https://github.com/mickael-menu/zk/issues/48#issuecomment-871457287

nathanlesage:
Since @JulienMir has unfortunately not been active on this PR for quite a while, this PR has now been superseded by 4c033817, which adds a graph view, building a little bit on the work he has done, but with some improvements with regard to data handling.

A graph view is now part of Zettlr.

Issue #859, opened 2020-05-31
[DISCUSSION/FEATURE/REFACTOR] Decentralize IPC modules


nathanlesage:
As I mentioned in the past, after the FSAL rewrite has been done, it's time to prepare the next big refactoring process. I personally think that after that one, the app is decentralized and modularized to a degree that scalability is once again reached sufficiently. (If I'm wrong and there are other modules that are difficult to scale, please feel free to have a look!)

## What's this about?

Basically, it concerns a full rewrite of the inter process communication logic, which is currently heavily centralized in two main classes: `zettlr-ipc.js` and `zettlr-rendereripc.js`. There are multiple problems with this.

One: When I began writing the app, I did not anticipate it would at some point include multiple windows, so the idea was to rout all communication to interface-like class objects that could do the heavy lifting so that from throughout both renderer and main process one could simply send messages and be sure they'd land on the other end. However, there are two problems:

1. Every message that arrives FROM the counterpart needs to be fed through the main object in each process, so even if a message is needed in the editor alone, it'll need to pass through the renderer object. The same holds true for the main process: Even if an object solely concerns the citations, it'll need to pass through the main process handler.
2. Additionally, this dependence on one specific class has led to a curious behaviour I recently noticed when I switched the behaviour of the `Close Window` menu item to `Close Tab, and close window if there are no more tabs`, as — no matter which window was actually focused — the message would always be deferred to the main window, where it would close a tab — even if your intention was to close the log window.

## How should we do it?

So here's my idea of what we might do, where your input is warmly welcome:

1. Create two providers (i.e. still have one object, but that object shall be completely agnostic to what is being passed to it), which'll mostly work as the current ones — with two specific changes: First, it will only take `on`-listeners, and diseminate the events to these without interfering, instead of trying to handle them itself, and second, it will allow either to target "the current focused" window or a specific one (at least in the main process, the renderer IPC calls will always be pointing towards the one single main process)
2. Create a sensible model of what kinds of channels we need to patch data through
3. During implementation, get rid of all of these multiple-messages being sent

### Communication Channels

So here's what I think we need concerning channels (if you are new to IPC, channels equal the different events an EventEmitter can emit, so we can think of any string we like)

* `command` (main only) --> Will reach the main application which will try to find commands that respond to the given command
* `citeproc` --> Will only facilitate communication between the renderer's editor and the sidebar, e.g. something encapsulated in the `citeproc-provider.js`
[...]

PR #927, opened 2020-06-15
[WIP] Custom keymaps

AlexandreDubray:
## Description

The goal of this PR is to provide to the user a way to customize the keymaps. The idea is to provide a general interface to change the bindings for the general behaviour of Zettlr/CodeMirror as well as for the new modes introduced in #683 .

## Changes

At the moment the keymaps are handled in a similar way as the preferences. There is a provider that is responsible for creating, setting and saving the mapping. Since these changes are relatively independent (the main interactions with the rest of the code is just message handling) it should not break anything.

## Additional information

There is still some work to do on this in order to be viable. This PR will also serve as a backlog on what still need to be done (I will keep adding things as I progress in the code or receive suggestions).

Todo:
- [ ] Define a list of assignable function
- [ ] Check that the mapping is consistent and correct
- [ ] Enhance the UI (simple table at the moment, with text input. Not practical)
- [ ] handle vim keymap
- [ ] handle Emac keymap (not sure how it's done since I do not use it)
- [ ] Test the whole thing (: 

Maybe we can open a discussion issue for people who might have ideas they want to discuss?

Tested on:
OS: Archlinux, kernel 5.6.15
branch: dev


nathanlesage:
Thanks for opening this! Let me just attempt to uncover the intricacies and quirks of how the code currently works; because maybe this helps spot stupidities from me and also prevent overlooking something!

> Define a list of assignable functions

For this, a look at the shortcuts may help. User-run actions are stored in `source/main/commands` (most of them require some user-input which is asked for in the `ZettlrBody`-class which subsequently sends the provided data to main to actually run the command). Further actions are stored in the CodeMirror addon files for the SublimeText-addon (some actions are already used by Zettlr, others not --> see the `generate-keymap.js`-file). The same applies to the vim and emacs-mode, I suppose, albeit I am genuinely frightened by the size of the vim mode source code.

> Check that the mapping is consistent and correct

Given that the mapping is (currently) not even consistent in Zettlr itself, here's where things are defined currently:

1. Most shortcuts are defined by the menu (as soon as you assign an `accelerator` to a menu item, it basically becomes a application-wide shortcut), the rest by CodeMirror (built-in shortcuts as defined in the default/vim/emacs keymap, plus then overrides provided for by the `generate-keymap.js`-file PLUS overrides in some of the plugins — don't ask me why I thought this would be a cool idea)
2. Then, a few shortcuts are defined hidden away in the monstrosity that calls itself constructor of the `ZettlrEditor`-class (mostly visible by the code asking for the state of the `metaKey`, `ctrlKey` or `altKey`, e.g.)
3. Then, more shortcuts are defined in the `ZettlrBody`-class (such as focussing of either the editor or the file list)
4. Also `ZettlrToolbar` defines shortcuts
5. Not to mention the vue-plugins (in `resources/vue`)

It's a mess, I know, but I guess this happens with "organic" growth …

***

With regard to the shortcuts, at least concerning CodeMirror, how this *mostly* works (I am very sure some plugins don't respect that either) is:

1. A plugin defines some function
2. The plugin registers this as a CodeMirror "command"
3. Either the plugin (in the plugin file), CodeMirror itself (in a keymap) or Zettlr (in the `extraKeys` generated by said function) provides a binding, that is: map a key to the function

So I think what might help would be to take the `generate-keymap.js`-function and simply map EVERYTHING that there is to it (imagine sinister laughing here), with un-assigned functions being set to `false` (which equals "Don't trigger anything for this" in CodeMirror-speak).

The question would be: Do you also want to make the globals overrideable? Because then this would require changes to the menu generator (which I'll torture in the coming days myself)

Then just one thing concerning the setter `set (fun, key)` — The fun sounds like a pun =D 

Also I see that you're making use of the IPC classes. I am torn with this, because as to issue #859 we need to completely tear them apart, and want to decentralize, so the interesting question would be: Do we already begin decentralizing it by creating the channel `keymap` (or `keybinds`???) here or wait for the "full" rewrite? @tobiasdiez is currently on vacation, but he wanted to take care of this.

AlexandreDubray:
Ok, I see. Indeed, it looks like it's a big mess^^ But it's fine, we can figure something out.
I think too that we should have a single file handling all the bindings (regardless of wether they apply on the editor or the menus).
At the moment I handle this in the `keymaps-provider.js` file (I copied what I saw for the configurations). You think it is better to code this in the `generate-keymaps.js` file?

Concerning the IPC classes, I saw that you wanted to rework them. For the moment, it might be better to continue as it is now. I fell that it will be easier for me to adapt my code after the rework than to have something mid-way between centralized and decentralized.

nathanlesage:
> single file handling all the bindings (regardless of wether they apply on the editor or the menus)

Nah, that won't be possible (because the menus are instantiated in the main process, while the editor lives in the sandboxed renderer), so two files are the least, I think … or do you mean like keeping the associations in the main process and have the renderer request them and apply them?

> You think it is better to code this in the generate-keymaps.js file?

Mom, let me have a look … Well, it's basically the same, but I couldn't think of anything better, so sure, it's fine!

> I fell that it will be easier for me to adapt my code after the rework than to have something mid-way between centralized and decentralized.

Alright, I had the same thought, perfect!

AlexandreDubray:
> or do you mean like keeping the associations in the main process and have the renderer request them and apply them?

Yep something like that. Is it technically possible? (or is it ok with the vision you have of the codebase :p)
I think that for scalability and maintainability it is better to have only one place too look at. 
So typically eveything related to defining/checking the keymaps would be in `keymaps-provider.js` (for example) and if the renderer needs the keymaps, they would request it by sending a message.

I feel that it would ease the understanding of the code.

nathanlesage:
> I think that for scalability and maintainability it is better to have only one place too look at.

Yes!

> So typically eveything related to defining/checking the keymaps would be in keymaps-provider.js (for example) and if the renderer needs the keymaps, they would request it by sending a message.

This sounds good! Then we could even implement some easy cascades that would notify everything that wants to know. I think that's how you've currently implemented that, right?

AlexandreDubray:
Yes that was my idea (even though I did not know that the bindings were scattered all over the code!^^).
But as I discover the codebase and the Electron framework at the same time, I might still have some
things that need to be tune, especially with the IPC and how the changes are communicated:-)

nathanlesage:
Yes, of course; take your time! :)

AlexandreDubray:
Ok so I think I got the basic layout for the custom keymaps.
There is still some things about which I'm not quite happy in the code (for example the fact that the keybindings are define by strings, which leads to ugly code since some componenet seems to *need* bindings define with `-` separator and other `+` separator), but the core of the code is there I guess.

Since @nathanlesage is quite busy at the moment, maybe @tobiasdiez could you give a quick look at this and see if the logic seems good?

IpeTek:
I'm so grateful to all of you having worked on this! Looking forward to the functionality.

tkna91:
@AlexandreDubray Thanks for all the work you put into so many of them.
I am hoping to be able to customize keyboard shotcuts on Zettlr, but it seems quite difficult?

faviobecker:
- Also interested in this.

Issue #915, opened 2020-06-10
[BUG] PDF export fails when image scaling parameter is provided


obehrto:
## Description
PDF export fails when an image scaling parameter such as width=25% is provided. Other exports such as docx and HTML work and produce the expected image output. The error message is:

`Error producing PDF.
! Argument of \Gin@iii has an extra }.
<inserted text> 
                \par 
l.299 }`

## Reproducing
Include an image as, for example, `![](image.png){width=25%}` and generate PDF.

## Expected behaviour
Generate PDF with expected image scaling.

## Platform
 - OS and version: Windows 10
 - Zettlr Version: 1.70-beta.9

nathanlesage:
Would you mind testing this out with the [default Pandoc template](https://github.com/jgm/pandoc/blob/master/data/templates/default.latex)?

obehrto:
Yes, when using the default.tex template it works perfectly and the scaling is correct. I am also using `--filter pandoc-xnos` if that is of interest.

nathanlesage:
> es, when using the default.tex template it works perfectly and the scaling is correct.

That's good; means I have to update Zettlr's default template. If anybody wants this to be quicker, feel free to research which changes need to be done and feel free to submit a PR!

nathanlesage:
Fixed in PR #928

PR #928, opened 2020-06-15
Fix issue #915

obehrto:
Added image scaling from pandoc default latex template. Fixes issue #915

## Description
Replaces the image scaling code with code from pandoc default.latex

## Changes
Replaced image scaling with code from pandoc default.latex

## Additional information

Windows 10

### DIFF
--- a/source/main/assets/export.tex
+++ b/source/main/assets/export.tex
@@ -47,23 +47,15 @@ $endif$
 % _italics_ will become underlines, which is not desirable.
 \usepackage[normalem]{ulem}
 
-% Pandoc uses includegraphics. But as some images may be too big, we have to
-% replace the command with another one that ensures proper scaling of images.
-% Therefore first save the old includegraphics command somewhere else.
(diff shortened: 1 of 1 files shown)


nathanlesage:
Thank you! Quick question: My original comment said that I apparently replaced the default command because very big images overflowed on the template — did you test if super large images overflow with the pandoc default, or does it work out of the box now?

obehrto:
Yes, the overflow prevention is retained. The behavior is now as it should be. If no scaling parameter is provided oversize images are scaled down to 100% column width but if they are even larger than the text area on a page are also scaled vertically to prevent vertical overflow. These automatic maximum values can be overwritten by providing both horizontal and/or vertical scaling parameters such as `{width=XX}` or `{height=XX}` or `{width=XX height=XX}` in Zettlr.

nathanlesage:
Perfect! Thank you very much for clarifying it! :) Aaaaaand merge!

PR #938, opened 2020-06-16
WIP: Use electron-webpack for building the app

Aigeruth:
## Description

**This is a proof of concept and work in progress PR to use electron-webpack to build both Electron main and renderer bundles.**

**WARNING**: As this change is still WIP, some parts of the application is still broken do the changes in how packages and files referenced.

Intended outcome of this change is to reduce the overall release size. I built the application on the develop branch and on this branch as well:

```bash
$ du -sh release-develop/mac/Zettlr.app
275M    release-develop/mac/Zettlr.app
$ du -sh release/mac/Zettlr.app        
239M    release/mac/Zettlr.app
```

*Note*: as this is a proof of concept change, it only inlines the mermaid-js dependency.

Related to Zettlr/Zettlr#921

## Changes
This changes introduced bundles for Jascript and CSS for both Electron main and renderer.

## Additional information

Tested on: macOS


package.json
```diff
@@ -26,15 +26,22 @@
...
+  "electronWebpack": {
+    "main": {
+      "sourceDirectory": "source/main"
+    },
+    "renderer": {
+      "sourceDirectory": "source/renderer",
+      "template": "source/renderer/assets/index.htm"
+    },
+    "whiteListedModules": [
+      "mermaid"
```
Aigeruth, line 42: This line is responsible for inlining dependencies that makes possible to reduce the shipped binary size.

package.json
```diff
@@ -59,9 +66,11 @@
       }
     ],
     "directories": {
-      "output": "release",
-      "app": "source"
+      "output": "release"
     },
+    "files": [
+      "!**/node_modules/mermaid${/*}"
```
Aigeruth, line 72: This line is responsible for excluding the mermaid-js related files from the released version.

Eventually the goal should be to exclude the whole `node_modules` directory.

source/main/index.js
```diff
@@ -43,13 +43,19 @@ const { app } = require('electron')
...
-const Zettlr = require('./main/zettlr.js')
+const Zettlr = require('./zettlr.js')
 // Helper function to extract files to open from process.argv
-const extractFilesFromArgv = require('./common/util/extract-files-from-argv')
+const extractFilesFromArgv = require('../common/util/extract-files-from-argv')
 
 // Introduce v8 code caching
 require('v8-compile-cache')
 
+if (module.hot) {
```
Aigeruth, line 53: This makes possible to use hot-reloading during development.

source/main/index.js
```diff
@@ -62,18 +68,10 @@ let zettlr
...
- * @type {Boolean}
- */
-let isFirstInstance = app.requestSingleInstanceLock()
-
-/**
- * Exit immediately if this is a second instance of Zettlr.
- * @param  {Boolean} isFirstInstance Whether or not this is a second instance.
- */
-if (!isFirstInstance) app.exit(0)
+if (app !== undefined && !app.requestSingleInstanceLock()) {
```
Aigeruth, line 71: Probably unnecessary change in the final version.

source/main/index.js
```diff
@@ -123,20 +121,8 @@ app.on('open-file', (e, p) => {
  * else. It is necessary to wait for the ready event, because prior, some APIs
  * may not work correctly.
  */
-app.whenReady().then(() => {
+app.on('ready', () => {
   global.log.info('Electron reports ready state. Instantiating main process...')
-
-  try {
```
Aigeruth, line 129: `electron-webpack-vue` takes care of this.

source/main/zettlr-window.js
```diff
@@ -102,15 +105,20 @@ class ZettlrWindow {
...
     this._win = new BrowserWindow(winConf)
-
-    // Then activate listeners.
-    // and load the index.html of the app.
-    this._win.loadURL(url.format({
-      pathname: path.join(__dirname, '../renderer/assets/index.htm'),
-      protocol: 'file:',
-      slashes: true
-    }))
+    if (isDevelopment) {
```
Aigeruth, line 110: This is for hot-reloading during development.

source/renderer/assets/index.htm
```diff
@@ -4,39 +4,8 @@
     <meta charset="UTF-8">
     <!-- Allow script execution only from the app itself or inline stuff to
          supress the Electron default warning (and also make it safer here) -->
-    <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-inline'">
+    <meta http-equiv="Content-Security-Policy" content="script-src 'self' 'unsafe-inline' 'unsafe-eval'">
     <title>Zettlr</title>
-
```
Aigeruth, line 9: Everything is bundled by webpack now.

nathanlesage:
Hold a moment. So do I correctly understand it, that the overall aim of your PR is to basically trash all of `node_modules` unnecessary stuff (e.g. when developers commit the whole documentation to NPM instead of removing it beforehand, etc.)?

Or does it also include a rewrite of the application logic?

I am basically currently averse to minimizing the shipped code, because it increases transparency of the code (+ due to module caching Zettlr performs whitespaces do not have any impact on the runtime). I am, however in favor of an approach that solely cleans up the node_modules.

Furthermore, one additional question would be: Electron-builder apparently has removed the necessity to have a two-package-structure, which would be great, but I haven't had the time to test that, but Zettlr needs to switch to that too — are there possible caveats with electron webpack?

So much for the first overview over this issue! I'm happy to see how this progresses! :)

Aigeruth:
This change does not impact application logic at all. It modifies some of the import paths but, it does not require change in the logic.

The `app.asar` is currently 100+MB, because it contains most of the `node_models` directory. (Electron build is kind enough to remove development dependencies by default). This causes problems, because e.g. mermaid-js ships 3 different builds plus map files, but Zettlr only uses one of them.

```
$ du -h source/node_modules/mermaid/dist/*
 16K	source/node_modules/mermaid/dist/index.html
408K	source/node_modules/mermaid/dist/mermaid.core.js
688K	source/node_modules/mermaid/dist/mermaid.core.js.map
4.4M	source/node_modules/mermaid/dist/mermaid.js
4.8M	source/node_modules/mermaid/dist/mermaid.js.map
1.1M	source/node_modules/mermaid/dist/mermaid.min.js
5.1M	source/node_modules/mermaid/dist/mermaid.min.js.map
```

Removing the complete `node_modules` directory can be a longer term goal, but it does not have to happen in a single PR, however, the application has to be switched to `electron-webpack` in one step and that results in a substantial change.

> Furthermore, one additional question would be: Electron-builder apparently has removed the necessity to have a two-package-structure, which would be great, but I haven't had the time to test that, but Zettlr needs to switch to that too — are there possible caveats with electron webpack?

I have removed the `source/package.json` as part of this change.

> I am basically currently averse to minimizing the shipped code, it increases transparency of the code

I'm not sure what you mean by this. All source are bundled together in the `app.asar` and it has to be manually extracted for inspection. I find much more convenient to browse the code on GitHub or in Visual Studio Code than extracting it by:

```
npx asar extract /Applications/Zettlr.app/Contents/Resources/app.asar asar-unpack
```

It is also possible to ship a `.js.map` file, so the code does not look minified when the Developer Tools are open.

Using a webpack pipeline has other benefits like using [tree shaking](https://webpack.js.org/guides/tree-shaking/) for eliminating unnecessary code, reducing memory usage, etc. It also enabled better hot reloading during development, and potential usage of TypeScript for static type-checking. You might have already noticed that it also helped eliminating the global `$`, which makes easier to follow the code.

nathanlesage:
Ah, I see what you mean. 

> I find much more convenient to browse the code on GitHub

Yes, but I'm thinking of the edge cases here where people want to look up something quickly and may be offline. But that's subjective, of course.

> It is also possible to ship a .js.map file, so the code does not look minified when the Developer Tools are open.

That would be a necessity, yes. This project heavily relies on users being able to look up some things quickly without having to build it locally.

> You might have already noticed that it also helped eliminating the global $, which makes easier to follow the code.

N.B. to that: I want to get rid of jQuery completely asap, you might have noticed that I started refactoring parts of the code to use vanilla JS.

Aigeruth:
> I want to get rid of jQuery completely asap,  you might have noticed that I started refactoring parts of the code to use vanilla JS.

I did not look into the git history to notice this, I only noticed that the sidebar uses Vue.js. Do you intend to replace jQuery + handlebars?

What time frame do you have in mind for migrating away from jQuery? I think, I have replaced at least ~20 references (files) to `$`. This seems like a significant project to me.

nathanlesage:
> What time frame do you have in mind for migrating away from jQuery?

None, at the moment. The very far-far-away plan is to fully migrate the renderer to vue.js and the main process to TypeScript, but this is a significant endeavour, which is why I want to reserve this for maybe next year. It'll be heavy, but until then, the thing I want to achieve is at least get rid of jQuery, which is a huge project, yes, but the good thing is that we can do this iterative. The main point is that I realized at some time that we don't need jQuery because we need to support only one browser at a time either way.

package.json
```diff
@@ -26,15 +26,22 @@
     "lang:refresh": "node scripts/refresh-language.js",
     "csl:refresh": "node scripts/update-csl.js",
     "reveal:build": "node scripts/update-reveal.js",
-    "wp:dev": "cross-env NODE_ENV=development webpack --progress --colors",
-    "wp:prod": "cross-env NODE_ENV=production webpack --progress --colors",
-    "watch:gulp": "gulp watch",
```
tobiasdiez, line 31: I think the gulpfile (as well as a few related dev dependencies) can be removed as well, right?

Is electron-webpack taking care of everything done in these scripts? E.g. linting the files etc

source/main/index.js
```diff
@@ -123,20 +121,8 @@ app.on('open-file', (e, p) => {
  * else. It is necessary to wait for the ready event, because prior, some APIs
  * may not work correctly.
  */
-app.whenReady().then(() => {
+app.on('ready', () => {
```
tobiasdiez, line 124: The whenReady() method is (apparently) the preferred way: https://github.com/electron/electron/pull/21972

package.json
Aigeruth: Yes, you are right. I had this it removed locally, I just haven't updated it. I pushed it now.

source/main/index.js
nathanlesage: I already changed these things in upstream, so a git pull would resolve this in a better way, I'd suggest!

Aigeruth: I have rebased my change on the latest develop branch.

nathanlesage: Perfect! Short question, as I don't run into that problem: How difficult is rebasing and pulling in the latest changes? Is it easy, or bothersome for you?

Aigeruth: It is easy until there is a conflict, e.g. a file is modified on `develop` and my change also tries to change that file. At that point, it will require the same conflict resolution process as merging `develop` into my branch, but this keeps things cleaner.

nathanlesage: Yes, this makes sense. But in this regard, now we know that all branches are up to date with develop, which'll break fewer things, hopefully! Thanks for the info :)

nathanlesage:
Just saw that this wasn't marked as a draft; I hope I didn't break anything (if I did please scold me)

Aigeruth:
Closing this in favor of #1055.

PR #989, opened 2020-07-04
Update README

doup:
## Description

Update README instructions for setting the project the first time.

## Changes

Removed stale docs, Yarn and Node 12+ are **required**.

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/erbium
--- a/README.md
+++ b/README.md
@@ -69,19 +69,19 @@ If you have downloaded Zettlr, [head over to our website](https://docs.zettlr.co
 
 ## Contributing
 
-Zettlr is an [Electron](https://www.electronjs.org/)-based app, so to start developing, you'll need to have a [NodeJS](https://nodejs.org/)-stack on your computer installed. Make sure Node and preferably [Yarn](https://yarnpkg.com/en/) are installed, which is the recommended package manager.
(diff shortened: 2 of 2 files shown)


nathanlesage:
Thanks for opening the PR!

First, concerning `npm`, you are right; we had to drop npm-support from throughout the workflow files as well. However, as to your Node-requirement, here's [a snippet from the Electron docs](https://www.electronjs.org/docs/tutorial/development-environment#developer-environment):

> First, install a recent version of Node.js. **We recommend that you install either the latest LTS or Current version available**. Visit the Node.js download page and select the macOS Installer.

Hence, locking down Node.js to `erbium` using the `.nvmrc` is something I strongly disapprove, because people will assume this, and as I personally keep node updated at all times, I know that many people will rely upon the `.nvmrc` and might at some point wonder why their Node is apparently too old for the given environment b/c the `.nvmrc` is outdated.

Long story short: I'd suggest removing the `.nvmrc` and instead append to the README explanations the recommendation from the Electron docs.

nathanlesage:
Short update, I just read the [`nvm` docs](https://github.com/nvm-sh/nvm#nvmrc), and apparently we can create an `.nvmrc`-file that always keeps itself updated by always using the latest `lts`:

```bash
$ echo "node" > .nvmrc # to default to the latest version
```

The question would only remain: How many people actually rely on nvm? New config files tend to add chore workload, if I'm not mistaken? I have never come to the point where I needed two node versions installed …

nathanlesage:
@doup any updates on this?

Issue #786, opened 2020-05-13
[Enhancement] UX small  enhancements  request thread


danieltomasz:
Instead of creating every time new issues to the horror of @nathanlesage  I will put here inconsistencies I noticed or small things that could improve workflow with the app

### Design/UX update
1. [UX]  Rename file/note bubble is too narrow, it doesn't display the full name of the note

###  Enhancements
2. [Enhancement] it is possible to trigger rename bubble centred on a note in sidebar view? when using rename right-click option on the note

3. ~~[enhancement] Open plain https link with alt. In this moment Zettlr open links only when in Markdown tags [1.7.2]~~

6. ~~Add "Open directory [in file manager]"  option when right-clicking on the directory in the sidebar~~

7. When moving files between directories and files exist add an option to ask if overwrite (with info which file is later edited)

### General/ possible bugs
4. Zettlr doesn't remember the width of the sidebar when restarted, every time I need to resize it (Zettlr 1.7.2)

5. ~~"New virtual directory" still as an option when clicking on a folder in the sidebar. As I understood the possibility to create virtual dir has bee removed in this release? Choosing this option it doesn't do anything anyway~~

nathanlesage:
> Instead of creating every time new issues to the horror of @nathanlesage

❤️

danieltomasz:
5.  "New virtual directory" still as an option when clicking on a folder in the sidebar. As I understood the possibility to create virt dir has bee removed in this release?

nathanlesage:
someone also mentioned (while we're at UX) to enable dark styling of the dialogs. Just did that:

![image](https://user-images.githubusercontent.com/17251683/81818126-e833d780-952d-11ea-8274-7087fea1772a.png)

Virtual directory thingy is also gone in three, two …

danieltomasz:
When the Zettlr window size is smaller than The Preferences Pane, the preferences pane start apperaring in different part of the screen when changing different Preferences tab

<img width="992" alt="image" src="https://user-images.githubusercontent.com/7980381/83328578-47ab1a80-a284-11ea-800e-9f3dba23e1b7.png">
<img width="1001" alt="image" src="https://user-images.githubusercontent.com/7980381/83328585-57c2fa00-a284-11ea-8c3b-ee29a1b4e1a2.png">

<img width="992" alt="image" src="https://user-images.githubusercontent.com/7980381/83328614-8ccf4c80-a284-11ea-928f-063fd399e93e.png">

[...]

PR #990, opened 2020-07-04
Small UX improvements

doup:
## Description

- Increase top-toolbar search input `x` button hit-area #786 (`Search field improvements`)
- Increase editor header-tag & block toggle buttons hit-area #786 (`Improve section controls`)

### DIFF
--- a/resources/less/geometry/codemirror.less
+++ b/resources/less/geometry/codemirror.less
@@ -153,10 +153,15 @@
     border: none;
     left: 0 !important;
   }
-  .CodeMirror-foldgutter { width: .7em; }
+  .CodeMirror-foldgutter { width: 2em; }
   .CodeMirror-foldgutter-open, .CodeMirror-foldgutter-folded {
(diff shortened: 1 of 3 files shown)


doup:
I've done just a couple of them. If I'm in the mood I'll continue another day with some of the other issues. 😅

nathanlesage:
I'll be reviewing these as soon as I can! And better have multiple smaller PRs for these enhancements; mainly because I always have a high anxiety when it comes to adapting the CSS, because the source files are just a mess. But thanks already for the super good review of the UX over in the issue!

(if you feel like it, be sure that I'd send you a bouquet if you refactor these 💐 )

nathanlesage:
Almost perfect, just had the chance to test it out! The increased sizes do look better in fact!

image

Only I'm not really fond of the non-square-ish design of the new heading characters; previously they were more square, and I'd like to retain that somehow. … opinions?

PR #1048, opened 2020-07-11
Prototype of new IPC modules

tobiasdiez:
This is a prototype of an idea for the new IPC module. The IPC component is working, but a few other things are still broken. So for now this PR is to get feedback if I'm on the right track or if you thought about something completely different.

With these changes, the implementation of a new IPC module looks as follows (see the citeproc-provider):
- Create a new interface specifying the methods that are available: e.g `IpcCiteService`
- Create an implementation of this interface, and call `IpcModule.registerMain` to register it as a handler in the main process: e.g. `citeproc-provider`
- To call methods from the renderer, use `IpcModule.forRenderer` to get a wrapper implementing the interface (and which proxies all calls to the correct IPC methods), e.g. `IpcModule.forRenderer<IpcCiteService>().getCitation(...)`.

This has the advantage that one does not need any magic strings (i.e. if one changes the `IpcCiteService` interface one gets compile errors if one doesn't change the main and renderer implementation accordingly). Moreover, the IPC-specific implementation is completely hidden in the new class `IpcModule` and the rest of the code doesn't need to care about IPC at all.

Inspired by:
- https://github.com/vivimice/electron-ipcfy
- https://blog.logrocket.com/electron-ipc-response-request-architecture-with-typescript/
- https://foxypanda.me/clean-interprocess-communication-in-electorn/
- https://github.com/sindresorhus/electron-better-ipc
- https://dev.to/deiucanta/electron-typed-ipc-31ph
- https://vottv2.z22.web.core.windows.net/files/src_electron_common_ipcMainProxy_js/index.html

In order to make this work, I choose to use typescript (in order to get everything type-safe and use generics etc). This move required to change the other files in this PR (build script, handling of assets etc). But if I understood you correctly, you wanted to move to typescript anyway in the long run.

In addition, we could use something like https://github.com/kimamula/ts-transformer-keys to restrict the possible api functions that are registered via the context bridge.


nathanlesage:
> But if I understood you correctly, you wanted to move to typescript anyway in the long run.

Yes! I'm more and more growing angry over the non-typings of JavaScript (+ using TypeScript would allow proper module minimization, tree shaking, etc.)

Let me have a look …

.eslintignore
```diff
@@ -10,3 +10,6 @@ source/common/assets/tpl
 # Ignore jQuery UI and the copied indentlist-addon
 source/common/assets/js/jquery-ui.min.js
 source/renderer/assets/codemirror/indentlist.js
+
+# Ignore build output
+dist
```
nathanlesage, line 15: I would argue we can directly stuff the TypeScript output into the correct directories?

.gitignore
```diff
@@ -17,6 +17,9 @@ yarn.log
 # Don't commit ESLint error reports
 eslint_report.htm
 
+# Ignore built files
+dist/**/*
```
nathanlesage, line 21: Same as above? I'd like to have as less additional stuff to do as possible, so I'd rather have the build vomit everything into the right directory directly (we could create two directives for debug with types, and for release only with sourcemaps)

.vscode/launch.json
```diff
@@ -10,7 +10,7 @@
       "windows": {
         "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/electron.cmd"
       },
-      "args" : ["."],
+      "args" : ["./dist/main.js"],
```
nathanlesage, line 13: Would then need to be in some way adapted (e.g. `./source/main/modules/ipc/main.js`)

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "target": "es5",
+        "module": "commonjs",
+        "allowJs": true,
+        "strict": true,
+        "esModuleInterop": true,
+        "outDir": "dist",
```
nathanlesage, line 8: This line would need to be concerning the out-file (at least until we got everything where we need it)

source/main.ts
```diff
@@ -181,7 +185,7 @@ app.on('activate', function () {
  * Hook into the unhandledRejection-event to prevent nasty error messages when
  * a Promise is rejected somewhere.
  */
-process.on('unhandledRejection', (err) => {
+processInfo.on('unhandledRejection', (err) => {
```
nathanlesage, line 188: This doesn't work at all, I always wanted to do this properly, but haven't had the time :D

source/main.ts
```diff
@@ -40,7 +40,7 @@ global.log = {
 
 // We need the app and process modules.
 const { app } = require('electron')
-const process = require('process')
+const processInfo = require('process')
```
nathanlesage, line 43: Why did you rename this as processInfo?

source/IpcModule.ts
```diff
@@ -0,0 +1,57 @@
+import { ipcRenderer, ipcMain } from 'electron'
+
+type Promisified<T extends Function> =
```
nathanlesage, line 3: Just so I know: TypeScript doesn't support async/await …?

source/IpcModule.ts
```diff
@@ -0,0 +1,57 @@
...
+        [K in keyof D]: D[K] extends Function ? Promisified<D[K]> : never;
+    }
+
+export abstract class IpcModule {
+    public static forRenderer<T>(): Ipcfied<T> {
+        return new Proxy({}, {
+            get (target: any, prop: PropertyKey, receiver: any): any {
+                const methodName = prop.toString();
+                return async function(...args: any[]) {
+                    return ipcRenderer.invoke(methodName, args);
```
nathanlesage, line 22: Very good that you additionally switched to the sandboxing/invoking, as we'd need to fully sandbox the renderer either way

source/main/providers/citeproc-provider.ts
```diff
@@ -494,3 +456,9 @@ class CiteprocProvider {
 }
 
 module.exports = new CiteprocProvider()
+
+global.citeproc = module.exports
+
+// Listen for synchronous citation messages from the renderer
+// Citeproc calls (either single citation or a whole cluster)
+IpcModule.registerMain<IpcCiteService>(module.exports)
```
nathanlesage, line 464: Something regarding this implementation bugs me, but I can't really pinpoint it … maybe it's that it's very intransparent and not much verbose …

(+ here are all trailing newlines missing)

gulpfile.js
```diff
@@ -67,12 +67,50 @@ function handlebarTemplates () {
...
+    .pipe(sourcemaps.write())
+    .pipe(gulp.dest('dist'))
+}
+
+/**
+ * Copies all vendor-specific assets to the "dist" folder
+ */
+function copyVendorAssets () {
+  return gulp.src([
+    'node_modules/katex/dist/katex.min.css',
```
nathanlesage, line 90: this whole stuff should be coordinated with #938 because I already see a clash of modules here

source/IpcModule.ts
```diff
@@ -0,0 +1,57 @@
...
+                }
+            }
+        });
+    }
+
+    public static registerMain<T>(service: T) {
+        // Get all methods of the service
+        const userFunctions = IpcModule.getAllUserFunctions(service)
```
nathanlesage, line 30: Do I get this right that you simply extract EVERYTHING that can serve as a getter from the class? Isn't that leaking unintentional functionality as well? o_O

source/IpcModule.ts
```diff
@@ -0,0 +1,57 @@
...
+                return await service[methodName](someArgument)
+            })
+        }
+    }
+
+    private static getAllUserFunctions(x: any) {
+        const isGetter = ( x, name:string ) => ( Object.getOwnPropertyDescriptor(x, name) || {} ).get
+        const isFunction = ( x, name:string ) => typeof x[name] === "function";
+
+        // Recursively, get all properties of the object (and its prototype hierarchy)
```
nathanlesage, line 44: I would argue that you now don't rely on magic strings anymore, but now it looks almost as if you're replacing magic strings with magic methods, and I don't really see how this fits? But maybe I'm not getting the TypeScript implications that much from just looking at this PR …

nathanlesage:
Ah, one more thing: Is two-way-communication possible with this thing? (Does it work renderer->main using function args and main->renderer using returns?)

.eslintignore
tobiasdiez: It is possible to emit the compiled javascript files into source. But then you run into problems with git and the other post-processing steps.  I think this is actually cleaner this way: have all source files of whatever type (js, ts, css, img, ...) in the "source" folder and upon compilation these source files are processed (by tsc, webpack, gulp, whatever) to yield the correct "compiled" version in the "dist" folder.
This way you get a clear and clean distinction between original source files and compiled files.

source/main.ts
tobiasdiez: Typescript complained that `process` was already defined elsewhere... (didn't investigate where)

source/IpcModule.ts
tobiasdiez: It does! The `Promisfied` stuff here is only for convince: if the original method (in the interface) returns "string" then the "ipcfied" method returns "Promise<string>" (essentially since `ipcRenderer.invoke` returns a promise. These type declarations here take care of this automatic conversion.

source/main/providers/citeproc-provider.ts
tobiasdiez: I thought the short and simple code was actually an advantage ;-)

gulpfile.js
tobiasdiez: Yeah...I'm not yet convinced that the "electron-webpack" project is the right way to go...but I don't have any experience with this and was just judging from the activity of the project. Also I don't like that it relies on so many assumptions where the code needs to lie etc. What's your opinion?

source/IpcModule.ts
tobiasdiez: Yes, it exposes all methods of the service. If this is too much, then the caller can restrict it by using
```
registerMain(
 new class implements T {
    //...only implement the required methods by forwarding them to wherever you want
})
```

source/IpcModule.ts
tobiasdiez: You are right. Instead of magic strings we now rely on the "magic methods", namely the ones defined in the interface (which specifies the "contract" between renderer and main). The advantage is that typescript (and the IDE) will complain if the method signatures are not correct - the magic strings version only complains upon runtime.

tobiasdiez:
Thanks for the quick and helpful feedback!

>     1. I very much like the general idea you're going here with, but I think that it is overly complicated
Complicated in which sense?
 
>     2. Concerning progress: If we are to gradually adopt TypeScript for the whole thing, which I'd like, I would like to get the opposite approach: Not compiling everything with the Typescript, but rather compile single components using TypeScript and putting them into appropriate directories within Source. This way we change as few things as possible, which'll help the stability of the overall project. I'd like to have this the same way as, e.g., the vue component (all source files in resources, albeit we should maybe add another top-level-directory for the Typescript stuff and move everything compilable in there …?)

Since conversion of normal js to ts mostly means that you simply rename the file, I'm not sure if a new folder structure is needed. Moreover, you would need to replicate the whole folder structure again, since there might be ts files in common, main and renderer. Which advantages do you see in this approach?

>     3. Concerning Vue, webpack and the other stuff we also have: I wasn't able to research whether it's possible to write vue-components in TypeScript. If so, then perfect, but if not, we should in general go towards an approach of using a vue/webpack-toolchain for the different renderers and a TypeScript-toolchain for the main process

I have no experience here, but with this PR typescript takes care of transpiling the ts files to normal js, and similarily webpack compiles the vue files to normal js as well. So if you want to use typescript in vue, then this needs to be added in the webpack toolchain.

>     4. In general, but this is arguably personal preference, I'd like the IPCs not tied to the module classes (and in general, I think that the IPC implementation should be functional and not class-based, as this looks way too Java-esque for my taste; see e.g. the FSAL which I intentionally implemented with only one main class for the interface, but then functional components for everything else, I'd like the IPC implementation functional as well, as long as there are not good reasons that I just overlook here, which is more than possible as I still think I didn't 100 % get all of the code on this first run)

I'm not sure what you mean by "functional" vs "class-based". The current code of course allows you to use the [composite design pattern](https://en.wikipedia.org/wiki/Composite_pattern) to define the IPC method, e.g.
```
registerMain(
 new class implements ComplicatedInterfaceWithManyMethods {
    public methodA() {
         return classA.methodA()
    }

    public methodB() {
         return classB.methodB()
    }
})
```

gulpfile.js
Aigeruth: Yes, this change will clash with the webpack change.

@tobiasdiez, I'm not sure what you mean by assumptions about code structure. Webpack can locate packages installed by npm/yarn automatically and for me that means less assumptions than having explicit references to node_modules in the code, because that relies on relative path to something. We might look at different things, so just let me know. I'm happy to discuss this more in depth separately.

As for the electron-webpack change, I have an updated version locally that I haven't pushed yet. The remaining broken thing is the theme switching. I'm thinking about opening a precursor PR for restructuring themes to make that shift easier and not further inflate the scope of the electron-webpack change. What do you two think about phasing out Less and using CSS variables only? (I have not seen other Less specific features used, but quite possible I missed something.)

As for a longer term perspective, as @tobiasdiez mentioned, Electron Forge is probably the right way to head and I missed it when I just searched for webpack and Electron. However, Electron Forge should be used with webpack as well, so switching from electron-webpack should be fairly straight forward as the main migration cost comes from standardising import paths across the codebase.

tobiasdiez: Thanks @Aigeruth for chiming in! I really like your PR #938 and realized to late that the typescript-related changes here solve similar problems as you did. So we should definitely coordinate how to proceed from here, otherwise we duplicate work for no reason.

Concerning the "electron-webpack" framework, I have the impression that it relies a lot on 'convention over configuration', i.e. certain files have to lie in a certain location since otherwise nothing works. The gulp-approach on the other hand is really flexible (at the cost of being more verbose). But I realize that this initial skepticism might come from my unfamiliarity with the webpack world.

I just created a new test project using `yarn create electron-app my-new-app --template=typescript-webpack` (which uses electron forge with typescript and webpack support). This looks very similar to what you have in #938, so I would suggest to convert your PR to use electron-forge, and I will wait with this PR here until yours is merged. If you want we can also work together on #938. What you do you think?

Aigeruth: Sounds good, I'm happy collaborate on this migration as it is a bigger piece than I originally anticipated. I'm keen on breaking #938 into multiple PRs because it is just too risky rolling a PR that modifies 200-300 files.

I have raised #1055 to outline my thinking on how we could gradually prepare for the Electron Forge migration and make it a series of smaller changes. This could enable easier collaboration as well. We could even create a project board for this. Please, let me know what you think.

I see what you mean. I generally prefer 'convention over configuration' because it means that the convention is more likely to be documented and potential contributors are more likely to be familiar with it than something that I could create with a configuring something from scratch, therefore it enable better/faster collaboration in my view. It also allows me to focus on the important bits and not on the configuration. I have this somewhat under the same category as linting where I happily trade my personal style/taste for a standard way for collaboration.

.eslintignore
nathanlesage: I see what you mean, but first, the `source`-directory is the one `electron-builder` expects the files to be in and we can add corresponding gitignore-rules, which work for the vue sources easily.

So as it is now it's easier to include into the recommended electron-workflow (plus we still would have a separation of source and compiled files). My aim here currently is to have the source-directory only contain compiled files and assets in the future, and have all the "real" source files in resources, maybe `lib` (which would be the TypeScript equivalent, if I'm not mistaken?). We can discuss the file structure, though!

source/main.ts
nathanlesage: Huh, then we should investigate that, because `processInfo` seems like a misnomer as it doesn't only contain information …

source/main/providers/citeproc-provider.ts
nathanlesage: Yeah, short code is always good, but I realized in the past three years that two or three more lines sometimes helps make the code more readable – and especially since we're finally talking about compiled code, these lines would get shortened either way :)

gulpfile.js
nathanlesage: Good points, and I think what you two are describing here is indeed something we should have a longer chat about, as I assume both the migration as well as a potential rollback would be super time-consuming, so we should discuss to make sure the right approach is taken — what would you say we meet for a talk on Jitsi this week?

Please just send me a mail so that we can take this off GitHub :) (hendrik@zettlr.com)

source/IpcModule.ts
nathanlesage: That is certainly true, but wouldn't it not maybe make more sense to define a generic interface instead of providing interfaces? I'm not sure but I'm thinking of a method within the provider class that just returns an object with the correct methods and signature (in a similar vain as `global.ipc.on` as we do it now). This would keep the necessary runtime code to a minimum … or do I miss some great benefit here?

nathanlesage:
> Complicated in which sense?

I think I'm not too fond of using many classes (but I'm biased here based off a LOT of talks at NDC and the like, that oftentimes try to make sure not to use too much class-based approaches). Maybe if I have the time to look at it more closely I would be able to propose something! But now that we're in general hopping on the TypeScript- and efficiency-game with the other PR by @Aigeruth we may want to discuss this either way – I left a comment that we should all meet up ideally this week or next week to discuss this and begin drafting a roadmap, because maybe what would be better is to slowly migrate everything to TypeScript, and clean up the whole code. Maybe this would even allow jumping to v2.0.0 – but as I said, I would very much like it if we had a chat on this. You were right the last time we talked that sometimes a video-call might be good to discuss such things!

.eslintignore
tobiasdiez: Electron-forge should make this pretty convenient: all the source files are put into /source (or the somewhat more common /src by default) and forge then takes care of the proper post-processing (and puts everything in the /.webpack folder). I'll pause this PR for now until the migration to forge is complete.

source/IpcModule.ts
tobiasdiez: So you would like to have one big file in witch all ipc commands are defined? Can you maybe post a pseudo-code example of what you envision in a perfect world (without any restrictions by js or the like) so that I know in which direction you are thinking?

tobiasdiez:
@nathanlesage 
I think, I'll have a bit of time next week to work on my open PRs here. Since you are currently reconstructing many things, I was wondering if you have any design constraints or idea how you would like to have the IPC implemented? 
This PR contains of course a lot of unnecessary changes right now, I'll clean it up soon.

tobiasdiez:
I've now removed all the unncessary changes from the first draft and updated the PR.

The strategy is the same and a typescript interface is used to describe the IPC interface, providing a type-safe communication that is not much different to calling the service directly. The only difference is that you use `citeService = IpcModule.forRenderer<IpcCiteService>()` instead of `citeService = new IpcCiteService()` to get an instance of the service.

@nathanlesage Review and feedback very much appreciate before I continue fixing the remaining compilation errors and refactor the complete citation provider.

nathanlesage (commented):
Okay, I have spent 30 minutes looking at this code golf, and while I'm still not quite sure what it exactly does, I do like the idea. However, I would suggest we find a way to implement this in a way that's more readable.

We can certainly reduce this code using the following assumptions/simplifications:

- In the renderer we always invoke, in main we always handle, and all of that must be handled by promises, so we don't have to check whether or not these things return a promise, because all of those do.
- Instead of throwing darts at a secondary instance of the providers (which will yield horrible bugs because the providers assume there is just one instance at all times — we can certainly argue that we might want to do that in Functional style as well) in order to pick the relevant functions, we should be more verbose, that is, _explicitly_ state which operations are supported.
- Each provider has exactly one "method" (e.g. `citeproc-provider`) and several "sub-methods"; and these sub-methods are _always_ in the `command` property, whereas any args are _always_ in the `payload`-property

I think given that we can simplify this method. I would even argue we should register an _object_ with `registerMain` that contains callbacks. Example:

```typescript
ipcModule.registerMain({
  'some-event': async (a, b): Promise<T> => {
    // call some internals here and return something
  }
})
```

This way we are able to still make use of interfaces to enable type checking but can boil down the unreadable interfaces to something manageable. What do you think?

tobiasdiez:
Thanks for your feedback!

> In the renderer we always invoke, in main we always handle, and all of that must be handled by promises, so we don't have to check whether or not these things return a promise, because all of those do.

You mean that every method in the interface `IpcCiteService` already returns a promise? That's certainly possible. What's the advantage though? Ok, one then doesn't need the somewhat strange-looking `Promised` and `Ipcfied` types, but instead you have an assumption about your providers that is not statically verified by typescript (or you have a type that checks that all functions are indeed returning promises).

>  explicitly state which operations are supported.

This is the purpose of the interface, e.g.  `IpcCiteService`. This interface clearly says which commands are provided by this particular provider. So I'm not sure what you mean here.

> Each provider has exactly one "method" (e.g. citeproc-provider) and several "sub-methods"; and these sub-methods are always in the command property, whereas any args are always in the payload-property.

This is somewhat similar to how it is implemented right now, except that the root method of the provider is a class instead of method as you propose. Of course, you could replace the class by a method that simply returns an object; but that would be just a different form of a constructor, or not?

Your example is already possible with the current code with only a minimal change:
```js
interface IpcTestService {
    someevent: (a: string, b: string) => Promise<string>
}

IpcModule.registerMain<IpcTestService>({
  someevent: async (a: string, b: string): Promise<string> => {
    // call some internals here and return something
    return 'test'
  }
})
```
The point here is that the object "implements" the `IpcTestService` interface. It doesn't matter if you construct it as an instance of a class, or as a simple object. But I agree that might be become handy for smaller interfaces where you don't really need a provider class.

nathanlesage:
The code you just provided is much better! We still have typechecking, but are more flexible, it's extremely readable now, and people only need to know what an Interface looks like and will understand what the code is doing.

And yes, we can safely assume all functions have to return a Promise and thus remove these barely readable interfaces, because if they don't Electron will crash (or at least throw an error), so that typechecking is ... "runtime checking"

The other benefit of passing an object into the registration aside from preventing side effects is that the registration will basically look almost like the current implementation, and only add Type checking.

tobiasdiez:
Okay, so should I delete the Ipcfied and Promisfied types? I don't really see the disadvantage of having them since they are there only to ensure proper typing and don't influence the runtime behavior at all. Note that they are types and not interfaces. I sadly cannot change the fact that typescripts type definitions always look a bit cryptic once you go beyond the basic use cases.

tobiasdiez:
Okay, this is now ready for review.

nathanlesage:
Due to various reasons (the biggest being that I'm currently receiving quite a lot of mails from people doing funny things with the renderers and running the calculator from iFrames), I had to move quickly and completely shut off node integration and turn on context isolation for all renderers. This means that a few things with this PR are now broken. But I haven't changed the internal structure; the only thing we might now want to change with this is to move the API to the preload script as I've already mentioned once.

I'm really sorry for having broken the PR, but it's kind of a stressful situation right now and I'm really hating myself for the bad choices I made in the past, hope you're not too angry!

Here's what I changed (+ a few things in the affected files, mainly just replacing the requires with calls to window.ipc/window.path): https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/preload/index.ts

tobiasdiez:
No worries. I changed the code to use the context bridge now. What do you think?

Currently it doesn't work and throws the following error during start:
```
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "D:/Programming/Zettlr/node_modules/electron/dist/LICENSE"
 * "D:/Programming/Zettlr/node_modules/electron/dist/LICENSES.chromium.html"
 * "D:/Programming/Zettlr/node_modules/electron/dist/chrome_100_percent.pak"
 * and more ...
 @ ./source/IpcModule.ts 4:19-38
```
Any idea of how to fix this? The pointer `IpcModule.ts 4:19-38` leads to nowhere...

tobiasdiez:
Let's close this in favor of  #3087.

Issue #1055, opened 2020-07-12
Migrating to Electron Forge


Aigeruth:
**This is a WORK IN PROGRESS issue to outline migration to Electron Forge based #938.**

# Goal

Zettlr uses [Electron Forge](https://www.electronforge.io/).

[Electron Forge](https://www.electronforge.io/) is one of the recommended build tool by [the official Electron website](https://www.electronjs.org/docs/tutorial/boilerplates-and-clis#electron-forge) and it uses webpack as bundler. It is more actively maintained than [electron-webpack](https://webpack.electron.build/) that could be used with [Electron Builder](https://www.electron.build/).

As we discussed on #938 and #1048, Zettlr should move towards Electron Forge.

## Using imports instead of global variables

Some of the packages are used in a way that assumes their presence in the global or window scope. This will not be compatible or easy to maintain with webpack, so as a preparation for this migration, they should be removed.

- [x] Import tippy.js explicitly #1056 
- [x] Import popper.js explicitly #1056 
- [ ] Import `$` for jQuery everywhere (or remove jQuery completely) #1062 
  - [ ] Prevent accidental use of jQuery #1071 
- [x] Import jquery-ui #1057 
    - [x] jquery-ui tabs
    - [x] jquery-ui sortable
    
## Moving the single package.json file
 
 Using two `package.json` files has not been required since [Electron 8](https://www.electron.build/tutorials/two-package-structure). Webpack would rely on a single `package.json`, so doing this upfront can reduce the size of that change.

- [x] Using `app.getVersion()` for version information #1090 
- [x] Utilising RequireJS for locating and importing libraries:
    - [x] CodeMirror (it is currently imported with relative path that explicitly contains `node_modules`) #1095
- [ ] Rest is done as part of https://github.com/Zettlr/Zettlr/pull/1060
 
## Improve code structure to work with webpack
- [ ] Replace `_autoload.js`

## Using a single index.html
 
Project has multiple HTML files currently:

- `source/renderer/assets/index.htm`
- `source/quicklook/index.htm`
- `source/print/index.htm`
- `source/log-viewer/index.htm`

These could be merged together a single HTML file with a different entry point for Javascript. It can be beneficial as there are overlapping parts between these HTML files. This will bring Zettlr closer to be a single page application that starts up based on routing, and can make easier to extend the usage of Vue.

This part is not mandatory as Electron Forge seem to support [multiple entry points](https://www.electronforge.io/config/plugins/webpack#configuration).

[...]

PR #1056, opened 2020-07-12
Remove tippy.js from global scope

Aigeruth:
This is related to #1055.

## Changes

Use tippy.js through imports instead of global scope to make it more explicit and easier to understand.

## Additional information

Tested on: macOS


nathanlesage:
Ehm … how did you get this working? I had to remove the `requires` and pull in tippy using a script-tag, because they somehow stopped shipping the CJS modules o_O Are you sure this works?

Aigeruth:
I tested it locally and worked, but please, do checkout my branch locally and test. It is always good to have things tested on multiple machines.

PR #1057, opened 2020-07-12
Import jquery-ui as CommonJS module

Aigeruth:
This is related to #1055.

## Changes

Import jQuery UI as a CommonJS module instead of as a UMD bundle.

## Additional information

Tested on: macOS


nathanlesage:
I've yet to read issue #1055 (I'm slowly working my way up the notification list …), but I would like to just drop in here that I would like to get rid of anything jQuery-related either way. I've already begun implementing some stuff in vanillaJS for the document tabs (there's no jQueryUI involved), and we don't really need jQuery either way because we only need to support one single browser. So this will allow us to further straighten out the jQuery stuff. I used it in the beginning since it was the only tool I knew of, and we all know that if you have a hammer every problem looks like a nail …

So I'd like to propose that we leave it like that and remove jQuery and jQueryUI from wherever we can and simply remove it completely!

Aigeruth:
@nathanlesage, sorry, I'm not sure I fully follow. Do you have a change locally that has a tabs implementation without jQuery UI? Or there is code already committed to Zettlr that uses a different implementation for tabs than jQuery UI and could be reused in these use-cases? If any of these, we can move forward that direction. If not, I would still recommend incorporating this change as it makes this dependency more explicit and easier to understand.

I think, phasing out jQuery is a significant undertaking and could take a while. It would also benefit from moving jQuery to a local scope from a `<script/>` tag as contributors will be able to locate the touch points individually and remove jQuery as a dependency step by step. This lowers the barrier of entry for these changes, prevents accidental usage of jQuery, and decrease the risk of the whole migration.

nathanlesage:
> Or there is code already committed to Zettlr that uses a different implementation for tabs than jQuery UI and could be reused in these use-cases?

Yes, I've manually implemented a vanilla-tab structure in the editor tabs --> https://github.com/Zettlr/Zettlr/blob/develop/source/renderer/util/editor-tabs.js

We could certainly move the tabbing-functionality to its own class so that we can reuse that in different parts of the app.

Concerning the phasing out of jQuery: I actually only use it to query some elements, and there are direct mappings to vanilla functions. It's just a little bit tedious. As with regard to jQuery UI, until we can find a way to make the TOC sortable better, this might be a good idea.

So following proposal: I'll make sure to tackle jQuery this week so that we can get rid of all of this, and then we can move on from this, I'll be testing your PR and merge it once it works!

nathanlesage:
Alright, I'll be opening a separate PR to finally phase out jQuery and jQueryUI and get to that!

nathanlesage:
P.S., after installing from the new lockfile, it threw this error:

image

I'm not sure what caused this though …

Aigeruth:
**Re:** error above, what is the best way to reproduce this error? Is it the TOC feature?

---

**Re:** jQuery removal

I would take a step back before we jump on replacing jQuery with custom implementations.

I understand that some functions might have a 1:1 replacement, but it is likely that some require some logic to be implemented. This means adding more custom code and thus adding risk. I think, replacing imports involves less risk and it is easier to roll back if anything goes wrong.

One more angle to consider: anything that replaces jQuery will likely be temporary if the end vision is to use Vue for everything as Vue will very likely have its own approach to it. I can see how removing jQuery might be a win of having less dependencies, but will it make a difference in the end game?

Just an idea, but it might be enough if there is an eslint rule that prevents adding new code using jQuery. This could allow a gradual migration where features can be still added to Zettlr, releases can be made and nothing needs to block.

It sounds like moving completely away from jQuery UI sortable might be difficult anyway and a gradual approach removes the immediate need of changing something that works. I think, it is also challenging to create an abstraction for tab support that will work with Vue without actually using Vue.

nathanlesage:
>  Is it the TOC feature? I'm happy to take a look.

Yes, it's the only place the sortables are being used – I've had this kind of errors in the past, which is why I pushed this, and also tippy, to the global context. I do not like that either, but it was the only way it worked :/

> I would take a step back before we jump on replacing jQuery with custom implementations.

Please see #1062, which I'll be implementing.

> I understand that some functions might have a 1:1 replacement, but it is likely that some require some logic to be implemented. This means adding more custom code and thus adding risk. I think, replacing imports involves less risk and it is easier to roll back if anything goes wrong.

I fully understand your point, and in most circumstances I would say "Yes!", but in the particular case of JavaScript, the only real reason for jQuery is actually to query elements, and as `querySelectorAll` is working perfectly, we do not need it for that. I know the codebase in and out and I know there are exactly two things that depend on jQuery for which we need some more code: The tabs in the dialogs, and the sortable. Everything else is already running on vanillaJS. I did not start the jQuery-removal-inquisition yesterday. If you look especially at the newer modules, there is no jQuery in them at all. For instance the Table Editor is already 100 % vanilla JS.

> One more angle to consider: anything that replaces jQuery will likely be temporary if the end vision is to use Vue for everything as Vue will very likely have its own approach to it. I can see how removing jQuery might be a win of having less dependencies, but will it make a difference in the end game?

It will make a difference for you and @tobiasdiez, because jQuery, as mentioned before, is super nasty to handle in an Electron context (just like tippy.js), and hence we can get this off our mind. Switching to Vue basically just means one additional step – I am currently replacing a lot of `$(#some-id)` with `document.getElementById(some-id)`, and with regard to vue we can completely ditch this (as it will become `<element v-on:click="handlerCode">`), but the major win will be a cleaner code base, plus we can see if we can in fact remove jQuery for good.

> I think, it is also challenging to create an abstraction for tab support that will work with Vue without actually using Vue.

Exactly, which is why I will keep the jQueryUI in the code base for the time being. But not the jQuery instances. Yes, jQuery will still be some where, but it's out of the way then.

Aigeruth:
I have cloned the repository again and tried it on the `develop` branch by doing:

```
$ yarn install
$ cd source && yarn install
$ yarn start
```

Then I click on the 'Show the table of contents' icon on the top bar, this error does not show up in the console for me.

```
➜  Zettlr-test git:(develop) ✗ node -v
v14.5.0
➜  Zettlr-test git:(develop) ✗ yarn -v
1.22.4
```

I use macOS 15.5.5.

Do you have a complex test file? I have only created a simple one with a few sections and subsections.

nathanlesage:
> Do you have a complex test file? I have only created a simple one with a few sections and subsections.

I simply used the `yarn test-gui` command for that, opened a random document and tried it. Maybe it's a bad dependency somewhere?

Aigeruth:
Oh, I see. It happens when you click on an item in the TOC and not when the TOC rendered. I'll take a look. This is probably just a missing import.

PR #1059, opened 2020-07-13
Make links in references clickable

maxedmands:
## Description

Made URLs in references into clickable links that open in the system's browser.

## Changes

* Enable the citeproc "development extension" called `wrap_url_and_doi`
* Handle clicks on any non-attachment link in the attachments pane by opening the URL in external browser
* Update CSS so that the links show up nicely in the pane

Screen Shot 2020-07-12 at 7 20 30 PM
Screen Shot 2020-07-12 at 7 20 44 PM

## Additional information

The fact that this is using a citeproc "development extension" that doesn't seem to be documented (people refer to it in issues, but it's nowhere in the documentation!) is worrisome. If you don't want to merge this that's OK with me. But I definitely find this change to be very helpful.

Tested on:
* MacOS Catalina 10.15.4 (19E287)


source/renderer/zettlr-attachments.js
```diff
@@ -188,6 +189,8 @@ class ZettlrAttachments {
               console.error('Could not open attachment:' + potentialError)
             }
           })
+      } else {
+        shell.openExternal(elem.attr('href'))
```
tobiasdiez, line 193: Please add a short comment explaining that this code is invoked if the user clicks on a link coming directly from bib data (e.g. url or doi).

maxedmands: Sounds good, done!

resources/less/theme-berlin/light/attachments.less
```diff
@@ -13,4 +13,8 @@
             background-color: var(--grey-2);
         }
     }
+
+    .csl-bib-body a {
+        color: var(--blue-0)
```
nathanlesage, line 18: Do the rules look good in all five themes in light & dark?

source/main/providers/citeproc-provider.js
```diff
@@ -278,6 +278,7 @@ class CiteprocProvider {
       // is only for preview purposes, it should follow the language like the
       // rest of the interface.
       this._engine = new citeproc.Engine(this._sys, this._mainStyle, this._lang)
+      this._engine.opt.development_extensions.wrap_url_and_doi = true
```
nathanlesage, line 281: As it looks like a `development_extension`, is there documentation on the stability of this approach?

maxedmands: As far as I can tell there is no official documentation here, this whole thing is incredibly sketchy to me. But it works!

Here's the code in citeproc-js that implements it:
https://github.com/Juris-M/citeproc-js/blob/3389e9b23a1873c1bb0af16eef7533aefd669701/src/node_text.js#L320-L365

resources/less/theme-berlin/light/attachments.less
maxedmands: Subjectively, I think they look fine in all five themes :)

nathanlesage: Perfect, thank you!

source/main/providers/citeproc-provider.js
nathanlesage: Hmhmhmhm … I'm a little bit unsure. I REALLY see the wish (and all the benefits) of making URLs clickable, but I'm afraid we're breaking something using this (more or less) undocumented code …

So I basically see two options:

1. Implement something that simply auto-links any link in the rendered bibliography (hence, in the `zettlr-attachments.js`-file), which would be fine, because then it's "our" code (and we don't depend on another library)
2. Implement it this way, but then we need a switch, because I don't want to risk relying on features from a third-party-library that are clearly labeled unstable (because this way, in case something breaks we can simply tell people that they should uncheck that option in the preference until a fix)

So one of the two approaches would work, but I'll leave the one you choose open to you because it's your PR! :)

I personally am biased towards option 1 (which would be much more stable and easier to implement) – here's how we might achieve that:

https://github.com/Zettlr/Zettlr/blob/e64a4a7b033f6db552e0cada428c4bd275ddadd3/source/main/commands/update-check.js#L110-L116

Just instead of transforming already HTML-links take something that checks URLs!

maxedmands: Makes sense! I don't know if I'm going to have the time to implement this, but I donate my PR to y'all if you're interested in taking up the banner :)

nathanlesage: Take your time, it's your hard work, and this means you're in control :) It'll take some time either way until the next version will be released!

nathanlesage:
You know what? Let's just do it, and hope that citeproc upholds to German quality standards …

Issue #746, opened 2020-05-08
[BUG] Too many "file has been modified remotely" notifications


Zverik:
## Description

When I was testing syncing apps, I've seen many "remotely modified" notifications. After settling on one, I get like one a month, perfectly valid.

After upgrading to 1.7.0 beta, I see these notifications very often. It is completely unrelated to remote modifications, since I'm offline. There is something wrong in the saving code.

## Reproducing

* Edit a file.
* Switch to another app.
* Retry until you see the message.

## Expected behaviour

Not a single "file has been modified remotely" notification as long as I'm neither syncing with other machines, nor modifying files in different apps.

## Platform
 - OS and version: macOS 10.14.5
 - Zettlr Version: 1.7.0-beta1

## Additional information

I can debug this, but there's nothing useful in the console.

nathanlesage:
Mh, it might be that the logic I use for ignoring the events being generated by the app still has some bugs to it, and doesn't ignore the events correctly on all systems. I'm running the app constantly with the console open, where correct `Ignore`-statements are displayed each time a "remote" change is being ignored as the origin is the app itself …

Zverik:
Just got the notification again — typed some words, switched to another window, the popup appeared after a few seconds. Here are log contents:

```
[12:42:48]	Trying to run command through Application: file-save

[12:42:48]	<<< IPC OUT: tags-database
1: {"text":"1","count":2,"className":""}
...
wfh: {"text":"wfh","count":1,"className":""}

[12:42:48]	<<< IPC OUT: tags-database
1: {"text":"1","count":2,"className":""}
...
wfh: {"text":"wfh","count":1,"className":""}

[12:42:48]	<<< IPC OUT: mark-clean
hash: -268821158

[12:42:48]	<<< IPC OUT: file-replace
hash: -268821158
file: {"parent":83715665,"dir":"/Users/ilazverev/Sync/Zettel/Lyft","path":"/Users/ilazverev/Sync/Zettel/Lyft/lyft-2020-05-11.md","name":"lyft-2020-05-11.md","hash":-268821158,"ext":".md","id":"20200504194317","tags":["lyft","weekly"],"type":"file","wordCount":157,"charCount":937,"modtime":1589190071931.57,"creationtime":1588610574867.0217,"linefeed":"\n","modified":false}

[12:42:48]	<<< IPC OUT: tags-database
1: {"text":"1","count":2,"className":""}
...
wfh: {"text":"wfh","count":1,"className":""}

[...]

PR #1060, opened 2020-07-13
Migrate to electron-forge

### LINKED ISSUE #1055: Migrating to Electron Forge

tobiasdiez:
This is a prototype for the final migration to electron-forge. I more or less used `electron-forge import` to let electron-forge migrate the configuration, and then fine-tuned a few things (mainly copied from the electron-forge with typescript + webpack support template).

For now `yarn start` starts successful, but the renderer is not yet working. The main idea of this PR is to have a convenient way to track progress towards https://github.com/Zettlr/Zettlr/issues/1055, and to serve as a base for the developer call. 

Changes:
- Replace electron-builder by electron-forge
- Start using typescript (as proof-of-concept)
- Move resources/vue to source/renderer/sidebar
- Move resources/less to source/common/assets/less/
- Use webpack's handlebar-loader to load templates
- Package using `electron-forge` (but continue using `electron-builder` to create the installer - at least for now)
- Load files through "safe-file" custom protocol (which is safer than disabling websecurity application-wide)
- Remove cache-query (Date.Now) from external files, since this lead to problems and wasn't needed in my tests (ie images were reloaded correctly after switching between files)

Fixes #1055.

Todo:
- [x] Make themes work
- [x] Fix `_autoload.js`: WARNING in ./source/main/commands/_autoload.js
Module not found: Error: Can't resolve 'D:\Programming\Projects\Zettlr\source\main\commands\' in 'D:\Programming\Projects\Zettlr\source\main\commands'
- [x] Fix loading of `custom.css` and embedded images: Not allowed to load local resource: file:///C:/Users/Tobi/AppData/Roaming/zettlr/custom.css?1595773341352
- [x] Preferences do not load
- [x] References do not work


tobiasdiez:
So the electron-forge part seems to work (in principle).

What is left are webpack specific things. For example, code using `__dirname` currently breaks the build as the file cannot be found. I also had problems with dynamic require statements such as
https://github.com/Zettlr/Zettlr/blob/7fdbb22c466d8ee5ca66bfaa983dcbcb878ce1f8/source/renderer/assets/codemirror/autoload.js#L35-L38

@Aigeruth how did you solved these things in your PR? (feel free to directly commit to this PR if you have an easy solution).

Aigeruth:
The autoloader failed due to a HTML file in one of the directories, so I [skipped that](https://github.com/Zettlr/Zettlr/pull/938/files#diff-eae83ba97215eeb96fe946e22c0fba1b):

```javascript
if (mode === 'rpm') {
    continue
}
```

The more interesting one was the command autoloader that I replaced with a static [index module](https://github.com/Zettlr/Zettlr/pull/938/files#diff-6c505ec978c78648a7884f5d2b4c46df).

tobiasdiez:
Thanks. I saw you uncommented the dynamic import for the modes. I did so now as well, but this needs to be fixed at some point.

Moreover, I've now included support for the vue component (i.e. the sidebar). There are still runtime errors but they are only related to stuff that you guys are already working on (e.g. tippy, jquery etc). So I'll pause this PR here until the other ones are  merged.

tobiasdiez:
Update: it loads!
image
Still a bit of work necessary to fix the remaining errors. @nathanlesage if you have the time, then I would appreciate to get your initial feedback. Do you like the structure etc?

package.json
```diff
@@ -12,28 +12,47 @@
...
-    "watch:gulp": "gulp watch",
-    "watch:wp": "cross-env NODE_ENV=development webpack --progress --colors --watch",
-    "watch": "npm-run-all --parallel watch:*",
     "test": "mocha",
     "test-gui": "node scripts/test-gui.js",
-    "lint": "eslint --ext js,vue -f html -o eslint_report.htm source resources"
+    "lint": "eslint --ext js,vue,ts -f html -o eslint_report.htm source resources"
+  },
+  "config": {
+    "forge": {
```
Aigeruth, line 33: Just a thought: would it make sense to use the config file format and separate the Electron Forge configuration out of the `package.json` file? It might be easier to read as it is a Javascript file and not a JSON file. It might also result in in less merge conflicts. What do you think?

https://www.electronforge.io/configuration

source/common/assets/less/geometry/codemirror.less
```diff
@@ -32,7 +32,7 @@
 
   &.CodeMirror-empty {
     // If the editor is empty, display a nice background image
-    background-image: url('../img/feather.svg');
+    background-image: url('../../img/feather.svg');
```
Aigeruth, line 35: As moving files around result in a large change, I'm wondering if it is possible to separate out moving the Less files under the `source` directory from the webpack (Electron Forge) change, to make this PR smaller and easier to review. What do you think?

source/common/assets/css/.gitignore
```diff
@@ -1,5 +0,0 @@
-# Ignore all compiled CSS files
```
Aigeruth, line 1: 👍

source/common/assets/.gitignore
```diff
@@ -1,2 +0,0 @@
-# Don't commit the handlebars runtime to VCS
```
Aigeruth, line 1: 👍

source/common/assets/tpl/.gitignore
```diff
@@ -1,5 +0,0 @@
-# Ignore all compiled template files
```
Aigeruth, line 1: 👍

package.json
tobiasdiez: This would be possible indeed. I don't have a strong preference for either version. @nathanlesage you?

source/common/assets/less/geometry/codemirror.less
tobiasdiez: Sadly not, at least not easily without replicating what webpack does. The less files previously were compiled to css by hand, and then put into a `source/renderer/assets/css`. All references to other files are relative to this folder. Webpack requires that the references are relative to the original file.

tobiasdiez:
So, it works in principle. There are still a few minor things to fix. I've changed the PR description above to reflect the current state. @nathanlesage @Aigeruth any input concerning the open points is highly appreciated.

package.json
nathanlesage: Everything that's out of the `package.json` is a good config, I don't like the `electron-builder`-config in there either

source/common/assets/less/geometry/codemirror.less
nathanlesage: Phew, at least for once it's good that I don't make use of all too many external resources …

package.json
tobiasdiez: Ok, done!

tobiasdiez:
@nathanlesage any idea why the custom.css / images do not load? Moreover, the references are not shown, but also no error message is thrown. Any ideas why this happens?

nathanlesage:
@tobiasdiez Good question … I need to fix a lot of the stuff that's been lying around before I can get to viewing it so I'm not sure when I'll be able to!

tobiasdiez:
No worries, I looked into it again and could resolve both issues. Only theme support is missing.

tobiasdiez:
Finally! Theme switching now also works. This was the last item on my todo list, which means this PR is now ready for review.

I clicked every ui-button I could find, but it is very likely that I missed something. For this reason, I propose the following strategy:
- You review and test as normally
- I fix the issues you found and change the code according to your feedback
- Merge this PR
- Immediately release a beta, or at least let more people test it via twitter / forum etc

What do you think?

nathanlesage:
This is a good idea, but we really need to make sure that the tooling works afterwards; I had quite a struggle yesterday getting the build workflow to run again, but yeah, I'll see to it. Currently I'm a little bit exhausted as I finally managed to fix #746, so I hope I'll find the energy to do so next week!

source/renderer/zettlr-body.js
```diff
@@ -150,17 +149,40 @@ class ZettlrBody {
...
+      case 'frankfurt':
+        themeToUse = frankfurt
+        break
+      case 'karl-marx-stadt':
+        themeToUse = karlMarxStadt
+        break
+      case 'bordeaux':
+        themeToUse = bordeaux
+        break
+    }
```
Aigeruth, line 178: What do you think about something like this?

```suggestion
    const themes = {
        berlin: require('./../common/assets/less/theme-berlin/theme-main.less'),
        bielefeld: require('./../common/assets/less/theme-bielefeld/theme-main.less'),
        frankfurt: require('./../common/assets/less/theme-frankfurt/theme-main.less'),
        karlMarxStadt: require('./../common/assets/less/theme-karl-marx-stadt/theme-main.less'),
        bordeaux: require('./../common/assets/less/theme-bordeaux/theme-main.less'),
    }')

    themes.keys().forEach(theme => newTheme === theme ? themes[theme].use() : themes[theme].unuse())
```

source/renderer/zettlr-body.js
```diff
@@ -150,17 +149,40 @@ class ZettlrBody {
...
+
+    // Unload all themes except the new one
+    [ berlin, bielefeld, frankfurt, karlMarxStadt, bordeaux ]
+      .filter(theme => theme !== themeToUse)
+      .forEach(theme => theme.unuse())
+    // Load the new theme
+    themeToUse.use()
```
Aigeruth, line 185: Nice find! Is this a webpack built-in function?

Aigeruth: > ⚠️ Behavior is undefined when unuse is called more often than use. Don't do that.

-- https://github.com/webpack-contrib/style-loader#lazystyletag

Have you experienced any problem so far?

source/renderer/zettlr-body.js
tobiasdiez: Thanks for the good suggestion. This makes it indeed more readable and easier to understand!

source/renderer/zettlr-body.js
tobiasdiez: No, so far everything works. I added a comment that points to this warning as a reminder if we experience problems in the future.

source/renderer/util/load-icons.js
```diff
@@ -14,24 +14,14 @@
...
-      let content = await fs.readFile(fullPath, { encoding: 'utf8' })
-      let icon = {}
-      icon[path.basename(file, ext)] = content
-      clarityIcons.add(icon)
-    }
-  }
+async function loadCustomIcons () {
+  /* eslint-disable import/no-webpack-loader-syntax */
+  clarityIcons.add({
+    'code-alt.svg': require('!!svg-inline-loader!./../assets/icons/clarity-custom/code-alt.svg'),
```
nathanlesage, line 21: Love it!

But we should definitely add a comment to tell people why we are using `!!svg-inline-loader!` prepended to the path. This will make it easier for other people understand this beautiful (yet full-of-assumptions) code!

source/renderer/zettlr-body.js
```diff
@@ -151,17 +149,24 @@ class ZettlrBody {
...
-
     // On config change, change the theme according to the settings
-    let href = $('link#theme-css').attr('href')
-    href = href.replace(/bielefeld|berlin|frankfurt|karl-marx-stadt|bordeaux/, newTheme)
-    $('link#theme-css').attr('href', href)
-    this._renderer.getEditor().refresh()
-    this._currentTheme = newTheme
+    let newThemeName = global.config.get('display.theme')
+
+    const themes = {
```
nathanlesage, line 155: I am sure I do not fully understand this code, but it looks like "bad practice".

I am missing the place where the LESS files are converted to CSS. And if it might introduce problems if unuse is used more often than the require, we should not do this.

But just for clarification:

1. Webpack compiles the templates to CSS and stored them as a JS-module within the final renderer bundle
2. "require" just puts the theme in the variables
3. Webpack immediately adds the CSS as links into the head of the file (that is: all five of them)
4. You remove those that are not correct

This really does not work good. Rather, I would like to propose the following two alternatives:

1. Put ALL themes into one big CSS module (if everything is within one file either way, we can completely throw overboard the idea of separating the file in the first place, and only keep them separated in the source), but namespace them (e.g. `body.theme-berlin` and `body.theme-berlin.dark`) and deal with this using classes
2. Remove that fancy Webpack magic (which adds a lot of intransparency to the build process either way and is one of the major reasons I really do not like working with any JavaScript bundler …)

Either way, the logic here is completely opaque to the non-initiated and we need to comment what is exactly happening here and why that is so.

source/renderer/zettlr-body.js
```diff
@@ -183,7 +188,8 @@ class ZettlrBody {
       return global.ipc.send('file-new', { 'name': generateFileName(), 'hash': dir.hash })
     }
 
-    let cnt = makeTemplate('popup', 'textfield', {
+    const textfieldTemplate = require('./../../resources/templates/popup/textfield.handlebars')
```
nathanlesage, line 191: For clarification: How is the handlebars runtime bundled with these? Is it so that webpack will pre-compile the `handlebars` during bundling, and this "require" actually requires the runtime with one of the templates? Or how does this work?

This is a strong indication that we need to comment this as well.

source/renderer/zettlr-renderer.js
```diff
@@ -120,7 +121,7 @@ class ZettlrRenderer {
       // Apply the custom CSS stylesheet to the head element
       global.ipc.send('get-custom-css-path', {}, (ret) => {
         let lnk = $('<link>').attr('rel', 'stylesheet')
-        lnk.attr('href', 'file://' + ret + '?' + Date.now())
+        lnk.attr('href', 'safe-file://' + ret)
```
nathanlesage, line 124: See above my question why the cache-breaker is apparently not needed anymore, and why we need `safe-file`, as this was never needed beforehand.

webpack.rules.js
```diff
@@ -0,0 +1,82 @@
...
+    use: {
+      loader: 'raw-loader',
+      options: {
+        // Do not wrap in js module
+        esModule: false
+      }
+    }
+  },
+  {
+    // Handle audio files: just copy them
```
nathanlesage, line 62: Maybe swap the audio with the citeproc section (for cleaner loader-stuff)

webpack.rules.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+  },
+  {
+    test: /(.ts|.tsx)$/,
+    exclude: /(node_modules|\.webpack)/,
+    use: {
+      loader: 'ts-loader',
+      options: {
+        transpileOnly: true
```
nathanlesage, line 78: Do I understand it correctly that this will only be necessary until we have everything ported to TS?

webpack.plugins.js
```diff
@@ -0,0 +1,9 @@
+const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin')
+const VueLoaderPlugin = require('vue-loader/lib/plugin')
+
+module.exports = [
+  new ForkTsCheckerWebpackPlugin(),
```
nathanlesage, line 5: Why are these two modules loaded within a different file, and not in the main `plugins`-property?

webpack.renderer.config.js
```diff
@@ -0,0 +1,56 @@
...
+})
+rules.push({
+  test: /\.handlebars$/,
+  use: [{
+    loader: 'handlebars-loader',
+    options: {
+      // Automatically load referenced images
+      inlineRequires: '/img/',
+      // Use custom Handlebars runtime with extra helpers registered
+      runtime: path.join(__dirname, 'source/common/zettlr-template.js')
```
nathanlesage, line 36: Short reminder that I said somewhere above that I think we should rename this thing to better reflect what it does now! :)

source/common/assets/less/main.less
```diff
@@ -0,0 +1,14 @@
...
+@import '~@clr/icons/clr-icons.css';
+
+// Tippy
+@import '~tippy.js/dist/tippy.css';
+
+// App layout
+@import './geometry/geometry-main.less';
+
+// Main theme
+@import './theme-berlin/theme-main.less';
```
nathanlesage, line 14: Why do you only import the main theme here? I assume this will be resolved once we also speak about the way the theme-switching is currently handled! ;)

package.json
```diff
@@ -12,29 +12,29 @@
...
+    "package:win-arm": "electron-forge package --platform=win32 --arch=arm64",
+    "package:linux": "electron-forge package --platform=linux",
+    "release": "yarn package && electron-builder",
+    "release:mac": "yarn package:mac && electron-builder --mac --publish \"never\" --prepackaged out/TODO",
+    "release:win": "yarn package:win && electron-builder --win --x64 --ia32 --publish \"never\" --prepackaged out/zettlr-win32-x64",
+    "release:win-arm": "yarn package:win-arm && electron-builder --win --ia32 --arm64 --publish \"never\" --prepackaged out/TODO",
+    "release:linux": "yarn package:linux && electron-builder --linux --publish \"never\" --prepackaged out/TODO",
     "lang:refresh": "node scripts/refresh-language.js",
     "csl:refresh": "node scripts/update-csl.js",
     "reveal:build": "node scripts/update-reveal.js",
```
nathanlesage, line 31: Memo to ourselves: Adapt the available commands in the Readme as well as the documentation

package.json
```diff
@@ -60,16 +60,15 @@
...
+      "output": "release"
     },
     "afterPack": "./scripts/afterPack.js",
     "afterSign": "./scripts/afterSign.js",
     "mac": {
       "category": "public.app-category.productivity",
       "target": "dmg",
       "artifactName": "Zettlr-${version}-${arch}.${ext}",
-      "icon": "resources/icons/icns/icon.icns",
+      "icon": "resources/icons/icon.icns",
```
nathanlesage, line 71: One super important question I just realized: Electron forge also requires the icons; so does this mean that the prepackaged app produced by electron-forge contains ALL of Windows 64bit, Windows 32 bit, Windows 64bit ARM? This might actually be a problem, and we need to check that!

forge.config.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  'packagerConfig': {
+    'asar': true,
+    'icon': './resources/icons/icon' // Automatically adds file extension based on OS
```
nathanlesage, line 4: If I understand it correctly, and forge produces EXEs, we should see to checking that the distinction 64/32/arm works on Windows

nathanlesage (approved):
Heya,

I *finally* managed the code review! It's indeed a lot, but it turns out if you're in the right mood, it's actually easy to understand (the past two months I simply didn't have the mental resources to do this). I am much more confident that it works good, but as I already indicated, there are some open questions. I'm posting them here for the public interest, but I am looking forward to talking to you about them and giving birth to a – hopefully – faster and leaner build!

- [x] Does it run locally on all platforms?
  - -> See one below
- [x] Does electron-forge work?
   - -> See one below
- [x] Does electron-builder work?
    - -> See one below
- [ ] Does it release correct packages that can be opened/installed/used on all platforms (special emphasis to the Windows-question)?
  - Tested on
     - [x] Windows
     - [ ] Linux
     - [x] Mac
- [x] The theming is not good currently (see my comment on one of the LESS files and the configChange function)
   - -> Remember current theme, and only call unuse on this one
- [x] How does handlebar-loading actually work?
   - -> Discussed in call
- [x] `require`s vs. `import`?
   - -> Do as part of typescript migration: https://github.com/Zettlr/Zettlr/issues/1186
- [x] How does the Windows build process work and not screw up the ARM vs. x86 separation?! (Underlying question: What does electron forge actually spit out?)
   - -> Discussed in call
- [x] How is the grouping dependencies vs. devDependencies being decided upon? Can‘t we basically just throw everything in the dev dependencies? If so we might want to think about semantically separating them in a way that makes sense.
  - -> See https://classic.yarnpkg.com/en/docs/dependency-types
- [x] Why do we need `scripts/build-app.js`? I didn't see it being required anywhere
  - -> I also couldn't find any references, is it currently used in the develop branch?
- [ ] Do the workflows run? (This is something we have to test out after merging)
   - -> Do after merging
- [x] Commenting, commenting, commenting. The changes perform a lot of "magic" under the hood and rely upon an insane amount of assumptions of what will be happening that is very specific not just to the JavaScript community, which has it's thing of not being transparent with their code, but even more specifically the Electron-TypeScript-Webpack-Forge-Builder toolchain that we use. And I feel that we might want to explain that to others (I can do most of it to lower your workload, except where I was like "Nani the fuck? 🙂" myself)
   - -> @nathanlesage has a look 
- [x] Unmet peer dependency warnings
   - -> No idea where they are coming from. electron-forge includes webpack. Should we add webpack explicitly as dev dependency? 
- [ ] Application doesn't shut down easily during development, as the nodeJS main process takes an awful long time to shut down
   - -> Not reproducible on Windows, Seems to be a temporary? problem on Mac
- [x] ESLint spuckt einen monströsen Haufen Warnings aus — sehe ich das richtig, dass das daran liegt, dass ESLint für TypeScript konfiguriert ist? Wenn ja, würde ich sagen, ist es zwar miese Arbeit, ich würde aber sagen nicht die Warnings suppressen, sondern stattdessen da überall mal durchgehen und die Warnings alle mitigaten (sind sehr viele vor allem über implicit type conversions).
   - -> Do after merging: https://github.com/Zettlr/Zettlr/issues/1185
- [x] Wir müssen `scripts/test-gui.js` noch anpassen!
   - -> Try to fix it, but doesn't work (get ENOENT failure, but this also happened with the old version that called "electron" directly)
  - Idea: check-in test resources + config file, remove script completely and replace it by a vs code launch config
- [x] ARM und Linux prepackaged stuff muss noch in den `release`-commands angepasst werden.
  - -> Done
- [ ] macOS-build fails (kann wegen des Leerzeichens sein): Command failed: codesign --sign 1D16A3B8ED07139F88D1598423DD5646340A4445 --force --timestamp --options runtime --entitlements ./scripts/assets/entitlements.plist ./release/mac/Zettlr.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Helpers/chrome_crashpad_handler
Forge baut aber korrekt. Nur es fehlen alle assets (Schriften, SVGs, etc.)
   - -> Tested again under Windows, and everything works. Strange
- [ ] Was ist mit dem sandboxing fix (`scripts/afterPack.js`)? Der muss auch migriert werden, sonst funktioniert Linux nicht out of the box.
   - -> @nathanlesage has a look
- [x] Quicklook, Print und LogViewer windows werden derzeit glaub ich nicht respektivert und noch nicht mitgebundled. Yup, die Dinger sind komplett broken.
   - -> Should work again


I'm putting an "approve" here because I do feel very confident that most of this is actually pretty decent, and I'm a fan. There are actually only two or three construction sites (themes, LESS, CodeMirror plugins) where I can already see how we might solve it!

source/renderer/sidebar/store.js
```diff
@@ -11,9 +11,9 @@
  * END HEADER
  */
 
-const Vue = require('vue')
-const objectToArray = require('../../source/common/util/object-to-array')
-const findObject = require('../../source/common/util/find-object')
+const Vue = require('vue').default
```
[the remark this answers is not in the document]
tobiasdiez: Removing the `default` yields Uncaught TypeError: Vue.set is not a function.

Aigeruth, line 14:
```suggestion
import Vue from 'vue'
```

I think, these two are equivalent.

source/renderer/sidebar/tree-item.vue
```diff
@@ -95,8 +95,8 @@
 
 <script>
 // Tree View item component
-const findObject = require('../../source/common/util/find-object.js')
-const { trans } = require('../../source/common/lang/i18n')
+const findObject = require('../../common/util/find-object.js')
+const { trans } = require('../../common/lang/i18n')
```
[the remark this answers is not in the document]
Aigeruth: IMHO, it should be not break anything. However, I would consider doing the `require` -> `import` migration in a separate pull request by adding the [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import). It might come with some autofixing capabilities.

source/main/providers/citeproc-provider.js
```diff
@@ -43,9 +43,8 @@ class CiteprocProvider {
     this._mainLibrary = ''
     // The Zettlr internal preview of these citations will always use Chicago,
     // because (a) it's just a preview, and (b) Chicago is the default of Pandoc.
-    this._styleID = 'chicago-author-date'
     this._lang = global.config.get('appLang')
-    this._mainStyle = fs.readFileSync(path.join(__dirname, `../assets/csl-styles/${this._styleID}.csl`), 'utf8')
+    this._mainStyle = require('./../assets/csl-styles/chicago-author-date.csl')
```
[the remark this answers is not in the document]
tobiasdiez: Tobi

source/renderer/assets/codemirror/zettlr-mode-multiplex.js
```diff
@@ -1,7 +1,7 @@
 /* global CodeMirror define */
 // ZETTLR SPELLCHECKER PLUGIN
 
-const highlightingModes = require('../../../common/data').highlightingModes;
+const highlightingModes = require('../../../common/data.json').highlightingModes;
```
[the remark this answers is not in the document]
tobiasdiez: Tobi

source/renderer/assets/codemirror/zettlr-plugin-render-images.js
```diff
@@ -31,9 +31,9 @@
    * Creates a definite absolute URL if the information suffices.
    * @param {string} base The base path to be used
    * @param {string} fragment The URL to be converted, either relative or absolute
-   * @returns {string} The converted absolute URL with a cachefree-parameter.
+   * @returns {string} The converted absolute URL.
    */
-  function makeAbsoluteCachefreeURL (base, fragment) {
+  function makeAbsoluteURL (base, fragment) {
```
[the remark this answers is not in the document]
tobiasdiez: Tobi

source/renderer/zettlr-body.js
tobiasdiez: Tobi: store current theme + unuse it

webpack.plugins.js
tobiasdiez: Tobi: include in renderer

package.json
tobiasdiez: Tobi: Drop 32, electron-forge-maker-nsis for future

source/renderer/zettlr-body.js
tobiasdiez: Done

source/renderer/assets/codemirror/zettlr-mode-multiplex.js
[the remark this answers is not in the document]
tobiasdiez: Moved the array from the json file to here, as dicsussed in the call.

webpack.plugins.js
tobiasdiez: Done

source/renderer/assets/codemirror/zettlr-plugin-render-images.js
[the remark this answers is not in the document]
tobiasdiez: I now prevented caching using html headers (for the safe-file protocol). Works nicely for me.

source/renderer/dialog/custom-css.js
```diff
@@ -56,7 +56,7 @@ class CustomCSS extends ZettlrDialog {
             // We have to create the element
             global.ipc.send('get-custom-css-path', {}, (ret) => {
               let lnk = $('<link>').attr('rel', 'stylesheet')
-              lnk.attr('href', 'file://' + ret + '?' + Date.now())
+              lnk.attr('href', 'safe-file://' + ret)
```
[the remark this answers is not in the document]
tobiasdiez: See above: now handled in file protocol

source/renderer/zettlr-renderer.js
tobiasdiez: See above: caching is now handled via headers for the safe-file protocol

source/global.d.ts
```diff
@@ -0,0 +1,21 @@
...
+      notify: any
+      notifyError: any
+      ipc: any
+      citeproc: any // CiteprocProvider
+      config: any
+      application: any
+      typo: any
+      filesToOpen: any
+      preBootLog: any
+      tippy: any
```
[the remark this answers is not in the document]
tobiasdiez: Good point. Tracked in https://github.com/Zettlr/Zettlr/issues/1186.

source/renderer/sidebar/store.js
tobiasdiez: It's probably best to do this as part of the typescript migration https://github.com/Zettlr/Zettlr/issues/1186 (for which imports are the preferred way)

tobiasdiez:
Ok, I went through everything and fixed (almost) all of the open points. Added remarks to your big list in https://github.com/Zettlr/Zettlr/pull/1060#pullrequestreview-462775513, and added short remarks in the other PR comments.

So from my side this should be ready to go now.

nathanlesage:
I've migrated everything from the checklist to their own respective issues that can be dealt with after this PR. However, as discussed, we need it to work without any visible errors, before I will merge this, because errors will pop up any way once we try to run the build pipeline, and I would like to only focus on them, so everything we can do now we should be doing now and not later! :)

Here's the abridged version of the list of things todo before merge:

- [ ] Does it release correct packages that can be opened/installed/used on all platforms (special emphasis to the Windows-question)? (I will check why the macOS build is failing, and provide feedback, as to Linux, I'll test out if the build works, and if I can get VirtualBox running, also test if they install)
- [ ]  Application doesn't shut down easily during development, as the nodeJS main process takes an awful long time to shut down (I will check this)
- [ ] macOS-build fails (kann wegen des Leerzeichens sein) (I will check this, as I have the certificate in my key chain and can investigate)
- [ ] Was ist mit dem sandboxing fix (scripts/afterPack.js)? (I will investigate what we need to adapt, if anything)

tobiasdiez:
Nice! I just fixed the ESlint errors, so one thing less to do. Yeah!

nathanlesage:
Soooo, I'm utterly sorry for leaving this aside for the past week, my head was/is filled with other stuff, but I'm finally at it!

So the macOS build failure was apparently in part because I didn't have Wifi on the train in Lower Saxony, because now it built just fine. However, we need to point explicitly to the app-Bundle: `--prepackaged out/Zettlr-darwin-x64/Zettlr.app` otherwise the licenses will be included in the DMG, but that is really a minor thing for later.

(We still might want to add this in the future, but we never had that so we don't need that right now)

Soo, then to the linux test (on a shitty Xubuntu VM, but better than nothing, I guess): It frankly didn't work. The VM always hung up on me whenever I tried starting it, but there seems to be still some error in the build process.

So to everyone reading this, if anyone could test it out, it would be awesome!

As for the rest: I think there's everything else in place, and we can finally merge this thing after Linux builds are confirmed! I'll do this tomorrow, as I'm too tired right now, but I think tomorrow we can merge!

tobiasdiez:
No worries! I've found some other procrastination projects this week ;-)

But this sounds perfect! I don't have a linux machine lying around, so this is hard for me to test.

tobiasdiez:
@nathanlesage any progress with the linux builds?

nathanlesage:
Hey @tobiasdiez, I was still not able to test this out properly, but decided to just merge this PR later this evening. I will be releasing a 1.8 alpha immediately after so that we can just test it in the wild.

Sorry btw for going dark for a week, something positive has happened in my private life which has required most of my time ;)

nathanlesage:
So it's merged, now we can look out for bugs and errors and fix them before releasing a `1.8.0-alpha`!

tobiasdiez:
Thanks @nathanlesage. Let me know if you or somebody else encounters any issues and I'll try to fix them as soon as possible.

PR #1062, opened 2020-07-13
Phase out jQuery

nathanlesage:
## Description

In response to the great initiatives from @Aigeruth and @tobiasdiez in attempting to clean up the application codebase, I'd pay my fair share by finally doing some chores I've had planned all along; especially in order to make the PRs #1060 and #938 easier to implement as jQuery does prove to be somewhat a hassle and is not really necessary for an Electron application.

## Changes

Remove all jQuery code and migrate to vanilla JS

## Additional information

It's about time to ditch this old behemoth. Ideally, this also makes the app faster as there is no boilerplate code running in between the native function calls, plus we get rid of some kilobytes of dependencies that attaches itself to the global object

Tested on: macOS 10.15.5


source/print/zettlr-print-window.js
```diff
@@ -49,46 +50,78 @@ class ZettlrPrintWindow {
...
+    this._iframe.setAttribute('src', name)
     this._reposition() // Initial reposition
   }
 
   _act () {
     // Activate the window controls.
-    $('.windows-window-controls .close').click((e) => { ipc.send('message', { 'command': 'win-close', content: {} }) })
-    $('.windows-window-controls .resize').click((e) => { ipc.send('message', { 'command': 'win-maximise', content: {} }) })
-    $('.windows-window-controls .minimise').click((e) => { ipc.send('message', { 'command': 'win-minimise', content: {} }) })
+    let winClose = document.querySelectorAll('.windows-window-controls .close')[0]
```
Aigeruth, line 81: It seems like you are looking for only the first instance of the selector. Have you considered using [querySelector](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)?

```suggestion
    const winClose = document.querySelector('.windows-window-controls .close')
```

Also, you don't plan to reassign the type of the `winClose` variable, I recommend using `const` instead of `let`.

PR #1064, opened 2020-07-13
Fix Table of Contents

Aigeruth:
This addresses a regresssion introduced in #1057: Table of Contents
allows navigating in the document again.

Tested on: macOS 10.15.5 with `yarn test-gui`


nathanlesage:
❤️

PR #1066, opened 2020-07-14
Set desktop Linux desktop icon in BrowserWindow config

13rac1:
## Description

Linux executables do not allow embedding an icon. The icon must instead
be set using the icon option in the BrowserWindow constructor. See
electron-packager docs:
https://github.com/electron/electron-packager/blob/61f6c0c3/docs/api.md#icon

Related: https://github.com/electron/electron-packager/issues/128

## Changes

Added 128x128 logo path to BrowserWindow config for three windows: main, quicklook, and print. Used the image from https://github.com/Zettlr/Zettlr/blob/b73f41b1/resources/icons/png/128x128.png which seems a good balance of filesize vs quality for the usecase.

### Before
image

### After

image

This icon is also used for alt-tab program switching and multi-workspace previews.
image

Tested on: Linux Mint 19.3 MATE


nathanlesage:
Looks good, just one additional request: Would you mind renaming the icon to 128x128.png? This way it would be more in sync with the resources directory, plus if needed we can add more of the icon sizes, depending on what we need. After this, it's good to merge!

13rac1:
Done! I've reported this same issue to Slack, now that I see how simple the solution is.

nathanlesage:
> Done! I've reported this same issue to Slack, now that I see how simple the solution is.

You might also want to point the electron team to this; as I recall that there was some change in the docs of `BrowserWindow`, which prompted me to remove the icon actually a few weeks ago. I had this very setting in the app, but removed it as it didn't seem necessary.

PR #1071, opened 2020-07-14
Add minimal set of jQuery linting rules

Aigeruth:
GitHub published this plugin when they migrated away from jQuery in
2018. It contains more rules that can be enabled gradually and it
recommends native alternatives.

Tested on: macOS 10.15.5

[GitHub's blog post](https://github.blog/2018-09-06-removing-jquery-from-github-frontend/)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,6 +4,9 @@
         "es6": true,
         "node": true
     },
+    "plugins": [
+        "jquery"
+    ],
     "extends": [
         "standard",
         "plugin:vue/vue3-recommended"
@@ -44,6 +47,19 @@
           },
           "svg": "any",
           "math": "any"
-        }]
+        }],
+        "jquery/no-ajax": 2,
+        "jquery/no-ajax-events": 2,
+        "jquery/no-animate": 2,
+        "jquery/no-bind": 2,
+        "jquery/no-fade": 2,
+        "jquery/no-load": 2,
+        "jquery/no-param": 2,
+        "jquery/no-serialize": 2,
+        "jquery/no-sizzle": 2,
+        "jquery/no-slide": 2,
+        "jquery/no-toggle": 2,
+        "jquery/no-when": 2,
+        "jquery/no-wrap": 2
     }
 }
--- a/package.json
+++ b/package.json
@@ -194,6 +194,7 @@
     "eslint-config-standard": "^14.1.1",
     "eslint-loader": "^4.0.2",
     "eslint-plugin-import": "^2.22.0",
+    "eslint-plugin-jquery": "^1.5.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-standard": "^4.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2987,6 +2987,11 @@ eslint-plugin-import@^2.22.0:
     resolve "^1.17.0"
     tsconfig-paths "^3.9.0"
 
+eslint-plugin-jquery@^1.5.1:
+  version "1.5.1"
+  resolved "https://registry.yarnpkg.com/eslint-plugin-jquery/-/eslint-plugin-jquery-1.5.1.tgz#d6bac643acf9484ce76394e27e2b07baca06662e"
+  integrity sha512-L7v1eaK5t80C0lvUXPFP9MKnBOqPSKhCOYyzy4LZ0+iK+TJwN8S9gAkzzP1AOhypRIwA88HF6phQ9C7jnOpW8w==
+
 eslint-plugin-node@^11.1.0:
   version "11.1.0"
   resolved "https://registry.yarnpkg.com/eslint-plugin-node/-/eslint-plugin-node-11.1.0.tgz#c95544416ee4ada26740a30474eefc5402dc671d"


nathanlesage:
There is not much jQuery in the codebase either way (checked yesterday, plus due to that the PR will make it next week), so I do not yet see the benefits of adding more rules for a week.

Instead of opening new PRs, I would really like to speak with you + Tobias on the two bigger PRs, because you seem to be rushing things way too far without a clear goal in mind, and I would like to coordinate that process first to not introduce problems into the codebase. So I would really like to reiterate my wish to have a call with you both first, to make sure we are all on the same page. Would this work for you?

Aigeruth:
When all rules are enabled in this plugin, it reports 235 instances of jQuery being used. That means a large enough change to me to do it as a series of steps to reduce the risk or enable collaboration between contributors.

In my opinion, this rule set should live longer in the codebase than a week. It is one thing to remove the *current* usage of jQuery, but contributors might try to use it again. I understand that those new usages are likely to be caught during code review, but it is even nicer for them to be notified upfront before opening a PR. It basically codifies expectations.

These small PRs that I have open so far are all about reducing the size of #938 and #1060. #1055 outlines the steps that can be taken safely with minimum risk to enable Electron Forge or electron-webpack. Smaller changes are easier to review, easier to test, and easier to revert when needed. I have also done most of these changes in #938 because they were needed, however they can be also done without being in #938 and that makes things much easier.

nathanlesage:
> it reports 235 instances of jQuery being used.

Yeah, be that as it may, but none of these make use of extended jQuery-functionality that couldn't be easily replaced.

> It is one thing to remove the current usage of jQuery, but contributors might try to use it again.

This, however, is a much better argument, and that is something I hadn't thought about.

> These small PRs that I have open so far are all about reducing the size of #938 and #1060.

I am not saying anything against these smaller PRs, as they are indeed safe to review. However, both #938 and #1060 will change the file structure a lot, and I would simply like to talk about where to put what, which is easier in person than using GitHub. I cannot accept PRs that fundamentally change the way Zettlr is built without thoroughly understanding what is going on, as I will be the one who has to manage further changes and I would like to feel comfortable with all the changes, which I personally feel is extremely hard using the automated review feature here on GitHub … my sole headache with these PRs is the amount they will change workflows, and while I am fully aware that this is necessary, I need to understand the fundamentals of this.

***

Besides of that, a few comments on the files, but as I mentioned, I now see that this is a good idea to keep these rules until we switch to Vue completely, and after the comments have been resolved, we can merge this.

source/renderer/dialog/preferences.js
```diff
@@ -125,11 +125,7 @@ class PreferencesDialog extends ZettlrDialog {
     $('.dicts-list-search').on('keyup', (e) => {
       let val = $('.dicts-list-search').val().toLowerCase()
       $('.dicts-list').find('li').each(function (i) {
-        if ($(this).text().toLowerCase().indexOf(val) === -1) {
-          $(this).hide()
-        } else {
-          $(this).show()
-        }
+        this.style.display = this.textContent.toLocaleLowerCase().includes(val) ? 'block' : 'none'
```
nathanlesage, line 128: This is an interesting case, because I am extremely sure we shouldn't do it this way, but rather with helper classes – this way we don't need to expect the original `display` to be set to `block`, as this would not necessarily work on inline-elements, so please revert that change, I'll see to including a corresponding helper class in my PR #1062

source/renderer/dialog/tag-cloud.js
```diff
@@ -42,9 +42,9 @@ class TagCloud extends ZettlrDialog {
       $('.dialog .tag').each(function (index) {
         res = res.toLowerCase()
         if ($(this).text().toLowerCase().indexOf(res) < 0) {
-          $(this).hide()
+          this.style.display = 'none'
         } else {
-          $(this).show()
+          this.style.display = 'block'
```
nathanlesage, line 47: Same as above

source/renderer/dialog/preferences.js
Aigeruth: I picked this approach, because my understanding is that this is exactly what jQuery does under the hood: https://github.com/jquery/jquery/blob/master/src/css/showHide.js#L22

I wanted to keep the implementation as close to it as possible to reduce risk and as slim as possible. Happy to use another approach if there is a helper available.

Aigeruth:
> I cannot accept PRs that fundamentally change the way Zettlr is built without thoroughly understanding what is going on, as I will be the one who has to manage further changes and I would like to feel comfortable with all the changes, which I personally feel is extremely hard using the automated review feature here on GitHub … my sole headache with these PRs is the amount they will change workflows, and while I am fully aware that this is necessary, I need to understand the fundamentals of this.

I fully with you on this, that's why I have stopped #938 and created #1055. It is just not good for anyone to have a PR open with 200-300 files changes, because it is not easy to grasp. My intent is go over #1055 and do everything that is not major structural change step by step. Ideally, that should leave the last PR adding webpack as a bundler being pretty small and easy to understand.

GitHub has [a useful extension](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) for Visual Studio Code that allows reviewing PRs in VSCode. It gives wider context for code. Have you tried that before?

tobiasdiez:
Good idea to have some rules in place that prevent a spread of jQuery code.

Concerning the migration, jQuery is actually not a big problem. The only thing that needs to be done is to convert the `\* global $ *\` to proper imports of jQuery so that webpack resolves it correctly. I would suggest to do this for now and then remove/replace the jQuery code later (the migration to forge is already a big project, so it probably better not to mix it with too many other big changes). For this PR, this means I would restrict it to the usage of the eslint plugin and revert the changes of hide/show commands for now.

Aigeruth:
> Concerning the migration, jQuery is actually not a big problem. The only thing that needs to be done is to convert the \* global $ *\ to proper imports of jQuery so that webpack resolves it correctly.

That's correct. It is straight forward: it touched 29 files add adding 1 line to them.

---

I have removed the `no-hide` and `no-show` rules`. I have added a set of rules to start with that passes without any code modifications.

Aigeruth:
@nathanlesage, would you like me to update anything else on this PR? I think, it is ready to go in as is. Once it is merged, I can create follow-up PRs adding one rule at a time to phase out jQuery gradually. It would help you keeping #1062 small and not as an ongoing PR. What do you think?

nathanlesage:
Hey, sorry for taking so long, the past week was horribly busy, so I didn't manage much, but I'll go through the PRs here tomorrow. I also think the PR looks good so far, but I wanna do it all tomorrow with a clean head :)

nathanlesage:
Soooo, sorry for the utterly inexcusable delay, here we go!

PR #1076, opened 2020-07-16
Middle Mouse closes Editor Tabs

Kaan0200:
## Description
Adds ability to close tabs with middle-mouse click.  This is the same behavior that exists on google chrome, visual studio code, and a number of other programs.

## Changes
- Attached onauxclick function onto the tab element
- Added logic to notice if "auxclick" and button is 1 (middle mouse) are clicked when doing click action and then attempt to close tab.

## Additional information
Tested on: Windows 10

### DIFF
--- a/source/renderer/util/editor-tabs.js
+++ b/source/renderer/util/editor-tabs.js
@@ -102,7 +102,6 @@ module.exports = class EditorTabs {
   }
 
   syncFiles (files, openFile) {
-    console.log(`[TABS] Synching ${files.length} files`)
     // First reset the whole tab bar
     for (let instance of this._tippyInstances) {
       instance.destroy()
--- a/source/renderer/util/editor-tabs.js
+++ b/source/renderer/util/editor-tabs.js
@@ -29,7 +29,10 @@ module.exports = class EditorTabs {
     this._tippyInstances = []
 
     // Listen to the important events
-    this._div.onclick = (event) => { this._onClick(event) }
+    this._div.onclick = (event) => this._onClick(event)
+    // Listen for non-primary clicks
+    this._div.onauxclick = (event) => this._onClick(event)
+
     this._div.ondragstart = (evt) => {
       // The user has initated a drag operation, so we need some variables
       // we'll be accessing throughout the drag operation: The currently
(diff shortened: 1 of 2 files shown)


source/renderer/util/editor-tabs.js
```diff
@@ -29,7 +29,10 @@ module.exports = class EditorTabs {
     this._tippyInstances = []
 
     // Listen to the important events
-    this._div.onclick = (event) => { this._onClick(event) }
+    this._div.onclick = (event) => this._onClick(event)
+    // Listen for non-primary clicks
+    this._div.onauxclick = (event) => this._onClick(event)
```
nathanlesage, line 34: Please make sure to add the square brackets to these one-liners. I know they are not necessary, but it emphasises that we're dealing with a function here!

source/renderer/util/editor-tabs.js
```diff
@@ -200,7 +203,12 @@ module.exports = class EditorTabs {
 
     // If given, call the callback
     if (this._intentCallback) {
-      this._intentCallback(hash, (closeIntent) ? 'close' : 'select')
+      // middle mouse click closes tab
+      if (event.type === 'auxclick' && event.button === 1) {
+        this._intentCallback(hash, 'close')
+      } else {
+        this._intentCallback(hash, (closeIntent) ? 'close' : 'select')
```
nathanlesage, line 210: Instead of this if/else-switch, we can make it slightly more readable, by assigning `let middleClick` (or something similar) the result of the if condition and then in the ternary if/else query for `closeIntent || middleClick`.

source/renderer/util/editor-tabs.js
Kaan0200: Interesting, I would have argued that the square brackets indicate we're "ok" with putting logic related to the onclick up here, instead of indicating that the onclick logic should only reside in the lower _internal onClick.

But I'm just having a discussion, for the sake of discussion, I'll change it. 👍

Kaan0200: done.

source/renderer/util/editor-tabs.js
Kaan0200: and done.

nathanlesage:
Thanks! Perfect! :)

PR #1077, opened 2020-07-16
Fix broken citation rendering

maxedmands:
## Description

Previously, in-text citations consisting of only an `@` (not surrounded by [brackets]) didn't render at all -- they just completely disappeared after the cursor was no longer on top of them.

Kapture 2020-07-15 at 23 33 16

Turns out this was because we [recently introduced another capture group in the `citationRE`](https://github.com/Zettlr/Zettlr/commit/b1513e2fa3d05c026760bc21c302076a9a828947#diff-1ca6a5f8f19ef9ca9a775b923bc11084L18-R19) -- so we have to skip that capture group when referring to it later on.

## Changes

Refer to the correct capture group when attempting to pull the citation out of the provided regular expression.

## Additional information

I didn't add anything to the changelog since the bug this fixes was introduced after the most recent release.

Tested on: MacOS Catalina 10.15.4 (19E287)

### DIFF
--- a/source/renderer/assets/codemirror/zettlr-plugin-render-citations.js
+++ b/source/renderer/assets/codemirror/zettlr-plugin-render-citations.js
@@ -57,7 +57,7 @@
 
       // Run through all links on this line
       while ((match = citationRE.exec(line)) != null) {
-        let citation = match[1] || match[2]
+        let citation = match[1] || match[3]
 
         // Now get the precise beginning of the match and its end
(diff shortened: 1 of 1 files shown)


nathanlesage:
Just saw this PR! Sorry for not reacting earlier, but I noticed this myself and already fixed this in upstream! But thank you very much for spotting this! I'm sorry I didn't see it earlier :)

PR #1090, opened 2020-07-18
Use app.getVersion()

Aigeruth:
This allows accessing the version number independently from the
package.json file. It is important as the source/package.json is not
present in a non-two package.json setup.

Tested on: macOS 15.15.6

Related to #1055

### DIFF
--- a/source/renderer/dialog/about.js
+++ b/source/renderer/dialog/about.js
@@ -13,6 +13,7 @@
  * END HEADER
  */
 
+const { remote } = require('electron')
 const ZettlrDialog = require('./zettlr-dialog.js')
 const { trans } = require('../../common/lang/i18n')
 const formatDate = require('../../common/util/format-date')
@@ -37,7 +38,7 @@ class AboutDialog extends ZettlrDialog {
 
   preInit (data) {
     process.getCPUUsage() // First call returns null, so we have to call it twice
-    data.version = require('../../package.json').version
+    data.version = remote.app.getVersion()
     data.uuid = global.config.get('uuid')
 
     // Debug info: Versions, argv, env, and overall process uptime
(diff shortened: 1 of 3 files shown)


source/renderer/dialog/about.js
```diff
@@ -37,7 +38,7 @@ class AboutDialog extends ZettlrDialog {
 
   preInit (data) {
     process.getCPUUsage() // First call returns null, so we have to call it twice
-    data.version = require('../../package.json').version
+    data.version = remote.app.getVersion()
```
tobiasdiez, line 41: `remote` is deprecated. Is there another way (instead of using IPC methods) to get the app version in the renderer?

Aigeruth: `remote` is marked deprecated based on my understanding and it will be needed to be turned on explicitly from Electron 10, but there is no alternative available yet. A user-land based extra package might be scheduled to Electron 12. This is just my understanding based on electron/electron#21408, please, double check it.

I'm not aware of any other method that would work right now, but I'm open to any suggestions. After switching to webpack, I expect that it should be possible to create a constant during build time that contains the version number and eliminate the need for this call.

Aigeruth: Actually, there might be a package already: https://github.com/electron/remote

tobiasdiez: Yeah, the plan is it to move `remote` to a separate package that needs to be imported (via yarn). The intention seems to be that applications shouldn't use it (but it's still there in case you don't want to refactor or can life with the negative sides). 

Anyway, the only other working solution I could find is using the IPC system (which is blocked by #1048, which itself depends on #1055). So let's first get forge right, and then worry about the proper implementation of the remote stuff.

nathanlesage:
Hey, thanks for taking so long for getting back to you!

Good idea, albeit @tobiasdiez we should keep in mind that the remote-module is deprecated, so we shalln't forget to replace the remote-call with an IPC-call …

PR #1095, opened 2020-07-20
Use RequireJS to locate CodeMirror

Aigeruth:
Utilising RequireJS for imports means that these files will not rely on
the two package.js structure. This enables easier migration to webpack
in the future.

Relates to #1055.

Tested on: macOS 10.15.6


nathanlesage:
Good catch! I'll make sure to test it out on the weekend, when I have some free time!

nathanlesage:
Aaaand merging

PR #1096, opened 2020-07-21
Make all paths for inserted images relative & POSIX

yhancik:
## Description
All images paths created when pasting or dropping an image are now **relative** and **POSIX** (using forward slashes). This unifies images paths across platforms (in particular Windows) and insertion method.
In addition, image captions now only contain the filename+extension of the image, instead of the full path.

## Changes
Both during the image-pasting process (`save-image-from-clipboard`) and image-dropping (`this._cm.on('drop')`), the image path is modified *right* before its insertion (`this._app.ipc.send('insert-text')` when pasting, `this._cm.replaceSelection(str)` when dropping). This way, it doesn't interfere with other FS operations that might take place beforehand.

## Additional information
I don't know if it's worth importing `os` to only transform the path on Windows. I assume the path transformation using Node's Path library is optimal enough, but I'm not a Javascript optimisation expert ;) I guess the impact is low as we don't insert hundreds of images at once.

It goes as follow
- `pathToInsert.split(path.win32.sep)` : will split Windows paths using the Windows standard separator (`\`). POSIX paths using forward slashes are not affected.
- `path.posix.join()` :  joins all path segments using the POSIX separator, then normalises the resulting path (as per Node documentation)

Tested on: Windows 10
(erm, tested as in "I tried it and it works, I passed POSIX paths to it manually and it worked to", I, uh, actually never used a testing suite :p)


nathanlesage:
Heya,

thanks for the PR! As mentioned in my comment to yours, I'd suggest we keep the path names as they are (the less we mingle with them the less we have to worry about \*the future\*™), BUT I did not know that I added the full path to the image captions, which obviously doesn't make sense and should be changed!

So if normal image paths in URLs work even on Windows, I'd say to keep the changes minimal. And I just realized we don't even need a backslash-escaper, as there are no backslashes allowed in file names (because that would render it a directory with a file).

How about that? :)

***

N.B.: Why the f*** did I use `p.substr(p.lastIndexOf('/') + 1)}` and not simply `path.basename` in the first place …?

yhancik:
> So if normal image paths in URLs work even on Windows

What is "normal" in this context? ;)  But I trust your judgement anyway. 

Now that I know where to change it, I can always build my own custom version of Zettlr fitting my needs ;)
("my needs" is mostly using Zettlr to edit .md files to online use - on Github, on BeakerBrowser, or a static-site generator, all using relative posix paths)

nathanlesage:
> What is "normal" in this context?

… well, Windows :D 

I mean, I'd like to have many things as close to the library's definitions, as these people have more time and resources on their hands to catch errors and fix bugs, so by relying on, in this case for example, `path`, we can save ourselves a potential source of error that we have to deal with!

tobiasdiez:
The question is which format the path to images have in the markdown document, right?
So whether a) 
```
[this is an image](/path/to/img.png)
```
or b)
```
[this is an image](\path\to\img.png)
```
Then I'm strongly in favor of normalizing them to format a) on all systems (and regardless of what the `path` gives). I couldn't find what the "official" markdown specification is saying about this, but version a) is definitely was Github and other popular markdown tools are using as @yhancik said, see e.g. https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-readmes#relative-links-and-image-paths-in-readme-files. I think it's important to use similar conventions.

nathanlesage:
Hmm yeah, I see what you mean … then the question would be: Are these the only instances where such a path normalization would need to happen on Windows?

yhancik:
The only one that comes to mind is dragging attachments into the edit panel.

Dragging from the sidebar only inserts an internal link with the ID, right? (or did I miss an option somewhere?)

I don't know if there are other cases where Zettlr would write a path for you? 🤔

nathanlesage:
Well, I guess we'll find out eventually if something breaks 😄 

But one last thing: Would you mind wrapping the Windows-code into `if process.platform === 'win32'`? This way we only need to do so on Windows platforms!

yhancik:
Done @nathanlesage 

For consistency sake, I also made images paths relative when you drag & drop an attachment from the sidebar.

I haven't touched paths created when you drag & drop a non-image file from the sidebar, because that would require Zettlr to handle relative paths when you click on a file link. Which is not the case right now, afaik.

I also noticed that paths/filenames with parenthesis break the `` syntax, so I added a small `replace()` to fix that. Let me know if it should rather be in a different pull request, though.

Finally, as you'll see I have split each path operation in consecutive assignations, instead of having one line of code do everything. Let me know whatever you prefer in terms of coding style. I kept it like this to make changes clear :)

yhancik:
That would make sense. I'm only fiddling with the existing code right now, and I admit I don't really have a global view of how the whole project is organised... any recommendation where that method could go?

nathanlesage:
> I also noticed that paths/filenames with parenthesis break the  syntax, so I added a small replace() to fix that. Let me know if it should rather be in a different pull request, though.

Rather, I'd like you to copy over the code from the render-link codemirror plugin, because there I finally solved that damned brackets-problem (this way it'd be consistent)

https://github.com/Zettlr/Zettlr/blob/ab41a04a58d65743889e3fa0af0213594643c294/source/renderer/assets/codemirror/zettlr-plugin-render-links.js#L81-L109

> Finally, as you'll see I have split each path operation in consecutive assignations, instead of having one line of code do everything. Let me know whatever you prefer in terms of coding style. I kept it like this to make changes clear :)

I don't see this in your comments, I must've missed this!

yhancik:
Ok so I moved the path manipulation & link generation to `source/common/util/generate-file-link.js`.

I made it a generalistic function that can be used either for normal file links (`[]()`) or for images (``).

I'm not calling the function for image attachments (in `source/renderer/zettlr-attachments.js`), as `path.basename()` is sufficient to make links relative.

I removed the parenthesis processing, as it should indeed be fixed in `zettlr-plugin-render-images.js`, based on the code in `zettlr-plugin-render-links.js`. I'll make another PR for that if I get a chance to look into it :)

source/renderer/zettlr-attachments.js
```diff
@@ -106,6 +106,10 @@ class ZettlrAttachments {
         if (FILETYPES_IMG.includes(a.ext.toLowerCase())) {
           // Override the drag data with a link to the image
           let uri = decodeURIComponent(a.path)
+
+          // Make the link relative instead of absolute
+          uri = path.basename(uri)
```
nathanlesage, line 111: This won't work because people can have a totally different directory selected, so it needs to be something like (I think) `path.relative(markdownBaseLink, uri)`

Soooo it does make sense to call your new function here, and not mingle with the absolute path here, as this is in the drag-data anyway. This will break in a lot of circumstances!

yhancik:
Hey @nathanlesage .. I'm sorry I've been a bit out of this for a while. 

Considered there have been many changes since I started this (especially with 1.8), is it still worth pushing changes to this PR, or am I just too far behind and it would be best to make a new one?

nathanlesage:
So, sorry for not merging the PR for such a long time, I've been pretty much overworked. If you could resolve the merge conflicts that have arisen by now, I'll immediately merge the PR!

yhancik:
@nathanlesage finally resolved the conflicts and updated for 1.8!

yhancik:
@nathanlesage I hope we'll have something like that in 2.0 ;)

nathanlesage:
Ouh damn, completely forgot about that PR o.O Sorry about that! Will have a look, and now it's pinned so it won't be closed …

source/common/util/generate-file-link.js
```diff
@@ -0,0 +1,39 @@
...
+ * CVM-Role:        <none>
+ * Maintainer:      Yhancik
+ * License:         GNU GPL v3
+ *
+ * Description:     Generates Markdown file link or image
+ *
+ * END HEADER
+ */
+
+const path = require('path')
```
nathanlesage, line 14: Since we had to disable node integration, this won't work in the renderer process. Instead, what we can do now is use `const path = window.path`. (And that won't work in the main process, so since this function is only called from within the MarkdownEditor instance, I'd rather put this function there. Then we can also upgrade it to ES6 using `export default function` and using import statements)

source/common/util/generate-file-link.js
```diff
@@ -0,0 +1,39 @@
...
+  // Transforms Win32 paths (backslashes) into Posix paths (fwd slashes)
+  if (process.platform === 'win32') {
+    pathToInsert = path.posix.join(...pathToInsert.split(path.win32.sep))
+  }
+
+  // Generates the Markdown, using the filename as caption
+  text = `[${path.posix.basename(pathToInsert)}](${pathToInsert})\n`
+
+  // Adding ! if we're making an image link
+  if (isImg) text = '!' + text
```
nathanlesage, line 36: I think it might be more elegant to just return that (e.g. `return (isImg) '!' + text`) and make `text` a `const`

source/main/commands/save-image-from-clipboard.ts
```diff
@@ -119,10 +120,10 @@ export default class SaveImage extends ZettlrCommand {
     fs.writeFile(imagePath, image.toPNG(), (err) => {
       if (err) return global.notify.normal(trans('system.error.could_not_save_image'))
       // Insert a relative path instead of an absolute one
-      let pathToInsert = path.relative(path.dirname(activeFile.path), imagePath)
+      let textToInsert = generateFileLink(imagePath, path.dirname(activeFile.path), true)
```
nathanlesage, line 123: Mh, maybe rename that to `generateRelativeMarkdownLink`…?

source/common/util/generate-file-link.js
nathanlesage: Update: Just saw that the function is being required in both main and renderer process; so I think we'll need the following (ugly) construct:

`const path = (typeof window !== 'undefined') ? window.path : require('path')`

nathanlesage (approved):
After WAY too long, apologies for taking that amount of time, but it's almost ready to merge. There are a few merge conflicts, but the code itself should work perfectly, and I can't think of anything more elegant to implement; so once the conflicts and the one problem in the utility function are being addressed we can finally merge this …

Sorry again for the huge delay!

yhancik:
No worries, only happy it's eventually integrated into Zettlr :D

nathanlesage:
Since this PR is relatively old, I have now had a look and I think mostly the issues here have been addressed in commit e6561822

PR #1101, opened 2020-07-23
Allows ARM64 Builds on Windows

### LINKED ISSUE #890: [ENHANCEMENT] Offer Windows ARM builds
## Description

I had expected the installer to install a 32 bit version based on the download link title but it seems it installs the 64 bit version even on ARM. This results in a windows error on opening "This app can't run on your PC".

## Reproducing

Run the latest installer on an ARM based windows device. Tested with the latest stable and beta.

## Expected behaviour

Installer should detect the architecture and install the 32 bit version until an ARM build is available.

## Platform
 - OS and version: Windows 10 Home V2004
 - Zettlr Version: 1.6.0 & 1.7.0-beta.9
 - Screen Resolution: 2880x1920

smitec:
## Description

Adds an ARM64 target for windows.

## Changes
- Adds ARM64 to the NSIS config.
- Fixes a resource location that was incorrect for windows builds.

## Additional information
Installs correctly from the installer when tested on an ARM windows build.

Tested on: Windows 10 Home on Surface Pro X

Fixes #890

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,6 +90,17 @@ jobs:
         with:
           name: Zettlr-${{ steps.pkg.outputs.version }}.exe
           path: ./release/Zettlr-${{ steps.pkg.outputs.version }}.exe
+      - name: Build Windows NSIS installer arm64
+        if: matrix.os == 'windows-latest' # Only if the job runs on Windows
+        # Make sure to sign the windows installer using the corresponding SSL EV.
+        run: CSC_LINK=${{ secrets.WIN_CERT }}  CSC_KEY_PASSWORD=${{ secrets.WIN_CERT_PASS }} npm run release:win-arm
+        shell: bash # I'm sorry PowerShell, but I'm not rewriting code for you.
+      - name: Cache Windows release arm64
+        if: matrix.os == 'windows-latest' # Only if the job runs on Windows
+        uses: actions/upload-artifact@v1
+        with:
+          name: Zettlr-${{ steps.pkg.outputs.version }}-arm64.exe
+          path: ./release/Zettlr-${{ steps.pkg.outputs.version }}.exe
       # MACOS ================================================================
       - name: Build macOS DMG file
         if: matrix.os == 'macos-latest' # Only if the job runs on macOS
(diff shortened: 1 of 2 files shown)


tobiasdiez:
I tried installing under Windows x64, and it still works. The images also seem to be correct now:
image
image

One disadvantage (I think) is that the Windows installer increased in size and is now at 210mb (up from 135mb, which was my last installer that I built two months ago).

nathanlesage:
Thanks for fixing the image paths!

However, a 200MB installer …? I'm afraid people are going to start throwing things at me if we simply bundle everything in one single installer :S 

I mean, 32bit and 64bit should be bundled together, but I think ARM (and possibly a portable version) should be bundled in a second installer (which also makes sense for users plus more verbose that there is actually an ARM-compatible binary in there) …

The thing is, I looked it up and I think electron-builder doesn't have the ability to just spit out two binaries by itself, so maybe we can migrate _again_ to custom build scripts and feed it to the builder separately?

tobiasdiez:
Maybe it's possible with forge?

nathanlesage:
> Maybe it's possible with forge?

That would be beautiful!

smitec:
Made some changes to allow a single ARM build, installer comes out at about 68MB.

For an unknown reason it needed to include the ia32 as well to get nsis to build it correctly.

Updated the build scripts to include it as its own EXE but those may need a careful check.

nathanlesage:
> Made some changes to allow a single ARM build, installer comes out at about 68MB.

Bueno!

> For an unknown reason it needed to include the ia32 as well to get nsis to build it correctly.

Maybe it wants to include both a 32bit and 64bit ARM …? (Is ARM even produced in a 32bit arch?) Unless it's a bug. But the way you found to ensure electron builder outputs two binaries is clever, I like that approach.

So, before merging, I'd just like you (and possibly other people with Windows) to just double-check the installers, ideally so that they work on 32bit-computers, 64bit-computers and ARM-based computers. If this turns out good, I'll gladly merge this! :) 

I'll also promote this on Twitter, maybe we get some more testers.

smitec:
I've tested the build of `release:win-arm` again on the surface book and can confirm that is working.
I can also confirm that `release:win` works on by desktop under x64.

I don't have a 32 bit machine to test on at the moment.

nathanlesage:
> I don't have a 32 bit machine to test on at the moment.

Well, you have the 32bit-flag, so if the installer is about the same size as the current ones on the releases, this should indicate that there's both the 64bit and the 32bit binary in there. Is the built installer the same size as, e.g., the 1.7.2 release-EXE? If so, I don't see any problems merging and offering ARM builds beginning with the next iteration :)

smitec:
Looks to be the same, 134Mb.

nathanlesage:
Perfect, ありがとう！

PR #1108, opened 2020-07-23
Changed 'four' to 'five' for number of themes

jsstanley:
## Description
Zettlr 1.7.1 now has five themes instead of four (🎉) so the project's README needed updating.

## Changes
I updated the number of themes mentioned in the README from 'four' to 'five'.


nathanlesage:
Whoops! Good catch, thanks a lot! :)

Issue #1127, opened 2020-07-28
[ENHANCEMENT] Allow save as JPG in addition to PNG in Image Paste from Clipboard dialog


13rac1:
## Description

`save-image-from-clipboard` currently only saves PNG format images: https://github.com/Zettlr/Zettlr/blob/34d302cf66b336af7703272f90b3d4eec22c73f5/source/main/commands/save-image-from-clipboard.js#L44-L46
https://github.com/Zettlr/Zettlr/blob/34d302cf66b336af7703272f90b3d4eec22c73f5/source/main/commands/save-image-from-clipboard.js#L105

It is useful to optionally save images to JPG. I tend to use both to avoid the disk space waste of PNG when it is unnecessary. I'm currently changing the format manually after pasting.

## Proposed Changes

UI/UX options:
1. Two buttons: `Save as PNG` and `Save as JPG` on the _Insert Image from Clipboard_ dialog.
2. Radio button _Format_ on the _Insert Image from Clipboard_ dialog.
3. (simplest) Allowing `.jpg` extensions in the filename field.

Electron's `NativeImage` class supports `toJPG()`: https://www.electronjs.org/docs/api/native-image#imagetojpegquality So there is no significant work required in that regard.

![image](https://user-images.githubusercontent.com/195061/88705331-3899ea80-d0c4-11ea-87b5-9c2d1f6ee98a.png)

## Caveats

Options 1 and 2 increase the UI complexity. Option 3 creates a hidden feature, but might be the best option if the help text makes it clear and `.png` is always shown by default. All options would probably require a "JPEG Quality" option in the settings, but a 85% default is sufficient.

13rac1:
No bot. I will get back to this.

13rac1:
No bot. I will get back to this.

nathanlesage:
I find it cute to a certain extent how everyone is hushing the bot away like some dog :D

nathanlesage:
Done! Now, if you provide either extension, the app will save it as either PNG or JPEG, depending on the extension.

PR #1128, opened 2020-07-28
Support saving as JPEG in Insert Image from Clipboard dialog

13rac1:
## Description

When `save-image-from-clipboard` is called with a file name ending with a 'jpg' extension, then the file will be saved as a JPEG image. PNG remains the default format. This is the bare minimum change. Help text should be added to the dialog to make this feature more clear, could be handled in this PR or separately.

## Changes

Adds a file extension check to `main/commands/save-image-from-clipboard.js` to allow saving as JPG.

Example logs:
```
Preparing to save: 20200709110510.md; filepath: /home/user/zettelkasten
Saving image as: /home/user/zettelkasten/b3367857ad1f2c65b9a3a63a923dd68d2.jpg
```

Tested on: Ubuntu 20.04 / Mint 20

For #1127

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - The application now uses `app.getVersion()` instead of requiring the `package.json`-file (thanks to @Aigeruth for implementing).
 - CodeMirror is now required directly within the plugins and is independent of the location of `node_modules`.
 - Zettlr is now also available for ARM64 Windows (thanks to @smitec for implementing).
+- Insert Image from Clipboard saves as JPEG when the 'jpg' file extension is specified.
 
 # 1.7.1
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
Better idea with regard to the important "hidden option" remark you made: I actually do suggest adding a radio button, and have it override a possible extension. Would be easy to implement (would require a new translation string but that shouldn't pose a problem) and this way all users would have a visual indication as to what is actually happening. We always have to think in favour of not so technical users!

13rac1:
> We always have to think in favour of not so technical users!

Definitely, but I wanted to discuss because the additional UI elements increase UX complexity.

nathanlesage:
Absolutely, and I appreciate this! As I said, I think an additional "File Format: (X) PNG ( ) JPG"-radio button should do the trick. PNG/JPG wouldn't need to be translated, the file format thingy would need (which means I'd need the identifier for it)

nathanlesage:
Sorry overlooked the stalebot notification!

nathanlesage:
Hey, sorry for the late answer: If you resolve the conflicts, I can merge the PR finally!

PR #1132, opened 2020-07-29
Show eslint errors in github ui

tobiasdiez:
## Description
The eslint workflow fails in my other PR, but also the upload of the html is not working making it hard to see whats wrong (especially when it's bed time and one shouldn't browse ones PRs anyway).

## Changes
Use [eslint github action](https://github.com/marketplace/actions/eslint-action), which adds the errors directly in the diff and provides a nice overview. See for example here:
https://github.com/googleinterns/step31-2020/pull/68/files
https://github.com/googleinterns/step31-2020/pull/68/checks?check_run_id=925231660

## Additional information


nathanlesage:
Alright, thanks!

PR #1136, opened 2020-07-30
[Codemirror] Util function for parentheses detection in links/images paths

yhancik:
## Description
Images with parentheses in their path weren't rendering properly.
In response to https://github.com/Zettlr/Zettlr/pull/1096#issuecomment-663783451, I used & adapted the existing parentheses-detection code (used for links rendering), so it can be used both for links and images.

## Changes
- created the renderer util function `parenthesisDetection` to handle parentheses detection in links and images
- it uses the original parentheses detection code from `zettlr-plugin-render-links.js`, fixing an issue where the first closing parenthesis in the path was skipped
    - indeed `[link](https://www.google.com/search?q=(inside)outside)` 
    - was rendered as a link to `https://www.google.com/search?q=(insideoutside)`)
- `zettlr-plugin-render-links.js` has been adapted to use `parenthesisDetection`
- `parenthesisDetection` has been added to `zettlr-plugin-render-images.js` so images also support paths with parentheses

## Additional information
It has the following limitations:
1. A link must always contain a matching pair
→ a path like `path(likethis` will not work.

2. Multiples parentheses must be enclosed in each other
→ a path like `link(first)(second)` will not work.

I still think percent-encoding parentheses (`%28` `%29`) in the code, when links & images are inserted would be safer. See how an editor like Brackets will percent-encode spaces as `%20` in autocompleted paths.
That being said, HTML is not Markdown and it might indeed impede readability.

**TODO**: 

- [ ] when the cursor is on an image and the code is shown, syntax highlighting is not applied properly to the end of the image code. I suppose code rendering is using the regex only and not benefiting from the parenthesis detection in image rendering. But where is it, then?

Tested on: Windows 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## GUI and Functionality
 
 - Fixed a bug causing Export Project to fail.
+- Fixed parentheses detection in links and images paths.
 
 # 1.7.2
 
--- a/source/renderer/assets/codemirror/zettlr-plugin-render-images.js
+++ b/source/renderer/assets/codemirror/zettlr-plugin-render-images.js
@@ -12,6 +12,9 @@
 })(function (CodeMirror) {
   'use strict'
 
+  const path = require('path')
+  const parenthesisDetection = require('../../util/parenthesis-detection')
+
   // GENERAL PLUGIN VARIABLES
 
   // Image detection regex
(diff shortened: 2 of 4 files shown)


nathanlesage:
Since this PR is now two years old with no movement, I'm going to close it. Feel free to complain!

PR #1168, opened 2020-08-06
Tag case insensitivity and hyphenation

ClintH:
I recently migrated to Zettlr from a home-brew system. My inconsistencies with tag capitalisation made the 'tag cloud' dialog messy and prevented proper searching (since #TAG wouldn't match usage of #tag). Further, some tags in my YAML front matter had spaces. Although they would show up properly in the tag cloud, it doesn't work to search them. 

Tags parsed from documents were already lower-cased. This patch also lower-cases tags when read in from YAML front matter. This makes the tag cloud display show unique tags regardless of case. 

Tag search terms are also lower-cased. Previously, a search for #tag and #TAG would only match their exact case.

Tags from YAML frontmatter are hyphenated if they contain spaces. This prevents search from breaking ('#world war' doesn't actually match usage of 'world war' as a YAML keyword). The usage of a hyphen might better off be a configuration option? Some might prefer an underscore, or period?

## Changes
Added 'toLocalLowerCase()' calls and .split(' ').join('-) at a few places.

## Additional information
Not sure if this issue is something others get annoyed by, or whether it's worth merging into the project. I needed it though :)

Tested on: Windows 10 v2004, Node 14.7.0, yarn 1.22.4

### DIFF
--- a/source/main/modules/fsal/fsal-file.js
+++ b/source/main/modules/fsal/fsal-file.js
@@ -222,13 +222,13 @@ function parseFileContents (file, content) {
   while ((match = tagRE.exec(mdWithoutCode)) != null) {
     let tag = match[1]
     tag = tag.replace(/#/g, '') // Prevent headings levels 2-6 from showing up in the tag list
-    if (tag.length > 0) file.tags.push(match[1].toLocaleLowerCase())
+    if (tag.length > 0) file.tags.push(match[1].toLocaleLowerCase(global.config.get('appLang')))
   }
 
   // Merge possible keywords from the frontmatter
   if (file.frontmatter && file.frontmatter.hasOwnProperty('keywords')) {
     // Lower-case (as with regex-extracted tags) and replace spaces with dashes
-    file.frontmatter.keywords = file.frontmatter.keywords.map(k => k.toLocaleLowerCase().split(' ').join('-'))
+    file.frontmatter.keywords = file.frontmatter.keywords.map(k => k.toLocaleLowerCase(global.config.get('appLang')))
 
     file.tags = file.tags.concat(file.frontmatter.keywords)
   }
--- a/source/main/modules/fsal/fsal-file.js
+++ b/source/main/modules/fsal/fsal-file.js
@@ -227,9 +227,8 @@ function parseFileContents (file, content) {
 
   // Merge possible keywords from the frontmatter
   if (file.frontmatter && file.frontmatter.hasOwnProperty('keywords')) {
-    // Lower-case (as with regex-extracted tags) and replace spaces with dashes
+    // Lower-case (as with regex-extracted tags)
     file.frontmatter.keywords = file.frontmatter.keywords.map(k => k.toLocaleLowerCase(global.config.get('appLang')))
-
     file.tags = file.tags.concat(file.frontmatter.keywords)
   }
 
--- a/source/main/modules/fsal/fsal-file.js
+++ b/source/main/modules/fsal/fsal-file.js
@@ -222,11 +222,14 @@ function parseFileContents (file, content) {
   while ((match = tagRE.exec(mdWithoutCode)) != null) {
     let tag = match[1]
     tag = tag.replace(/#/g, '') // Prevent headings levels 2-6 from showing up in the tag list
-    if (tag.length > 0) file.tags.push(match[1].toLowerCase())
+    if (tag.length > 0) file.tags.push(match[1].toLocaleLowerCase())
   }
 
   // Merge possible keywords from the frontmatter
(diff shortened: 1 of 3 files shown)


nathanlesage:
Thanks  for the PR! I'll have a look at it at some point (we're currently trying the big #1060 thingy :D)! But I think it's basically okay to merge, I'll get back to you with a better review soon :)

ClintH:
Great! Happy for the feedback when you've got time.

source/main/modules/fsal/fsal-file.js
```diff
@@ -222,11 +222,14 @@ function parseFileContents (file, content) {
   while ((match = tagRE.exec(mdWithoutCode)) != null) {
     let tag = match[1]
     tag = tag.replace(/#/g, '') // Prevent headings levels 2-6 from showing up in the tag list
-    if (tag.length > 0) file.tags.push(match[1].toLowerCase())
+    if (tag.length > 0) file.tags.push(match[1].toLocaleLowerCase())
```
nathanlesage, line 225: Great find! But you should provide the user's language for that to work properly (I think it's `global.config.get('app_lang')`)

source/main/modules/fsal/fsal-file.js
```diff
@@ -222,11 +222,14 @@ function parseFileContents (file, content) {
...
   }
 
   // Merge possible keywords from the frontmatter
   if (file.frontmatter && file.frontmatter.hasOwnProperty('keywords')) {
+    // Lower-case (as with regex-extracted tags) and replace spaces with dashes
```
nathanlesage, line 230: This will confuse users as it will mess with people's expectations, so I would really like you to remove this `map`. We need a better way to do that (and I may have one, see my yet-to-be-written comment at the end)

source/main/modules/fsal/search-file.js
```diff
@@ -54,7 +54,7 @@ module.exports = function searchFile (fileObject, terms, cnt) {
     if (t.operator === 'AND') {
       if (fileObject.name.toLowerCase().indexOf(t.word.toLowerCase()) > -1 || fileObject.tags.includes(t.word.toLowerCase())) {
         matches++
-      } else if (t.word[0] === '#' && fileObject.tags.includes(t.word.substr(1))) {
+      } else if (t.word[0] === '#' && fileObject.tags.includes(t.word.substr(1).toLocaleLowerCase())) {
```
nathanlesage, line 57: See my first comment

source/main/modules/fsal/search-file.js
```diff
@@ -65,7 +65,7 @@ module.exports = function searchFile (fileObject, terms, cnt) {
           matches++
           // Break because only one match necessary
           break
-        } else if (wd[0] === '#' && fileObject.tags.includes(wd.toLowerCase().substr(1))) {
+        } else if (wd[0] === '#' && fileObject.tags.includes(wd.toLowerCase().substr(1).toLocaleLowerCase())) {
```
nathanlesage, line 68: See my first comment

source/main/modules/fsal/fsal-file.js
ClintH: Are both lower-casing and hyphenation a problem, or just the hyphenation? I'd retain the map to do the lower-casing.

ClintH:
Thanks for the comments, Nathan.

I was wondering, does the project have a commitment on tag form or semantics? If the Twitter idea of hashtags are followed, they should be a) case insensitive, b) contain no spaces or other punctuation and c) can't start with numerals. At the moment there's some inconsistencies.

If the Twitter model is followed, the second question is whether should Zettlr be generous with input? ie. ignore case and punctuation for search and grouping of tags. Internally, tags would be normalised: '#corona-virus' and '#CoronaVirus' would be the same, as would 'corona virus' appearing in the YAML keyword list. This would also affect the cloud tag display, but original tags would still be in the Markdown. Searching for '#corona virus' would search for the tag 'corona' and word 'virus' as separate terms, but enclosing it in quotes would be the same as searching for '#coronavirus' or '#corona-virus'. 

Being stricter with tags would improve tag extraction from documents. Avoiding "... #1 favourite..."  erroneously having '#1' extracted as a tag.

I'll change the PR to stick to the lower-casing, as that seems OK, but if there's support for normalising tags I can chip away on that separately.

cdaven:
Is this really still waiting for changes?

nathanlesage:
Bonjour, there are some complex merge conflicts; I'd like to merge finally (after way too long). Would you mind fixing the merge conflicts? Then I'll immediately merge.

ClintH:
Sure, I'll look into it!

PR #1170, opened 2020-08-07
HTML Template: Update Figure CSS

Kangie:
## Description
Figure captions should be, by default, centred under the caption.

## Changes
This PR adds some CSS to the default HTML export template to handle `figure` and `figure figcaption`

To do this, `figure` needs to be rendered as an inline-block.

## Additional information
This isn't perfect - if the figure isn't wide enough to centre the caption, it will be aligned with the left edge of the figure. 

I consider this to be an edge case though - Having a long caption under a very narrow image isn't something you come across too often, and it falls back to the old behaviour when this is the case.

Please squash this PR when you merge - I forgot to update the changelog with my initial commit!

Tested on: 
Win10: 20.04 (Zettlr 1.7.4 via custom CSS)
Gentoo Linux: Compiled from current head of develop branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.7.5
+
+## Export
+
+- HTML Export now centres figure captions underneath figures.
+
 # 1.7.4
 
 ## GUI and Functionality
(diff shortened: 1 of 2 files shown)


Kangie:
Closes #1171

nathanlesage:
Looks good! Here we go

PR #1173, opened 2020-08-07
Added 'space-before-function-paren' ESLint rule

ClintH:
## Description
The ESLint rules didn't reflect the project's coding style of having a space after function names and before parentheses. This improves the linter coverage. (And what was important for me was preventing VS Code from wrongly auto-formatting code when saving files)

## Changes
Added the 'space-before-function-paren' rule.

## Additional information
VS Code also reformatted the remainder of the .eslintrc.json file, it seems to improve readability. If this is not preferred I can edit to avoid this.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,6 +16,10 @@
         "ecmaVersion": 2018
     },
     "rules": {
+        "space-before-function-paren": [
+            2,
+            "always"
+        ],
         // The following rule-changes to JSStandard Coding Style are tradition,
         // as they were included with the default configuration of Atom's ESLint
         // plugin, so we'll keep them here for the time being.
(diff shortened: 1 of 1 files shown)


nathanlesage:
The formatting is fine, a little bit more verbose but that's never a bad idea. Concerning the rule: Isn't that standard ESLint? That's why I never added it, my ESLint always complained with me when I forgot the space in a function declaration …

ClintH:
I'm not sure, but I needed to add it in order to stop VS Code from clobbering the formatting. I figured it might be the case for other contributors too.

nathanlesage:
This is safe again one of those pesky VSCode config things, but it's good to have that rule made explicit anyway!

PR #1182, opened 2020-08-10
Remove usage of jQuery .val() #1055

Aigeruth:
**Note**: The goal of this change is to remove the `.val()` usage, but removing every jQuery usage is not a goal in this change.

Added jquery/no-val eslint rule to prevent future usage.

Tested on: macOS 10.15.6

### DIFF
--- a/source/renderer/dialog/paste-image.js
+++ b/source/renderer/dialog/paste-image.js
@@ -47,15 +47,55 @@ class PasteImage extends ZettlrDialog {
     return data
   }
 
+  get imageNameElement () {
+    return document.getElementById('img-name')
+  }
+
+  get imageName () {
+    return this.imageNameElement.value
+  }
+
+  get imageWidthElement () {
+    return document.getElementById('img-width')
+  }
+
+  get imageWidth () {
+    return parseInt(this.imageWidthElement.value || 0)
+  }
+
+  set imageWidth (width) {
+    this.imageWidthElement.value = width
+  }
+
+  get imageHeightElement () {
+    return document.getElementById('img-height')
+  }
+
+  get imageHeight () {
+    return parseInt(this.imageHeightElement.value || 0)
+  }
+
+  set imageHeight (height) {
+    this.imageHeightElement.value = height
+  }
+
+  get aspectRatioElement () {
+    return document.getElementById('aspect-ratio')
+  }
+
+  get aspectRatio () {
+    return this.aspectRatioElement.value
+  }
+
   postAct () {
     // Activate the sending buttons
     $('#save-cwd, #save-other').on('click', (e) => {
       // The content is either save-cwd or save-other
       global.ipc.send('save-image-from-clipboard', {
         'mode': $(e.target).attr('id'),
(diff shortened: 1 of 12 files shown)


source/renderer/zettlr-body.js
```diff
@@ -657,39 +657,39 @@ class ZettlrBody {
...
+    const replaceWithElement = document.getElementById('replaceWhat')
+    const replaceWith = () => replaceWithElement.value
     // If a regular expression was restored to the find popup, make sure to set
     // the respective class.
-    if (regexRE.test($('#searchWhat').val())) {
-      $('#searchWhat').addClass('regexp')
-      $('#replaceWhat').addClass('regexp')
+    if (regexRE.test(searchFor())) {
+      searchForElement.addClass('regexp')
+      replaceWithElement.addClass('regexp')
```
nathanlesage, line 668: Doesn't it need to be `searchForElement.classList.add('regexp')`? I couldn't find a vanilla `addClass`-function in the JS docs?

source/renderer/zettlr-body.js
```diff
@@ -657,39 +657,39 @@ class ZettlrBody {
...
-      if (regexRE.test($('#searchWhat').val())) {
-        $('#searchWhat').addClass('regexp')
-        $('#replaceWhat').addClass('regexp')
-      } else {
-        $('#searchWhat').removeClass('regexp')
-        $('#replaceWhat').removeClass('regexp')
-      }
+    searchForElement.addEventListener('keyup', (e) => {
+      this._findPopup.searchVal = searchFor()
+      const isInDocument = regexRE.test(searchFor())
```
nathanlesage, line 682: It's not `isInDocument`, but rather `isRegularExpression` (or, `isRegExp` for short) as it only checks if the search value is a regular expression or not

source/renderer/zettlr-body.js
Aigeruth: Good catch!

source/renderer/dialog/paste-image.js
```diff
@@ -47,15 +47,55 @@ class PasteImage extends ZettlrDialog {
     return data
   }
 
+  get imageNameElement () {
+    return document.getElementById('img-name')
+  }
+
+  get imageName () {
+    return this.imageNameElement.value
```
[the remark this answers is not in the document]
Aigeruth: In this particular case, it is separated to keep consistent with the pattern that other attributes use even though `imageNameElement` is only used once.

I separated the `.value` from the HTML element query, so other attributes (e.g. `classList`) of the element can be accessed easily in the future. Having HTML reference encapsulated in `*Element` getters keeps the number of reference to 1, and that makes it very easy (and safer) to change.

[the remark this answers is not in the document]
nathanlesage: Perfect, thank you!

Aigeruth:
I've updated my change. Let me know if anything else needs to be updated. 👍

nathanlesage:
Looks good now! Thanks again!

Issue #1233, opened 2020-09-03
[ENHANCEMENT]: Figure alignment in HTML Export template


Kangie:
## Description
Figure are not consistently centred in the HTML export.

## Proposed Changes
Add some additional CSS into the template so that all figures are centred in HTML export.
## Additional Information
I'll have a dig into this over the next week or so, just adding the enhancement so that I don't forget / can link to a PR.
I should have looked into it a bit deeper when I was doing the figure caption centring!

PR #1234, opened 2020-09-03
HTML Export: Properly centre figures and figure captions.

Kangie:
Remove some redundant CSS (inline block for figures) and instead use `text-align` in both figure and figcaption, providing a more consistent experience on HTML export, particularly where the figures aren't particularly wide.

Closes Zettlr/Zettlr#1233

Tested on: Gentoo Linux (current head of develop)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fixed the fold-gutter being too close to the text.
 - The editor link autocompletion now respects the setting to use headings level 1 instead of YAML frontmatter titles where possible.
 - The paste image dialog now also provides the original image size as a default value, so that you simply can use the arrow buttons on the field to adjust the image size.
+- HTML export should now centre both figures and figure captions.
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
Do you have some screenshots showing before/after?

Kangie:
@nathanlesage I didn't generate any, but I'll attach some to the PR for current Develop behaviour vs this PR after dinner. :)

nathanlesage:
Ah, I think I see what you mean. So, this PR will be an improvement to the previous one? Well, then if you're happy with how it works so far, there's nothing against merging! ;)

Kangie:
Yeah, should be. It's a lot more consistent now across the board. 

As we're using the Pandoc `implicit_figures` extension, if a user wants something to not be handled as a figure they can use the following syntax to make them inline (pulled from pandoc doco, but I have tested this!):

```MarkDown
![This image won't be a figure](/url/of/image.png)\

![This image also won't be a figure](/url/of/image.png)\

![This image will be a figure](/url/of/image.png)
```
This will basically place the `img` tag on a line with a `p` tag (so, inline because it's no longer an image by itself on a line with alt text).

nathanlesage:
Alright, thanks for clarifying; merging …!

PR #1243, opened 2020-09-05
Remove usage of jQuery .text() #1055

Aigeruth:
Added jquery/no-text eslint rule to prevent future usage.

Rest of the added rules reported no usage, they are there to prevent
future usage.

Tested on: 10.15.6


source/renderer/assets/codemirror/zettlr-plugin-render-images.js
```diff
@@ -26,28 +26,6 @@
...
 
-  /**
-   * Creates a definite absolute URL if the information suffices.
-   * @param {string} base The base path to be used
-   * @param {string} fragment The URL to be converted, either relative or absolute
-   * @returns {string} The converted absolute URL.
-   */
-  function makeAbsoluteURL (base, fragment) {
```
Aigeruth, line 35: ESLint highlighted this as unused function.

nathanlesage: Yeah, please don't delete, there's something off since we switched to forge, so we need to check what we need and what not!

Aigeruth: No problem. I've added it back. It makes the linting step fail, but I assume that's okay.

Kangie: I've submitted a PR that should incidentally fix this for you.

source/renderer/dialog/preferences.js
```diff
@@ -19,19 +19,24 @@ const validate = require('../../common/validate.js')
...
     this._boundCallback = this.afterDownload.bind(this)
     this._textTimeout = null
+  }
 
-    // Build the loading spinner that we need for the downloading indication
-    this._spinner = $('<div>').addClass('sk-three-bounce')
-    for (let i = 1; i < 4; i++) {
-      this._spinner.append($('<div>').addClass('sk-bounce' + i).addClass('sk-child'))
-    }
+  get spinner () {
```
nathanlesage, line 32: Is this function in the constructor …?

source/renderer/util/template.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
nathanlesage, line 1: Please rename the file to 'render-template.js', please feel free to also add unit-tests.

source/renderer/zettlr-attachments.js
```diff
@@ -39,50 +40,64 @@ const FILETYPES_IMG = [
...
     this._open = !this._open
+    this.container.classList.toggle('open', this._open)
   }
 
   /**
     * Refreshes the list with new attachments on dir change.
     */
   refresh () {
-    this._fileContainer.text('')
+    if (!this.fileContainer) {
```
nathanlesage, line 88: Please reduce this statement to a one-liner!

nathanlesage (approved):
Looks good, just a few minor things! Did you especially test the language downloader and the references-list rendering?

source/renderer/dialog/preferences.js
Aigeruth: No, it is outside of the constructor and it is a property getter on the class.

source/renderer/util/template.js
Aigeruth: 👍 I'll look into that.

source/renderer/zettlr-attachments.js
Aigeruth: I'm usually all about shorter code, but I strongly prefer being explicit with blocks for conditions, because they can lead to bugs and security issues. One verify famous one is the `goto fail;` one:

- https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/
- https://www.martinfowler.com/articles/testing-culture.html

Let me know what you think and if you still prefer to have it as a one-liner.

source/renderer/assets/codemirror/zettlr-plugin-render-images.js
nathanlesage: Yes it does!

source/renderer/zettlr-attachments.js
cdaven: While this is none of my business ...

Personally, I think that omitting brackets for one-line blocks (and even removing the newline) _sometimes_ makes the code look cooler (maybe even cleaner), but not all the time (e.g. if the line becomes very long), and there's no other advantage. Well, sure, you get to save two keystrokes. But this "rule" introduces arbitrariness, which I don't like at all.

Always using brackets for blocks and always inserting a newline after a condition is a rule that can be used consistently, and will almost always make the code easier to read. In my humble opinion. :-)

nathanlesage: > but I strongly prefer being explicit with blocks for conditions, because they can lead to bugs and security issues

Agreed. I personally only tend to do so with these `return`/`continue`-expressions, but as @cdaven said, it's a little bit arbitrary. The thing is that I've used these one-liners all around the code and my inner OCD prevents me from simply leaving it like that. So maybe we can begin to look if we see these one-liners more often and convert them as we encounter them …?

Aigeruth: I've looked into enforcing brackets for `if` statements for an different codebase in the past, and the ESLint [curly](https://eslint.org/docs/rules/curly) rule with the `"all"` setting is the best way to do it.

source/renderer/util/template.js
Aigeruth: I've added some tests.

Have you considered using Jest? https://jestjs.io/

source/renderer/zettlr-attachments.js
```diff
@@ -39,84 +40,107 @@ const FILETYPES_IMG = [
...
-      this._fileContainer.append($('<p>').text(trans('gui.no_attachments')))
+      this.fileContainer.append(renderTemplate(`<p>${trans('gui.no_attachments')}</p>`))
     } else {
       this._attachments = this._renderer.getCurrentDir().attachments
       let fileExtIcon = clarityIcons.get('file-ext')
       for (let a of this._attachments) {
-        let svg = ''
-        if (fileExtIcon) svg = fileExtIcon.replace('EXT', path.extname(a.path).slice(1, 4))
-
-        let link = $('<a>').html(svg + ' ' + a.name)
```
Aigeruth, line 93: I removed jQuery used here in my last update.

Aigeruth:
> Did you especially test the language downloader and the references-list rendering?

Yes, I tested the downloader with multiple languages.

Yes, I tested both adding attachments and making references based on a Zotero export.

source/renderer/util/template.js
nathanlesage: Thanks for the test! Concerning jest: No, not yet, but we can certainly discuss the benefits over mocha in a separate issue!

source/renderer/zettlr-attachments.js
nathanlesage: Okay then let's leave it like this, and enforce that setting at a later time (but please first let us deal with some of the high amount of issues … I want the linter to return green for the tests asap)

nathanlesage (approved):
Let's go!

PR #1247, opened 2020-09-07
Fix image rendering in Electron forge branch.

### LINKED ISSUE #1228: [DEVELOP] Image preview rendering broken in the Electron forge branch
Pinging @tobiasdiez 

Image rendering is broken, neither the `safe-file`-protocol nor the regular way works out, there's something afoul, need to check this.

Kangie:
#  Description
Fix image rendering in electron forge branch.

# Changes
Replaced `makeAbsoluteCachefreeURL` with `makeAbsoluteURL`
makeAbsoluteURL now returns all 'file:' protocol URIs as 'safe-file:', which we have a handler for.

## Additional information
Fixes #1228

Tested on: 
Gentoo Linux.

### DIFF
--- a/source/common/util/make-absolute-url.js
+++ b/source/common/util/make-absolute-url.js
@@ -7,7 +7,7 @@ const path = require('path')
 * @param {string} fragment The URL to be converted, either relative or absolute
 * @returns {string} The converted absolute URL with a cachefree-parameter.
 */
-module.exports = function makeAbsoluteCachefreeURL (base, fragment) {
+module.exports = function makeAbsoluteURL (base, fragment) {
   let urlObject
   try {
     // If it's already a correct URL, we are almost done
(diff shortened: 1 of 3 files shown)


nathanlesage (changes requested):
Thanks for implementing! Just one question: can you include an image, then modify it externally and remove the image preview and have Zettlr re—render it with the changes visible? (This was the reason we had the "cache-breaker" for)

Plus please adapt the corresponding unit-test for the function, which will now break.

Kangie:
Hi Nathan,

The `safe-file://` handler says that it prevents caching of local files, and testing on Linux and Windows indicates that this works for both absolute and relative paths with these changes.

I'll have a look at the unit tests and update my PR!

Kangie:
@nathanlesange

There's actually no unit test for this function. There's one for the function that runs to make paths absolute before export, and one to make valid URIs, but not one for this. I suspect that's because, until recently, this function was part of zettlr-plugin-render-images. (Also, we should standardise terminololgy - URL or URI?)

So, nothing fails. This function returns `safe-file:` because we're not allowed to use `file://` from the renderer.
We should probably develop a test at some point, but I'm honestly pretty fried right now.

Will you accept this PR as-is with the promise of a unit test when I have more free time?

nathanlesage:
Ahhhh right, I've lost the overview during the forge migration - then let's keep it in mind - in general I'd like to have unit tests for everything in the util directory, but that's work in progress, so all good if the other changes are in :)

Kangie:
Yeah, go ahead. This will fix the linting errors in all other PRs as a happy side-effect.

I'm comfortable with the way that images are being handled, and spent all day today working with content using this code.

nathanlesage (approved):

PR #1250, opened 2020-09-08
Adds a 'close all' context menu to tab header. Fixes #1226

### LINKED ISSUE #1226: [ENHANCEMENT]: Close all tabs option in tab context menu.
## Description
Add additional options to the context menu for opened file tabs.

## Proposed Changes
Add additional context menu options to the tab for open files:
- Close all files
- Close all other files

Potentially: Close all files to the Left/Right, depending on feedback. 
This one isn't super useful for me - I'd prefer to be able to close the 100 file I've worked on all at once, but I've seen the feature elsewhere, so if other users will find it useful it might be worthwhile.

Couldn't find an existing enhancement for this after a quick search, apologies if this one already exists!

anoopengineer:
## Description

This is my first PR. Please be gentle.
Adds a new menu option "Close all" which closes all open tab. Fixes #1226 

## Changes
Added a new menu context json in `renderer/assets/context/` directory specifically for context menu for tab header. Earlier the context menu was file.json which is shared with the context menu for the file/directory tree. I feel like adding a new context json is more cleaner that introducing a new scope and making it work with exiting json files. In the future this will enable us to add tab specific menu options like Close all to right, Close others etc

## Additional information
Fixes #1226

All tests passed.
Tested on: Arch Linux

### DIFF
--- a/source/renderer/assets/context/tabs.json
+++ b/source/renderer/assets/context/tabs.json
@@ -1,0 +1,21 @@
+[
+  {
+    "label": "menu.rename_file",
+    "command": "file-rename"
+  },
+  {
+    "label": "menu.delete_file",
+    "command": "file-delete"
+  },
+  {
+    "label": "menu.duplicate_file",
+    "command": "file-duplicate"
+  },
+  {
+    "type": "separator"
+  },
+  {
+    "label": "menu.close_all_tabs",
+    "command": "file-close-all"
+  }
+]
--- a/source/renderer/util/editor-tabs.js
+++ b/source/renderer/util/editor-tabs.js
@@ -131,10 +131,12 @@ module.exports = class EditorTabs {
     // Now make sure that the active tab is visible and scroll if necessary.
     let tabbarWidth = this._div.offsetWidth
     let activeElem = this._div.getElementsByClassName('active')[0]
-    if (activeElem.offsetLeft + activeElem.offsetWidth > tabbarWidth) {
-      this._div.scrollLeft += activeElem.offsetLeft + activeElem.offsetWidth - tabbarWidth
-    } else if (activeElem.offsetLeft < this._div.scrollLeft) {
-      this._div.scrollLeft = activeElem.offsetLeft
+    if (typeof activeElem !== 'undefined') {
+      if (activeElem.offsetLeft + activeElem.offsetWidth > tabbarWidth) {
+        this._div.scrollLeft += activeElem.offsetLeft + activeElem.offsetWidth - tabbarWidth
+      } else if (activeElem.offsetLeft < this._div.scrollLeft) {
+        this._div.scrollLeft = activeElem.offsetLeft
+      }
     }
 
     // After synchronising, enable the tippy
--- a/source/renderer/zettlr-editor.js
+++ b/source/renderer/zettlr-editor.js
@@ -779,6 +779,15 @@ class ZettlrEditor {
   }
 
   /**
+   * Closes all open files
+   */
+  closeAll () {
+    for (let file of this._openFiles) {
+      global.ipc.send('file-close', { 'hash': file.fileObject.hash })
+    }
+  }
+
+  /**
    * Attempts to close an open tab, and returns true if it did.
    * @returns {boolean} True, if a tab has been closed, otherwise false.
    */
--- a/source/renderer/zettlr-context.js
+++ b/source/renderer/zettlr-context.js
@@ -24,7 +24,8 @@ const TEMPLATES = {
   'directory': require('./assets/context/directory.json'),
   'editor': require('./assets/context/editor.json'),
   'file': require('./assets/context/file.json'),
-  'text': require('./assets/context/text.json')
+  'text': require('./assets/context/text.json'),
+  'tabs': require('./assets/context/tabs.json')
 }
 
 /**
@@ -256,7 +257,7 @@ class ZettlrCon {
       // 2. The file name
       // 3. The document itself
       if (elem.hasClass('filename') || elem.hasClass('close')) elem = elem.parent()
-      template = TEMPLATES.file
+      template = TEMPLATES.tabs
       shouldSelectWordUnderCursor = false
       hash = elem.attr('data-hash')
       if (elem.attr('data-id')) attr.push({ 'data-id': elem.attr('data-id') })
--- a/source/renderer/zettlr-renderer.js
+++ b/source/renderer/zettlr-renderer.js
@@ -560,6 +560,13 @@ class ZettlrRenderer {
   }
 
   /**
+   * Closes all open files
+   */
+  closeAllFiles () {
+    this._editor.closeAll()
+  }
+
+  /**
    * Request the renaming of a file
    * @param  {ZettlrFile} f The file, whose name should be changed
    */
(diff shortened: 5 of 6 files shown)


source/renderer/assets/context/tabs.json
```diff
@@ -0,0 +1,21 @@
...
+    "command": "file-duplicate"
+  },
+  {
+    "type": "separator"
+  },
+  {
+    "label": "menu.close_all",
+    "command": "file-close-all"
+  }
+]
```
Kangie, line 21: Nitpick: EOF Newline here please.

Kangie (changes requested):
Looks good. I'm excited for the new functionality. Only a minor nitpick, really!

Edit: It might be worth using `git rebase` to squash all of your commits once you've fixed this, and force-push the new commit to your branch. 

Up to you, the squash can be done when the PR is merged, too!

source/renderer/util/editor-tabs.js
```diff
@@ -131,10 +131,12 @@ module.exports = class EditorTabs {
     // Now make sure that the active tab is visible and scroll if necessary.
     let tabbarWidth = this._div.offsetWidth
     let activeElem = this._div.getElementsByClassName('active')[0]
-    if (activeElem.offsetLeft + activeElem.offsetWidth > tabbarWidth) {
-      this._div.scrollLeft += activeElem.offsetLeft + activeElem.offsetWidth - tabbarWidth
-    } else if (activeElem.offsetLeft < this._div.scrollLeft) {
-      this._div.scrollLeft = activeElem.offsetLeft
+    if (typeof activeElem !== 'undefined') {
```
nathanlesage, line 134: Good catch!

source/renderer/assets/context/tabs.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  {
+    "label": "menu.duplicate_file",
+    "command": "file-duplicate"
+  },
+  {
+    "type": "separator"
+  },
+  {
+    "label": "menu.close_all",
```
nathanlesage, line 18: Please rename to, e.g., `menu.close_all_tabs`, because otherwise we would run into problems with regard to windows, and notify me of your chosen identifier, so that I can add it to Zettlr Translate!

nathanlesage (changes requested):
Very good, but please undo your modifications to the language files; we need to add those using the translation system. Apart from that: I really like the idea of adding a context menu, and when the other commands work as expected, I'm happy to merge!

source/renderer/assets/context/tabs.json
anoopengineer: Well there is indeed a new line at the end of the file as shown by the file hexdump below. I am guessing Github diff view doesn't show that?

```bash
hexdump -c  source/renderer/assets/context/tabs.json
0000000   [  \n           {  \n                   "   l   a   b   e   l
0000010   "   :       "   m   e   n   u   .   r   e   n   a   m   e   _
0000020   f   i   l   e   "   ,  \n                   "   c   o   m   m
0000030   a   n   d   "   :       "   f   i   l   e   -   r   e   n   a
0000040   m   e   "  \n           }   ,  \n           {  \n            
0000050       "   l   a   b   e   l   "   :       "   m   e   n   u   .
0000060   d   e   l   e   t   e   _   f   i   l   e   "   ,  \n        
0000070           "   c   o   m   m   a   n   d   "   :       "   f   i
0000080   l   e   -   d   e   l   e   t   e   "  \n           }   ,  \n
0000090           {  \n                   "   l   a   b   e   l   "   :
00000a0       "   m   e   n   u   .   d   u   p   l   i   c   a   t   e
00000b0   _   f   i   l   e   "   ,  \n                   "   c   o   m
00000c0   m   a   n   d   "   :       "   f   i   l   e   -   d   u   p
00000d0   l   i   c   a   t   e   "  \n           }   ,  \n           {
00000e0  \n                   "   t   y   p   e   "   :       "   s   e
00000f0   p   a   r   a   t   o   r   "  \n           }   ,  \n        
0000100   {  \n                   "   l   a   b   e   l   "   :       "
0000110   m   e   n   u   .   c   l   o   s   e   _   a   l   l   _   t
0000120   a   b   s   "   ,  \n                   "   c   o   m   m   a
0000130   n   d   "   :       "   f   i   l   e   -   c   l   o   s   e
0000140   -   a   l   l   "  \n           }  \n   ]  \n                
000014c

```

source/renderer/util/editor-tabs.js
anoopengineer: :+1:

source/renderer/assets/context/tabs.json
anoopengineer: Done. Went with your suggestion of `menu.close_all_tabs`.

Kangie (commented):
Looks good!

source/renderer/zettlr-editor.js
```diff
@@ -778,6 +778,16 @@ class ZettlrEditor {
     return this
   }
 
+  /**
+   * Closes all open files
+   */
+  closeAll () {
+    this._currentHash = null
```
nathanlesage, line 785: I am pretty sure we need to remove this statement, as it should be reset whenever a tab is being closed (which is happening if you send a file-close command – what is happening is that (a) renderer sends command (b) main checks if we can close the file and perform a cleanup (c) main sends a command to actually close the file to the renderer (d) the editor will reset the `this._currentHash` to `null` if the last tab has been closed.

nathanlesage (changes requested):
One last line, and then we can merge it! I will already add the new translation identifier to [Zettlr Translate](https://translate.zettlr.com/) so that we can already begin to translate it!

Thank you for your efforts! :)

source/renderer/zettlr-editor.js
anoopengineer: I added that specifically to handle an edge condition. Steps to reproduce:
1. Open 3-4 tabs
1. Close Zettlr
1. Relaunch Zettlr
1. Zettlr will come up with previous tabs with one tab highlighted.
1. Right click on the highlighted tab and select close all

You'll see that the last tab still remains open. 

I added setting `_currentHash` to `null` to handle this. But your comment made me dig in a bit deeper. To me it looks like there is a bug elsewhere in the system. When closeAll fires off  IPC for `n` files, only `n-1` calls to close() are received causing last file to remain open. This is happening only when you close Zettlr with many tabs and relaunch it. I am not sure what might be causing this.

nathanlesage: > To me it looks like there is a bug elsewhere in the system.

Very likely!

> When closeAll fires off IPC for n files, only n-1 calls to close() are received causing last file to remain open. This is happening only when you close Zettlr with many tabs and relaunch it. I am not sure what might be causing this.

Thank you for digging into this! Let's suggest the following, as we still have time until the next iteration will be released:

1. You remove that statement, as it's not supposed to be in there
2. I will open a bug report with your findings so that we can handle this correctly :) 

Thank you very much for this detailed bug description, because it'll make removing the bug very easy!

anoopengineer: Thank you. I have removed that line.

nathanlesage (approved):
And here we go!

PR #1258, opened 2020-09-10
Disable disk cache for http requests.

Kangie:
This globally disables the on-disk cache for http requests within electron.

Conveniently, https images still work on the develop branch. :D 

This definitely changes the caching behavour of http/https images; large images now visibly take time to be received and rendered by Zettlr once you force the action.

Tested on: Gentoo Linux

### DIFF
--- a/source/main.ts
+++ b/source/main.ts
@@ -24,6 +24,9 @@ import extractFilesFromArgv from './common/util/extract-files-from-argv'
 // Developer tools
 import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
 
+// Disable the disk cache for http requests.
+app.commandLine.appendSwitch('disable-http-cache')
+
 global.preBootLog = [{
   'level': 2, // Info
   // eslint-disable-next-line no-irregular-whitespace


nathanlesage:
I like! But how visibly do they take? Do you think this might be seen as a nuisance?

Kangie:
Honestly, it depends on the speed of your connection and the distance from the server. The biggest issue as far as I see it is the time to first byte - for me in Australia, hitting servers in Europe or the Americas can be ~600ms. If content is hosted on a CDN it's much faster. I barely notice a difference grabbing an image hosted by google, for example.

Particularly large images might be an issue, in terms of UX. Downloading the same huge image 4-5 times might be an extra 20-30 MB+ of data usage for those on restricted connections. 

On the other hand, if you're working with images that large and aren't caching them yourself anyway, you're probably doing it wrong.

The delay before the image is completely rendered is no longer than inserting the image initially (before it's cached) currently.

I'll leave the decision up to you!

nathanlesage:
Hm, this sounds like a case for a settings toggle - the question would be how to implement it, as the config is loaded much later :/

Kangie:
Maybe something like the [safe-file handler](https://github.com/Zettlr/Zettlr/blob/b1e1e36d962ad1d20a61cca46817208fab487cdd/source/main.ts#L147)?

nathanlesage:
Oh, I didn't see that :D But yeah, this does look good … but this would mean we need a protocol handler for http and https, right …? This might even introduce more flexibility and security as we could add more checks for certain stuff the renderer is requesting as we go, right …?

Kangie:
Got some free time coming up, so I'll see what I can do about revising this PR. :)

nathanlesage:
Ahh, yeah, thanks! Sorry I simply forgot the PR over other work and stuff 🙈

Kangie:
I know the feeling!

nathanlesage:
Hey, I'm just revising all the PRs in the hope that I can merge some of them – I have an idea: Let's close this PR for now, but if someone complains that their remote images are not cached, we know where to go for it!

PR #1269, opened 2020-09-14
Add %y and %X placeholders

cdaven:
## Description

Add `%y` and `%X` placeholders for generating Zettelkasten ID and names for new files.

- `%y` is the year with 2 digits
- `%X` is the Unix epoch/timestamp (10 digits)

To allow an ID with only Unix epoch, had to lower the minimum string length requirement of zkn.idGen to 2 characters (for "%X").

As an aside, I don't know how much you want to limit or allow when it comes to generating Zettelkasten IDs. Of course, the default 14 digit timestamp is not as likely to be confused with something else when parsing the full text of notes. But, using two digits for years will "save" two unnecessary characters, and the Unix epoch can represent the same point in time with only 10 characters.

## Additional Information

I have not created language strings for the preferences, as I'm not sure how to go about doing this. Please advice.

Tested on: Windows 10

### DIFF
--- a/resources/templates/dialog/preferences.handlebars
+++ b/resources/templates/dialog/preferences.handlebars
@@ -422,12 +422,14 @@
             {{i18n "dialog.preferences.zkn.intro"}}
             <ul>
               <li><strong>%Y</strong>: {{i18n "dialog.preferences.zkn.var_year"}}</li>
+              <li><strong>%y</strong>: {{i18n "dialog.preferences.zkn.var_year_two"}}</li>
               <li><strong>%M</strong>: {{i18n "dialog.preferences.zkn.var_month"}}</li>
               <li><strong>%D</strong>: {{i18n "dialog.preferences.zkn.var_day"}}</li>
               <li><strong>%W</strong>: {{i18n "dialog.preferences.zkn.var_week_number"}}</li>
               <li><strong>%h</strong>: {{i18n "dialog.preferences.zkn.var_hour"}}</li>
               <li><strong>%m</strong>: {{i18n "dialog.preferences.zkn.var_minute"}}</li>
               <li><strong>%s</strong>: {{i18n "dialog.preferences.zkn.var_second"}}</li>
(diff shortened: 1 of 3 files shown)


nathanlesage:
> and the Unix epoch can represent the same point in time with only 10 characters.

The Unix epoch will run into a huge problem in 18 years, so I'm not sure we should use it (most programs continue to switch away from saving the current date as an integer, and instead use strings, which is much more future-proof). With regard to the two-year-variable: Sure thing! The language strings look good so far.

Did you double-check to add the new variable(s) everywhere in the dialogs where these "Variables: %x, %y, %m" thingy is? I remember I added the support for this in multiple dialogs.

> I have not created language strings for the preferences, as I'm not sure how to go about doing this. Please advice.

Looks perfectly fine. Once your PR is ready to merge, simply tell me all language identifiers that you created, so that I can add them to [Zettlr Translate](https://translate.zettlr.com/)!

Kangie:
> The Unix epoch will run into a huge problem in 18 years, so I'm not sure we should use it.

Arguably, that's an issue with 32-bit `time_t` integer, rather than with the Unix epoch. Representing a 64-bit `time_t` value using 10 decimal digits (so up to 9,999,999,999 seconds) will get you to Saturday, 1 January 2287 before it has to roll over. 

If Zettlr is still in use in 2287, we can fix it then.

cdaven:
> The Unix epoch will run into a huge problem in 18 years

In this case, no. Javascript doesn't even have integers, only `number`, which is "a double-precision 64-bit binary format IEEE 754 value, like double in Java or C#."

Try this on jsconsole.com or in Node:

```js
var d = Date.parse("2999-12-31 23:59:59.123");
new Date(d - 1000).toISOString();
```

`d` is 32503676399123, and subtracting 1000 ms from this gives "2999-12-31T22:59:58.123Z", which is exacly 1 second earlier.

I will have a look at your other comments later, thank your!

nathanlesage:
Alright, you got me there, let's add this to the backlog for January 2700 😄

PR #1270, opened 2020-09-14
Minor changes to language strings

ivernus:
## Description

Minor changes to language files to modify startup alert.

## Changes

Modified language strings to show "You have the newest version..." rather than "You got the latest version..."


nathanlesage:
Please do not modify these files, as they are pulled in from our translation service, so your changes would be overwritten in the next few days. Instead, please go to [Zettlr Translate](https://translate.zettlr.com/) to help translating the app! Thank you for your engagement!

Issue #1264, opened 2020-09-11
[DEVELOP] CSS-Loader and Less-Loader cannot be upgraded


nathanlesage:
Pinging @tobiasdiez @Aigeruth 

---

I just bumped all dependencies; everything works out, but I couldn't upgrade the `less-loader` (would result in the LESS code being unable to be compiled as the corresponding module cannot be found by forge) nor the `css-loader` (would not properly include the CSS files we need for the virtual scroller to work properly). Would be cool if we could check why that is, but unfortunately I didn't have time to investigate.

tobiasdiez:
Probably one of the breaking changes in https://github.com/webpack-contrib/css-loader/releases/tag/v4.0.0. I'll try to investigate this, but not before next wee.

nathanlesage:
Thanks so much! If I can help, just ping me :)

PR #1279, opened 2020-09-18
Add husky and lint-staged

Aigeruth:
[Husky](https://github.com/typicode/husky) is a package to automatically setup pre-commit hooks. It supports
various linters like eslint.

[lint-staged](https://github.com/okonet/lint-staged) package scopes eslint to staged files only. This can improve
performance and eliminate the noise from unchanged files.

These packages have no impact on functionality, but they can improve developer experience.

_Note:_ I don't expect this setup to automatically limit the scope of the ESLint GitHub Actions only to the files that are changed in a PR. This is something that I intend to further investigate.

Tested on: macOS 10.15.6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ yarn.log
 
 # Don't commit ESLint error reports
 eslint_report.htm
+.eslintcache
 
 # Ignore electron-forge build output
 out/
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "reveal:build": "node scripts/update-reveal.js",
     "test": "mocha",
     "test-gui": "node scripts/test-gui.js",
-    "lint": "eslint --ext js,vue,ts source"
+    "lint": "eslint 'source/**/*.{js,ts,vue}'"
   },
   "config": {
     "forge": "./forge.config.js"
@@ -211,8 +211,10 @@
     "fork-ts-checker-webpack-plugin": "^5.1.0",
     "handlebars": "^4.7.6",
     "handlebars-loader": "^1.7.1",
+    "husky": ">=4",
     "jsdom": "^16.4.0",
     "less-loader": "6.2.0",
+    "lint-staged": ">=10",
     "lorem-ipsum": "^2.0.3",
     "mocha": "^8.1.3",
     "node-loader": "^1.0.1",
@@ -226,5 +228,13 @@
     "uglify-js": "^3.10.4",
     "vue-loader": "^15.9.3",
     "vue-template-compiler": "^2.6.12"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,vue,ts}": "eslint --cache --fix"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1000,7 +1000,7 @@ ansi-colors@^3.0.0:
   resolved "https://registry.yarnpkg.com/ansi-colors/-/ansi-colors-3.2.4.tgz#e3a3da4bfbae6c86a9c285625de124a234026fbf"
   integrity sha512-hHUXGagefjN2iRrID63xckIvotOXOojhQKWIPUZ4mNUZ9nLZW+7FMNoE1lOkEhNWYsx/7ysGIuJYCiMAA9FnrA==
 
-ansi-escapes@^4.2.1:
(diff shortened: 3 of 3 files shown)


tobiasdiez:
In the end, it comes down to developer preference of how you want to ensure that your code is free of lint errors. Some prefer to have your IDE show linter warnings in the code, some let their IDE reformat the code on git commit, some like husky/git precommit and some like to cleanup their code at the very end by running eslint manually. It's good to support all these development styles.

However, if I understood it correctly, husky is runs always with this configuration. I'm not a big fan of this, because it forces devs to use precommit. Can it be configured that devs have to enable husky?

nathanlesage:
As we already have a lot of problems regarding the electron-forge migration that we need to mitigate I don't want to add even more new packages. I want to get back to my policy of only allowing new dependencies in the package.json if they are without alternative.

However, as we already have eslint, and it is working, I don't see a reason why we would need to change that. Furthermore, there is not much ground to gain, as ESLint is pretty fast and takes only a few seconds to work, so there's really no need to add something further where we have to keep configurations and dependencies up to date.

First, please let us deal with the issues already at hand, before introducing potential new ones.

Aigeruth:
What I picked up based on your previous comments on other PRs that eslint generates some noise currently that you would like to address this. I agree with you that having extra linting errors on PRs that are not related might lead to other slipping through and this is a risk.

Husky and lint-staged should scope eslint to the changed files only automatically, so new contributors don't face the existing linting failures locally. The other value that Husky potentially brings, that it sets the pre-commit hook up automatically, so potentially, every PR gets linted before it is opened. That should reduce things to review on your side and free up some time.

Yes, Husky and lint-staged are extra dependencies, however, they are development dependencies only and won't be part of the shipped package. As most of the configuration for these to out of box, I don't see much risk in hurting the developer experience and user experience is completely intact.

If you think, this change doesn't help in any way, that's okay. Let me know if I can take a look on any of the electron-forge issues. I'm only aware of #1264, so I will start there.

nathanlesage:
> What I picked up based on your previous comments on other PRs that eslint generates some noise currently that you would like to address this.

Yes, but not by simply hiding it and not checking said files, but by finalizing the migration to electron-forge and TypeScript. I do not want to hide errors, as this is arguably the worst way to deal with errors, but keep them and address them gradually.

> The other value that Husky potentially brings, that it sets the pre-commit hook up automatically, so potentially, every PR gets linted before it is opened. That should reduce things to review on your side and free up some time.

Well, the linter runs after a PR is opened, and this is more welcoming, as this first and foremost enables the user to open it, we can address linter errors later. Furthermore, having these errors adds more urgency to addressing these issues. So it is temporary and will return to normal once we finalize the issue.

> however, they are development dependencies only and won't be part of the shipped package.

This is no valid argument, as we also need to keep the devDependencies updated, and this does not eliminates the work for me to learn the correct workings. I have had a lot of problems figuring out the config yesterday of electron-forge and electron-builder, so I want to reduce the amount of config files as much as possible, and not add more configurations to it.

Every new configuration increases the mental load on me more and more, and as I'm still the one in charge, I have to decide that I cannot cope with this amount of overview.

> Let me know if I can take a look on any of the electron-forge issues. I'm only aware of #1264, so I will start there.

Thank you for your help, this will help a lot :)

Aigeruth:
The linter should run both locally and from the GitHub Actions. Running locally gives feedback to developers quicker and removes the noise on the open PR. However, we shouldn't rely on people not forgetting to run linters/tests locally or configuration their development environment to do so. Running linters from GitHub Actions is simply a must, so I won't go into that further. 😄 

> but keep them and address them gradually.

This is what the scoping achieves in my opinion. Linting errors are not actual bugs or errors (they might be potential problems), so fixing them gradually is absolutely acceptable. As you codebase evolves, these problematic places will be touched and the linting problems will be address. However, linting errors shouldn't scream on every change and potentially discourage contributions.

Working down on all errors can be still done. It is possible to run eslint for all errors and have dedicated PRs for that. Scoping just allows separating code healing PRs and feature PRs in less noise way.

> we also need to keep the devDependencies updated

Have you tried [Dependabot](https://dependabot.com/) or [renovate-bot](https://renovate.whitesourcesoftware.com/)? I use Dependabot for a few other projects. It can be even configured to auto-merge development dependencies. If find it very useful personally.

---

I just wanted to offer some options here, but if you're happy with the current approach, that's okay too. 👍

PR #1339, opened 2020-10-10
Upgrade less-loader to 7.0.2

Aigeruth:
less-loader does not have less as a dependency anymore, so it has to be
added as a development dependency.

Related to #1264

Tested on: 10.15.7


nathanlesage:
Ah thanks a lot for looking into it -- would you mind fixing the less dependency to the latest (e.g. `^7.0.3`)? This maintains consistency with the rest of package.json!

Aigeruth:
`7.x.x` and `^7.0.3` will be resolved to the same package version: latest patch version within the specified major version (7). IMHO, `7.x.x` is a more intuitive syntax and removes the need to remember what the `^` means. It also has the benefit of not needing to update the `package.json` as frequently, because minor and patch version change will be only reflected in `yarn.lock` and `package-lock.json` and this works well with auto-updaters like Dependabot.

I can update the syntax for consistency.

nathanlesage:
I know but this way the package json is a little bit more verbose! We can surely switch this consistently in the future. Thank you!

PR #1347, opened 2020-10-13
Fix #1346

### LINKED ISSUE #1346: [ENHANCEMENT / BUG?] Override pre-existing text in global search bar with highlighted text when pressing CTRL+Shift+F
## Description

**Current state**: After you searched something globally, the search term(s) remain in the search bar. Now, when you highlight some text in some document, and then press `Cmd/Ctrl+Shift+F` to search globally for it, the content of the search bar will not be updated with the highlighted text. _(Not sure if this behaviour is a bug or not)_

## Proposed Changes

When pressing `Cmd/Ctrl+Shift+F` with some text highlighted, override the content of the global search bar with that of the highlighted text.

## Caveats
I believe none.

## Additional Information
Tested with Zettlr **1.7.5** on Linux Mint 19.2

kyaso:
## Description

This PR fixes #1346. Now, whenever the **searchbar** is focussed and there is some highlighted text in the document, the content of the searchbar will be updated with the highlighted text; previously this only happened when the searchbar was empty.

## Changes

In file `zettlr-toolbar.js` where the event listener `focus` for the searchbar is defined:
- Removed check for empty searchbar content
- Searchbar content (`searchBarInput.value`) is now changed when the first element of `selections`, `selections[0]`, is a string of non-zero length
    - Previously, it was checked whether the array `selections` _itself_ was of non-zero length
        - However, for some reason this check apparently evaluated as true even when nothing was highlighted
- `searchBarInput.select()` now gets called unconditionally
    - When the searchbar content is non-empty, the content gets selected
    - When the content is empty, nothing will happen

## Additional information

Tested on: Linux Mint 19.2

### DIFF
--- a/source/renderer/zettlr-toolbar.js
+++ b/source/renderer/zettlr-toolbar.js
@@ -205,14 +205,12 @@ class ZettlrToolbar {
 
     this.searchBarInput.addEventListener('focus', () => {
       this._autocomplete = this._renderer.getFilesInDirectory()
-      if (this.searchBarInput.value === '') {
-        // Let's prefill this with the selection from the editor if possible.
-        let selections = this._renderer.getEditor().getSelections()
-        if (selections.length > 0) {
-          this.searchBarInput.value = selections[0]
-          this.searchBarInput.select() // Ease of access
-        }
+      // Let's prefill this with the selection from the editor if possible.
+      let selections = this._renderer.getEditor().getSelections()
+      if (selections[0].length > 0) {
+        this.searchBarInput.value = selections[0]
       }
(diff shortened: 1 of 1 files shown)


Kangie:
Looks good. Linting errors are unrelated to your work, and the logic is sound.

Kangie (approved):

cdaven:
I don't understand why this approved PR was never merged, but I did another version of the same fix in #1589.

nathanlesage:
> I don't understand why this approved PR was never merged

Simple: No time, I'm completely overworked. Thanks for reopening.

PR #1373, opened 2020-10-23
Update translation for "edit in place"

jotak:
French translation on tutorial: "Changer le texte de place" means "move the text around" which is different from edit in place, so fixing this text.

  - [ ] I documented all behaviour as far as I could do.
  - [ ] I have tested this extensively and paid extra attention to
  - [ ] I have made use of ESLint using the provided configuration from the
  - [ ] I matched my code-style to the repository (as far as possible).
  - [ ] As far as JS-files are concerned, I made sure to copy (in case of new files)

## Additional information
Just a documentation change, I didn't build or test locally.

cc @framatophe


nathanlesage:
Thank you for the fix! :)

nathanlesage (approved):

PR #1392, opened 2020-11-01
Add %y and %X placeholders

cdaven:
_Sorry, I had to close the previous PR #1269 and recreate it._

## Description

Add `%y` and `%X` placeholders for generating Zettelkasten ID and names for new files.

- `%y` is the year with 2 digits
- `%X` is the Unix epoch/timestamp (10 digits)

To allow an ID with only Unix epoch, had to lower the minimum string length requirement of zkn.idGen to 2 characters (for "%X").

As an aside, I don't know how much you want to limit or allow when it comes to generating Zettelkasten IDs. Of course, the default 14 digit timestamp is not as likely to be confused with something else when parsing the full text of notes. But, using two digits for years will "save" two unnecessary characters, and the Unix epoch can represent the same point in time with only 10 characters.

## Additional Information

I have not created language strings for the preferences, as I'm not sure how to go about doing this. Please advice.

Tested on: Windows 10

### DIFF
--- a/resources/templates/dialog/preferences.handlebars
+++ b/resources/templates/dialog/preferences.handlebars
@@ -422,12 +422,14 @@
             {{i18n "dialog.preferences.zkn.intro"}}
             <ul>
               <li><strong>%Y</strong>: {{i18n "dialog.preferences.zkn.var_year"}}</li>
+              <li><strong>%y</strong>: {{i18n "dialog.preferences.zkn.var_year_two"}}</li>
               <li><strong>%M</strong>: {{i18n "dialog.preferences.zkn.var_month"}}</li>
               <li><strong>%D</strong>: {{i18n "dialog.preferences.zkn.var_day"}}</li>
               <li><strong>%W</strong>: {{i18n "dialog.preferences.zkn.var_week_number"}}</li>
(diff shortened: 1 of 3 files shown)


cdaven:
@nathanlesage, new language strings for Zettlr Translate:

- dialog.preferences.zkn.var_year_two (suggestion in English: "Year (two digits)")
- dialog.preferences.zkn.var_unix_epoch (suggestion in English: "Unix time")

I have not modified source/common/assets/tpl/dialog/preferences.handlebars.js, which contains `<span class=\"info\">Variables: %id, %Y, %M, %D, %W, %h, %m, %s, %uuid4</span>`, as it seems to be auto-generated from resources/templates/dialog/preferences.handlebars, but I don't know how.

Kangie (approved):

nathanlesage:
Apologies for the delay, merging!

> I have not modified source/common/assets/tpl/dialog/preferences.handlebars.js, which contains <span class=\"info\">Variables: %id, %Y, %M, %D, %W, %h, %m, %s, %uuid4</span>, as it seems to be auto-generated from resources/templates/dialog/preferences.handlebars, but I don't know how.

No, that has to be adapted likewise, will do that quickly

Issue #759, opened 2020-05-09
[BUG] - Keyboard shortcuts not working


TheAlmightyDada:
## Description

The shortcuts to focus on the Tree and navigate the tree via the arrow keys are not working.

## Reproducing

Press Ctrl+Alt+T while on the editor.

## Expected behaviour

Shortcut Ctrl+Alt+T should focus on the Tree view and I should be able to navigate files with the arrow keys

Ctrl+Alt+E should focus on the editor. 

## Platform
 - OS and version: Windows 10 Pro; 1903
 - Zettlr Version: 1.6.0

nathanlesage:
> Shortcut Ctrl+Alt+T should focus on the Tree view

This was always working, and never broken

> I should be able to navigate files with the arrow keys

This, however, did not work. I just fixed that!

TheAlmightyDada:
Hi Nathan,

Must be something on my end then?

When I hit "Ctrl + Alt + E" I get this: "é"

When I hit "Ctrl + Alt + T" with the editor in focus, nothing happens.

Thanks for taking a look at the arrows, I am now on the beta 1.7 release.

nathanlesage:
> When I hit "Ctrl + Alt + E" I get this: "é"

Wait a moment … Ctrl+Alt+E? I'm not quite sure if the shortcut wasn't Ctrl+Shift+E…? If the latter works, it's an error in the documentation!

TheAlmightyDada:
I don't think those shortcuts are actually listed in the documentation. I found them while looking at #165 

Thanks for having a look, your efforts have made Zettlr an invaluable tool, day to day.  `Shift + Ctrl + T/E` absolutely do work. 

I'm a little wet behind the ears with FOSS - do I have to wait until the next update for the Up/Down functions?

nathanlesage:
> Shift + Ctrl + T/E absolutely do work.

ah, good to know! If they aren't mentioned in the documentation, please feel free to add them on our documentation repo! :)

And concerning when the functionality is available: If you want to test out the (not completely stable) beta version, there's just a fresh new one on the releases page available! This includes the fix already!

esm7:
Shouldn't this be in 1.7.4? At least for me (under Linux) these shortcuts don't do anything :(

cdaven:
[...]

PR #1394, opened 2020-11-01
Move global keyboard shortcuts to main menu

cdaven:
## Description

Bug fix for #759 (keyboard shortcuts not working _on Windows_).

## Changes

The suggested fix is to remove the "global GUI shortcuts" (keydown event listener), which is never triggered on Windows, and replace with menu items with accelerators. This is suggested by [Electron tutorial](https://www.electronjs.org/docs/tutorial/keyboard-shortcuts).

I added two menu items in the View menu, between Toggle Sidebar and Reset Zoom:

- Focus on File Manager (Ctrl/Cmd+Shift+T)
- Focus on Text Editor (Ctrl/Cmd+Shift+E)

Please note that the F2 keyboard shortcut has now been removed, since the menu item "Rename file" already has Ctrl/Cmd+R as accelerator. I don't know if this is a problem or how to fix it.

## Additional information

New strings for Zettlr Translate:

- menu.focus_filemanager
- menu.focus_editor

Tested on: Windows 10

### DIFF
--- a/source/app/service-providers/assets/menu.darwin.json
+++ b/source/app/service-providers/assets/menu.darwin.json
@@ -289,6 +289,19 @@
         "type": "separator"
       },
       {
+        "label": "menu.focus_filemanager",
+        "accelerator": "Cmd+Shift+T",
+        "command": "focus-file-manager"
+      },
+      {
+        "label": "menu.focus_editor",
+        "accelerator": "Cmd+Shift+E",
+        "command": "focus-text-editor"
+      },
+      {
+        "type": "separator"
+      },
+      {
         "label": "menu.reset_zoom",
         "accelerator": "Cmd+0",
         "command": "zoom-reset"
--- a/source/app/service-providers/assets/menu.win32.json
+++ b/source/app/service-providers/assets/menu.win32.json
@@ -239,6 +239,19 @@
         "type": "separator"
       },
       {
+        "label": "menu.focus_filemanager",
+        "accelerator": "Ctrl+Shift+T",
+        "command": "focus-file-manager"
+      },
+      {
+        "label": "menu.focus_editor",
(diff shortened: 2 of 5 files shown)


cdaven:
Could probably use or implement something in file-manager.vue to focus the file-list, instead of using the global ID selector?

nathanlesage:
I've not added those shortcuts to the global menu out of two reasons:

1. I'm pretty sure it's _not_ best practice to add each and every shortcut to the menu
2. Which shortcuts should be present in the menu and which ones don't?

Furthermore, I'd not listen stoically to the Electron team when it comes to best practices, as this sometimes is not the best for any application. I've had this experience several times in the past.

***

With regard to those shortcuts not working on Windows, this might have a trivial reason as they also didn't work on macOS when I tested them out just now.

So, in fact we might simply fix the obvious bug first (which I just did). I personally do _not_ want to add those commands to the menu, because it is counter-intuitive to actually _click_ those items in order to focus one of the two regions. In this case, a user wanting to focus one of those will _always_ resort to the menu item, so in this case the menu item would only be a proxy.

So, please, try out my fix, it should work now, as I've moved the event listener to the window and also migrated it away from jQuery. For macOS, it is now working, and it should also on Windows and Linux, independent of the actual key code (apparently, on a German keyboard layout, pressing `Cmd/Ctrl` together with Shift and a letter makes `event.key` spit out the lowercase variant, whereas on some others it might spit out the capital letter. I've taken that into account, but, as I said, please also test it out, because I want to be sure that it works now!

nathanlesage:
N.B.: The fix with a method instead of the global ID selector: This does sound good, and we can certainly accomodate for that!

cdaven:
Ok! I understand your reasoning.

PR #1398, opened 2020-11-03
Fix file-list focus bug

cdaven:
## Description

Based on #1394, this makes sure that we don't focus the file-list when it's not visible. Seems to be an old bug.

Reproduce the bug:

1. Select "Combined" file manager mode
2. Press Cmd/Ctrl+Shift+T

The file manager will disappear, and an empty file list will be shown with "No directory selected".

## Additional information

Tested on: Windows 10


nathanlesage (approved):
I didn't realize this was happening, as I almost never use combined mode -- thanks for the fix!

PR #1421, opened 2020-11-12
Show tooltip with title on hover over citation previews

mathcals:
## Description

Add a title text to the citation preview element, containing the titles of the cited sources, such that a tooltip is shown on hover. This makes it easy to see the title of a source without having to look at the bibliography. 

## Changes

- Added a `titleText` property to the payload of `get-citation` messages
- Add a title attribute to the citation preview (i.e. the `citeproc-citation` codemirror widget)

Tested on: Ubuntu 20.04


nathanlesage:
It is as if you could read my mind: I wanted to add something like that months ago!

However, one question: Instead of the titles – how about we simply add the fully rendered bibliographical reference? This was what I wanted to show in those tooltips myself, and by utilising tippy.js, we have a good chance to even make this look pretty good (as we can use HTML, hence, a list). What do you think? Then we wouldn't need an additional new function in the provider, but instead could rather utilise the makeBibliography function, just passing the citations from the one in question?

mathcals:
Sounds like a good idea! Then we could also add a link to the corresponding attachements in this tooltip, similar to how this is done for hyperlinks.

nathanlesage:
> Then we could also add a link to the corresponding attachements in this tooltip, similar to how this is done for hyperlinks.

Hmmm, I'm worried that this might be an inch too much info for a tooltip, don't you think? 🧐

mathcals:
I implemented it and I think the tooltips do not look cluttered, but I am curious about your opinion.

nathanlesage:
Hey, sorry for the late response, I'll be looking over that thing soon (if not, feel free to comment here again to ping me), and then hopefully I can merge it!

mathcals:
No problem. I really appreciate the effort you put into this project btw. Zettlr is a very useful tool.

nathanlesage:
Hey, would you mind resolving the conflicting files, re-test it again and then I can merge it!

mathcals:
:+1: I resolved the merge conflict (was only CHANGELOG.md) and re-tested. Everything seems to work fine.

nathanlesage:
Thank you very much :) Just one thing I just now realised: Would you mind adding the ipcRenderer-call from the zettlr-rendereripc to the citeproc-hook directly? I know it's my fault that I just saw it now, but I'm planning on removing both zettlr-ipc and zettlr-rendereripc completely and decentralising this stuff. That would be great! And sorry again for coming around the corner after you've resolved all this stuff 🙈

mathcals:
Uhm, I added the getCitationReference method to global.citeproc in citeproc-provider.js now, but I am not sure if this what you mean. If not, could you clarify what you want me to change?

mathcals:
@nathanlesage I still think this would be a nice feature. I am happy to add it, but I am not sure if the change I made is what you mean. Could you take a look at it and clarify what you would like me to change?

PR #1435, opened 2020-11-19
Latex Template: Define `\VERB` that pandoc uses for the inline_code_attributes extension

Kangie:
## Description
Add `\VERB` command to latex template. Pandoc uses this when inline code is given an attribute (e.g. `` `test`{.rest} ``) and exports will fail if this command is not defined. 

## Changes
Added `\VERB` and `\VerbBar` commands to default export template.
## Additional information

Tested on: Gentoo Linux; Current head of develop. Pandoc 2.9.21

### DIFF
--- a/source/main/assets/export.tex
+++ b/source/main/assets/export.tex
@@ -104,56 +104,12 @@ pdfproducer={Zettlr with Pandoc and XeLaTeX},
 pdfborder={0 0 0}
 }
 
-% The following preconditions are necessary to have Pandoc produce highlighted
-% code without throwing errors like mad.
-% \defaultfontfeatures{Mapping=tex-text,Scale=MatchLowercase}
-\usepackage{color}
-\usepackage{fancyvrb}
+% Setup pandoc code highlighting
+% pandoc uses an if statement, but until we have code
+% to enable or disable highlighting we'll ignore that.
 
-% Define the Shaded environment Pandoc uses
-\newenvironment{Shaded}{}{}
+$highlighting-macros$
 
-% Define the Highlighting environment for Pandoc
-\DefineShortVerb[commandchars=\\\{\}]{\|}
-\DefineVerbatimEnvironment{Highlighting}{Verbatim}{commandchars=\\\{\}}
-
-% Define the VERB environment that pandoc uses for inline_code_attributes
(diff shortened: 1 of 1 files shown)


nathanlesage:
I can't find this in the default Pandoc LaTeX template – is this an error on their side, that they forgot this or is there something I'm missing? <.<

https://github.com/jgm/pandoc/blob/master/data/templates/default.latex

Kangie:
It only appears in your pandoc latex output when you attempt to use the syntax. It confused me a lot. 

I don't know how it's being inserted in the preamble (it magically appears much like the syntax highlighting parts that aren't in there either!), but you can validate the behaviour by attempting to generate a pdf with the pandoc default.latex template and the Zettlr template using the content: 

```
`test`{.rest} 
```

nathanlesage:
> I think I would be worth replacing the baked in list of highlighting elements below my changes with an appropriate if statement so that pandoc inserts the content.

Yes, I'd also like to only include statements copied and pasted from the Pandoc default template, because this way we know it'll work. In the end, the default Zettlr template is just a stripped down version of that one ¯\\\_(ツ)\_/¯

Kangie:
OK, I've pushed a new commit that replaces the current code highlighting block with pandoc's `$highlighting-macros$` replacement. This actually includes the `\VERB` command that I was having trouble with, and should work with any pandoc version that supports it (so, anything more recent that ~1.6(?).

Have tested the export using pandoc, on a standalone system, and Zettlr+pandoc on my development system - Everything seems fine. :)

Suggest that when you squash+merge this PR you use the last commit's description.

nathanlesage:
Any Pandoc version means…? Would be great to have that!

Kangie:
Git blame says that its been in the pandoc default latex template since [before 2.0.5](https://github.com/jgm/pandoc-templates/blame/ea3841573fe83d8e101d70cbd78a1d6aa4bd6a21/default.latex#L256), so at least the last 3 years. I can't find an exact version, but it shouldn't cause issues like the citeproc change. :)

nathanlesage:
Alright, let's see that it works! Boy, I'm so happy once we finally bundle Pandoc, because then we have SO MANY MORE possibilities of finally getting all those templates to something cool!

PR #1443, opened 2020-11-23
Latex Template: Update for new Citeproc

Kangie:
##  Description

Latex Template: Update for new Citeproc.
* Also merge in some updates that make numbered citation styles render properly.

## Changes
Update `$if(csl-refs)$` block to be consistent with current pandoc default.latex.

This *will not* work with old version of pandoc (<=2.11 required). Note that the current template does *not* work with pandoc >=2.11.

Newer versions of pandoc will require that citeproc be explicitly called with `-C` or `--citeproc`; Citeproc will not be automatically run if a bibliography is passed to pandoc.

## Additional information
This probably shouldn't be merged as is - the default pandoc command will not work!

Tested on: Windows 10 (20.04); Head of Develop.

### DIFF
--- a/source/main/assets/export.tex
+++ b/source/main/assets/export.tex
@@ -163,10 +163,24 @@ pdfborder={0 0 0}
 $if(csl-refs)$
 \newlength{\cslhangindent}
 \setlength{\cslhangindent}{1.5em}
-\newenvironment{cslreferences}%
-{$if(csl-hanging-indent)$\setlength{\parindent}{0pt}%
-\everypar{\setlength{\hangindent}{\cslhangindent}}\ignorespaces$endif$}%
-{\par}
+\newlength{\csllabelwidth}
+\setlength{\csllabelwidth}{3em}
+\newenvironment{CSLReferences}[2] % #1 hanging-ident, #2 entry spacing
+ {% don't indent paragraphs
+  \setlength{\parindent}{0pt}
+  % turn on hanging indent if param 1 is 1
+  \ifodd #1 \everypar{\setlength{\hangindent}{\cslhangindent}}\ignorespaces\fi
+  % set entry spacing
+  \ifnum #2 > 0
+  \setlength{\parskip}{#2\baselineskip}
+  \fi
+ }%
+ {}
+\usepackage{calc}
+\newcommand{\CSLBlock}[1]{#1\hfill\break}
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hey, thanks for the initiative! The main idea right now, which I've also discussed with Pandoc directly, is to simply bundle Pandoc with Zettlr, which'll make a lot of those things easier – once we exactly know which Pandoc version Zettlr will use we can first completely overhaul the Zettlr template so that it supports all the stuff the default Pandoc template supports, and also make sure that we don't have these Pandoc 2.10 ≠ Pandoc 2.11 interdependency issues!

Brli:
I'm quite against bundle your own pandoc, instead, as dev who utilize the tool a lot, you should encourage users to update this app, so to trim down the possible flaw in their system.

Kangie:
> instead, as dev who utilize the tool a lot, you should encourage users to update this app, so to trim down the possible flaw in their system.

The issue with this is that large distributions (i.e. Debian/Ubuntu) tend to be way behind in versions as a result of the way their software distribution has been setup. They do not update for features, only bugs/security issues once the package has been accepted for a release. 

@nathanlesage have you put any thought into how to accomplish this? I've been thinking along the lines of a docker container...

nathanlesage:
With bundling Pandoc? Well, just download the prebuilt releases from GitHub and stuff them into the app, there's not much about it. No need to add a Docker container (plus how would you justify installing a 2 GB package in the background for a 15mb binary? :D)

> I'm quite against bundle your own pandoc, instead, as dev who utilize the tool a lot, you should encourage users to update this app, so to trim down the possible flaw in their system.

I value your valor, but this is not what I intend to do here: Bundling Pandoc will ensure all features work and prevent stuff like the 2.10-2.11 desaster we had. It does not work to simply ask everyone to update (see @Kangie's reply as well!)

nathanlesage:
And with regard to the template: The current one works with 2.11 like a charm, so let's wait a few seconds (My Homebrew updated to Pandoc 2.11 a few days ago and I had no errors yet …)

Kangie:
I'd expect it to error if you try and use citeproc to include references.

You'll also note that citeproc has to be called differently!

nathanlesage:
And here we go, I think Pandoc 2.11 has so far spread throughout most package repositories. I'm getting more and more requests from the non-tech side of the userbase wondering what those backslashy errors are that Pandoc is spitting out 👀

PR #1455, opened 2020-11-27
Add mhchem rendering support

likeadoc:
## Description

Added mhchem rendering support for KaTex as per https://katex.org/docs/node.html.

Tested on: 
Arch Linux, Windows


nathanlesage:
Looks good, thanks!

PR #1466, opened 2020-11-28
Fix language files search path

xatier:
e5d807a moved i18n into the lang directory within language pack, which
caused the incorrect search path of enumLangFiles helper.

## Description

I believe this is missed in e5d807a. This would cause issues if the user drops a customized translation into the webpack directory.

## Changes

Fix the search path, to be consistent with `getTranslationMetadata`.


xatier:
Rebase for eb979a3f4a9.

xatier:
@nathanlesage would you please take a look? :eyes:

nathanlesage:
Whoops! Thanks for spotting this!

PR #1494, opened 2020-12-06
Clean up machine-generated unused files

aidanhs:
## Description

This cleans up a number of machine-generated files that have a single commit on them in September and nothing else since. I did not observe any breakage as a result of removing them.

I didn't add anything to the changelog as it's just internal removal :)

Tested on: Ubuntu 18.04


nathanlesage:
Sharp eye, these were remnants from the time before forge!

PR #1498, opened 2020-12-07
use zeit upstream of webpack-asset-relocator-loader

Brli:
## Description

Use `@zeit/webpack-asset-relocator-loader` to replace the forked one which is out-of-date.

## Changes

package.json
webpack.rules.js

no breaking api change.

## Additional information

also open PR [upstream](https://github.com/electron-userland/electron-forge/pull/2067)

Arch Linux


nathanlesage:
Boy, this whole bubble has too many modules and no organisation …

Thank you for spotting this!

PR #1500, opened 2020-12-08
Fix file open extensions (remove dot prefix)

jnystad:
## Description
Current behaviour removes all files on windows (due to double dot).

## Changes
Removed dots in extensions list provided to file open dialog.

## Additional information
Electron expects extensions without dot: https://github.com/electron/electron/blob/v10.1.5/docs/api/dialog.md#dialogshowopendialogbrowserwindow-options

I am not in a position to test this on other platforms, but I assume following electron docs is the safest bet.

Tested on: Windows

### DIFF
--- a/source/main/zettlr.ts
+++ b/source/main/zettlr.ts
@@ -396,9 +396,9 @@ export default class Zettlr {
     // TODO: Move this to a command
     // The user wants to open another file or directory.
     const rmdSupport = global.config.get('enableRMarkdown') as boolean
-    const extensions = [ '.markdown', '.md', '.txt' ]
+    const extensions = [ 'markdown', 'md', 'txt' ]
     if (rmdSupport) {
-      extensions.push('.rmd')
+      extensions.push('rmd')
     }
 
     const filter = [{ 'name': trans('system.files'), 'extensions': extensions }]


nathanlesage:
> I am not in a position to test this on other platforms, but I assume following electron docs is the safest bet.

Absolutely. My last information was that these filters expect them WITH a dot, so they must've changed this in the meantime.

However, to keep this safe to use in general, I would like to ask you to provide a `map` on the filter directly in the function: https://github.com/Zettlr/Zettlr/blob/develop/source/main/modules/window-manager/dialog/ask-file.ts

As the common way of representing file extensions in node.js is _with_ the dot, and as there are a lot of places where extensions are specified to be fed into that function, it would be best to simply check all file extensions if they have the preceding dot, and remove that just before actually calling the function. This way it will work in other functions just as well, and we can migrate more safely (plus we can continue to stuff extensions pulled from nodeJS's `path.extname`-function into filters without thinking about normalising them everywhere).

Would you mind doing that? This would be awesome! Furthermore, feel free to also add a Changelog entry indicating your change (in the "Under the Hood" section as nothing changes for the end user), adding your GitHub handle as well? This would be awesome :)

nathanlesage:
Sorry, I just forgot about your PR and the fix is already in the code base 🙈

PR #1542, opened 2020-12-20
Move regexes to source/common/regular-expressions

Kangie:
## Description

Moved all of the regexes that I could find to source/common/regular-expressions
## Changes

Regexes removed from source files in favour of calls to functions in source/common/regular-expressions
## Additional information

I'm pretty sure that I got all of the regexes, but I'm 100% sure that I missed something with a non-standard name.

I've done my best to keep the names consistent, but where the same RE was called within two different functions I've made a call. 

List RE functions have all been given consistent names.

At some point, it would be a good idea to alphabetise the list of functions in source/common/regular-expressions

I've tested as many of the functions as I can; various syntax highlighting / markup things all seem to be working:

- Lists
- Images
- URLs
- Footnotes
- References

Let me know if you have any feedback; this should make all of the Regexes easier to maintain.

Gentoo Linux.

### DIFF
--- a/source/common/regular-expressions.js
+++ b/source/common/regular-expressions.js
@@ -97,6 +97,13 @@ module.exports = {
   /**
    * Returns a regular expression that matches MarkDown footnote references.
    *
+   * This matches footnote links in the style of [^<text>] but only if it includes
+   * numbers, letters (without umlauts) and _ as well as - chars.
+   * The second or third capturing group contains the identifier. The other
+   * will then be `undefined`
+   * This also matches strings where ] is the end of a string, but
+   * still ensures, ":" is NOT behind the closing bracket.
+   *
    * @return  {RegExp}              The compiled Regular Expression
    */
   'getFnRE': function () {
@@ -118,6 +125,11 @@ module.exports = {
 
   /**
    * Returns a regular expression that matches MarkDown footnote references.
+   *
+   * This matches footnote references in the style of [^<text>]: <reference text>
+   * This matches the same type of footnotes as the fnRE and includes two
+   * capturing groups: match[1] holds the identifier, match[2] the reference text.
+   * group 1: footnote number; group 2: text
    *
    * @param   {boolean}  multiline  Whether to match multiline
    *
--- a/source/common/regular-expressions.js
+++ b/source/common/regular-expressions.js
@@ -253,12 +253,6 @@ module.exports = {
   /**
    * Returns a regular expression that matches inline Maths.
    *
-   * Matches all inlines according to the Pandoc documentation
-   * on its tex_math_dollars-extension.
-   * More information: https://pandoc.org/MANUAL.html#math
-   * First alternative is only for single-character-equations
-   * such as $x$. All others are captured by the second alternative.
-   *
    * @param   {boolean} global      whether the expression should be global
    *
    * @return  {RegExp}              The compiled Regular Expression
@@ -270,6 +264,27 @@ module.exports = {
       flag)
   },
 
+  /**
+   * Returns a regular expression that matches inline Maths.
+   *
+   * Used to render inline math.
+   *
+   * Matches all inlines according to the Pandoc documentation
+   * on its tex_math_dollars-extension.
+   * More information: https://pandoc.org/MANUAL.html#math
+   * First alternative is only for single-character-equations
+   * such as $x$. All others are captured by the second alternative.
+   *
+   * @param   {boolean} global      whether the expression should be global
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getInlineMathRenderRE': function (global = false) {
+    let flag = (global) ? 'g' : ''
+    return RegExp(
+      /(?<!\\)\${1,2}([^\s\\])\${1,2}(?!\d)|(?<!\\)\${1,2}([^\s].*?[^\s\\])\${1,2}(?!\d)/.source,
+      flag)
+  },
   /**
    * Returns a regular expression that matches MarkDown links.
    *
--- a/source/common/regular-expressions.js
+++ b/source/common/regular-expressions.js
@@ -27,7 +27,11 @@ module.exports = {
   },
 
   /**
-   * Returns a regular expression that matches MarkDown Blocks (???).
+   * Returns a regular expression that matches MarkDown Blocks.
+   *
+   * First capturing group: preceding whitespace. Second cap.: line contents
+   * Non-capturing group in the middle: all block elements.
+   * Non-capturing group afterwards: catches all whitespace
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
@@ -39,6 +43,11 @@ module.exports = {
 
   /**
    * Returns a regular expression that matches MarkDown Citations.
+   *
+   * Should match everything permittible -- first alternative are the huge
+   * blocks, second alternative are the simple @ID-things, both recognised by
+   * Pandoc citeproc.
+   * citationRE is taken from the Citr library (the extraction regex)
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
@@ -155,7 +164,7 @@ module.exports = {
   },
 
   /**
-   * Returns a regular expression that matches MarkDown highlighting (???).
+   * Returns a regular expression that matches Zettlr's MarkDown highlighting extension.
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
@@ -298,6 +307,8 @@ module.exports = {
   /**
    * Returns a regular expression that matches unordered Markdown lists.
    *
+   * Includes a capture group to capture list tokens.
+   *
    * @return  {RegExp}              The compiled Regular Expression
    */
   'getListTokenRE': function () {
@@ -319,6 +330,8 @@ module.exports = {
 
   /**
    * Returns a regular expression that matches unordered MarkDown lists.
+   *
+   * Used in CodeMirror.
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
@@ -366,7 +379,7 @@ module.exports = {
   },
 
   /**
-   * Returns a regular expression that matches inline frames.
+   * Returns a regular expression that matches MarkDown table headings.
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
@@ -399,7 +412,16 @@ module.exports = {
   },
 
   /**
-   * Returns a regular expression that matches inline frames.
+   * Returns a regular expression that matches:
+   *
+   * 1. Underscore strong
+   * 2. Underscore emphasis
+   * 3. Asterisk strong
+   * 4. Asterisk emphasis
+   * 5. Heading levels 1-6 (mark)
+   * 6. Heading levels 1-6 (content)
+   * 7. Blockquotes
+   * 8. Inline code
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
--- a/source/common/regular-expressions.js
+++ b/source/common/regular-expressions.js
@@ -253,12 +253,21 @@ module.exports = {
   /**
    * Returns a regular expression that matches inline Maths.
    *
+   * Matches all inlines according to the Pandoc documentation
+   * on its tex_math_dollars-extension.
+   * More information: https://pandoc.org/MANUAL.html#math
+   * First alternative is only for single-character-equations
+   * such as $x$. All others are captured by the second alternative.
+   *
+   * @param   {boolean} global      whether the expression should be global
+   *
    * @return  {RegExp}              The compiled Regular Expression
    */
-  'getInlineMathRE': function () {
+  'getInlineMathRE': function (global = false) {
+    let flag = (global) ? 'g' : ''
     return RegExp(
-      /^(?:\${1,2}[^\s\\]\${1,2}(?!\d)|\${1,2}[^\s].*?[^\s\\]\${1,2}(?!\d))/.source
-    )
+      /^(?:\${1,2}[^\s\\]\${1,2}(?!\d)|\${1,2}[^\s].*?[^\s\\]\${1,2}(?!\d))/.source,
+      flag)
   },
 
   /**
@@ -343,7 +352,7 @@ module.exports = {
   /**
    * Returns a regular expression that matches unordered MarkDown lists.
    *
-   * Used in CodeMirror.
+   * Used in CodeMirror MarkDown shortcuts.
    *
    * @return  {RegExp}              The compiled Regular Expression
    */
--- a/source/common/regular-expressions.js
+++ b/source/common/regular-expressions.js
@@ -14,49 +14,157 @@
  */
 
 module.exports = {
+
   /**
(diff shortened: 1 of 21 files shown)


source/common/regular-expressions.js
```diff
@@ -14,6 +14,190 @@
...
+  },
+
+  /**
+   * Returns a regular expression that matches MarkDown Headings.
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getHeadingRE': function () {
+    return RegExp(
+      /(#+)\s+/.source
```
nathanlesage, line 153: Don't we need to assure start-of-line with ^?

source/common/regular-expressions.js
```diff
@@ -14,6 +14,190 @@
...
+    )
+  },
+
+  /**
+   * Returns a regular expression that matches MarkDown highlighting (???).
```
nathanlesage, line 158: To your question marks: Yeah, this is non-standard syntax that I've added based on a request. It's supported almost nowhere, but the intention is to get certain strips of Markdown displayed textmarker-like (I think Zettlr's themes use yellow background for this)

source/common/regular-expressions.js
```diff
@@ -14,6 +14,190 @@
...
+  /**
+   * Returns a regular expression that matches file IDs as in the settings
+   *
+   * @param   {boolean}  multiline  Whether to match multiline
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getIDRE': function (multiline = false) {
+    let flag = (multiline) ? 'm' : ''
+    let idRegExpString = global.config.get('zkn.idRE')
```
nathanlesage, line 177: Did you make sure this function doesn't get called before the configuration provider is loaded? (Just asking because I'm always afraid of such race conditions :D)

source/common/regular-expressions.js
```diff
@@ -14,6 +14,190 @@
...
+  },
+
+  /**
+   * Returns a regular expression that matches inline frames.
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getIframeRE': function () {
+    return RegExp(
+      /^<iframe.*?>.*?<\/iframe>$/.source,
```
nathanlesage, line 197: Note to myself: We definitely need a better way to get iFrames, because right now it's still that users can actually open a door to malware, because this will just load *any* iFrame, not just "good" ones (e.g. YouTube, Vimeo, etc.)

source/common/regular-expressions.js
```diff
@@ -43,6 +228,117 @@ module.exports = {
...
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getListTaskListRE': function () {
+    return RegExp(
+      /^(\s*)(- \[[x ]\])(\s*)/.source
+    )
+  },
+
+  /**
+   * Returns a regular expression that matches unordered Markdown lists.
```
nathanlesage, line 299: * and includes a capturing group to capture the list tokens, right?

source/common/regular-expressions.js
```diff
@@ -43,6 +228,117 @@ module.exports = {
...
+      /[*+-]\s/.source
+    )
+  },
+
+  /**
+   * Returns a regular expression that matches unordered MarkDown lists.
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getListUnorderedCMRE': function () {
```
nathanlesage, line 325: CM?

source/common/regular-expressions.js
```diff
@@ -43,6 +228,117 @@ module.exports = {
...
+  },
+
+  /**
+   * Returns a regular expression that matches MarkDown files.
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getMarkDownFileRE': function () {
+    return RegExp(
+      /.+\.(?:md|markdown|txt)$/.source,
```
nathanlesage, line 338: RMarkdown is missing here (rmd). I've let that one out of many regular expressions until now because I read that Adobe uses .rmd as some binary file format as well, but I'm close to simplifying the code to just allow RMarkdown by default (which would remove the corresponding setting as well) …

source/common/regular-expressions.js
```diff
@@ -57,25 +353,71 @@ module.exports = {
...
-      idRegExpString = `(${idRegExpString})`
-    }
+  'getTableRE': function () {
+    return RegExp(
+      /^\|.+\|$/.source,
+      'i')
+  },
 
+  /**
+   * Returns a regular expression that matches inline frames.
```
nathanlesage, line 369: Are you sure that's what this RegExp does? :D

source/common/regular-expressions.js
```diff
@@ -57,25 +353,71 @@ module.exports = {
...
...+)\]\((.+?)\)|(((?:(?:aaas?|about|acap|adiumxtra|af[ps]|aim|apt|attachment|aw|beshare|bitcoin|bolo|callto|cap|chrome(?:-extension)?|cid|coap|com-eventbrite-attendee|content|crid|cvs|data|dav|dict|dlna-(?:playcontainer|playsingle)|dns|doi|dtn|dvb|ed2k|facetime|feed|file|finger|fish|ftp|geo|gg|git|gizmoproject|go|gopher|gtalk|h323|hcp|https?|iax|icap|icon|im|imap|info|ipn|ipp|irc[6s]?|iris(?:\.beep|\.lwz|\.xpc|\.xpcs)?|itms|jar|javascript|jms|keyparc|lastfm|ldaps?|magnet|mailto|maps|market|message|mid|mms|ms-help|msnim|msrps?|mtqp|mumble|mupdate|mvn|news|nfs|nih?|nntp|notes|oid|opaquelocktoken|palm|paparazzi|platform|pop|pres|proxy|psyc|query|res(?:ource)?|rmi|rsync|rtmp|rtsp|secondlife|service|session|sftp|sgn|shttp|sieve|sips?|skype|sm[bs]|snmp|soap\.beeps?|soldat|spotify|ssh|steam|svn|tag|teamspeak|tel(?:net)?|tftp|things|thismessage|tip|tn3270|tv|udp|unreal|urn|ut2004|vemmi|ventrilo|view-source|webcal|wss?|wtai|wyciwyg|xcon(?:-userid)?|xfire|xmlrpc\.beeps?|xmpp|xri|ymsgr|z39\.50[rs]?):(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.]|[a-z0-9.-]+[.][a-z]{2,4}\/)(?:[^\s()<>]|\([^\s()<>]*\))+(?:\([^\s()<>]*\)|[^\s`*!()[\]{};:'".,<>?«»“”‘’])))|([a-z0-9.\-_+]+?@[a-z0-9.\-_+]+\.[a-z]{2,7})$/.source,
```
nathanlesage, line 397: Note to myself: We definitely should get to simplifying this monster 👀

source/common/regular-expressions.js
```diff
@@ -57,25 +353,71 @@ module.exports = {
...
+  },
+
+  /**
+   * Returns a regular expression that matches Zettelkasten IDs.
+   *
+   * @return  {RegExp}              The compiled Regular Expression
+   */
+  'getZknTagRE': function () {
+    return RegExp(
+      /##?[^\s,.:;…!?"'`»«“”‘’—–@$%&*#^+~÷\\/|<=>[\](){}]+#?/.source,
```
nathanlesage, line 419: I remember that I've added a negative lookbehind to make sure the hashtag is either on the beginning of a line or preceded by a space, but then I don't know where this regexp is being used …

In general: It's a very good idea to collect all REs here, because there's so many potential issues you can spot with this!

source/renderer/modules/markdown-editor/plugins/render-citations.js
```diff
@@ -45,7 +46,7 @@
   // blocks, second alternative are the simple @ID-things, both recognised by
   // Pandoc citeproc.
   // citationRE is taken from the Citr library (the extraction regex)
```
nathanlesage, line 48: It might make sense to add this comment above the getCitationRE-func in the regexp-file, so that it's clear where it comes from (and, whenever I should change this RE, I know that I must update this as well)

nathanlesage (approved):
Amazing work, thank you so much! In general I think this is exactly how it needs to be :)

Just one question: Would you mind double-checking each affected file that the functionality still works out? (Please also feel free to add corresponding GUI-checks to the scripts/test-gui - folder)

If everything works out, then we can merge it :) This code is platform-independent so it suffices if it works out on Windows!

source/common/regular-expressions.js
Kangie: I'll validate this against the current source tomorrow, but I did my best to make exact copies of the existing REs. Better to make changes once I'm sure I haven't broken anything!

source/common/regular-expressions.js
Kangie: Fantastic. I'll remember that next time I want you to implement something that's not in common mark :P

I'll update the docs!

source/common/regular-expressions.js
Kangie: Pretty sure that one was already in this file, but I'll take a look.

Kangie:
> Amazing work, thank you so much! In general I think this is exactly how it needs to be :)

I was just following your lead; at some point you'd put a few common REs here. This way they're all in one place when I need to steal one for working with MarkDown!

Thanks for the feedback! I haven't tested everything yet, but everything that I've tested so far has worked. I'll continue to work on it and implement any further feedback. 

I may enable some additional linting rules with your approval - enforce function alphabetisation in the long term, and make it a warning for now.

source/common/regular-expressions.js
Kangie: Yes. 100%. The function name is merely obfuscation.

source/common/regular-expressions.js
Kangie: Codemirror. 
This RE is only used by codemirror stuff. It could probably be consolidated at some point, as there's another RE that does pretty much the same thing, but I don't want to change things until I know this hasn't broken anything.

nathanlesage:
Perfect! I know testing is cumbersome as you have to trigger all functions at least once, so I hope you're not mad if I want you to do the testing this time :D

As for alphabetisation rules: I'm not extremely sure if that's useful in any case: for instance the service providers all feature a shutdown-function and I like them to be at the same position under the constructor, which wouldn't work in that case. But then, I never tried out something else so I might very well like it :D

source/common/regular-expressions.js
Kangie: The current regex doesn't do that, so I won't change this for fear of unintended consequences, unless you'd prefer that I update it in this PR?

source/common/regular-expressions.js
Kangie: This code is unchanged.

source/common/regular-expressions.js
Kangie: Won't change for now, unless you insist, same as above. :)

source/common/regular-expressions.js
Kangie: I like needing to scroll for five minutes to parse a regex.

source/common/regular-expressions.js
Kangie: `.../markdown-editor/modes/spellchecker.js` and `.../markdown-editor/modes/markdown-zkn.js` used to have independent implementations of this regex.

Kangie:
Tests: 

| Module | Result |
| ----------- | --------- |
| make-valid-uri | http/https/file URIs work |
| dir-project-export | Unable to test (my pandoc install is broken pending some dependencies from distro. ) |
| footnotes | Success! Tooltips! |
| autocomplete | Success! Syntax highlighting for code works. This syntax isn't implemented though: {.tex .numberlines} |
| markdown-zkn |  I don't use/understand this functionality well enough to test properly. All the REs that it calls do seem to work though|
| spellchecker | Unknown words within code/footnote/zkn ignored |
| continuelist | Fail. Then pass. :D |
| indentlist | Pass. Numbering doesn't restart though - did it ever? I Haven't written a new ordered list in Zettlr in a couple of months. |
| markdown-shortcuts | Pass. Everything seems to highlight successfully. |
| render-citation | Short form works - I assume that the long form works too. |
| render-iframes | Not sure how this one is supposed to work - have never used an iframe in Zettlr. Can't make anything happen, will test against develop tomorrow. |
| render-math | multiline works, inline regex matches but rendering is broken.  |
| render-tables | Tables seem to render in WYSIWYG. Headings are properly bold. |
| render-tasks | Regex matches, but rendering is broken | 
| wysiwyg | All of the expected formatting seems to work | 

So to merge, Validate / Fix: 

- [ ] Task List
- [ ] Inline Maths
- [ ] iframes

Pretty happy with that overall. :)

I probably won't write tests at this point, but with everything consolidated that's a lot easier than it was before.

Kangie:
Iframes behave the same way on develop, so I'm calling that a pass. Task lists still aren't working... but everything else is. The regex is unchanged. Why?!

I'll look into it tomorrow hopefully. If you can spot the obvious error that I've made let me know!

Kangie:
OK, Rebased to fix the merge conflict and tidy up the history. Should be good to go. 

Lint error seems to have been introduced in the tables space, but not in any module that I've touched.

nathanlesage:
Yeah, linter error was my fault, but I've fixed it just now. I'll be fetching your branch later today and just test it out quickly, but I don't think that all too much should go wrong! That's, after all, what the `test-gui` is for, maybe I'll find some edge cases validating new contents to those files …

> I think I've spent 16 of the past 24 hours working with JS for one project or another...

Reminds me very much of myself …

Kangie:
Rebased to fix merge conflict. Looks like export-project-dir is now typescript (along with a lot of other commands!). 

Fixed up the RE import in export-project-dir, and the linting in dir-delete.

source/common/regular-expressions.js
nathanlesage: No, no, if that breaks things, better not. I concur with you to better play it safe, especially with these big PRs!

source/common/regular-expressions.js
nathanlesage: Yes, you're right! Let's see that as a note to myself, and probably you, if you get bored on the holidays 😆

source/common/regular-expressions.js
nathanlesage: That's bad – did you merge them? (Because, after all, both are modes, so both must work on exactly the same basis)

nathanlesage:
> Rebased to fix merge conflict. Looks like export-project-dir is now typescript (along with a lot of other commands!).

Actually, all. I just got the motivation to do that work, and right now I'm happy to having pulled through with this. Migrating to TypeScript is cumbersome as hell. But, alas, I found a few undiscovered bugs, so I count that as a success.

I was about to comment here to tell you that I stumbled upon a regex within a command, but you're fast, I'm really impressed! :)

nathanlesage:
Sooooooo, just pulled your PR and went through the functionality of the files where you've removed the regexps from, and I couldn't find any error, so merging!

You're the best ❤️

Issue #1430, opened 2020-11-17
[ENHANCEMENT] Allow user to disable mouse zoom


cianoc:
## Description
Add an option in preferences that allows the user to disable mouse zoom (CMD/CTRL - mouse scroll).

## Justification
When using the Apple magic mouse in Zettlr it is almost impossible to click on a link without also zooming in/out - as the slightest lateral movement on the mouse will trigger 'scrolling'. Allowing the user to optionally disable this would remove a source of considerable frustration.

baubaid:
Adding my "vote" for this enhancement. I've already done this accidentally more times than I care to count.

tap349:
Even when using trackpad I get accidental zooming in and out when scrolling. It's very distracting.

nathanlesage:
Took a long time but finally here we go!

PR #1557, opened 2020-12-23
Added scroll to zoom setting toggle

borkowmichal:
## Description

I added a settings toggle that allows to disable the "cmd/ctrl + scroll" zoom behaviour in the editor.
I happen to trigger this accidentally a lot and I saw that someone posted an issue for this ( #1430 ) so I thought I may propose this small enhancement I made in my own version

## Changes
Added:
- the option in the settings menu (in advanced)
- the i18n texts (I used an automatic translation for the german version)
- a check in the zoom hook for the value of the setting

## Additional information

I don't know how you like to keep the settings organised, I named it "enableScrollToZoom" and put it under "editor".  
As for the toggle itself I put it in advanced section.
Just let me know of the naming / organisation you'd like to see !

Tested on: macos

### DIFF
--- a/resources/templates/dialog/preferences.handlebars
+++ b/resources/templates/dialog/preferences.handlebars
@@ -921,6 +921,13 @@
             </label>
             <label for="system.deleteOnFail">{{i18n "dialog.preferences.delete_on_fail"}}</label>
           </div>
+          <div class="cb-group">
+              <label class="cb-toggle">
+                  <input type="checkbox" name="editor.enableScrollToZoom" value="yes" id="editor.enableScrollToZoom" {{#if editor.enableScrollToZoom}}checked="checked"{{/if}}>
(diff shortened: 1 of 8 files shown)


nathanlesage:
Looks good to me, just a few small things:

1. I would think we might want to put the setting into the editor tab, as it's related to that one
2. The translation files may not be touched, I have to add that to [Zettlr Translate](https://translate.zettlr.com/) instead. Please double-check your localisation identifier and, once you're happy, I can add it there.
3. We must cache the config value, because the wheel-event handler is called **a lot**, that is, several hundreds of times per second, and the "global.config.get"-handler always has to call the main process, which will induce a lagged experience, which is not wanted. From main there will be an event once the config has changed and which we can listen for in order to update the setting.

If these points are addressed, I will happily merge this one!

PR #1563, opened 2020-12-24
Alternative macOS app icon

marc2o:
Donation to the Zettlr project with icon source files.

## Description
Alternative .icns and .png source files for app icon on macOS.

## Changes
No changes, but additional files.

## Additional information
The vector source files are on Figma:https://www.figma.com/community/file/923604358234136278/Zettlr-macOS-Icon

macOS Big Sur 11.1


nathanlesage:
Thank you very much! ❤️

Just two small things: As you added to the icon maker explanations on how to create icns from an iconset -- do you have that? It would make sense to provide the iconset folder as well!

And further, as the PNG icons are also used within the app, I personally think we should only replace the macOS-icon and retain the flat version for Windows and Linux as I think this might look better there! But as I'm only on Mac, I'll reconsider if others think the new icon would look better there as well :)

marc2o:
The iconset is the »icon.iconset« folder with the png files. That's the folder I used to compile the icns file. Is that what you meant?

nathanlesage:
Yes, exactly :)

nathanlesage:
Soo, sorry for the delay, but here we go :)

PR #1565, opened 2020-12-25
Fix for pandoc command error.

Brli:
## Description

The new pandoc parameter is problematic, there are three parameter involved in the PR:
- `--citeproc`
- `--bibliography="/path/to/file.bib"`
- `--csl=/path/to/style.csl`

There are two case of note file involved:

1. markdown with citation notes.
2. notes without.

pandoc export to pdf command failed at default 1.8.4 setup, aka. `--citeproc --bibliography "$bibliography$" $cslstyle$`

Thus there are 2x3x3=18 combinations to test, let's use 1 and 0 and metric here to indicate whether a parameter exist or not, and group by the type of note files, then the default setup is indicated as (1, [1, "1or0"], "[1,1]or[0,0]":

### normal file

- don't need any of the parameter, so, go with `(0, [0, 0], [0, 0])`
- works fine under `(1, [1, 1], [0, 0])`
- **not working with `(1, [1, 1], [1, 0])` and `(1, [1, 0], [1, 0])`**

### file with citation

- best work with `(1, [1, 1], [1, 1])`
- works under `(1, [1, 1], [0, 0])`
- failed with error at `(1, [1, 1], [1, 0])`, `(1, [1, 0], [0, 0])`, `(1, [1, 0], [1, 0])`, `(1, [1, 0], [1, 1])`
- doesn't render citation at  `(1, [0, 0], [1, 0])`, `(1, [0, 0], [1, 1])`,  `(1, [0, 0], [0, 0])`
- meaningless situations: (0, [1or0, 1or0], [1or, 1or0]) -> 5 types

### Conclusion

Based on above discussion, the most applicable case would be `(1, [1, 1], [0, 0])`.

Thus I combined `--citeproc` into `$bibliography$` and by test existence of bibliography file to trigger both `--citeproc` and `--bibliography` on.

## Changes

The `run-pandoc.js`, the pandoc command.

Require manual intervention to reset once from advanced tab of preference.

## Additional information


nathanlesage:
Thank you for fixing this accident of mine – looks all good, except I would like to add an additional command to check if the CSL library actually exists (`isFile`) – and also, please reinstate the CSL checker (with the lonnger if statement in front of it), because that will throw less errors on the linter and also make the code more readable. Other than that, it's ready to merge!

Brli:
done and also detach the bibliography part as it is also a bit too long IMO.

source/main/modules/export/run-pandoc.js
```diff
@@ -47,6 +47,12 @@ module.exports = async function (options) {
     throw new Error(trans('system.error.no_xelatex_message'), trans('system.error.no_xelatex_title'))
   }
 
+  // Include CSL library if exist
+  let bibliography = ''
+  if (global.config.get('export.cslLibrary') && isFile(options.cslLibrary)) {
```
nathanlesage, line 52: Just saw this, and that it's my fault: If there is a cslLibrary in the option, we should use that one, else if there is a global config value for that, use that, and only if both fail we shouldn't use anything.

The same holds true for the bibliography. Sorry, was my mistake, but I didn't have the time to track down the options cslLibrary (and also cslStyle where I made that mistake as well)!

nathanlesage:
Just noticed one small thing – sorry for the extra work! If you could then quickly resolve the Changelog conflict, we're good to go :)

Brli:
> Just noticed one small thing – sorry for the extra work! If you could then quickly resolve the Changelog conflict, we're good to go :)

IMHO, you shouldn't ask PR to include changelog change...as it should be maintained by cooperator to avoid conflicts like this...

source/main/modules/export/run-pandoc.js
Brli: IMO, you should copy the default values to user config dir, and stick the app with it (aka. update config only in that json without touch default) to maintain hierarchy, instead of messing with if-else clause checking method all around codebase.

Brli: BTW, AFAIK from the code, cslLibrary is the path to bibliography, which is fuzzing.

nathanlesage:
> IMHO, you shouldn't ask PR to include changelog change...as it should be maintained by cooperator to avoid conflicts like this...

True, but when I started with Zettlr, there were a lot of cool things I didn't know such as producing a Changelog by extracting the commit messages … But changing this is low priority, so it's easier since you did it either way to solve the conflict ¯\\\_(ツ)\_/¯

Brli:
meh, sorry to know that, anyway...thanks for the merge.

PR #1566, opened 2020-12-26
Update packages in pandoc template so that table generation works in >=2.11.3

Kangie:


Kangie:
So, the new version of pandoc changed the packages required to make tables.

This will fix that. It **should** be backwards compatible (it's just adding a new package after all), but I haven't tested it as I'm working on bleeding-edge pandoc stuff.

nathanlesage:
Let's just go for it!

PR #1567, opened 2020-12-26
Standalone Template

Kangie:
This works with the tutorial directory and the following frontmatter:

```yaml
---
title: "Pandoc and LaTeX Guide"
author:
  - The Zettlr Team
date: 2020-06-23
keywords:
  - Zettlr
  - Tutorial
subject: Zettlr Tutorial

title-page: true
block-subheadings: true
table-row-highlighting: true

---
```
You'll find that the title (etc) is overwritten because pandoc processes every YAML block that it encounters in order, but it provides a good example of what can be done. Most characters *should* work, but I haven't tested much other than latin characters and a few symbols at this point.

The default fonts for LaTeX export have been changed to Noto Sans (and variants) for mainfont (mostly to get glyph coverage), and Liberation Sans for headings. I've used Inconsolata (you will need v [2.0](https://github.com/googlefonts/Inconsolata/releases/tag/v2.012) for XeTeX) for the monotype font.

The template is pretty configurable, check out the comment block at the top - Some of the variables in the example yaml block are optional, but Title, Author, Date, and Keywords are not.

Some things aren't working yet, like for some reason the table row highlighting (that I've implemented so many times) and the footer page numbering. I figure that this is enough of a start to get some feedback on.

I've done my best to document what each option does, so that users can easily customise to suit their tastes.

Todo: 

- [x] Create a vector background image that can be optionally enabled to make the titlepage a little more flashy.
- [x] Draft flag implementation - should put the documentclass into `draft` mode and put a 'draft' watermark across each page. 
- [x] Fine Tune Image Placement (Zettlr differs from the pandoc default template, and I'm not sure why).
- [x] Fix Footer
- [x] Fix Table Row Highlighting (should alternate grey and white/transparent rows when enabled)

I note that pandoc on my system is broken and I've been using a docker-based pipeline to do my builds. The template does work. It may not work with Zettlr (yet), though I can't see any reason that it shouldn't.

### DIFF
--- a/source/main/assets/export-beamer.tex
+++ b/source/main/assets/export-beamer.tex
@@ -1,0 +1,483 @@
+% Lackadaisical / Zettlr Beamer Template
+% Use to generate presentations from MarkDown via Pandoc
+% Heavily based on the pandoc default template, but customised for beamer (i.e. removed all if/else statements relating to beamer)
+
+\PassOptionsToPackage{hyphens}{url}
+
+\documentclass[
+$if(fontsize)$
+  $fontsize$,
+$endif$
+$if(papersize)$
+  $papersize$paper,
+$endif$
+  ignorenonframetext,
+$if(handout)$
+  handout,
+$endif$
+$if(aspectratio)$
+  aspectratio=$aspectratio$,
+$endif$
+$for(classoption)$
+  $classoption$$sep$,
+$endfor$
+xcolor={table},]{beamer}
+
+% Background image for each slide.
+$if(background-image)$
+\usebackgroundtemplate{%
+\includegraphics[width=\paperwidth, height=\paperheight]{/working/templates/template_resources/lackadaisical_beamer_background.pdf}%
+$endif$
+}
+
+% Navigation symbols should be suppressed by default, but can be enabled.
+\setbeamercolor{caption name}{fg=normal text.fg}
+\beamertemplatenavigationsymbols$if(navigation)$$navigation$$else$empty$endif$
+$for(beameroption)$
+\setbeameroption{$beameroption$}
+$endfor$
+
+% Set beamer font theme early
+\usefonttheme{serif}
+
+\widowpenalties 1 10000
+\raggedbottom
+
+% Define some colours and then tell beamer to use them.
+\definecolor{blue}{HTML}{0D324D}
+\definecolor{burgandy}{HTML}{880D1E}
+\definecolor{green}{HTML}{6A8E7F}
+\definecolor{white}{RGB}{251,252,255}
+
+\definecolor{foreground}{RGB}{2, 2, 2} % Black
+\definecolor{background}{RGB}{251,252,255} % White
+\definecolor{grey}{RGB}{155,155,155}
+\definecolor{subtitle}{RGB}{102,255,204}
+%\definecolor{hilight}{RGB}{102,255,204}
+%\definecolor{vhilight}{RGB}{255,111,207}
+
+% Titles on every page
+\setbeamercolor{titlelike}{fg=burgandy}
+% Titles witin blocks
+\setbeamercolor{block title}{fg=blue}
+
+% Titlepage items.
+\setbeamercolor{subtitle}{fg=green}
+\setbeamercolor{institute}{fg=grey}
+
+% Body text
+\setbeamercolor{normal text}{fg=foreground,bg=background}
+
+% Lists
+\setbeamercolor{item}{fg=blue} % Bullet colour
+\setbeamercolor{itemize items}{fg=blue}
+\setbeamercolor{subitem}{fg=grey}
+\setbeamercolor{itemize/enumerate subbody}{fg=grey}
+\setbeamertemplate{itemize subitem}{{\textendash}} % Use en-dash for subitems.
+\setbeamerfont{itemize/enumerate subbody}{size=\footnotesize} % make them smaller.
+\setbeamerfont{itemize/enumerate subitem}{size=\footnotesize}
+
+
+$if(section-titles)$
+\setbeamertemplate{part page}{
+  \centering
+  \begin{beamercolorbox}[sep=16pt,center]{part title}
+    \usebeamerfont{part title}\insertpart\par
+  \end{beamercolorbox}
+}
+\setbeamertemplate{section page}{
+  \centering
+  \begin{beamercolorbox}[sep=12pt,center]{part title}
+    \usebeamerfont{section title}\insertsection\par
+  \end{beamercolorbox}
+}
+\setbeamertemplate{subsection page}{
+  \centering
+  \begin{beamercolorbox}[sep=8pt,center]{part title}
+    \usebeamerfont{subsection title}\insertsubsection\par
+  \end{beamercolorbox}
+}
+\AtBeginPart{
+  \frame{\partpage}
+}
+\AtBeginSection{
+  \ifbibliography
+  \else
+    \frame{\sectionpage}
+  \fi
+}
+\AtBeginSubsection{
+  \frame{\subsectionpage}
+}
+$endif$
+
+% ulem is needed for underlining and strikethrough text commands.
+% the option "normalem" is needed to preserve _italics_. Without the option,
+% _italics_ will become underlines, which is not desirable.
+\usepackage[normalem]{ulem}
+
+
+$if(theme)$
+\usetheme[$for(themeoptions)$$themeoptions$$sep$,$endfor$]{$theme$}
+$endif$
+$if(colortheme)$
+\usecolortheme{$colortheme$}
+$endif$
+$if(fonttheme)$
+\usefonttheme{$fonttheme$}
+$endif$
+$if(mainfont)$
+\usefonttheme{serif} % use mainfont rather than sansfont for slide text
+$endif$
+$if(innertheme)$
+\useinnertheme{$innertheme$}
+$endif$
+$if(outertheme)$
+\useoutertheme{$outertheme$}
+$endif$
+
+%
+% Configure Fonts
+%
+
+% fontspec provides an easy interface to XeTeX's font configuration.
+\usepackage{fontspec}
+
+$if(monofont)$
+  \setmonofont{$monofont$}
+$else$
+  \setmonofont{Inconsolata}[Scale=MatchLowercase]
+$endif$
+
+\defaultfontfeatures{ Ligatures=TeX } % This is actually a default option for main and sans font, but applies to all fonts set after this point.
+$if(mainfont)$
+  \setmainfont{$mainfont$}
+$else$
+  \setmainfont[%
+    BoldFont={Noto Sans Bold},
+    ItalicFont={Noto Sans Italic},
+    BoldItalicFont={Noto Sans Bold Italic}
+  ]{Noto Sans}
+    % We can use ucharclasses to setup font transitions based on unicode blocks.
+  \usepackage[CJK, Latin, Mathematics, Malayalam, Thai, Sinhala, Symbols]{ucharclasses}
+
+  \newfontfamily{\cjkfont}{NotoSansCJK-Regular.ttc}
+  \newfontfamily{\japanesefont}{NotoSansCJK-Regular.ttc}
+  \newfontfamily{\latinfont}{Noto Sans}
+  \newfontfamily{\malayalamfont}{Noto Sans Malayalam}
+  \newfontfamily{\sinhalafont}{Noto Sans Sinhala}
+  \newfontfamily{\thaifont}{Noto Sans Thai}
+  \newfontfamily{\unifiedCJKfont}{NotoSansCJK-Regular.ttc}
+  \newfontfamily{\mathsfont}{Noto Sans Math}
+  \newfontfamily{\symbolfont}{Noto Sans Symbols}
+
+
+% We're beginning and ending groups here so that the transition works anywhere.
+% This means that if e.g. a symbol is used within monospaced text that it will return to being monospaced
+% instead of going back to our default\main font!
+
+  \setTransitionsForCJK{\begingroup\cjkfont}{\endgroup}
+  \setTransitionsForJapanese{\begingroup\japanesefont}{\endgroup}
+  \setTransitionsForMathematics{\begingroup\mathsfont}{\endgroup}
+  \setTransitionsForSymbols{\begingroup\symbolfont}{\endgroup}
+
+  \setTransitionFrom{Malayalam}{\endgroup}
+  \setTransitionFrom{Sinhala}{\endgroup}
+  \setTransitionFrom{Thai}{\endgroup}
+  \setTransitionTo{Malayalam}{\begingroup\malayalamfont}
+  \setTransitionTo{Sinhala}{\begingroup\sinhalafont}
+  \setTransitionTo{Thai}{\begingroup\thaifont}
+$endif$
+
+$if(sansfont)$
+  \setsansfont{$sansfont$}
+$else$
+  \setsansfont{Liberation Sans}
+$endif$
+
+
+\usepackage{unicode-math}
+
+%
+% Required for pandoc to generate tables.
+%
+
+$if(tables)$
+  \usepackage{longtable,booktabs,array,caption}
+  % Make caption package work with longtable
+  \makeatletter
+  \def\fnum@table{\tablename~\thetable}
+  \makeatother
+$endif$
+
+$if(table-row-highlighting)$
+  \let\oldtabular\tabular
+  \let\endoldtabular\endtabular
+  \renewenvironment{tabular}{%
+    \rowcolors{2}{green}{}
+    \oldtabular
+  }{\endoldtabular}
+
+  \let\oldlongtable\longtable
+  \let\endoldlongtable\endlongtable
+  \renewenvironment{longtable}{%
+    \rowcolors{2}{green}{}
+    \oldlongtable
+  }{\endoldlongtable}
+$endif$
+
+
+% Pandoc puts lists within \tightlist. We need to provide that command else the conversion will fail.
+\providecommand{\tightlist}{
+  \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
+
+% Use upquote if available, for straight quotes in verbatim environments
+\IfFileExists{upquote.sty}{\usepackage{upquote}}{}
+\IfFileExists{microtype.sty}{% use microtype if available
+  \usepackage[$for(microtypeoptions)$$microtypeoptions$$sep$,$endfor$]{microtype}
+  \UseMicrotypeSet[protrusion]{basicmath} % disable protrusion for tt fonts
+}{}
+
+\newif\ifbibliography
+
+$if(listings)$
+\usepackage{listings}
+\newcommand{\passthrough}[1]{#1}
+\lstset{defaultdialect=[5.3]Lua}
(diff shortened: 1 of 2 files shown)


Kangie:
@nathanlesage pandoc uses [this](https://github.com/jgm/pandoc/blob/47f435276a62b3eddac496de6a7e365631d57197/data/templates/default.latex#L291) code to set figure placement, but you've done it differently. Can you tell me more about why Zettlr did it this way?

```
 \floatplacement{figure}{H}
```

nathanlesage:
I didn't touch the Zettlr template in many months, so I have no idea as to "why" :D When I initially created that template I basically took the Pandoc template and removed everything that didn't seem utterly necessary -- a somewhat short-sighted decision. So probably it'll be best to just copy the Pandoc template again and start from scratch in customising it?

Btw if you want to test it out rightly, feel free to build the app locally; it will already bundle pandoc (run the get-pandoc.sh script before yarn package) - then it'll work

Kangie:
> Btw if you want to test it out rightly, feel free to build the app locally; it will already bundle pandoc (run the get-pandoc.sh script before yarn package) - then it'll work

I'm not having much luck with it. :(

```
[12:15:26]	Trying to run command through Application: dir-project-export
[12:15:26]	[Export] Using bundled Pandoc instead of system-wide!
[12:15:26]	Running Pandoc with command: /home/kangie/development/Zettlr/out/Zettlr-linux-x64/resources/pandoc "/home/kangie/writing/tutorial/en/export.tmp" -f markdown -t latex --template="/home/kangie/writing/tutorial/en/template.latex" --toc --toc-depth=2 $citeproc$ --pdf-engine=xelatex --mathjax -o "/home/kangie/writing/tutorial/en/en.pdf"
[12:15:26]	An error occurred on export: /home/kangie/development/Zettlr/out/Zettlr-linux-x64/resources/pandoc "/home/kangie/writing/tutorial/en/export.tmp" -f markdown -t latex --template="/home/kangie/writing/tutorial/en/template.latex" --toc --toc-depth=2 $citeproc$ --pdf-engine=xelatex --mathjax -o "/home/kangie/writing/tutorial/en/en.pdf"
title: Export failed
message: An error occurred on export: /home/kangie/development/Zettlr/out/Zettlr-linux-x64/resources/pandoc "/home/kangie/writing/tutorial/en/export.tmp" -f markdown -t latex --template="/home/kangie/writing/tutorial/en/template.latex" --toc --toc-depth=2 $citeproc$ --pdf-engine=xelatex --mathjax -o "/home/kangie/writing/tutorial/en/en.pdf"
additionalInfo: pandoc: $: openBinaryFile: does not exist (No such file or directory)
[12:16:07]	Trying to run command through Application: export
[12:16:07]	[Export] Using bundled Pandoc instead of system-wide!
[12:16:07]	Running Pandoc with command: /home/kangie/development/Zettlr/out/Zettlr-linux-x64/resources/pandoc "/tmp/export.tmp" -f markdown -t latex --template="/tmp/template.latex" $citeproc$ --pdf-engine=xelatex --mathjax -o "/tmp/000-Pandoc and LaTeX Guide.pdf"
[12:16:07]	An error occurred on export: /home/kangie/development/Zettlr/out/Zettlr-linux-x64/resources/pandoc "/tmp/export.tmp" -f markdown -t latex --template="/tmp/template.latex" $citeproc$ --pdf-engine=xelatex --mathjax -o "/tmp/000-Pandoc and LaTeX Guide.pdf"
title: Export failed
message: An error occurred on export: /home/kangie/development/Zettlr/out/Zettlr-linux-x64/resources/pandoc "/tmp/export.tmp" -f markdown -t latex --template="/tmp/template.latex" $citeproc$ --pdf-engine=xelatex --mathjax -o "/tmp/000-Pandoc and LaTeX Guide.pdf"
additionalInfo: pandoc: $: openBinaryFile: does not exist (No such file or directory)
```
Don't have ti

Kangie:
OK, So I've created a nice title page graphic that can be enabled. 

Bad news: XeLaTeX really does not like being passed this value as a string, and I don't want to hardcode the value.

Good news, if you edit the template to have the path to the PDF below the and enable the `title-page-background` metadata variable, there's now a vector title page graphic in Zettlr green. I can put some code around it that will suit images that *aren't* the size of the whole page, but if we don't intend to ship one I'd just as soon leave that to the user.

[lackadaisical_title.pdf](https://github.com/Zettlr/Zettlr/files/5744334/lackadaisical_title.pdf)

Please find an example attached below. There are a few characters where Noto doesn't include the glyph. They look like Emoji, and there's probably something that can be done for that, but for example content I'm not too worried about it.

[Pandoc_and_LaTeX_Guide.pdf](https://github.com/Zettlr/Zettlr/files/5744340/Pandoc_and_LaTeX_Guide.pdf)

I'll try and get my Zettlr-bundled (or system) pandoc working so that I can try some exports, but we should be pretty good.

I'll add another commit to remove the Zettlr hardcoded titlepage content, as it's not required with this template.

Kangie:
Oops. Here's a version without draft mode enabled: 

[Pandoc_and_LaTeX_Guide.pdf](https://github.com/Zettlr/Zettlr/files/5744374/Pandoc_and_LaTeX_Guide.pdf)


Please ignore the document revision history stuff, that's an artefact of the documentation build stuff I've been working on.

Kangie:
Remaining template issues (I'd love to add to this list, give me feedback!)

- [ ] Page numbers in footer are not properly spaced out.
- [ ] the `table-row-highlighting` option overflows the bounds of the table due to the way they are constructed by pandoc.

- I will fix 1, 
- There's no good option for 2. I personally work around it by having pandoc output latex, then run sed over the output, and then run latexMK to actually generate the PDF. That may be a little too in-depth for you though.

The date on the titlepage is actually my fault - My pipeline converts it to a javascript date object, and that is being written directly. I'll go update that...

Kangie:
@nathanlesage

I know that you're busy (and I'm going back to work in the next few days), but do you have any thoughts on the PDF export process? I don't want to make big changes that are inconsistent with the way that you want things to work. 

No hurry on this one, but I might be able to make a bit of a start on it.

Kangie:
OK, we now have a beamer template, and the LaTeX template looks pretty good. 

I think that the titlepage graphic for the LaTeX template isn't actually straight, so I'll have a go at updating that at some stage soon. It's not critical - we've still a bit of time left before the new exporter (etc).

The beamer template uploaded produces PDF output like the below. The attached zip includes the markdown source.
 
[Example_Presentation.pdf](https://github.com/Zettlr/Zettlr/files/5792146/Example_Presentation.pdf)
[beamer.zip](https://github.com/Zettlr/Zettlr/files/5792141/beamer.zip)

Caveats:

- I have not yet uploaded the vector images for the slide/title background (I figure they're not finalised yet)
- The title text should be aligned with the bottom, and it is not. Small issue, but I still haven't gotten around to it yet!
- Hyperref is unhappy, but I haven't noticed an actual issue and builds are successful. I'll try and track it down, but it's low priority and I have other things on my plate.

I still need to add other doctypes to latex, that will probably be at the end of the week / next week. It shouldn't be too much work work though, I'll just need to ensure that the frontmatter is available.

Kangie:
Things to tidy up:

Latex Template
- [ ] Draft watermark gives overfull hbox error on every page. (Not critical, but would be nice to fix)

Beamer Template
- [ ] Title Text alignment
- [ ] Hyperref complaints

Kangie:
OK, To use the templates read them to examine the variable documentation. They *should* work out of the box, but some features (e.g. titlepage) must be explicitly enabled.

The template resources below are what I intend to commit as default assets, but the colour schemes will be adjusted depending on feedback. Maybe we should page some actual designers. :P

[template_resources.tar.gz](https://github.com/Zettlr/Zettlr/files/5794202/template_resources.tar.gz)

You should probably specify a `$monofont$`, as the ones that I've used are unlikely to be compatible with the variable-width version found in most repos (XeTeX doesn't have support yet).

the `mainfont` is Noto Sans. If the `$mainfont$` variable is set, fontspec will attempt to use what it's given, else:

We use `ucharchasses` to transition between fontspec-configured Noto font variants, based on the code point of the glyph.

```LaTeX
  \newfontfamily{\cjkfont}{NotoSansCJK-Regular.ttc}
  \newfontfamily{\japanesefont}{NotoSansCJK-Regular.ttc}
  \newfontfamily{\latinfont}{Noto Sans}
  \newfontfamily{\malayalamfont}{Noto Sans Malayalam}
  \newfontfamily{\sinhalafont}{Noto Sans Sinhala}
  \newfontfamily{\thaifont}{Noto Sans Thai}
  \newfontfamily{\unifiedCJKfont}{NotoSansCJK-Regular.ttc}
  \newfontfamily{\mathsfont}{Noto Sans Math}
  \newfontfamily{\symbolfont}{Noto Sans Symbols}
```

This should provide a good level of coverage for font glyphs for a lot of users, but it's based on unicode code points. If we haven't specified a transition (e.g. emoji), it will be handled by fontspec normally. 

The transitions are based on [unicode blocks](https://en.wikipedia.org/wiki/Unicode_block). If a code point exists within the block and the font that we transition to does not have a glyph for that code point, XeTeX will error.

I've only experienced one outlier so far - for some reason the TradeMark symbol is in Noto Sans instead of Noto Sans Symbols. I'll see what can be done about that. 

Otherwise RTL support isn't in the template yet. I'll do it at some point.

Pandoc should have no trouble creating documents from there.

nathanlesage:
Hey, so first of all apologies for answering so late. I just took some time to carefully read what you've written here and I don't know how, but my brain forgot almost everything you have written. Apologies for this!!

Then, to your templates:

I think it's a great effort, but I see a few problems. You've mentioned elsewhere that the template does not require any non-default packages so if there are some packages that have to be installed after the basic installation, we can still direct people [to this FAQ-entry](https://docs.zettlr.com/en/faq/#on-export-to-pdf-i-constantly-get-error-messages). So that should not be any problem.

However, Zettlr is supposed to work (mostly) out of the box, and especially concerning fonts, I see the problem that we have to somehow ease the effort to install those fonts. This requies (a) detection whether the fonts are actually availabe, and (b), if they are not, offer a more or less easy installation.

After a little bit fiddling around, I can safely say that it's impossible with easy methods to check for installed fonts. Installing fonts to the user directory may be possible, but that might be seen as intrusive. Out of experience I would say that we actually need to be careful with that. So we'd need an easy way for users to install Noto (I think even without a good check we could still say "Hey, would you like to install the default font required for PDF documents?" and then shove the font files into their font directory). The same applies to all fonts, so we need SILs and have to double-check that we are allowed to programmatically install those … but I think we are allowed to.

Then, your templates look pretty awesome, and having replaced all values with frontmatter stuff, we can finally switch to generating metadata. In short, with regard to the "prepare-pdf" function, here's what I want to switch to:

1. "Prepare file", e.g. concatenate projets, convert image paths to absolute etc.
2. Write a defaults file using the information that the user has given in some settings dialogs and pass it to Pandoc. This way, frontmatters should be able to override everything, with sensible and customizable defaults
3. Check what we have:
    - revealJS: Then run Pandoc -> reveal-bundler
    - Textbundle/Pack: Then run the bundle-bundler
    - Other: Run Pandoc

So the prepare-pdf file will be superfluous afterwards, and instead we need a general purpose `writeDefaultsFile` function (or similar).

With regard to the design: It seems to me a tiny bit dated but it does have a consistent design, so principally I'm d'accord!

Kangie:
No worries on the delay. Plenty of time before we want to ship things and there's been other things happening here. 

I am open to persuasion on the font side of things - the ucharclasses stuff could easily be hidden behind a variable and recommended, with a sane fallback font. It starts to get a little complex, but I agree that users should not need to install anything on top of Zettlr to achieve PDF export!

The good news on the design side of things is that its mostly custom title page alignment and some assets (vector PDF images), so it's easy to change to something less dated. 

On the bundling fonts question, the OFL is very permissive, just attach the licence and attribution. 

With the exporter, your thoughts seem logical. The `export-project-dir` function would be a good base, I think. It already calls the concatenate function, handles absolute image paths, and does a regex on footnotes; which works as well for one file as for a directory. If we add frontmatter parsing and validation and defaults file writing to that we're most of the way to the exciting new world of Zettlr Exporting.

nathanlesage:
> and does a regex on footnotes

You won't believe this, but I was searching for this functionality for almost three months now, because I simply forgot 👀 

> but I agree that users should not need to install anything on top of Zettlr to achieve PDF export!

Well, users have to install a TeX distribution, so some effort is needed either way. But reducing this as much as possible would be nice!

> On the bundling fonts question, the OFL is very permissive, just attach the licence and attribution.

Alright, so just ship the fonts, and push them into the global font directory if the user wants so.

Overall, I'm really happy how it's progressing :)

Kangie:
@nathanlesage 

I'm writing a module at the moment to generate defaults files, so stay tuned. It may be as easy for you as `generateDefaultsFile(frontmatter, outputFile, writer)`, and then writing that out somewhere.

nathanlesage:
> @nathanlesage
> 
> I'm writing a module at the moment to generate defaults files, so stay tuned. It may be as easy for you as `generateDefaultsFile(frontmatter, outputFile, writer)`, and then writing that out somewhere.

Ehh, I would rather have `generateDefaultsFile(someVariables: Object, outputFile: string)` :D The frontmatters within the files should stay untouched so they can override the defaults. The defaults file should only be affected by:

1. Generic settings
2. Project-based settings

Everything else should be done using the frontmatter. But I'm looking forward to the module!

Kangie:
@nathanlesange

Here's what I've come up with:

- [Github](https://github.com/lackadaisical-documentation)
- [npm](https://www.npmjs.com/org/lackadaisical)

Better documentation is coming.

I may have understated the extent of the changes.

You'll be most interested in `defaults-generator` and `markdown-binder`.

You should be able to drop in the markdown binder in both the `dir-project-export` and quick export functions. The function will return: (JSONSchema)

```JSON
{
  "properties": {
    "documentContent": {
      "description": "Concatenated Markdown content.",
      "type": "string",
    },
    "documentFrontmatter": {
      "description": "Object containing frontmatter keys.",
      "type": "object"
    },
    "pandocConfiguration": {
      "description": "The content of a pandoc defaults file, validated against a built-in schema",
      "type": "object"
    }
  }
}
```
Check the package documentation or the `markdown-binder-cli` for an example of how I intend for the various functions to be used. I'll be maintaining these as part of my docker CI documentation system, so I figure that it's a good way to give back to you... and I get the benefit that my quick exports in Zettlr will use the same logic as my document builds internaly.

The `markdown-processor` is an enhanced version of the dir-project-export command, and handles the same logic (footnote collision and absolute image paths). It also inserts OS appropriate end-of-lines when concatenating.

I'll include an example below. It doesn't work in Zettlr right now, but I imagine that the modifications are trivial! You can try the markdown binder locally by:

1. 'yarn global add @lackadaisical/markdown-binder-cli`
2. `markdown-binder -i /path/to/directory/of/markdown/files -o /different/path/that/you/want/output/in`

It'll also provide a handy reference if you want to do the implementation before I get to it. I'm pretty beat!

The defaults-generator will generate a valid pandoc defaults file from document frontmatter parsed as an object. It includes a built in JSONSchema for pandoc defaults files, so each key that appears in the defaults file will have to pass validation else the function will throw an error.

It will also accept another object as 'Custom Metadata', basically defaults that the frontmatter will override. You can override the 'writer' and 'output-file' values when calling the function, so the user can't break your output that way.

In `dir-project-export` this code should replicate your existing export using the functions I've created (Note: this won't build right now. I'm not debugging it, it's 00:40 here. I'd love to see your attempt!)

```javascript
import os from 'os'
import bindMarkdown from '@lackadaisical/markdown-binder'

...

    // Reduce to files-only
    files = files.filter((elem) => { return elem.type === 'file' })

    let content = bindMarkdown({ 'file-list': files })

    // Mock a file object to which ZettlrExport has access
    let tempfile = {
      'path': path.join(dir.path, sanitize(config.title, { replacement: '-' })),
      'name': sanitize(config.title, { replacement: '-' }), // obvious filename
      'content': `---${os.EOL} ${yaml.dump(content.documentFrontmatter)} ---${os.EOL}${os.EOL}${content.documentContent}`
    }
```

The `writeOut` function in `markdown-binder` can be used as a guide for handling the return of `bindMarkdown`

I need to write a bunch more tests for the markdown-binder, and handle some rejected promises. I also need to write a table generator (because I don't like the way tablemark is doing things). I'm pretty confident that the defaults-generator will either work or throw a descriptive error. Let me know how you go?

Happy to accept any PRs / feedback / etc. Sorry if this doesn't make sense, I've been in crunch mode for the last couple weeks, now that I've hit this point I can finally relax a bit.  :)

Kangie:
OK, so the modules aren't (quite) working with Zettlr, but I'm making some progress. I'll also use this as an opportunity to learn typescript so I can make electron stop complaining about that... 

When I get it working (and you're happy with the methodology), I'll submit a new PR to replace the markdown processor and we can test that for a release or so before we start passing a configuration object to the function and change the pandoc command to use the returned defaults file.

nathanlesage:
Sweet, thanks for your work! It made me smile to see that you acquired the document headers from Zettlr 😊 

However, I think there are a few improvements we could make to this. (Like, directly write this in TypeScript and publish to NPM, I can certainly give you an intro for Zettlr/Citr, because that module works exactly like that [including handy GitHub Actions CI for auto-publishing]).

Would you possibly want to have a video call about this? I would really appreciate it if Zettlr could incorporate your work, and there are a lot of questions on my mind right now, and I think a quick call would be perfect to clear the way for anything? :) 

P.S.: I just threw out all of the 32 bit stuff, finally, and I am en route removing that evil Pandoc command setting!

Kangie:
> you acquired the document headers from Zettlr 😊

I stole the export function and your regexes as my first step, so I figured I'd take that too! 

> However, I think there are a few improvements we could make to this. (Like, directly write this in TypeScript and publish to NPM,

I actually spent today learning TypeScript to do just that. I need to make my tests work for the module that I ported to validate my work, but once it's done the rest should fall into place. 

There are packages currently published on NPM but I'm building manually for now. I know that the cli tool produces workable output for trivial test cases, but as the version numbers suggest it's still early days - it was all one module until Wednesday! 

I'll look into some sort of CI build, test, and publish workflow once it's all typescriptified.

> Would you possibly want to have a video call about this? I would really appreciate it if Zettlr could incorporate your work, and there are a lot of questions on my mind right now, and I think a quick call would be perfect to clear the way for anything? :)

I'd be happy to. I'm on Australian Eastern Daylight time. (GMT+11), so let's work out a time.

> P.S.: I just threw out all of the 32 bit stuff, finally, and I am en route removing that evil Pandoc command setting!

Your April estimate may have been a bit pessimistic!

nathanlesage:
> I'd be happy to. I'm on Australian Eastern Daylight time. (GMT+11), so let's work out a time that works for both of us.

According to this: https://www.timeanddate.com/worldclock/personal.html?cities=312,240

... it would be best for me in the morning and for you in the evening. Sooo … shall we say tomorrow at 10 AM my time // 9 PM your time?

Kangie:
Sounds good. I'll chat to you then. What's your preferred platform?

nathanlesage:
> Sounds good. I'll chat to you then. What's your preferred platform?

I do like Jitsi! Do you want to just email me a link at hendrik @ zettlr?

>  There are many options, and not all documented as well as I'd like!

Yeah, and the docs seem outdated, too, because the default file John and the others provide includes the option "no-highlight" which Pandoc didn't like :D 

I just implemented some rudimentary defaults-stuff in the newly written exporter, in case you're interested.

PR #1589, opened 2021-01-02
Always set global search field query on Ctrl+Shift+F

### LINKED ISSUE #1346: [ENHANCEMENT / BUG?] Override pre-existing text in global search bar with highlighted text when pressing CTRL+Shift+F

cdaven:
## Description

When pressing Cmd/Ctrl+Shift+F, the search field was previously filled with the selected text, but _only if empty_.

This change overwrites the search field with the selected text _every time_.

## Changes

Move the getSelections() call from onfocus handler to menu item command. This means that you can click the search field with the mouse, and the contents will _not_ be replaced, only selected.

## Additional information

Fixes #1346, replaces #1347

Tested on: Windows 10

### DIFF
--- a/source/renderer/zettlr-rendereripc.js
+++ b/source/renderer/zettlr-rendereripc.js
@@ -251,7 +251,7 @@ class ZettlrRendererIPC {
 
       // TODO: What the heck is that a kind of "name"?
       case 'dir-find':
-        // User wants to search in current directory.
+      // User wants to search in current directory.
         this._app.getToolbar().focusSearch()
         break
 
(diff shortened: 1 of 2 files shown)


cdaven:
I noticed the linting errors from the first commit, so I modified and moved the logic to `focusSearch()` instead of the case in `handleEvent()`.

nathanlesage:
Looks good, thanks and sorry for simply not being able to handle your previous PR.

Issue #1591, opened 2021-01-03
Writing in Zettlr painfully slow


ploum:
Got myself a new laptop and immediately installed Regolith Linux (basically Ubuntu). Installed the latest Zettlr.

Realized that writing is almost impossible with longer files (more than 2000 words, so not really huge). After some times, characters takes several seconds to appear.
I’ve try to set the watchdog option but it’s even worse. I’ve disabled the "update when file remotely modified" without success.

Quitting and restarting Zettlr fixes the issue for some tenths of minutes.

I have no idea how to investigate this.

(the laptop is brand new, with brand new OS and everything is uber smooth so it’s not performance problem)

ploum:
correction: it doesn’t work after a restart. It’s just that I was briefly editing shorter files. So Zettlr is unable to edit no-tiny files in all circumstances.

It’s very strange.

panaousisn:
I have been experiencing the same problem. Closing the app and re-starting it seems to fix it for a few seconds and then back to lag territory. I have no problems with very short files, but when a paragraph or file cross the 500 word mark, I experience major delays.

ploum:
Yep, that’s exactly my symptoms. What is your system?

ploum:
Ok, I’ve been using Zettlr with a top running next to it. When waiting between the key press and the display, Zettlr is using all the CPU it can.

Agias77:
I have the same problem. I have a MacBook Retina (Early-16 / MacBook 9,1).

The slowness appeared with the last 2/3 versions.

Strangely enough, slowness also appears on short texts.

tegularius-fr:
I have been experiencing the same behaviour recently. Kubuntu 20.10. Long lines (more than 500 chars) become pain at some point. And after the moment the line is more than 1000 chars (+\- 150-200 words) it is almost impossible to write something. CPU loads up to 102.3% at the moment.

nathanlesage:
The reason is the multiplexer which performs A LOT of regex matches for the Syntax Highlighter. Remedification only works with a full rewrite of the whole rendering mode, which takes considerable amount of time, and I don't know when I'll have that …

ploum:
Can we simply disable syntax highlighting as a workaround?

Le lun. 4 janv. 2021 à 00:05, Hendrik Erz <notifications@github.com> a
écrit :

nathanlesage:
> Can we simply disable syntax highlighting as a workaround?

[...]

Issue #1523, opened 2020-12-15
[DEVELOP] Performance Sink on Long Lines


nathanlesage:
A few have already complained about that on the issue tracker here, and I just experienced the same issue: On long lines, input can take seconds to actually appear, which is indeed a frustrating experience. I just had a 20 seconds blocking task running until the characters I wrote actually appeared.

I ran a quick performance check and found out that the long waiting times are caused by the multiplexing mode which Zettlr uses for codeblock syntax highlighting.

The culprit function that is being called dozens of times per keystroke (!) is the `indexOf`-function:

```javascript
  function indexOf(string, pattern, from, returnEnd) {
    if (typeof pattern == "string") {
      var found = string.indexOf(pattern, from);
      return returnEnd && found > -1 ? found + pattern.length : found;
    }
    var m = pattern.exec(from ? string.slice(from) : string);
    return m ? m.index + from + (returnEnd ? m[0].length : 0) : -1;
  }
```

Especially the `pattern.exec` function takes considerable time. I currently presume the reason is that the multiplexer mode has to go through all supported syntax highlighting modes (currently around 50) and test if it begins here. The multiplex mode does have its problems either way with the way Zettlr uses it, but now that I finally got a hint, here's what we need to do:

1. Think of a good method of integrating the multiplexer into the markdown-zkn mode Zettlr defines. Instead of having fifty regular expressions run over the same piece of code forever, we need one catch-all regex and then implement this functionality ourselves.
2. Migrate all syntax highlighting modes into the markdown-zkn mode
3. Remove the multiplex-mode from the loading chain so we can directly pull in the markdown mode

mattj23:
Forgive me if this is naive; I don't know JavaScript and I've not looked at Zettlr's codebase or CodeMirror's before yesterday and it's a lot to take in.  I also don't know Electron, so half of my struggle is just trying to understand what's happening and how things are supposed to flow.

This is what I _think_ I understand:

[...]

PR #1594, opened 2021-01-05
Modified multiplexer to solve issue 1523

mattj23:
## Description
I created a slightly modified version of the [CodeMirror multiplexer building function](https://github.com/codemirror/CodeMirror/blob/master/addon/mode/multiplex.js) in order to solve the long line performance issue noted in #1523 and #1591.

## Changes
The changes were extremely minimal.  The existing implementation constructed an array of mode objects from regular expressions and existing CodeMirror syntax highlighting modes and passed them into the `CodeMirror.multiplexingMode` building function.

I copied the building function directly from CodeMirror's addons and made a modification to the `token` function.  Taking advantage of the fact that 48 of the syntax highlighting modes all began with a triple backtick token, I placed a `.backTick` property on the modes which had this feature.  Then I modified the `token` function of the multiplexer to check the line for simple triple backticks before iterating through the pattern matching for every mode.  In the case that no triple backticks appear in the line, all of the marked modes are skipped.  This enormously reduces the burden of the syntax highlighting.

## Additional information
I attempted to leave as much working code untouched as possible.  I copied CodeMirror's code directly and did not fix any eslinter errors on it.  I did not change Zettlr code except to add the `.backTick` property.  Tests passed, and I was able to see the dramatic performance improvement on a previously problematic file.

Only one file was changed, `multiplex.js`. This should be able to be applied as a patch to an older version of Zettlr if desired.

Tested on: Manjaro Linux 20.2, x86_64


nathanlesage:
Oh my God, if that works, you're godsend! I'll make sure to test it out ASAP!

One more thing: We can further simplify the mode, because what we're actually looking for is that on the line of the three backticks nothing except I think leading spaces at allowed, so we can even check the `stream.sol()` function which should even reduce it further. I'll get back to you once I have been able to test it!

mattj23:
My girlfriend loves Zettlr and uses it to write her articles and her book. A small patch is the least I can do to contribute.  We are running a version with the patch applied directly back to 1.8.4/master on her Big Sur MacOS and the performance issue is gone. 

You have a decision to make about eslinter. The code that fails eslinter is a direct copy of code that is known to be working in Zettlr, but just wasn't directly in your repo. We can fix it to satisfy the linter but we'll be messing with something known to work. It's probably worth fixing, but we should review the changes carefully.

Let me know if you want me to clean it up.

Kangie:
I'm impressed. It looks good. I won't have a chance to test it in the immediate future, but will report back when I do (if this hasn't already been merged)!

Thanks so much for looking into this!

Kangie:
OK. Had to do some markdown editing. This looks great. Can't wait for it to get merged!

mattj23:
Were you able to test it? I'll fix the merge conflict tomorrow. Do you have any thoughts on cleaning up the Codemirror code to satisfy eslinter?

Kangie:
> Were you able to test it? I'll fix the merge conflict tomorrow. Do you have any thoughts on cleaning up the Codemirror code to satisfy eslinter?

I did some brief testing - Nothing seemed broken, and keystrokes were entered quickly, rather than at a 1/2 second (or more) delay.

Having quickly opened the file in code, the linter seems to be crying (mostly) about whitespace and semicolons. Those are pretty easy fixes, and also the "missing space before function parentheses", so I'd suggest going ahead and doing those. 

My suggestion? Run `yarn lint --fix` over the code - it'll automatically fix 95% of the errors and then you can handle the remaining few individually - It looks like they're mostly redeclared variables and type-safe equality operators from there. 

@nathanlesage has the final say on this though! It seems like a lot of errors to just ignore when most of the refactoring can be done automagically.

nathanlesage:
> Those are pretty easy fixes, and also the "missing space before function parentheses", so I'd suggest going ahead and doing those.

I'd also say this. Most of this is basically just JSStandard Code Style, so it will not affect the functionality for the most part.

> My suggestion? Run yarn lint --fix over the code - it'll automatically fix 95% of the errors and then you can handle the remaining few individually

Are you sure this is safe? I'm always a little hesitant in letting an algorithm fix things for me, because I know what horrible mistakes machines can make :D 

But if in doubt, I can do the adaptations as well, it'll probably take me about 5 minutes simply fix the file manually. The important part is that it works. I just tested it myself and couldn't spot any immediate errors.

I'll merge this AFTER tomorrows release just to be safe, but 1.9 is basically around the corner, and one release more or less won't hurt until then!

So I'm ready to merge.

Kangie:
> > My suggestion? Run yarn lint --fix over the code - it'll automatically fix 95% of the errors and then you can handle the remaining few individually
>
> Are you sure this is safe? I'm always a little hesitant in letting an algorithm fix things for me, because I know what horrible mistakes machines can make :D

100%. I wouldn't blindly rely on the fixes without validating them, but most of the auto fixes do a pretty good job in my experience - it generally only goes after easy things like spacing and semicolons.

> I'll merge this AFTER tomorrows release just to be safe, but 1.9 is basically around the corner, and one release more or less won't hurt until then!

You set the release schedule, but as this is a quite noticeable UX issue it might be better to get it in this release?

mattj23:
I ended up doing the corrections by hand.  It would probably be worth it for one of you to look at the git diff, since JS is completely new to me.  The only changes that I wasn't 100% certain about were the `===` evaluations (especially for `typeof <variable> === 'string'`, which doesn't entirely make sense to me based on some reading of what the `===` operator does, but I did find it recommended as the best practice for that operation.

It's a pretty self contained change that should be fairly easy to reason about (mostly because we really only added 6 or 7 lines of "new" code), but I don't have strong feelings about the release schedule.  My only thought is that because of the nature of the type of writing some Zettlr users do, it is a show-stopping UX issue for some people.  My girlfriend had described it as making Zettlr unusable on her current writing projects and had temporarily switched to Typora, which is why I decided to take a look at the code in the first place.  But she's currently running a patched version so we're fine here either way.

I'm happy to help out in any way I can.  You guys have a great project here.

Kangie (approved):

Kangie:
Can't see any issues with your linting fix. Changes look good, but I haven't checked this out and built it since yesterday. I assume that everything still works!

nathanlesage:
> Can't see any issues with your linting fix. Changes look good, but I haven't checked this out and built it since yesterday. I assume that everything still works!

Me too; let's just merge it, we have at least a week to see if any errors pop up! So after the merge, just build a version for yourselves and use it as if it were the regular 1.8.5, then we can easily spot problems during the next week!

PR #1607, opened 2021-01-07
suppression du passage resté en anglais

framatophe:
## Additional information


nathanlesage:
Thanks again

PR #1643, opened 2021-01-17
Fix #1637 (missing icons in deb)

### LINKED ISSUE #1637: Missing icons in 1.8.6 deb
## Description

The 1.8.6 deb does not install proper application icons. There is only /usr/share/icons/hicolor/0x0/apps/Zettlr.png instead of one file per resolution.

## Reproducing

Install the 1.8.6 deb on linux.

## Expected behaviour

The deb should install one properly scaled icon to all the different subfolders of /usr/share/icons/hicolor as 1.8.5 did and other previous versions did.

## Platform
 - OS and version: Debian testing
 - Zettlr Version: 1.8.6

kb173:
## Description

Fix icons in deb release as suggested in #1637.

## Changes

Change icon in the linux entry in `electron-builder.yml` to `icon: "./resources/icons/png"`, as it was before. This is also one of the recommended methods in https://www.electron.build/icons.

## Additional information

Tested on: Ubuntu 20.10 (Linux 5.8.0-38-generic, x86-64)


nathanlesage:
Hey, thanks! Would you mind to in this special circumstance opening this PR to the master branch, instead of develop which is the default? I think you have to reopen it for that to work.

Here's the description [from the docs](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request):

image

Then I can merge it immediately and release a hot fix!

**Why to the master branch?** This has to do with the fact that I just "crashed" develop and therefore I need that stuff in the master branch! Sorry for the inconvenience!

kb173:
Oh yeah, changing the branch in this PR had some side effects... I'll reopen, one moment!

Issue #1622, opened 2021-01-11
File gets deleted after changing file name's capitalization


chrstnbwnkl:
## Description

When trying to rename a md file to the same name, but with a different capitalization of letters (e.g. from test.md to Test.md), the file gets deleted.

## Reproducing

Create a .md file with a name like "test.md", then change its name to the same file name but with different capitalization (e.g. "Test.md". Click ok for the warning that the file already exists and will get overwritten. Instead of overwriting the file, it gets deleted.

## Expected behaviour

Rename the file to the new desired name, internally overwrite the file instead of deleting it.

## Platform
 - OS and version: MacOS Catalina 10.15.7
 - Zettlr Version: 1.8.5

crbald:
I can reproduce this (from uppercase to lowercase and the other way). The "Overwrite?" prompt appears and then the file is deleted.
Windows 10 1909, Zettlr 1.8.5

nathanlesage:
Ouh, I know where this is coming from – thanks for reporting.

PR #1650, opened 2021-01-18
Fixes 1622 - Prevent file from getting deleted if renamed as itself

MontelAle:
## Description

As illustrated in issue #1622 when renaming a file as itself (or as itself with a different capitalization) it gets deleted. 

This PR aimes at fixing this exact issue.

## Changes

This PR adds a condition to ensure that the original file (the one which is getting renamed) isn't counted (and found) as file that needs to be overwritten/deleted (to make "space" for the renamed file). 

In doing so it preserves the original file, thus allowing any file to be renamed as itself (or as itself with a different capitalization).

## Additional information

This approach should work on Linux (or any other OS with a case-sensitive File System). 

To roughly test this I made the Zettlr workspace folder case-sensitive (using `fsutil.exe file setCaseSensitiveInfo "path" enable`) and didn't encounter any issues.

Sadly, I was not able to run the mocha tests (using `yarn test`) even on the master branch. Therefore, more testing is obviously required (especially on Linux/OSX).

Tested on: Microsoft Windows 10 Pro (version 10.0.18363)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ No 32 bit builds are now available anymore. Only 64 bit (Intel and ARM) are supp
 - 32 bit AppImages and Windows are no longer supported.
 - Removed the Pandoc installation item from the help menu.
 - Moved the Pandoc and XeLaTeX settings to the export tab in preferences.
+- Fixed a bug that would delete file if it got renamed as itself
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
In order to make this work in every case, you should compare the descriptors for triply-equality. So instead of `file.name !== found.name` you should actually compare `file !== found` (the FSAL always spits out the actual object, not a clone so that the triple-equality condition holds true there). This way we know that the same file is being renamed towards a different capitalization. As it is now, this would overwrite `Some File.md` with `some file.md` because those names would result in true on a `file.name !== found.name`-comparison.

nathanlesage:
Here we go!

PR #1651, opened 2021-01-18
upstream module owner change

Brli:
## Description

change upstream.

## Changes

from @zeit/webpack-asset-relocator-loader to @vercel/
as per result in [npmjs.com](https://www.npmjs.com/search?q=webpack%2Dasset%2Drelocator%2Dloader)

and suggestion via the module during npm/yarn

## Additional information

none


nathanlesage:
Did you test packaging of resources, and that all functionality is being retained by this change in resultant packages? 0.8.0 to 1.2.1 is quite the jump with regard to module compatibility, I might think.

Brli:
yes, in fact, zettlr has little to none impact on this module as it seems not used anyway, currently.

it should stay as fallback support for possible assets inclusion in the future, however.

nathanlesage:
Here we go, sorry it took so long!

PR #1818, opened 2021-03-22
Defaults Writer

Kangie:
## Description

Implementation of @lackadaisical/defaults-writer

## Changes

- Tidied up dir-project export.
  + Now fires off all IO promises and awaits the result before looping
    over content.
  + Only replaces footnotes if there is more than one file.
  + Additional typing (types.d.ts).
- @lackadaisical/defaults-generator WIP implementation.

### DIFF
--- a/source/main/commands/dir-project-export.ts
+++ b/source/main/commands/dir-project-export.ts
@@ -12,14 +12,18 @@
  * END HEADER
  */
 
-// import path from 'path'
-// import sanitize from 'sanitize-filename'
+import { CodeFileMeta, MDFileMeta } from '../modules/fsal/types'
+
+import path from 'path'
+import sanitize from 'sanitize-filename'
 import ZettlrCommand from './zettlr-command'
-// import { makeExport } from '../modules/export'
+import { makeExport } from '../modules/export'
 import objectToArray from '../../common/util/object-to-array'
 import makeImgPathsAbsolute from '../../common/util/make-img-paths-absolute'
+import { EOL } from 'os'
 
 import { getFnExportRE } from '../../common/regular-expressions'
+import { ExporterOptions } from '../modules/export/types'
 
 export default class DirProjectExport extends ZettlrCommand {
   constructor (app: any) {
@@ -56,49 +60,50 @@ export default class DirProjectExport extends ZettlrCommand {
     // Import our RE
     let fnExportRE = getFnExportRE(true) // We want the multiline version.
 
-    // Concat the files
-    let contents: string[] = []
-    for (let file of files) {
-      const fileDescriptor = await this._app.getFileSystem().getFileContents(file)
-      let fileContents: string = fileDescriptor.content
-      // Directly make all image paths absolute to prevent errors if used
-      // in nested project directories (in which this._dir.path is not the
-      // same for all files). Also make the footnotes unique to prevent
-      // assigning errors.
-      fileContents = makeImgPathsAbsolute(file.dir, fileContents)
+    let promiseArr: Array<Promise<MDFileMeta | CodeFileMeta>> = []
+    files.forEach(element => {
+      promiseArr.push(this._app.getFileSystem().getFileContents(element))
+    })
 
-      fileContents = fileContents.replace(fnExportRE, (match, p1: string, offset, string) => `[^${String(file.hash)}${p1}]`)
+    const contentsArr = await Promise.all(promiseArr)
+    const fileContentArr: string[] = []
 
-      contents.push(fileContents)
-    }
-    // const finalContents = contents.join('\n\n')
+    contentsArr.forEach((value, index) => {
+      let fileContent = makeImgPathsAbsolute(path.dirname(files[index]), value.content)
+
+      if (!(contentsArr.length === 1)) {
+        fileContent = fileContent.replace(fnExportRE, (match, p1: string, offset, string) => `[^${String(files[index].hash)}${p1}]`)
+      }
+
+      fileContentArr.push(fileContent.trim() + `${EOL}${EOL}`)
+    })
 
     // Mock a file object to which ZettlrExport has access
-    // let tempfile = {
-    //   'path': path.join(dir.path, sanitize(config.title, { replacement: '-' })),
-    //   'name': sanitize(config.title, { replacement: '-' }), // obvious filename
-    //   'content': finalContents
-    // }
+    let tempfile = {
+      'path': path.join(dir.path, sanitize(config.title, { replacement: '-' })),
+      'name': sanitize(config.title, { replacement: '-' }), // obvious filename
+      'content': fileContentArr.join()
+    }
 
     // Start up the Exporter
-    // let opt = {
-    //   'format': config.format, // Which format: "html", "docx", "odt", "pdf"
-    //   'file': tempfile, // The file to be exported
-    //   'dest': dir.path, // On project exports, always dir path
-    //   'stripIDs': true,
-    //   'stripTags': true,
-    //   'stripLinks': 'full',
-    //   'pdf': config.pdf,
-    //   'title': config.title,
-    //   'date': '\\today',
-    //   'author': config.pdf.author,
-    //   'keywords': config.pdf.keywords,
-    //   'cslStyle': config.cslStyle
-    // }
+    let opt: ExporterOptions = {
+      'format': config.format, // Which format: "html", "docx", "odt", "pdf"
+      'file': tempfile, // The file to be exported
+      'dest': dir.path, // On project exports, always dir path
+      'stripIDs': true,
+      'stripTags': true,
+      'stripLinks': 'full',
+      'pdf': config.pdf,
+      'title': config.title,
+      'date': '\\today',
+      'author': config.pdf.author,
+      'keywords': config.pdf.keywords,
+      'cslStyle': config.cslStyle
+    }
 
     // Aaaand export.
     try {
-      // await makeExport(opt) DEBUG @Kangie please reenable, had to remove that to fix a linter error
+      await makeExport(opt)
     } catch (err) {
       global.log.error(err.message, err)
       global.application.displayErrorMessage(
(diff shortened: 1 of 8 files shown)


Kangie:
@nathanlesage, hoping to get feedback / direction / your thoughts here.

I've tidied up `dir-project-export` and added some functionality there: the frontmatter is now stripped and passed to the exporter to be used to construct a defaults file. It also does all of the I/O before it starts to concatenate files, should be a little more efficient.

The `defaults-writer` now accepts a frontmatter object as input (todo: what does dir-project-export return if there's no frontmatter?), and uses the [@lackadaisical/defaults-generator](https://github.com/lackadaisical-documentation/defaults-generator) to construct and validate a defaults file using your existing logic (for now) to pull settings.

At this point it looks like the function should query the Zettlr global export options (maybe it does already, it's late here and I'm not checking) and put that additional object through the defaults generator (frontmatter > project settings > zettlr global options). The module can already cater for that, so it shouldn't be too hard.

This also implements my updated template WIP - It's not perfect (and I didn't commit any assets to plug in), but it seemed like a good time to make it consistent with the HTML. I've actually had to learn a *lot* more recently about manually constructing title pages, so maybe I can do something to make it a bit prettier. I'll tidy up the comment documentation before we merge this PR (I know at least one if statement works opposite how I've documented it!)

source/main/commands/dir-project-export.ts
```diff
@@ -56,44 +61,48 @@ export default class DirProjectExport extends ZettlrCommand {
...
-    }
-    const finalContents = contents.join('\n\n')
+    contentsArr.forEach((value, index) => {
+      let fileContent = makeImgPathsAbsolute(path.dirname(files[index]), value.content)
+
+      if (!(contentsArr.length === 1)) {
+        fileContent = fileContent.replace(fnExportRE, (match, p1: string, offset, string) => `[^${String(files[index].hash)}${p1}]`)
+      }
 
+      fileContentArr.push(fileContent.trim() + `${EOL}${EOL}`)
```
nathanlesage, line 79: Pandoc is able to just take \n\n instead of, on Windows, using \r\n\r\n, right? If so, I would suggest we just keep it as simple as possible and just use \n. Removes one import and makes this all simpler!

source/main/commands/dir-project-export.ts
```diff
@@ -56,44 +61,48 @@ export default class DirProjectExport extends ZettlrCommand {
...
+    })
+
+    // TODO: This only extracts frontmatter at the top of the concatenated file. If there are other 'frontmatter' values in the document they should be extracted too!
+    const splitContent = matter(fileContentArr.join())
```
nathanlesage, line 83: Better: In the FSAL file loader, there is a function `extractYamlFrontmatter` (or similar, but the first word is definitely extract), which basically does exactly what this one does here.

I would suggest to replace the custom extract function with this one, and save it in the descriptors. This way you have the front matter for every file, and the only thing we need to change is change the exporter so that it accepts a "full" descriptor instead of the mock object. Also makes this function smaller and integrates it better with the FSAL! And it saves a lot of computing power since the frontmatter will be in the files at every point!

The only thing you'd need to take care of is that you write the extracted frontmatter into the `frontmatter`-prop of the file (or null, if the frontmatter is empty), so that no renderer code breaks (which expects the frontmatter properties to be properties on the `frontmatter`-property … there's too much property in here)

source/main/commands/dir-project-export.ts
```diff
@@ -56,44 +61,48 @@ export default class DirProjectExport extends ZettlrCommand {
...
     // Mock a file object to which ZettlrExport has access
     let tempfile = {
-      'path': path.join(dir.path, sanitize(config.title, { replacement: '-' })),
+      'content': splitContent.content,
+      'frontmatter': splitContent.data,
       'name': sanitize(config.title, { replacement: '-' }), // obvious filename
-      'content': finalContents
+      'path': path.join(dir.path, sanitize(config.title, { replacement: '-' }))
```
nathanlesage, line 89: See my comment above: Modify the exporter to accept an array of files rather than one file. We're writing the input files into the defaults file either way, so we can make use of this!

source/main/commands/dir-project-export.ts
```diff
@@ -56,44 +61,48 @@ export default class DirProjectExport extends ZettlrCommand {
...
     // Start up the Exporter
-    let opt = {
-      'format': config.format, // Which format: "html", "docx", "odt", "pdf"
-      'file': tempfile, // The file to be exported
+    let opt: ExporterOptions = {
+      'author': config.pdf.author,
+      'cslStyle': config.cslStyle,
+      'date': '\\today',
       'dest': dir.path, // On project exports, always dir path
+      'file': tempfile, // The file to be exported
```
nathanlesage, line 98: Change this to an array.

source/main/modules/export/index.ts
```diff
@@ -96,7 +96,7 @@ export default async function makeExport (options: ExporterOptions): Promise<Exp
     )
   } else {
     // Run Pandoc
-    const defaultsFile = await writeDefaults(options.format, options.sourceFile, options.targetFile)
+    const defaultsFile = await writeDefaults(options.format, options.file.frontmatter, options.sourceFile, options.targetFile)
```
nathanlesage, line 99: Just pass all file frontmatters (e.g. `options.files.map(file => file.frontmatter)`) and merge them?

source/main/modules/export/write-defaults.ts
```diff
@@ -26,14 +29,18 @@ export default async function writeDefaults (
     title: path.basename(sourceFile, path.extname(sourceFile))
   }
 
-  // Use an HTML template if applicable
+  // Pick the appropraite template, if applicable
   if (writer === 'html') {
     let tpl = await fs.readFile(path.join(__dirname, 'assets/export.tpl.htm'), { encoding: 'utf8' })
```
nathanlesage, line 34: What I wanna do with this code is write the templates to a custom resource dir for Zettlr (as I'm doing with the defaults files) and write this template into the HTML defaults file by default, so it can be easily adapted by the users. It would be cool to do this, and use the assets-provider for this! Gives us the ability to both use templates AND reference-docs without writing additional code!

source/main/modules/export/write-defaults.ts
```diff
@@ -26,14 +29,18 @@ export default async function writeDefaults (
...
     defaults.template = path.join(dataDir, 'template.tpl')
     await fs.writeFile(defaults.template, tpl, { encoding: 'utf8' })
   }
 
-  // TODO: LaTeX template
+  if (writer === 'latex') {
+    let tpl = await fs.readFile(path.join(__dirname, 'assets/export.tpl.latex'), { encoding: 'utf8' })
+    defaults.template = path.join(dataDir, 'template.tpl')
+    await fs.writeFile(defaults.template, tpl, { encoding: 'utf8' })
```
nathanlesage, line 42: See my comment above!

nathanlesage (commented):
I really like your take at this one! The major things I'd improve upon this – already very good – PR would be:

* Change the exporter so that it takes an array of input files, rather than a single one.
* Make the exporter to accept the MDMetaDescriptor and CodeFileMeta explicitly, which gives us access to everything we need, without any computational overhead
* Use this new library I didn't know about (gray-matter) instead of the custom "extractFrontmatter" function I wrote a few weeks ago. Makes everything fit nicely together

If these changes are done, we have the following possibilities:

* Just pass on the list of files we want to concatenate into the single output file. If we want to just compile one single file, simply pass a single file to the exporter
* Allows us to remove any logic of actually adapting the files prior (I think we don't even have to make the image paths absolute by ourselves, do we??)
* Makes the exporter able to handle arbitrary export requests and keeps it slim

But, so far: Thank you so much for your work! :)

Kangie:
Thanks for the feedback, looking forward to putting some more time into this over the week. :)

nathanlesage:
Heya! I just pushed a big commit where I rewrote the exporter engine. This does conflict with the current PR, yes, BUT I took into account a lot of stuff that's necessary for your defaults writer implementation.

Please feel free to have a look at that commit: https://github.com/Zettlr/Zettlr/commit/b4d9f279d8d1bd121f806efebcbd88b9e966a428

Here's (in general) the pipeline of how the exporter is supposed to work now:

First, you need a list of file descriptors (which contain the files' frontmatters, if they have some). You can just pass that to the exporter.

1. Call `makeExport` passing the source file descriptors and a target directory. You can add as many new properties as you see fit to that options object (see for everything the types.d.ts in the exporter module, I've made sure to comment everything
2. The exporter then calls the prepareFiles function. That one simply takes the files, prepares them as necessary, writes them to intermediary files and returns a list of corresponding file paths for each sourceFile.
3. Then the exporter runs the correct exporter as given in the "format" option, passing all infos.
4. The exporter plugin, then, calls the "writeDefaults" function and passes an arbitrary object with properties to add to the user-customised defaults file. There is your option to modify the defaults file as you see fit
5. Then Pandoc is being run using that defaults file, and everything is returned.

This way we have a much more modular structure for the exporter, and we can simply stuff in a list of files rather than one single file so that all heay loading is being done in the exporter itself. Now that it has coarsely the right structure I can continue working on other things, and will not touch the exporter anymore, but leave it to you! I hope I haven't completely undermined any of your ideas by this initiative!

If something is unclear just ask, but I hope I have done something useful for you here :)

Kangie:
I'll rebase when I get a chance and see what I can do. Appreciate your effort!

Kangie:
So, I've been a bit busy and haven't had a chance to look into any of this since your last post. It's still on my radar though, hoping to make a start early next week.

nathanlesage:
Don't rush, there's enough to do elsewhere :)

Kangie:
So I'm (finally) working on this PR again. The codebase has changed a lot in that time - you've been busy!

I've rebased, and am currently looking for the best location to plug in 'make a valid defaults file' as well as ensuring that streamlining the project export function hasn't caused any issues.

Kangie:
Rebased... but it's probably in a worse state than before. We'll just hide all of that behind a squash. ;)

I'll work from this point on getting this working with the new code and fixing any linting errors that result, let you know when I think it's ready.

nathanlesage:
Sounds great! :)

Issue #1880, opened 2021-04-16
[ENCHANTMENT] MDX support


krry:
## Description

As a zettlr, when I open a folder containing `.mdx` files in the file manager, I would like to see them treated as files, and see their contents displayed as those of a `.md` would be.

## Proposed Changes

This may entail letting the Zettlr app know that `mdx` files are markdown files. I do not care if the `x` aspects of mdx are handled richly, just plaintext for the beyond-markdown pieces would suffice.

## Caveats

[MDX](https://mdxjs.com/) may contain all manner of javascript.

nathanlesage:
Should be possible, yes, since we also support Rmd. However, it shall be noted that there won't be dedicated syntax highlighting for the JSX parts of MDX.

PR #1881, opened 2021-04-16
add mdx as a type of markdown file

krry:
## Description

Adds `.mdx` as a valid markdown filetype, making `.mdx` files appear in the file manager and load in the editor.
https://github.com/Zettlr/Zettlr/issues/1880

## Changes

Wherever there is a list of filetypes that contains `.md`, I added `.mdx`. There were three such lists.

## Additional information

I refrained from attempting to handle MDX files differently, because that opens a hellaciously bottomless can of worms. This small change however, allows me and anyone else using `.mdx` as a suffix for their zettelkasten note files to use Zettlr.

Mac OS 11 Big Sur


krry:
Unless this is resulting from a conflict with my local eslint setup, you are likely aware that the lint fails on the current `develop` branch. Assuringly, I see the same number of warnings in my PR branch as show up in the develop branch without my changes.

nathanlesage:
Thanks for opening the PR!

Just a few comments. First, we're not even at 2.0, so 2.0.1 doesn't make sense, rather, add that to the changelog entries for 2.0. Secondly, please try to make the changelog message less … "emotional" …? It's better to word it in a way that we've added Mdx support and be done with it. However, you may add your GitHub handle in a similar fashion to whenever I mentioned other contributions in the Changelog.

Then: I think you caught all lists, and that reminds me that I have to consolidate that list \~somewhere\~ … but that's for another day!

krry:
Thanks for the comments, @nathanlesage, I've updated the PR accordingly.

nathanlesage:
Perfect, merging

PR #1906, opened 2021-04-26
fix window control display on native appearance

Brli:
## Description

On Linux (and, should also on windows), the window control buttons are overlaying shown on tabbar and window manager when "Native Appearance" is enabled.

By checking the native appearance status first, instead of old platform-first approach, we are certain that the buttons won't show on native appearance.

## Additional information

Arch Linux

### DIFF
--- a/source/common/vue/window/Chrome.vue
+++ b/source/common/vue/window/Chrome.vue
@@ -201,10 +201,15 @@ export default {
       return this.menubar
     },
     showWindowControls: function () {
-      if (this.platform !== 'darwin') {
-        return true
-      } else {
+      // Window manager of Windows and Linux should have them built-in
+      // Thus we don't display them on native appearance
+      if (this.useNativeAppearance === true) {
         return false
+      }
+      if(this.platform === 'darwin') {
+        return false
+      } else {
+        return true
       }
     },
     platformTitlebarHeight: function () {


nathanlesage:
I see where you're getting at. However, a slight caveat: We cannot have the native Windows styling. macOS is the only operating system where the window controls are independent from the window chrome. This means, on Windows you either get a titlebar, a menubar AND the window controls, or none at all. And since titlebar+menubar+toolbar looks extremely ugly on Windows (and since most programs don't have a menu bar at all anymore) we must use a frameless window on Windows.

This is why I disabled the native appearance checkbox on Windows. I only retained it on Linux since there are so many different window managers out there I cannot accurately control the styling of the display. Furthermore, some window managers such as the native Ubuntu one from Canonical follow the macOS styling and have the menubar at the top, not on the window. Thus, on Linux it makes sense to enable users to choose to use the native window chrome instead of the custom one.

This means, the window controls should *always* be visible, except for two cases: It's macOS OR the user has selected they want "native appearance" while they are on Linux.

Thus, the if-statement should read the following:

```js
showWindowControls () {
  if (process.platform === 'linux' && this.useNativeAppearance === true) {
    return true
  } else {
    return false
  }
}
```

Brli:
Nah

> This means, the window controls should always be visible, except for two cases: It's macOS OR the user has selected they want "native appearance" while they are on Linux.

thus the condition line should go nested.

nathanlesage:
Precisely! :) Now we just need to update the comment and the PR is ready to merge

Brli:
Comment part, done!

nathanlesage:
Thanks! Merging …

nathanlesage:
Mh, yes, the settings are immediately applied, meaning that the WindowChrome will immediately hide all custom window chrome, and instead expect the native WindowChrome to take over. However, since that would require a "re-start" of the windows, it's not happening.

I see the following options here:

* Just don't allow native appearance and do everything by hand (more control, but also less chance of doing something cool for Linux users, given the plentitude of window managers)
* Add a check so that after certain settings have been altered, automatically restart the application (problem here: What to do with unsaved changes? I still need some things to implement there …)
* A (possibly) very elaborate but also difficult to implement solution: Loop over all open windows, save their current state, and re-start the window (a.k.a. close, and re-open at the same position immediately) – however, saving the state of the windows in order to re-open the correct tabs might be a problem …
* Don't apply a changed window appearance immediately, similar to the language settings. That might be the easiest, possibly via writing the nativeAppearance config setting into an environment variable rather than pulling it from config immediately, and have the window chrome read in the environment variable instead. This way, the old value will be retained for as long as the app runs and will only be changed on the next boot

PR #1935, opened 2021-05-08
Scrolling on Left - Not Working (In Progress)

RandBY:
- [YES] I documented all behavior as far as I could do. Team 9 should be in the comments of the lines changed. 
  - [YES] I target the develop-branch, and *not* the master branch.
  - [PARTIALLY] This change works partially since the button for the fold-gutter is not in the same place as where we are supposed to click.  Otherwise, the scrolling works on the left side. 
  - [UNSURE] I have made use of ESLint using the provided configuration from the
  - [NOT YET] we haven't added anything in changelog as our solution is not functioning fully. 
  - [YES] I matched my code-style to the repository (as far as possible).
  - [YES] I do agree that my code will be published under the GNU GPL v3 license.
  - We will sync the changes in the future within updated code but we are currently working on this. 

## Description
Before the scrolling on the left side was not working. To solve that, we initially added padding to the left inside codemirror-scroll but that wasn't working with the fold buttons. Therefore, we made changes to the codeMirror.less file. This is partially working as the scrolling works but when we click on the button it does't respond, but if we click on the extreme left side of the folding button margin, it works.  This is a continual development process as part of our university course, and the contribution was made by 4 team members. 

## Changes
**Update codemirror.less**
- Added margin-left of 10px in:
CodeMirror-foldgutter
CodeMirror-foldgutter-open
CodeMirror-foldgutter-folded
CodeMirror-foldmarker

**Updating Editor.vue to add scrollable left margin**
- Changed CodeMirror margin-left to 0em.

## Additional information
Scrolling on the left side is functional and the folding functionality is working; however, the folding functionality is clickable on the extreme left instead of where the folding buttons are located.

Tested on: Windows 10 , Ubuntu 18.04.5 and MAC Big Sur 11.3.1 m1 chip.

### DIFF
--- a/source/common/less/geometry/codemirror.less
+++ b/source/common/less/geometry/codemirror.less
@@ -10,6 +10,7 @@
   // fonts such as Hiragana/Katakana are correctly
   // aligned. Should also help accessibility-plugins.
   &.CodeMirror-wrap > div:first-child {
+    //margin-left: 200px; // Rand added 
     textarea {
       font-family: inherit;
       font-size: inherit;
@@ -17,7 +18,11 @@
       height: 1.4375em !important;
       bottom: -1.4375em !important;
       transform: scaleY(1.1);
+     // margin-left: 500px; // Rand added 
+   
     }
+
+
   }
 
   font-size: 1em;
@@ -158,14 +163,18 @@
     border: none;
     width: 2em;
     padding-right: 10px;
+    margin-left: 10px;// added margin to the left
   }
   .CodeMirror-foldgutter {
     width: 2em;
+    margin-left: 10px;//// added margin to the left
   }
   .CodeMirror-foldgutter-open, .CodeMirror-foldgutter-folded {
     text-align: right;
     cursor: pointer;
     font-size: 18px;
+    margin-left: 10px; // added margin to the left
+
 
     &:hover {
       color: var(--c-primary);
@@ -178,7 +187,9 @@
     border-radius: @border-radius;
     display: inline-block;
     font-size: 75%;
-    margin-left: 10px;
+    margin-left: 10px;// added margin to the left
+  
+    //padding-left: 50px;
     padding: 2px;
     cursor: pointer;
     vertical-align: baseline;
--- a/source/common/less/geometry/codemirror.less
+++ b/source/common/less/geometry/codemirror.less
@@ -10,7 +10,6 @@
   // fonts such as Hiragana/Katakana are correctly
   // aligned. Should also help accessibility-plugins.
   &.CodeMirror-wrap > div:first-child {
-    //margin-left: 200px; // Rand added 
     textarea {
       font-family: inherit;
       font-size: inherit;
@@ -18,7 +17,6 @@
       height: 1.4375em !important;
       bottom: -1.4375em !important;
       transform: scaleY(1.1);
-     // margin-left: 500px; // Rand added 
    
     }
 
@@ -163,17 +161,17 @@
     border: none;
     width: 2em;
     padding-right: 10px;
-    margin-left: 10px;// added margin to the left
+    margin-left: 10px;//Added margin to the left (team 9)
   }
   .CodeMirror-foldgutter {
     width: 2em;
-    margin-left: 10px;//// added margin to the left
+    margin-left: 10px;//Added margin to the left (team 9)
   }
   .CodeMirror-foldgutter-open, .CodeMirror-foldgutter-folded {
     text-align: right;
     cursor: pointer;
     font-size: 18px;
-    margin-left: 10px; // added margin to the left
+    margin-left: 10px; //Added margin to the left (team 9)
 
 
     &:hover {
@@ -187,7 +185,7 @@
     border-radius: @border-radius;
     display: inline-block;
     font-size: 75%;
-    margin-left: 10px;// added margin to the left
+    margin-left: 10px;//Added margin to the left (team 9)
   
     //padding-left: 50px;
     padding: 2px;
--- a/source/win-main/Editor.vue
+++ b/source/win-main/Editor.vue
@@ -1,6 +1,6 @@
 <template>
   <div id="editor" ref="editor" v-bind:style="{ 'font-size': `${fontSize}px` }">
-    <div v-show="showSearch" id="editor-search">
+    <div v-if="showSearch" id="editor-search">
       <div class="row">
         <input
           ref="search-input"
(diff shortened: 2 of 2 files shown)


nathanlesage:
Thank you for opening the PR! However, it seems your branch is behind develop a little bit, since a lot of changes I made to the Editor.vue component are reverted in the current state of your PR, so you might want to pull new commits to this branch at some point.

Rajkiran2804:
Hi, I am working in the same team as @RandBY, who was the one that made the pull request. With the recent changes, we have been trying to adapt our solution but it is not merging together.  We have tried alternative solutions to solve the issue but none of them seemed to work.
We tried making new changes to the code and the scrolling was working but unfortunately the folding was not working.  Could someone please provide us with an advice on what we're doing wrong?

nathanlesage:
@Rajkiran2804 Hey, sure. What do you mean with "it is not merging together"?

As I said in one of the emails, this issue is particularly pesky … :/

nathanlesage:
I may have had a sudden idea. So we know already that using pure CSS doesn't work in order to get CodeMirror to do what we want. So how about we attempt to do it using JavaScript?

We have access to the left side of the editor area via `this.$refs.editor` in the file `Editor.vue`.

HAH! Just tried it out and works like a charm. And all of that with not a single change to the CSS!

Please take a look at this commit: e1f58ae8

I added three events to the outer editor element (which encompasses the full editor area, not just the CodeMirror instance itself): One to toggle mouse wheel events, and two for mouse down and up.

I already implemented the scrolling, which works easy as follows:

1. Listen to events that fire when the user uses the mousewheel while on the editor element (that is: To the LEFT and OUTSIDE of the actual CodeMirror element)
2. Call a custom function (onEditorScroll)
3. Make sure the event.target is actually the editor element (many events "bubble up", so we also receive wheel events that are triggered on the CodeMirror instance itself, but we don't want to handle those because they are already handled)
4. Just apply the `deltaY` (the amount the user has scrolled vertically) to the CodeMirror-scroller instead of the editor element (which is not scrollable)

Now **what I'd like you to do is to implement the ability for the user to also *select* stuff** when they are to the left of the CodeMirror element. How would you do this? Well, there are a few functions that come in handy which you'll want to use. Here's how that should work out:

1. Whenever the user presses the LEFT mouse button (you need a check for that), you can set some variable in the `data`-property of the Editor.vue-component that contains the start line where the user has pressed the left mouse button
2. When the user then MOVES the mouse (that event is not yet implemented, but you should be able to easily do so) while the left mouse button is pressed (e.g. after the mousedown-event has fired && the button was in fact the left mouse button), you need to select all lines from the start line to the current cursor position
3. When the user RELEASES the left mouse button, stop adding any lines to the current selection and unset that start variable
4. When the user then presses the left mouse button AGAIN, set that start line variable again and do the process anew. Please remember that you should REPLACE any existing selection in this case

How do you find which line to select? There are some nice utility functions in CodeMirror, cm.cursorCoords, cm.charCoords, and cm.coordsChar which you can use in order to translate x/y-coordinates into locations within the editor. Then you can easily find out the line you need to select. You can access the editor instance from `Editor.vue` via `this.editor.codeMirror`. Here's the manual: https://codemirror.net/doc/manual.html

Some notes on how the user experience should be:

* If you press and hold the left mouse button, the line where your mouse is currently add should be selected in its entirety. Any other selected text from the editor should be deselected (that is: You need to replace the current selection with another one)
* If the user then moves the mouse while holding down the left mouse button, any line the mouse cursor enters should be added to the selection (but if the user moves down and then moves up again, the selection should be decreased – that means you need one variable that indicates the start line for the selection. Then at any mousemove you just select everything from that start line to the line where the mouse cursor is)
* When the user then releases the left mouse button, you should unset the variable that indicates the start line.
* When the start line variable is unset, you know that you should not select anything

sayoniroy:
> I may have had a sudden idea. So we know already that using pure CSS doesn't work in order to get CodeMirror to do what we want. So how about we attempt to do it using JavaScript?
> 
> We have access to the left side of the editor area via `this.$refs.editor` in the file `Editor.vue`.
> 
> HAH! Just tried it out and works like a charm. And all of that with not a single change to the CSS!
> 
> Please take a look at this commit: [e1f58ae](https://github.com/Zettlr/Zettlr/commit/e1f58ae85cffe9d251fa233a36a4f47d1ae4554f)
> 
> I added three events to the outer editor element (which encompasses the full editor area, not just the CodeMirror instance itself): One to toggle mouse wheel events, and two for mouse down and up.
> 
> I already implemented the scrolling, which works easy as follows:
> 
> 1. Listen to events that fire when the user uses the mousewheel while on the editor element (that is: To the LEFT and OUTSIDE of the actual CodeMirror element)
> 2. Call a custom function (onEditorScroll)
> 3. Make sure the event.target is actually the editor element (many events "bubble up", so we also receive wheel events that are triggered on the CodeMirror instance itself, but we don't want to handle those because they are already handled)
> 4. Just apply the `deltaY` (the amount the user has scrolled vertically) to the CodeMirror-scroller instead of the editor element (which is not scrollable)
> 
> Now **what I'd like you to do is to implement the ability for the user to also _select_ stuff** when they are to the left of the CodeMirror element. How would you do this? Well, there are a few functions that come in handy which you'll want to use. Here's how that should work out:
> 
> 1. Whenever the user presses the LEFT mouse button (you need a check for that), you can set some variable in the `data`-property of the Editor.vue-component that contains the start line where the user has pressed the left mouse button
> 2. When the user then MOVES the mouse (that event is not yet implemented, but you should be able to easily do so) while the left mouse button is pressed (e.g. after the mousedown-event has fired && the button was in fact the left mouse button), you need to select all lines from the start line to the current cursor position
> 3. When the user RELEASES the left mouse button, stop adding any lines to the current selection and unset that start variable
> 4. When the user then presses the left mouse button AGAIN, set that start line variable again and do the process anew. Please remember that you should REPLACE any existing selection in this case
> 
> How do you find which line to select? There are some nice utility functions in CodeMirror, cm.cursorCoords, cm.charCoords, and cm.coordsChar which you can use in order to translate x/y-coordinates into locations within the editor. Then you can easily find out the line you need to select. You can access the editor instance from `Editor.vue` via `this.editor.codeMirror`. Here's the manual: https://codemirror.net/doc/manual.html
> 
> Some notes on how the user experience should be:
> 
> * If you press and hold the left mouse button, the line where your mouse is currently add should be selected in its entirety. Any other selected text from the editor should be deselected (that is: You need to replace the current selection with another one)
> * If the user then moves the mouse while holding down the left mouse button, any line the mouse cursor enters should be added to the selection (but if the user moves down and then moves up again, the selection should be decreased – that means you need one variable that indicates the start line for the selection. Then at any mousemove you just select everything from that start line to the line where the mouse cursor is)
> * When the user then releases the left mouse button, you should unset the variable that indicates the start line.
> * When the start line variable is unset, you know that you should not select anything

Hey Hendrik, 

Congratulations on finally solving the issue. 

We will start looking into the newly assigned task. We are going to have a meeting soon and update you about the process. 

Thank you for giving us an ample amount of advice to get started.

sayoniroy:
@nathanlesage we have sent a new pull request adding the features you've requested. https://github.com/Zettlr/Zettlr/pull/2000

nathanlesage:
@sayoniroy Thank you! Closing this in favour of #2000

PR #1974, opened 2021-05-18
A few improvements for (unit) testing

tobiasdiez:
## Description
This PR contains the test-related changes from https://github.com/Zettlr/Zettlr/pull/648.

## Changes
- Use configuration file instead of command line arguments for tests (so that other tools like VS can pick these up easily)
- Add VS launch config to run tests in current file
- Create test setup file that creates the needed browser environment 

## Additional information

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,8 @@
+'use strict';
+
+module.exports = {
+  require: ['./test/setup.js'],
+  extension: ["ts"],
+  spec: ["test/**/*.js", "test/**/*.spec.ts"],
+  require: "ts-node/register"
+};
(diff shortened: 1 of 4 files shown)


nathanlesage:
What exactly does the setup.js do? I assume it's setting up some general things …? But it would be great if you could add in a few explaining comments!

tobiasdiez:
To be honest, I have no idea ;-)

In the context of #648, some tests were failing because of missing objects (especially in global). After googling I found https://github.com/enzymejs/enzyme/blob/master/docs/guides/jsdom.md and that solved the problem.

I've now added a few comments to the setup file.

nathanlesage:
Sorry for the delay, but having wrangled with several different webpack targets and the horrors of an undefined `window`-object, I now understand the file. Merging (in the hope that this'll also fix the test failings)!

PR #1976, opened 2021-05-18
Run lint workflow also on push to branches

tobiasdiez:
## Description
Currently the lint workflow is not triggered if you push to e.g. the develop branch. This might lead to problems as linting issues are not detected.

## Additional information


nathanlesage:
Good catch, thank you!

Issue #1977, opened 2021-05-18
Migration from 1.7.4 to develop throws error


tobiasdiez:
## Description
Trying to start the develop version (from source, using yarn start) while having a somewhat older Zettlr installed (1.7.4), throws the following error:
```
[23:38:58] Migrating from 1.7.4 to 1.8.8!
TypeError: replacements is not iterable
    at ConfigProvider.runMigrations (D:\Programming\Zettlr\.webpack\main\index.js:133743:29)
    at new ConfigProvider (D:\Programming\Zettlr\.webpack\main\index.js:133588:14)
    at Object.bootApplication (D:\Programming\Zettlr\.webpack\main\index.js:130980:22)
Done in 14.99s.
``` 

## Reproducing
Not sure how. Where can I find the file that is migrated?

## Expected behaviour
No error.

## Platform
 - OS and version: Win 10
 - Zettlr Version: 1.7.4 > develop

nathanlesage:
That's in the config provider, let me search that … here: https://github.com/Zettlr/Zettlr/blob/c54651390a971c0de884924c83709c937a9fd3d5/source/app/service-providers/config-provider.ts#L280

But then I'm like if you have an old version installed you can upgrade by installing successively newer versions; I don't think we will catch every possible migratory error. Maybe add a notice regarding updates?

tobiasdiez:
Thanks for the pointer. I've opened #1986 to fix this.

Maybe it would be worthwhile to save a config file when releasing and have automatic tests in place that make sure the migration is successful?

nathanlesage:
Thanks for the PR!

With regard to tests, I don't really know. I usually add migrations to the config as necessary, and I've been pretty clear on the "always update" and I kind of think that it's fully okay to have people update properly anyway, or?

tobiasdiez:
I don't think you can expect that people always continuously update. Sometimes they don't have the rights to do so (e.g. university admins need to update programs) or don't use the device that often (laptops) or just don't have the time to update straight away.

PR #1986, opened 2021-05-19
Improve config migration

tobiasdiez:
## Description
As described in #1977, upgrading from older versions throws the error `TypeError: replacements is not iterable` since back then the autocorrect replacements were stored in a different format. This is fixed. 

## Additional information

Tested on: Win 10


nathanlesage:
And here we go!

Issue #920, opened 2020-06-12
[ENHANCEMENT] Hover over hyperlink shows note in popup


ruizjme:
## Description
Wikipedia-like hover on hyperlink shows popup of linked note's contents

## Proposed Changes
When hovering over a link to another note in a note, a popup appears on the link showing the (beginning of) the note's contents. E.g.

![image](https://user-images.githubusercontent.com/14925094/84476368-5c9a9b80-acd1-11ea-829c-033f812e6950.png)

## Caveats
n/a

PhilHem:
I could imagine that this sort of feature is pretty dependant on how people take notes. Personally, I don't necessarily take notes that are structured like Wikipedia articles. In such a case, this feature would just put out garbage. 

I could imagine that this could possibly be an option if there was some sort of text environment, that is marked to be used for such a preview. 

For example `### This is a preview text section ###`.

This would however mess with the idea that the markdown files should mostly adhere to common markdown language. A philosophical break like that... Don't know if its warranted.

nathanlesage:
I like how apparently my subversive display of Marxist stuff around the net begins to influence people :D 

Nevertheless: In principle an interesting idea, the question would be what to display, because I want to keep the workload low while maintaining speed. Therefore, I neither can load the content of all notes into your RAM (well, I could, but then we would have created a Sarlacc pit for RAM modules) nor request files on hover (would take too long to display).

Compromise: I can show you something like what is displayed when hovering over the tabs in the current beta; would be a compromise between fast access and keeping the RAM out of harms way

Derkness:
We are a group of 5 students from Adelaide uni. Our names are Tom, Jason, Ben, Shrey, and Ryan, and as part of our course, “Software Process Improvement”, we are tasked with making a contribution to one real-world open-source project. We’d like to contribute to this issue. We will be working from April until the end of May 2021 during our semester 1. Our Pull requests should be coming steadily throughout April & May

If anyone has any tips on where to start with that, please feel free to contact us.

nathanlesage:
@Derkness Thank you for the help! @tobiasdiez since you're the assignee – what's the status here?

tobiasdiez:
I've not yet started working on this, so @Derkness and his team are more than welcome to start with it!

Derkness:
[...]

Issue #1454, opened 2020-11-27
[ENHANCEMENT] [1.8.0] Support ZKN links in hovering tooltip


kyaso:
## Description

First of all, Zettlr 1.8.0 is great so far! The tooltip for links is really helpful, thanks for implementing. What I would like to suggest with this issue is, to also enable the tooltip for **ZKN links** ([[id]]). This would speed up navigating through your Zettelkasten by a lot.

## Proposed Changes

- Show tooltip also for ZKN links, i.e. [[...]]
- It would be good if this worked irrespectively of the content between the [[ ]]
    - So for both [[20201127123559]] or [[my-file]] the tooltip would show up
- Clicking on the link of the tooltip would simply do the same as Ctrl/Cmd clicking
    - i.e. open the file (if it exists) + search globally for the id/text.

## Caveats

Not sure.

nathanlesage:
Fixed by #1988

PR #1988, opened 2021-05-19
Note hover tooltip

Derkness:
- [Need Guidance. Where to put this?] I documented all behaviour as far as I could do.
  - [Yes] I target the develop-branch, and *not* the master branch.
  - [Yes] I have tested this extensively and paid extra attention to
  - [Two warnings I can't think of how to remove ] I have made use of ESLint using the provided configuration from the
  - [Unsure where to put this in the file] I have added an entry to the CHANGELOG.md.
  - [I believe so] I matched my code-style to the repository (as far as possible).
  - [Yes] I do agree that my code will be published under the GNU GPL v3 license.
  - [Yes] As far as JS-files are concerned, I made sure to copy (in case of new files)
  - [Not fully sure what this means] I synced the latest commits to develop shortly before proposing

## Description

Hovering over an internal note link will display a tooltip. This contains: The file name, the first 200 charactrs, the word count, and the time since modification (in days, or hours, or minutes, choosing the largest digit greater than 0)

## Changes

I added a command, and linked it through index.ts

## Additional information

This is based on issue #920 (https://github.com/Zettlr/Zettlr/issues/920)
Testing:
- emojis (works perfectly fine)
- tabs and spaces are reduced to a single space to reduce clutter
- lines of code, test with <title>test</title> have the html tags removed
- bold and Italics shows the ** and _ just like the note doc
- dot point lists (shows the * but of course shows inline. This is fine)
- tables show as text seperated by '|'
- day/hour/minute works as expected, and each threshold is passed to the next 'tier' up
Concern:
- code comments seem to not be visible

Tested on: Windows 10 (current build)


kyaso:
I actually started working on something similar, what a coincidence! Probably shows that a tooltip for ZKN links can be a very handy feature.

If I understood your code correctly, a crucial feature (in my eyes) is missing: being able to click on the link to open the target file (+ global search, if enabled in options). That's actually what I was planning to implement. However, I haven't really started with that yet, so I unfortunately can't give you hints on that right now..

Also, does your approach work for ID-based links as well? Eg, [[20210520101431]].

kyaso:
Maybe I should also mention what I had in mind.

I was just going to implement a simple tooltip that just shows the ZKN link content. When you click on it, the same steps happen as if you CTRL/Cmd + clicked the link in the classical way (that is: open target file (if present), and do global search).

So it would work any kind of ZKN link. Eg, [[20210520101431]] (ID-based), [[my_file]] (file name based), or even just a link which acts as shortcut for global search [[search term(s)]].

Btw, there is also an issue #1454 already that asked for ZKN tooltips (mentioning here for reference)

Derkness:
Thanks for your review!
The issue we were tasked with solving (#920) was just to hover over a link and get a tooltip. Since the link can be ctrl+clicked normally with pre-existing systems, that wasn't included in the tooltip. However, I can add that feature if that is the general consensus of this community.

WIth the ID aspect, I will try to find a way to add that aswell. It currently does not work, thanks for spotting that :D

I'm not quite sure what you mean with respect to a global search though. The issue linked is just about seeing a preview of the document's information. Unless you mean searching for the file itself, which this does do I believe.

kyaso:
Ah, I wasn't aware of #920.

With global search I meant the search that gets initiated whenever you CTRL+click a ZKN link.

I believe, the current `develop` branch doesn't seem to trigger a global search on CTRL+click anymore (#1989), but it works in 1.8.x.

Derkness:
Tooltip previews for ID based ZKN links is now implemented, thanks for the spot Kyaso!

Derkness:
Furthermore, surrounding having a button in the tooltip to link. Would that be something the community desires? It seems redundant to me since clicking the link will already redirect to the note, but I can duplicate this if desired :D

kyaso:
> Furthermore, surrounding having a button in the tooltip to link. Would that be something the community desires? It seems redundant to me since clicking the link will already redirect to the note, but I can duplicate this if desired :D

I personally would love to see a clickable link at the bottom of the tooltip. That way, one doesn't have to do CTRL+click everytime, and navigating becomes much faster.

We already have that feature for external ("normal") links, so having it for ZKN links also, makes sense, IMO.

Let's see what @nathanlesage thinks of this (and the overall PR, of course).

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+      return
+    }
+
+    // If there's already a tippy on the instance, don't re-render it
+    if (a.hasOwnProperty('_tippy')) {
+      return
+    }
+
+    // Initialise displayed attributes to 'loading'
+    let title = 'loading'
```
nathanlesage, line 25: Please add a comment `// TODO: Translate!` after any one of these lines, just so that I can perform a big, broad search for all of the TODOs shortly before releasing 2.0 so I don't forget to translate those strings (the important part is the all-uppercase "TODO" keyword)

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+    let content = 'loading'
```
nathanlesage, line 26: btw wouldn't it make sense to add an ellipsis here to indicate a process …? Or, even better, a loading spinner?

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+    let wordCount = 'loading'
+    let time = 'loading'
+    // Create a tippy. This will display the loading values
```
nathanlesage, line 29: The tippy below won't display the loading values, no, it'll just display "Searching For File ...". Did you mean to say "This will contain the values **after they have been loaded**"?

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+    let tooltip = tippy(a, {
+      content: 'Searching For File...',
+      allowHTML: true, // Obviously
+      interactive: true,
+      placement: 'top-start', // Display at the beginning of the anchor
+      appendTo: document.body, // anchor
+      showOnCreate: true, // Immediately show the tooltip
+      arrow: false, // No arrow for these tooltips
+      onHidden (instance) {
+        instance.destroy() // Destroy the tippy instance.
```
nathanlesage, line 39: Actually, it might make sense not to destroy these tippys. I'm not sure about the performance impact of leaving the tippys, BUT I think this way we could save loading time if the user accidentally leaves the element and immediately hovers back over it (because if there's a `_tippy` associated with the link, then tippy will itself react to mousemove events and thus, re-display the tippy).

Would you mind taking a look into the Tippy documentation and maybe fiddle around a little bit with leaving the tippy vs. not leaving them? And also think about the potential usability impact if the notes get edited. If we leave the tippy instance and the user never scrolls, the loaded values might become outdated, the question would only be if that is ever a problem… I think not because as soon as the user switches documents (to edit the linked one) this `event.target` would get destroyed either way alongside the tippy and would be re-rendered either way.

So long story short, you may want to play around a little bit with some settings and tippy! You can also take a look at the formatting toolbar (it's either a hook or a plugin of the editor) which I implemented, since that one features some fancy tricks with the Tippy instance.

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+      }
+    })
+
+    // Find the file's absolute path
+    ipcRenderer.invoke('application', { command: 'file-path-find', payload: a.innerText })
+      .then((filepath) => {
```
nathanlesage, line 45: I would refactor so that the callback just calls a different function with the loaded file contents. Plus I just saw you're invoking two asynchronous calls in succession, and that's not only callback hell but a hell to maintain. I don't have an ad-hoc solution, but the good solution will definitely use functions and a better state management. Mh … we have to get to this.

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+        // If the file is found
+        if (filepath !== 'Not Found') {
```
nathanlesage, line 47: Don't do this. Just don't. That's what `undefined` or `null` are for. Or `false` (although in this case it MIGHT make sense to better use `undefined`. I'll comment that on the main processes' command once I get there in a few minutes)

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+          // Retrieve the file contets
+          ipcRenderer.invoke('application', { command: 'get-file-contents', payload: filepath })
+            .then((descriptorWithContent) => {
+              // Remove html from content so html is displayed normally. (delete html tags)
+              descriptorWithContent.content = descriptorWithContent.content.replace(/(<([^>]+)>)/gi, '')
```
nathanlesage, line 52: Don't do this. I've recently added an HTML sanitizer that works fine for this. In general, I'd suggest we display a rendered HTML source (a.k.a. run the whole content through the `md2html`-source). But we might want to take a look at that, since something like this got me a CVE a few days ago, so … remove that line for now, but we must talk about this.

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+              content = descriptorWithContent.content.substring(0, 50)
+              if (descriptorWithContent.content.length >= 50) {
+                content += '\n' + descriptorWithContent.content.substring(50, 100)
+                if (descriptorWithContent.content.length >= 100) {
+                  content += '\n' + descriptorWithContent.content.substring(100, 150)
+                  if (descriptorWithContent.content.length >= 150) {
+                    content += '\n' + descriptorWithContent.content.substring(150, 200)
+                    // Add an elipsis if there is more left
+                    if (descriptorWithContent.content.length >= 200) {
+                      content += '...'
```
nathanlesage, line 65: … what are you doing here? This looks both unnecessarily tedious, has to many nested `if`-statements and I'm not sure why we'd need this in the first place …?

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                  }
+                }
+              }
+              wordCount = descriptorWithContent.wordCount // The word count
+              title = descriptorWithContent.name // The file name
```
nathanlesage, line 71: In case you mean "title" in the general Zettlr one then you need to implement the triage of horror I added in other places, where first of all a YAML frontmatter title is being used, THEN the first h1 heading, and only THEN the filename (but without extension in most places), depending on certain configuration options.

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+
+              let dateDif = Date.now() - descriptorWithContent.modtime
+              time = (dateDif / (60000)) // The time since modification
```
nathanlesage, line 74: Erm, for this we have `luxon` bundled. You don't have to manually (+ non-localised) implement that thing. Take a look at the file list item Vue component, there I'm using that to precisely display a time difference.

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+                if (time !== '1 Minute') {
+                  time += 's'
+                }
+              }
+
+              // On ready, show a tooltip with the note contents
+              tooltip.setContent(`File Name: "${title}"<br>"${content}"<br>Word Count: ${wordCount}<br> ${time} Since Modification`)
+            }).catch(err => console.error('File content get error: ' + err))
+        } else {
+          tooltip.setContent('File Not Found')
```
nathanlesage, line 96: Add a TODO translate comment here as well, please

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,100 @@
...
+              } else {
+                time = Math.floor(time) + ' Minute'
+                if (time !== '1 Minute') {
+                  time += 's'
+                }
+              }
+
+              // On ready, show a tooltip with the note contents
+              tooltip.setContent(`File Name: "${title}"<br>"${content}"<br>Word Count: ${wordCount}<br> ${time} Since Modification`)
+            }).catch(err => console.error('File content get error: ' + err))
```
nathanlesage, line 94: There's a grammatical error in here. Please rephrase to "Could not load file for preview: "

source/main/commands/file-path-find.ts
```diff
@@ -0,0 +1,50 @@
...
+    file = this._app.getFileSystem().findExact((arg as string), 'id')
+    if (file !== undefined) {
+      return file.path
+    }
+    // It's not an ID, so search each type of file
+    for (let type of FILETYPES) {
+      file = this._app.getFileSystem().findExact((arg as string) + type, 'name')
+      if (file !== undefined) {
+        // If we find it, then return it
+        return file.path
```
nathanlesage, line 42: You know what I just realised? Up in the editor hook you are running two successive asynchronous IPC calls to first retrieve the file path and THEN the note's contents. However, why don't you combine these calls, and instead of `return file.path` use `return await this._app.getFileSystem().getFileContents(file)`? Then you have everything you need, skip the second IPC call altogether and can safely refactor the code in the callback into its own function.

source/main/commands/file-path-find.ts
```diff
@@ -0,0 +1,50 @@
...
+    let file
+    // It might be linked by ID
+    file = this._app.getFileSystem().findExact((arg as string), 'id')
+    if (file !== undefined) {
+      return file.path
+    }
+    // It's not an ID, so search each type of file
+    for (let type of FILETYPES) {
+      file = this._app.getFileSystem().findExact((arg as string) + type, 'name')
+      if (file !== undefined) {
```
nathanlesage, line 40: Are you sure `findExact` returns undefined if it doesn't find anything? (If yes, then please remind me to change that to `null`, that would be an error on my side)

source/main/commands/file-path-find.ts
```diff
@@ -0,0 +1,50 @@
...
+        // If we find it, then return it
+        return file.path
+      }
+    }
+    // We can't find it, so return Not Found
+    return 'Not Found'
```
nathanlesage, line 46: Here's where you should return `null`, `undefined`, or `false`, but not a "Not Found" string.

source/main/commands/file-path-find.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+module.exports = FilePathFind
```
nathanlesage, line 50: Remove this line, that's the old NodeJS module system, and you already have `export default class` above, as it's supposed to be.

Plus, I think it might sound nicer if you had it named `FileFindPath` (or, with my changes, a different, more fitting name) – don't feel _too_ obliged to copy my "But File/Dir has to go first!!"-approach if that makes it sound like complete nonsense …

CONTRIBUTING.md
```diff
@@ -13,6 +13,8 @@ Zettlr lives through contributions from the community. You do not have to know a
     - [Development](#development)
       - [Our Stack (with linked documentation)](#our-stack-with-linked-documentation)
       - [Issue Labels](#issue-labels)
+      - [Setup](#Setup)
+        - [Windows](#Windows)
```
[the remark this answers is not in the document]
Derkness: In the conversation from (#920) it was discussed I could add a quick guide on how I got the development setup working. I also added a bit of text, was my format done badly?

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: Sorry. The intented functionality is to just display "Searching For File...". the fact the values are initialised to 'loading' is a remnant form a previous build plan. I can set them to initialise to 'null'

CONTRIBUTING.md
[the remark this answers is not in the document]
nathanlesage: Aaaah that's where that comes from. Please remove these changes from this PR, open a new PR for that and let's keep the PRs separate.

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: This was because if I just assumed there was enough content and added newlines, I would get an extra space at the end oof the content. I will refactor this so it's a loop with just a scaling conditional check each time though.

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: in this case it's just the filename, to properly identify the file being referenced (mostly useful in the case where an ID was used, but also sueful in order to 'ground' yourself when looking at a large tooltip)

source/main/commands/file-path-find.ts
Derkness: I believe so. I initially set this check to 'null', however it was being triggered when there was no file as 'null' was the response of findExact.

source/main/commands/file-path-find.ts
Derkness: I agree. I will change the 'file-path-find' to achieve this. Probably renaming it for its new functionality. Would you like me to keep an old version of the file-path-find around so that other people have a command to just grab the absolute path of a file?

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: Would you have any tips on where to put the gif for this spinner?

nathanlesage: I don't know if we should use a GIF – we could do this readily with SVG or even JavaScript-animated HTML. After all, we do have the space, and a single, continuous animation shouldn't be too demanding (but the quality improvements are massive).

One quick Google search yielded, e.g., this one: https://loading.io/css/

And I remember I had a loading spinner animation in an older version of Zettlr, maybe we could recycle that one by having a look into the old versions, it should be ~somewhere~ there!

source/main/commands/file-path-find.ts
nathanlesage: Nah, I don't think we should do that. I've learned the hard way that you should only implement something as soon as you need it, and not in good faith that maybe someone uses it some day, so it's safe to just overwrite that command! (After all, it's super simple to implement that if we really need it)

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: Sure, I'll put one of those in. Would you be able to tell me where to put the css for the note-preview.js file I made?

nathanlesage: hm... "Somewhere reasonable", I don't have the source code in front of me, so I'll leave the decision to you for now. I mean moving around a file shouldn't be a deal breaker! :)

Derkness: I am a bit stuck with trying to get my css into my javascript. From my research, I can simply import the document. However this file must use the 'require' method, and even then I cannot find the file.

My chosen location for the file is a in a folder called css in the hooks folder. Therefore, the following code should work:

const { tippyCSS } = require('css/note-preview.css')

right? Let me know if I am way off here, im not quite sure with this stuff.

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: It seems that leaving the tippy up does not prevent the content being changed. Probably because, as you said, when navigating away to change the tippy is removed anyway. Because of this, I have removed the onHidden event all together.

I think that this means that when a note is edited in a second of Zettlr, then the tooltip in the first instance won't be updated, but that doesn't seem like a massive deal compared to the cost of re-creating each hover time.

Derkness:
Side note: As I was ticking off all the changes in my head, I was marking them as resolved to collapse them. I thought this was local to just me, but I think that the standard approach is you are meant to mark them as resolved when you're happy with the changes. So I apologise if I distrupted the standard there.

Derkness:
Hey guys, sorry to bother but I'm just wondering if I could get another review? I believe I have made the changs requested. The only thing remaning is the possible loading spinner, but I have posted a question about that in the discussion for it.

I only ask for an update as this is for a Uni Assignment, and it is due in 3 days.

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,95 @@
...
+
+    // Find the file's absolute path
+    ipcRenderer.invoke('application', { command: 'file-find-and-return', payload: a.innerText })
+      .then((descriptorWithContent) => {
+        // If the file is found
+        if (descriptorWithContent !== null) {
+          // Get the contents of the file such as:
+          // 4 lines of 50
+          for (let i = 0; i < 4; i++) {
+            content = descriptorWithContent.content.substring(0 * i, 50 * i)
```
nathanlesage, line 48: I still do not understand why you go through the hassle and inserting linebreaks every 50 characters here. Why not take a substring and then at least provide an algorithm for finding the next space accounting for possible Markdown markup?

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,95 @@
...
+          // use luxon to get a local time difference
+          let dateDif = DateTime.fromMillis(descriptorWithContent.modtime).diffNow([ 'days', 'hours', 'minutes', 'seconds' ]).toObject()
+
+          // Display this using top down logic, i.e. use days, and if not, hours, then minutes, then just now
+          if (dateDif['days'] * -1 >= 1) {
+            time = Math.floor(dateDif['days'] * -1) + ' Day'
+            if (dateDif['days'] * -1 > 1) {
+              time += 's'
+            }
+            time += ' ago'
```
nathanlesage, line 71: Alright, you're using Luxon, but you're still doing weird stuff. Point taken, the Luxon docs are hard, but have you tried this function (or something along these lines)?

https://github.com/Zettlr/Zettlr/blob/develop/source/common/util/format-date.js

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,95 @@
...
+      interactive: true,
+      placement: 'top-start', // Display at the beginning of the anchor
+      appendTo: document.body, // anchor
+      showOnCreate: true, // Immediately show the tooltip
+      arrow: false // No arrow for these tooltips
+    })
+
+    // Find the file's absolute path
+    ipcRenderer.invoke('application', { command: 'file-find-and-return', payload: a.innerText })
+      .then((descriptorWithContent) => {
```
nathanlesage, line 42: It would still be great if you could move all that logic to a separate function which you call here (either by simply passing it as a reference, or by calling it from within the callback).

nathanlesage (changes requested):
It's coming together.

One thing: I literally have no time and also to prepare for an exam on June 1st, so if you notice someone else waiting for a PR review, **please explain to your instructor that I cannot guarantee reviews by May 31st, but will do so asap afterwards**. This way people whose notifications are drowning in the 60 I have still open here won't get any problems with examination.

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: I did try to do that, but for some reason I can't get the function into my code. Not sure what im doing wrong, as all the found usages use the 'import' method, whereas I have to use the require method. I am not sure why this like isnt working:

const { formatDate } = require('../../../util/format-date.js')

and then running it as:

formatDate(dateDif)

I have tried to research this, but I can't find the words to describe exactly what I'm doing.

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: I honestly just assumed I had to manually set newlines. I didn't realise tippy autoformatted.

source/common/modules/markdown-editor/hooks/note-preview.js
nathanlesage: You should be able to use `import { formatDate } from '../../../util/format-date'`

Requires are deprecated either way.

Additionally, you need to pass a single Date object (or milliseconds; you can pass `descriptorWithContent.modtime` to that function), because `Date.now()` will automatically be instantiated within the function. You do _not_ need to pass a timedelta.

Derkness:
I have made the changes. Specifically I have turned find-file-and-return into find-file-and-return-meta-data. I figued this would be better because it makes the function more specific, and you mentioned before leaving loose programming things around for others to maybe use is a bad idea.

Derkness:
This was working when I submitted it, but I changed branches to contribute to #2042, and then on switch back everything is broken. I assume this is just something to do with branch switching considering the pushed versino works, but please let me know.

nathanlesage:
> Additionally, my instructor has allowed for our presentation to go ahead without the completion of this, however I will still stick around to contribute anyway.

That is very nice and I appreciate the spirit! I'm right now dedicated to code-reviewing all your PRs!

source/main/commands/file-find-and-return-meta-data.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+      * Removes a file.
+      * @param {String} evt The event name
+      * @param  {Object} arg the parameters of the file to be deleted
+      * @return {Boolean} Whether the file was successfully deleted.
+      */
+  async run (evt: string, arg: any): Promise<any> {
+    let file
+    let MetaData
```
nathanlesage, line 34: Please use camelCase for this variable

source/main/commands/file-find-and-return-meta-data.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (file !== undefined) {
+      MetaData = await this._app.getFileSystem().getFileContents(file)
+    }
+    // It's not an ID, so search each type of file
+    if (file === undefined) {
+      for (let type of FILETYPES) {
+        file = this._app.getFileSystem().findExact((arg as string) + type, 'name')
+        if (file !== undefined) {
+          // If we find it, then return it
+          MetaData = await this._app.getFileSystem().getFileContents(file)
```
nathanlesage, line 46: You should break here to save some resources

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,49 @@
...
+    ipcRenderer.invoke('application', { command: 'file-find-and-return-meta-data', payload: a.innerText })
+      .then((metaData) => {
+        // If the file is found
+        if (metaData !== null) {
+          // On ready, show a tooltip with the note contents
+          tooltip.setContent(`File Name: "${metaData[0]}"<br>"${metaData[1]}"<br>Word Count: ${metaData[2]}<br> Modified: ${metaData[3]}`)
+        } else {
+          tooltip.setContent('File Not Found') // TODO: Translate!
+        }
+      }).catch(err => console.error('File path find error: ' + err))
```
nathanlesage, line 47: It might make sense to just log the error here (the reason is that this way we'll have the stack trace)

source/common/modules/markdown-editor/hooks/note-preview.js
```diff
@@ -0,0 +1,49 @@
...
+
+    // Create a tippy. This will display the loading values
+    let tooltip = tippy(a, {
+      content: 'Searching For File...',
+      allowHTML: true, // Obviously
+      interactive: true,
+      placement: 'top-start', // Display at the beginning of the anchor
+      appendTo: document.body, // anchor
+      showOnCreate: true, // Immediately show the tooltip
+      arrow: false // No arrow for these tooltips
```
nathanlesage, line 34: Wouldn't an arrow pointing to the link contents make sense …?

nathanlesage (commented):
Sooo, here's finally the promised review! In general it looks pretty good; there are a few æsthetic things I've commented, but in general it looks good. I'll be testing this over the next week and then I think it can be merged!

source/common/modules/markdown-editor/hooks/note-preview.js
Derkness: I am happy to set the arrow to true, however I should warn that that is inconsistent with the current hyperlink tooltips.

nathanlesage: Yes, the hyperlink tooltips don't use that. But the formatting toolbar and the footnotes tooltip use it. I think we need to focus on the question: "Should it point to something?" and I think for those note tooltips it makes sense to point back to the link!

Derkness:
Awesome! Those final changes have been completed and everything should be done :D (i think :P)

nathanlesage:
Works like a charm! Merging …

Derkness:
Awesome! I just wanted to thank you on supporting me throughout this project. I don't think I would have been able to do it without your help, and it was a very good experience to learn how open source worked. I know when you started this project you didn't know it would be the basis of a Uni assignment, but the fact this project has had that effect is quite cool!

nathanlesage:
Absolutely! It was a pleasure! And of course, feel free to continue contributing whenever you have time!

nathanlesage:
Just wanted to let you know that I've just attempted at vastly improving the display of these tooltips:

image

Hope you like it! It's in https://github.com/Zettlr/Zettlr/commit/8266492c9a5e8a56b2985eb944ec9958e6ef302e

Derkness:
Awesome! That looks a lot nicer! I'm glad my backend was still usable after the redesign.

PR #2000, opened 2021-05-22
Add the feature of selecting text from left margin

AfraaAlobaid:
- [ YES] I documented all behaviour as far as I could do.
  - [YES ] I target the develop-branch, and *not* the master branch.
  - [ YES] I have tested this extensively and paid extra attention to
  - [NO ] I have made use of ESLint using the provided configuration from the
  - [NO ] I have added an entry to the CHANGELOG.md.
  - [YES ] I matched my code-style to the repository (as far as possible).
  - [ YES] I do agree that my code will be published under the GNU GPL v3 license.
  - [ YES] As far as JS-files are concerned, I made sure to copy (in case of new files)
  - [YES ] I synced the latest commits to develop shortly before proposing

## Description
We have added the feature that enables the user to select text by pressing the left mouse button on the left margin. 

## Changes
1. Whenever the user presses the LEFT mouse button (you need a check for that), you can set some variable in the data-property of the Editor.vue-component that contains the start line where the user has pressed the left mouse button.
2. When the user then MOVES the mouse (that event is not yet implemented, but you should be able to easily do so) while the left mouse button is pressed (e.g. after the mousedown-event has fired && the button was in fact the left mouse button), you need to select all lines from the start line to the current cursor position.
3. When the user RELEASES the left mouse button, stop adding any lines to the current selection and unset that start variable.
4. When the user then presses the left mouse button AGAIN, set that start line variable again and do the process anew. Please remember that you should REPLACE any existing selection in this case.

## Additional information
Tested on: 

Ubuntu 18.04.5
Windows 10
macOS Big Sur v 11.3.1

### DIFF
--- a/source/win-main/Editor.vue
+++ b/source/win-main/Editor.vue
@@ -591,10 +591,33 @@ export default {
      */
     editorMousedown (event) {
       if (event.target !== this.$refs.editor) {
-        // return // Only handle if the event's target is the editor itself
+        return 
+      }
+      // if the button is the mouse left button, start selecting lines
+      if(event.button === 0){
+
+        // set the start point of the selection to be where the mouse was clicked
+        let startPoint = this.editor.codeMirror.coordsChar({ left: event.pageX, top: event.pageY});
+
+        // set the end point to be the same y coordinate as the start point and add the width of client page
+        // to get the end of the line. Couldn't find a way from CodeMirror to get the end of the line 
+        // as they treat every line as the whole paragraph 
+        let endPoint = this.editor.codeMirror.coordsChar({ left: event.pageX + this.$refs.editor.clientWidth, top: event.pageY});
+
+        // apply the selection of a single line that corresponds to where the mouse was clicked
+        this.editor.codeMirror.setSelection(startPoint, endPoint);
+
+        // if the mouse is still clicked and moved down or up, change the selection to include the new lines
+        document.querySelector("#editor").onmousemove =  (e) => {
+          // get the point where the mouse has moved 
+          let addPoint = this.editor.codeMirror.coordsChar({ left: e.pageX, top: e.pageY});
+          // use the original start point where the mouse first was clicked 
+          // and change the end point to where the mouse has moved so far
+          this.editor.codeMirror.setSelection(startPoint, addPoint);
+        }
+
       }
 
-      // TODO: Enable selection of full lines on the editor instance
     },
     /**
      * Triggers when the user releases any mouse button
@@ -602,11 +625,20 @@ export default {
      * @param   {MouseEvent}  event  The mouse event
      */
     editorMouseup (event) {
-      if (event.target !== this.$refs.editor) {
-        // return // Only handle if the event's target is the editor itself
-      }
 
-      // TODO: Stop the selection of full lines on the editor instance
+      // we have commented this if condition because when the user presses the mouse from the 
+      // left margin and goes inside this.$refs.editor and releases, the event of mouse release
+      // is not handled 
+
+
+      // if (event.target !== this.$refs.editor) {
+      //   return 
+      // }
+
+      
+      // when the mouse is released, remove event handling of mouse move (no more lines to be added)
+      document.querySelector("#editor").onmousemove = null;
+
     }
   }
 }
--- a/source/win-main/Editor.vue
+++ b/source/win-main/Editor.vue
@@ -6,6 +6,7 @@
     v-on:wheel="onEditorScroll($event)"
     v-on:mousedown="editorMousedown($event)"
     v-on:mouseup="editorMouseup($event)"
+    v-on:mousemove="editorMousemove($event)"
   >
     <div v-show="showSearch" id="editor-search">
       <div class="row">
@@ -90,7 +91,8 @@ export default {
       replaceString: '', // Models the replace string
       findTimeout: undefined, // Holds a timeout so that not every single keypress results in a searchNext
       // END: Search options
-      activeDocument: null // Almost like activeFile, only with additional info
+      activeDocument: null, // Almost like activeFile, only with additional info
+      anchor: undefined
     }
   },
   computed: {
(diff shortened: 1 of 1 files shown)


source/win-main/Editor.vue
```diff
@@ -591,22 +591,54 @@ export default {
...
+      // if the button is the mouse left button, start selecting lines
+      if(event.button === 0){
+
+        // set the start point of the selection to be where the mouse was clicked
+        let startPoint = this.editor.codeMirror.coordsChar({ left: event.pageX, top: event.pageY});
+
+        // set the end point to be the same y coordinate as the start point and add the width of client page
+        // to get the end of the line. Couldn't find a way from CodeMirror to get the end of the line 
+        // as they treat every line as the whole paragraph 
+        let endPoint = this.editor.codeMirror.coordsChar({ left: event.pageX + this.$refs.editor.clientWidth, top: event.pageY});
```
nathanlesage, line 605: That's a very good catch, and from the looks of it precisely what would work!

source/win-main/Editor.vue
```diff
@@ -591,22 +591,54 @@ export default {
      */
     editorMousedown (event) {
       if (event.target !== this.$refs.editor) {
-        // return // Only handle if the event's target is the editor itself
+        return 
+      }
+      // if the button is the mouse left button, start selecting lines
+      if(event.button === 0){
```
nathanlesage, line 597: You can merge this if-statement with the one above (or do a check for event.button !== 0 and return) — that would improve readability and reduce the chance that any changes afterwards end up accidentally AFTER the closing bracket!

source/win-main/Editor.vue
```diff
@@ -591,22 +591,54 @@ export default {
...
+
+        // set the start point of the selection to be where the mouse was clicked
+        let startPoint = this.editor.codeMirror.coordsChar({ left: event.pageX, top: event.pageY});
```
nathanlesage, line 600: You can convert these to `const` (improves readability as it is clear those variables won't be changed)

nathanlesage: Ah, and: That startpoint should rather be called anchor (because that makes it clear that this is the anchor of the whole selection even as it is being resized) AND also saved as a variable of the Vue component (that is, set this in `data` above).

The reason is that you can then factor out the custom mousemove-event you are listening to after mousedown has been pressed. Inside the mousedown-event (which will be triggered every time) you can then make a check that the anchor is actually set. So here you should set it to the given coordinates as you do, and on mouseup you should set it to `undefined` or `null`. In the mousemove-handler (which should be implemented just like the mousedown and mouseup) you can then at the beginning just check for `anchor !== undefined`

source/win-main/Editor.vue
```diff
@@ -591,22 +591,54 @@ export default {
...
+        // set the end point to be the same y coordinate as the start point and add the width of client page
+        // to get the end of the line. Couldn't find a way from CodeMirror to get the end of the line 
+        // as they treat every line as the whole paragraph 
+        let endPoint = this.editor.codeMirror.coordsChar({ left: event.pageX + this.$refs.editor.clientWidth, top: event.pageY});
+
+        // apply the selection of a single line that corresponds to where the mouse was clicked
+        this.editor.codeMirror.setSelection(startPoint, endPoint);
+
+        // if the mouse is still clicked and moved down or up, change the selection to include the new lines
+        document.querySelector("#editor").onmousemove =  (e) => {
```
nathanlesage, line 611: As indicated above, please convert this into a vue-event (`v-on:mousemove`)

source/win-main/Editor.vue
```diff
@@ -591,22 +591,54 @@ export default {
...
      */
     editorMouseup (event) {
-      if (event.target !== this.$refs.editor) {
-        // return // Only handle if the event's target is the editor itself
-      }
 
-      // TODO: Stop the selection of full lines on the editor instance
+      // we have commented this if condition because when the user presses the mouse from the 
+      // left margin and goes inside this.$refs.editor and releases, the event of mouse release
+      // is not handled
```
nathanlesage, line 631: Good catch!

source/win-main/Editor.vue
```diff
@@ -591,22 +591,54 @@ export default {
...
+      // is not handled 
+
+
+      // if (event.target !== this.$refs.editor) {
+      //   return 
+      // }
+
+      
+      // when the mouse is released, remove event handling of mouse move (no more lines to be added)
+      document.querySelector("#editor").onmousemove = null;
```
nathanlesage, line 640: As stated above: Instead of removing the event listener, set the anchor to null here

nathanlesage (changes requested):
Very good work, you've even caught two things I didn't anticipate. Basically there's only one last thing to do, that is, move the event listener for mousemoves into its own dedicated vue-event, and then I'll test and merge. But it does look very good so far!

AfraaAlobaid:
Thank you very much for the quick feedback. In the last commit the anchor has been saved to data in the vue component and the mouse move event is changed to a vue event and handled in the same fashion as other events.

source/win-main/Editor.vue
```diff
@@ -590,23 +592,55 @@ export default {
...
+      let endPoint = this.editor.codeMirror.coordsChar({ left: event.pageX + this.$refs.editor.clientWidth, top: event.pageY});
+
+      // apply the selection of a single line that corresponds to where the mouse was clicked
+      this.editor.codeMirror.setSelection(this.anchor, endPoint);
+
+      // if the mouse is still clicked and moved down or up, change the selection to include the new lines
+    },
+
+    editorMousemove(event){
+      if(this.anchor == undefined){
```
nathanlesage, line 615: Here it's safe to perform a strict check (`===`).

nathanlesage:
Looks almost perfect, there are just some code styling issues – after function names or "if"s must come a space before the left bracket, and the curly brackets should also be a little bit spaced. Do you have ESLint in use? It should automatically flag those things so that you can adapt these and then the PR looks ready to test & merge.

AfraaAlobaid:
Thank you Hendrik, now it should be styled properly.

sayoniroy:
@nathanlesage Hi Hendrik, is there anything else that we need to do regarding this issue? We were wondering if we could get an update. Thanks 😊

source/win-main/Editor.vue
```diff
@@ -590,23 +592,51 @@ export default {
...
       }
 
-      // TODO: Enable selection of full lines on the editor instance
+      // set the start point of the selection to be where the mouse was clicked
+      this.anchor = this.editor.codeMirror.coordsChar({ left: event.pageX, top: event.pageY })
+
+      // set the end point to be the same y coordinate as the start point and add the width of client page
+      // to get the end of the line. Couldn't find a way from CodeMirror to get the end of the line
+      // as they treat every line as the whole paragraph
+      let endPoint = this.editor.codeMirror.coordsChar({ left: event.pageX + this.$refs.editor.clientWidth, top: event.pageY })
```
nathanlesage, line 606: It's better to make this a `const`

nathanlesage (commented):
This looks wonderful! I'll be testing out this PR during the next week, and then it's ready for merging :)

nathanlesage:
I just tested this PR, and it works like a charm! Thank you very much. As soon as you are happy, I will merge it!

nathanlesage:
Excuse my confusion, I just realised there's no more need for changes! Merging …

Issue #1989, opened 2021-05-19
[develop] Clicking on ZKN links does not perform global search anymore


kyaso:
## Description

I noticed in the current `develop` branch, that when clicking on ZKN links, no global search is performed. For tags it still works.

## Reproducing
Ctrl+click on any [[ZKN link]] --> target file should be opened, but no global search is started

## Expected behaviour
When clicking on a ZKN link, the target file should be opened and a global search for the ID (or whatever is between the [[ ]]) should be started (if enabled in options).

## Platform
 - OS and version: Manjaro Linux
 - Zettlr Version: `develop`

## Additional information
I looked through the code, and found this section in `Editor.vue`:
```vue
this.editor.on('zettelkasten-link', (linkContents) => {
      ipcRenderer.invoke('application', {
        command: 'force-open',
        payload: linkContents
      })
        .catch(err => console.error(err))
})
this.editor.on('zettelkasten-tag', (tag) => {
  this.$root.$emit('start-global-search', tag)
})
```

nathanlesage:
I didn't (yet) re-implement that one, so keeping it open! Feel free to add that `start-global-search` thingy in a PR making sure it respects the config setting which you mentioned (and which I already have forgotten, shame on me!)

nathanlesage:
Already fixed by #2003

PR #2003, opened 2021-05-23
Re-enable global search when following ZKN links

kyaso:
## Description
This PR addresses #1989.

## Changes
In `Editor.vue`, where the `zettelkasten-link` event is handled, additionally start a global search (provided the user has enabled this behaviour in the settings).

## Additional Info
I haven't added the fix to the CHANGELOG yet, shall I do it?

---

**Tested on**: Manjaro Linux


nathanlesage:
Sorry for the late response, but thanks for re-enabling that! Merging …

PR #1959, opened 2021-05-12
Tray QuickNote

cskeogh:
## Description
Add a tray in the system's notification area with menu items: `Show Zettlr` & `Quit`. Defaults to off, controlled by advanced preference: `Leave app running in the notification area` or `Show app in the notification area` (MacOS)

Tray on Windows 10

## Changes
Full detail of the changes are in [tray design](https://github.com/UofA-SPI21-team7/zettlr-assignment/wiki/Tray-design).

## Additional information
Tested on:

 - Windows 10
 - MacOS
 - Linux Fedora 33 Gnome Desktop
 - Linux Fedora 33 KDE Desktop
 - Debian 10 XFCE Desktop

### Timeline

Sorry, this pull request is 8 days behind on our timeline. Our timeline is:

- 12 Apr 21 – provide our design and user interface mock-ups
- 3 May 21 – provide our initial code pull request
- 17 May 21 – provide our updated code pull request, addressing review comments
- 24 May 21 – provide documentation pull request

### Checklist

  - [ ] I documented all behaviour as far as I could do.

 Our [team](https://github.com/UofA-SPI21-team7) will raise a documentation pull request by 24 May 2021

  - [x] I target the develop-branch, and *not* the master branch.
  - [x] I have tested this extensively and paid extra attention to
        potential cross-platform issues (e.g. Cmd/Ctrl/Super-key bindings)
  - [x] I have made use of ESLint using the provided configuration from the
        repository's .eslintrc.json file, and it did not complain.
  - [ ] I have added an entry to the CHANGELOG.md.

 Will add entry to CHANGELOG.md later if review is successful

  - [x] I matched my code-style to the repository (as far as possible).
  - [x] I do agree that my code will be published under the GNU GPL v3 license.
  - [x] As far as JS-files are concerned, I made sure to copy (in case of new files)
        or adapt (in case of existing files) the info in the header.
  - [x] I synced the latest commits to develop shortly before proposing
        so that no merge issues occur.

Closes #175

PR #2026, opened 2021-05-27
Tray QuickNote

### LINKED ISSUE #175: [Feature Request] Tray QuickNote
When implementing a Zettelkasten, what you might want to do is have the option to quickly enter new thoughts into it. Therefore, we'll implement a Tray icon enabling a quick entry. This feature is targeted towards the `1.4` release, if possible.

## Proposed solution

1. In the preferences, enable the Tray QuickNote and set several options for easy access, such as:
    - The folder in which they are placed automatically
    - File name patterns (such as: `%id %firstline.md` for example)
2. When you have a thought or find something online, simply click on the Tray icon to bring up the quick entry window
3. Enter the thought using standard Markdown, highlighted using Zettlr's theme
4. Add links, tags and what else you need
5. Possibly overwrite the target folder and the filename
6. Click "Create" and have the note added to your Zettelkasten.

## ToDos

- [ ] Add a window class
- [ ] Add a service provider for the Tray Window
- [ ] Add a command to enter the final note
- [ ] Solve the Tray limitations on Linux systems. See the [Tray Class Documentation](https://electronjs.org/docs/api/tray)
- [ ] Testing

cskeogh:
## Description
Add a tray in the system's notification area with menu items: `Show Zettlr` & `Quit`. Defaults to off, controlled by advanced preference: `Leave app running in the notification area` or `Show app in the notification area` (MacOS)

Tray on Windows 10

## Changes
Full detail of the changes are in [tray design](https://github.com/UofA-SPI21-team7/zettlr-assignment/wiki/Tray-design).

@nathanlesage Thank you for your great review on the first pull request #1959. [Our team](https://github.com/UofA-SPI21-team7) has worked hard and incorporated all the comments. Our team would love feedback and hopefully merge 😀 before 31-May-2021 because on 31-May-2021 our contributions will be graded by our University (@ctreude and supervisors).

## Translation Strings

After merging please add the following translation strings:

```json
{
    "dialog": {
        "preferences": {
            "leave_app_running_in_the_notification_area": "Leave app running in the notification area",
            "show_app_in_the_notification_area": "Show app in the notification area"
        }
    },
    "system": {
        "error": {
            "tray_not_supported": "Tray is not supported. Gnome Extension &quot;KStatusNotifierItem/AppIndicator Support&quot; is required for Tray support on the Gnome Desktop."
        }
    },
    "tray": {
        "show_zettlr": "Show Zettlr",
        "tooltip": "This is the Zettlr tray.\nSelect Show Zettlr to show the Zettlr app.\nSelect Quit to quit the Zettlr app."
    }
}
```

## Additional information
Tested on:

 - Windows 10
 - MacOS
 - Linux Fedora 33 Gnome Desktop
 - Linux Fedora 33 KDE Desktop
 - Debian 10 XFCE Desktop

### Checklist

  - [x] I documented all behaviour as far as I could do.
See our teams pull request https://github.com/Zettlr/zettlr-docs/pull/103
  - [x] I have tested this extensively and paid extra attention to
  - [x] I have made use of ESLint using the provided configuration from the
  - [x] I have added an entry to the CHANGELOG.md.
  - [x] I matched my code-style to the repository (as far as possible).
  - [x] As far as JS-files are concerned, I made sure to copy (in case of new files)

Closes #175


forge.config.js
```diff
@@ -51,6 +51,60 @@ async function downloadPandoc (platform, arch) {
...
+  return new Promise((resolve, reject) => {
+    const shellProcess = spawn('/sbin/ldconfig', ['-p'])
+    let out = ''
+    shellProcess.stdout.on('data', (data) => {
+      out += data.toString()
+    })
+    shellProcess.on('close', (code, signal) => {
+      if (code !== 0) {
+        reject(new Error(`Failed to run ldconfig: Process quit with code ${code}`))
+      } else {
```
nathanlesage, line 73: I don't think you need the else here, since a reject automatically aborts function execution, or? (Just asking because down in L84 you do it)

source/app/util/environment-check.ts
```diff
@@ -130,5 +132,59 @@ export default async function environmentCheck (): Promise<void> {
...
+ * an {Error} with the reason why.
+ * Returns true on Linux with other desktops. e.g. KDE, XFCE, etc.
+ *
+ * @return {*}  {Promise<boolean>} If supported, returns true. Never returns
+ *              false.
+ * @throws {Error} Details why the Tray is not supported; or
+ *                 Details the error if an error occurred while checking Tray
+ *                 support.
+ */
+export async function isTraySupported (): Promise<boolean> {
```
nathanlesage, line 161: I'm currently thinking whether we should put the function into its own file …? Mh, yeah, makes sense. (Btw in this case you don't need to export it, since it's only being used in this file ;)

source/common/vue/form/elements/Checkbox.vue
```diff
@@ -1,16 +1,26 @@
...
+    <div class="form-control cb-group">
+      <label class="checkbox"
+             v-bind:disabled="disabled"
       >
-      <span class="checkmark"></span>
-    </label>
-    <label v-if="label" v-bind:for="fieldID" v-html="label"></label>
+        <input
+          v-bind:id="fieldID"
+          type="checkbox" v-bind:name="name" value="yes"
```
nathanlesage, line 9: Would you mind putting all of these properties on their own respective line? I've just allowed this in the linter because sometimes something only has those three attributes; but this field has many more, so it's okay to have the two additional lines!

source/common/vue/form/elements/Checkbox.vue
```diff
@@ -1,16 +1,26 @@
...
+        >
+        <span class="checkmark"></span>
+      </label>
+      <label v-if="label" v-bind:for="fieldID" v-bind:disabled="disabled"
+             v-html="label"
+      >
+      </label>
+    </div>
+    <div v-if="info" class="form-control">
+      {{ info }}
```
nathanlesage, line 22: You're using this as an info to display the error, right? I have to take a look at this how it looks …

source/global.d.ts
```diff
@@ -30,6 +30,7 @@ declare module '*.png'
 interface Application {
   runCommand: (command: string, payload?: any) => Promise<any>
   isBooting: () => boolean
+  isQuiting: () => boolean
```
nathanlesage, line 33: Isn't it "Quitting" …?

source/main/modules/window-manager/index.ts
```diff
@@ -227,6 +237,13 @@ export default class WindowManager {
       return
     }
 
+    this._mainWindow.on('show', () => {
+      const leaveAppRunning = Boolean(global.config.get('system.leaveAppRunning'))
+      if (leaveAppRunning) {
+        global.tray.add(() => this.showAnyWindow(), () => app.quit())
```
nathanlesage, line 243: Mh. This might cause trouble if the user opens the main window several times (since the tray would be added on and on and on).

Rather, what would be cool, is if the TrayProvider itself had the possibility of showing and hiding the Try depending on the Preference setting. That is: The tray Icon immediately shows/hides when the user clicks the checkbox. I think otherwise we might run into logical errors. (Plus we could remove that small piece of logic from here, so no one needed to query the TrayProvider ever, would make it more concise, wouldn't it?)

source/main/zettlr.ts
```diff
@@ -149,9 +154,11 @@ export default class Zettlr {
     // listen to close-events on the main window, we should be able to handle
     // this, if we ever switched to the auto updater.
     app.on('before-quit', (event) => {
+      this.isQuiting = true
       if (!this._documentManager.isClean()) {
         // Immediately prevent quitting ...
         event.preventDefault()
+        this.isQuiting = false
```
nathanlesage, line 161: What is the logic of this variable, exactly? I'm just asking because I see what you want to achieve, but I'm curious if that's not possible with the state of the application itself …?

source/win-preferences/preferences.vue
```diff
@@ -267,6 +267,17 @@ export default {
...
       }
     })
+
+    if (process.env.ZETTLR_IS_TRAY_SUPPORTED === '0') {
+      const leaveAppRunningField = modelToField('system.leaveAppRunning', SCHEMA['tab-advanced'])
+      if (leaveAppRunningField !== undefined) {
+        global.config.set('system.leaveAppRunning', false)
+        leaveAppRunningField.disabled = true
+        if (process.env.ZETTLR_TRAY_ERROR !== undefined) {
+          leaveAppRunningField.info = '⚠️ ' + process.env.ZETTLR_TRAY_ERROR
```
nathanlesage, line 277: I wouldn't use Unicode symbols but rather a Clarity icon for that. That however would need to be implemented in the Vue component. But anyway, I think we can simply remove that icon, since the info string should be explaining quite a lot, shouldn't it?

nathanlesage (commented):
I'm more than happy with the work. The code is clean, concise, and does what it says on the can. There are a few minor things (again, I'm sorry, but I'm only planning on a review-by-review basis …) but while you're at it I will already in a quiet minute go and test this out, because I'm really hyped for the Tray Icon :)

source/common/vue/form/elements/Checkbox.vue
cskeogh: This is a screenshot on Windows. Do you like the look?
Preferences with Gnome error

source/main/modules/window-manager/index.ts
cskeogh: > Rather, what would be cool, is if the TrayProvider itself had the possibility of showing and hiding the Try depending on the Preference setting. That is: The tray Icon immediately shows/hides when the user clicks the checkbox.

Immediately show/hide the tray is implemented in [L183 of index.ts](https://github.com/Zettlr/Zettlr/pull/2026/files#diff-87c7699f8e2c0214b588787d0bb57543927f3f8857444550bf0dbb35d5294c06R183).

We will investigate and fix the logic error and moving code to TrayProvider.

source/common/vue/form/elements/Checkbox.vue
nathanlesage: Interesting that this icon works like an Emoji :O I did not expect that!

But yeah, in general this looks good – albeit I would make the text much smaller and also grey – now it looks as if it didn't belong to the checkbox. I've had something like that in previous versions which I not yet restored … wait let me search for a screenshot.

Here: https://docs.zettlr.com/en/reference/settings/#advanced

The Advanced screenshot features two inputs with such info strings – they are very close to the input they refer to and a little bit grey. This way this warning wouldn't draw so much attention (it's only important for those who want the tray anyway)

source/main/modules/window-manager/index.ts
nathanlesage: Ah, yes: This line should be in the TrayProvider. So that the Tray Provider itself listens to the config changes and shows/hides the Tray on its own. This would keep the window manager a little bit more lean. And, we have a lot of space in the TrayProvider either way, so this makes sense. Also, it encapsulates the functionality a little bit more in one place!

source/main/zettlr.ts
cskeogh: Can quit Zettlr by either:

1. titlebar `x`; or
2. `File` → `Quit`; `Ctrl-Q`; Tray → `Quit` etc.

`isQuiting` is set true when **2** occurs. Therefore do not prevent quitting when **2** occurs, see [index.ts:L279](https://github.com/Zettlr/Zettlr/pull/2026/files#diff-87c7699f8e2c0214b588787d0bb57543927f3f8857444550bf0dbb35d5294c06R279)

I tried other methods, but the other methods behave identically for **1** and **2**. The Electron | [app](https://www.electronjs.org/docs/api/app) API doesn't provide any state to help.

I will add a code comment to clarify this. Are you ok with the `isQuitting` implementation @nathanlesage ?

I will fix the misspelling.

Issue #1846, opened 2021-03-31
Question mark in Spanish


hiperterminal:
## Description

In Spanish, every opening question mark (¿) is tagged as a grammatical error

## Reproducing

Anytime you open a question mark and write any word, is marked as error. For example:

`¿Word `

## Expected behaviour

An open question mark is not a mistake in Spanish, should not be underlined in red.

![2021-03-31_15-32-Zettl-bug](https://user-images.githubusercontent.com/10172036/113152413-66789580-9236-11eb-9f60-5ea7ecd553e7.png)

## Platform
 - OS and version: Ubuntu 20.10
 - Zettlr Version: 1.8.7.

## Additional information

Many thanks!

nathanlesage:
There are a lot of characters which are not necessarily words which currently fall prey to the spellchecker, but I have no solution at hand right now how to mitigate this … so if anyone has, e.g., some great regular expressions to share with this regard, feel free to!

nathanlesage:
I did have a solution at hand to mitigate that. Don't ask me what took me so long.

The ¿-questionmark is now considered a delimiter, just like brackets or quotes, and that should do the trick …

PR #2124, opened 2021-06-24
Stop marking correct uses of quotation marks as misspelled words

potato-aardvark:
## Description

The spellchecker no longer marks certain instances of quotation marks as misspelled words. Also, the spellchecker no longer marks the use of the inverted exclamation point as incorrect (cf issue #1846). 

## Changes

1. Added the inverted exclamation point to the list of delimiters recognized by the spellchecker. 
2. Added smart double quotes to the list of delimiters. 
3. Previously, the spellchecker checked the first and last character of a token to see if it was a quotation mark, and if it found one, removes it. Now, it continues to check until the first and last remaining characters of a token are not quotation marks.
4. Previously, the spellchecker marked tokens consisting only of quotation marks as a misspelled word. Now, it ignores such tokens.

The result is that some constructions which used to be incorrectly marked as misspelled are recognized as correct, e.g.:
> “Smart-double-quoted phrase ending in punctuation.”
> ‘Smart-single-quoted phrase ending in punctuation.’
> 'Dumb-single-quoted phrase ending in punctuation.'
> “‘Nested quotes’ and variations”

No breaking changes.

## Additional information

  - [X] I documented all behaviour as far as I could do.
  - [X] I target the develop-branch, and *not* the master branch.
  - [X] I have tested this extensively and paid extra attention to
  - [X] I have made use of ESLint using the provided configuration from the
    - No new errors or warnings
  - [X] I have added an entry to the CHANGELOG.md.
  - [X] I matched my code-style to the repository (as far as possible).
  - [X] I do agree that my code will be published under the GNU GPL v3 license.
  - [X] As far as JS-files are concerned, I made sure to copy (in case of new files)
  - [X] I synced the latest commits to develop shortly before proposing

Tested on: <!-- OS including version -> Windows 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ The Custom CSS can now be edited directly in the assets dialog where you can als
 - MDX supported as a type of markdown file
 - New File and Edit File can now fast rename without selecting the extension
 - Add a tray to the system notification area, off by default. To activate, see Preferences → Advanced → "Leave app running in the notification area" (or "Show app in the notification area" when using MacOS).
+- Fixed a bug that would mark some quotation marks as misspelled.
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
A little bit late, but thanks for fixing!

Issue #2151, opened 2021-07-02
[DEVELOP] Enable overwriting of `userData` path


nathanlesage:
Following up on @waterfallpit, it's a good idea to somehow enable users to set the whole configuration directory to another one using a command-line flag.

This would kill two birds with one stone:

1. We can safely detach a production configuration with all of its data from testing the GUI
2. We can finally solve the (quiet, but still existing) demands for a portable app.

Should be fairly simple to implement; requires basically a new line of code in `main.ts` as well as some changes to the `test-gui`-script.

If anyone has a few free minutes, feel free to already do a PR. All the code's literally there and simply requires copy, pasting, and adapting. (+ a look into the electron docs how to change the `userData`-path globally)

waterfallpit:
@nathanlesage I implemented this via --config-dir (to keep --config switch available). So far only (and minor) issue is that `Dictionaries` dir is created in the default location. Relevant reported issue: https://github.com/electron/electron/issues/26039
I'll make a PR later this weekend. I'd like to test it some more.

nathanlesage:
@waterfallpit Thanks for taking a look! So I think we can definitely keep waiting until the corresponding fix is in an Electron release.

But generally, I'd say we can actually remove the `--config`-switch perspectively and replace it with `--data-dir` (or something similar) since it basically supersedes the `--config`-switch. I can take care of moving the code for the `test-gui`-command to use that new CLI switch, because I don't see much use in keeping the additional switch (which introduces considerable boilerplate anyways).

nathanlesage:
Sometimes it helps to scroll through old issues: This has already been implemented.

PR #2152, opened 2021-07-03
Custom data/configuration directory

waterfallpit:
## Description
It provides a way to specify a custom data directory via `--data-dir=<dir>` switch, which replaces the existing `--config=<file>` switch as suggested by @nathanlesage in #2151.

## Changes
I made changes to `main.ts` to add new functionality and to `config-provider.ts` to remove `--config` switch. Code block in`main.ts` has to come before other imports, even though ESLint will complain.

## Additional information
I would suggest to use non-default data dir by default when running `yarn test-gui`. With `--data-dir` switch a user can always specify a custom data dir and even point to the default dir (e.g., `~/.config/Zettlr`) should they choose to do so.

Tested on: Xubuntu and Windows 10.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The Custom CSS can now be edited directly in the assets dialog where you can als
 - **Feature**: Zettlr now supports bibliography files on a per-file basis. Just set the wanted bibliography in your YAML frontmatter, within the `bibliography`-property.
 - **Feature**: Now Zettlr can export to PDF even without any LaTeX-distribution installed on the system.
 - **Feature**: The footnote editing logic has been improved. Now, multiline footnotes are handled appropriately, and you can safely use multi-line footnotes alongside the in-place editing feature.
+- **Feature**: Custom data directory via `--data-dir` switch.
 - 32 bit AppImages and Windows are no longer supported.
 - Removed the Pandoc installation item from the help menu.
 - Moved the Pandoc and XeLaTeX settings to the export tab in preferences.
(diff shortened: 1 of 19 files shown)


nathanlesage:
> Code block in main.ts has to come before other imports, even though ESLint will complain.

Really? Why? o.O

nathanlesage:
whoops.

Just had a look, and since the flag is above the import of the lifecycle hooks, I think I have some call to `userData` before any function runs. In that case ESLint is right, and I made an error!

waterfallpit:
@nathanlesage At first I put it just after all the imports and at the first run the app created several dirs in the default data directory and then throw errors for some other dirs that were not present in the dir supplied via `--data-dir`. With some tweaks it worked partially, to be honest, since `config.json` (and some other files) were placed in supplied custom dir.

nathanlesage:
Yeah the main question I had was: Why doesn't it work *after* you import the lifecycle functions? And the only question I can imagine for that to be is that there is some call to `userData` in one of the files that gets executed immediately, rather than while calling. I just had a look but couldn't really find a cause for this.

In general we must be able to run this code _after_ the import statements. Having to run it before them indicates an error on my side.

waterfallpit:
@nathanlesage I moved the new code block from `main.ts` into `lifecycle.ts` and moved it among the imports. If I put it just before import of `environment-checkup`, it works as it should, if I put just after that import, it creates a bunch of subdirs in the default location (and some in the custom location), throws an error and exits.

Looking at the `environment-checkup.ts`, there are some calls to `app.getPath('userData')` for `const REQUIRED_DIRECTORIES` (line 47 and on), whose subdir names correspond to the subdirs created in the default location (they don't get created at that point yet, though).

nathanlesage:
> Moving that constant declaration into environmentCheck function solves the issue and thus the new code block can be placed after all the imports in main.ts. If you agree with that change, I can submit a new commit for this PR (presumably on Friday).

Ah, I knew there was something! Yes, please. This is just an old habit of mine when Zettlr was still small and I didn't knew what it meant to execute some function calls right at the top of an imported file. This is the mistake of mine I was mentioning :)

Thank you for taking the time to search for it!

If you like, you can also put the DELIM variable into the environment check function, since it's only used there as well (makes the whole file cleaner)

waterfallpit:
@nathanlesage Should be alright now, I moved all constant declarations inside the functions. ESLint is happy now. Tested it with `test-gui` and packaged version only on Linux, since I didn't change the code for handling the custom data dir (tested that one on Windows before too, just to be sure).

nathanlesage:
@waterfallpit Since that code is pretty much platform-independent if it works on one platform it should work on all others as well! Looks good! Merging …

Issue #498, opened 2020-02-02
[HELP WANTED] Custom Mermaid.js Styling for Zettlr


nathanlesage:
こんばんは、

I just finished initial implementation of Mermaid. I understand that this thing is huge, and that there is a lot of flexibility with regard to the charts. I personally don't use charts (maybe I'll start now since the option is there lol), which is why I would like to **ask for help**!

A lot of you have requested the Mermaid feature, and I've gladly implemented it, **but** it needs custom styling. This has actually been mentioned somewhere on Twitter I think (?), and it's necessary. Why? Because the built-in themes don't have night mode support. See proof:

![image](https://user-images.githubusercontent.com/17251683/73613099-783eb880-45f2-11ea-9159-187be66ff37b.png)

But I thought hey let's go one step further and provide a minimal base theme for that thing, using the Zettlr-internal colours and fitting the appearance into the app itself. Mostly, it should suffice that we define it *once* in the Berlin theme, because anything will propagate to the other themes. This would also enable users to further customise the styling using CustomCSS!

But **here's where you come in**: Do you use Mermaid in your daily work? Then I would like to ask you to share as many different things you can do with Mermaid so that I don't forget something and we can prevent bugs in the future!

The goal is to arrive at a full list of all things that need styling (elements, class names, and the like) and write two themes.

So in each of these steps, I gladly receive any help, whether be it deciding upon the actual styles, in sifting through the Mermaid themes to collect the info we need, or even coding the LESS into the stylesheet!

I'll be releasing another beta this week, which includes the basic mermaid support, and from then we can work together to make this happen!

Cheerio!

## Resources

- [Here is the link to all themes and styles that come with mermaid, which we need to override](https://github.com/mermaid-js/mermaid/tree/develop/src/themes)
- [This file is where we need to put the custom theme, except it's too big, then we need a new file. Look out for the class "mermaid-chart"!](https://github.com/Zettlr/Zettlr/blob/develop/resources/less/theme-berlin/light/codemirror.less)

maehr:
Hi @nathanlesage 

I haven't used mermaid excessively yet. But here is a real example https://gist.github.com/maehr/55be6c49ac414f3ec464c2728026e56c
Till now I only used PlantUML. But I am going to ditch it in favor of mermaid and vega-light.

Cheers m

nathanlesage:
Perfect, thanks! I'll include that in the test file! :)

nathanlesage:
[...]

PR #2199, opened 2021-07-18
Styling for Mermaid Diagrams

ThaddaeusWiedemer:
## Description
Better styling for mermaid diagrams that is also consistent with the selected theme and works with dark mode.

## Changes
I originally proposed a CSS file for better mermaid diagrams in [this Reddit post](https://www.reddit.com/r/Zettlr/comments/oinqg8/css_for_better_mermaid_diagrams/). I've since extended it to all mermaid diagram types, all Zettlr themes, and added support for dark mode.

Changes were made mainly in two places: I'm passing settings to the mermaid API in `render-mermaid.js` to directly control some of mermaid's settings. Elements that cannot be changed through the API are restyled in the `codemirror.less`-file of the Berlin light theme.

I've also included a test file in the `test-gui` folder to test all different elements of all types of mermaid diagrams in the future.

## Additional information
During implementation, I realized that some of the ideas from my original CSS would not translate well to all elements of other mermaid diagrams. My goals with the new style were:

- **Visual clarity and simplicity.** Elements should have high contrast and diagrams should be easy to read.
- **Consistency with Zettlr's style and theme.** I'm using the primary color of the theme for highlights and am re-using established color palettes.
- **Consistency across diagram types.** I use the same font, font-color, stroke-color, etc. across similar elements for all diagram types.

## Caveats
Mermaid is still growing and improving, so the API doesn't allow direct control of some elements. I was able to restyle most things anyway through CSS (sometimes resorting to rather hacky CSS selectors), but some things just can't be restyled. These include:

- The spacing of elements
- The width of bold arrows and arrows in general in class diagrams, relationship diagrams, and state diagrams
- The extremely small font size in class diagrams and Gantt charts
- The positioning of labels in pie charts
- The color of different sections in Gantt charts

Because all the colors need to be defined in the themes CSS, but it is most effective to change most of mermaid's settings through the API, I am accessing the CSS variables in `render-mermaid.js` through the `getComputedStyle()`-function. If there is a better way to pass these variables on, please let me know.

Additionally, the mermaid API can only take style variables on initialization, but the CSS variables can only be accessed after at least some elements have been rendered. Therefore, the mermaid API is now initialized in the `markdownRenderMermaid` function. I didn't notice any problems with that.

Lastly, changing to night mode or switching styles doesn't update the rendered diagrams. They would need to be rendered again, which right now means the file needs to be closed and re-opened. I think changing the style should trigger a re-render, but I don't know where to include that.

## Preview
Here are some impressions:

image

image

image

image

In night mode:

image

In theme Bielefeld:

image

In theme Frankfurt:

image

Tested on: Windows 10 20H2


ThaddaeusWiedemer:
I guess this PR would justify that #498 is closed.

source/common/modules/markdown-editor/plugins/render-mermaid.js
```diff
@@ -39,6 +38,109 @@ const mermaid = require('mermaid')
     let codeblock = [] // Holds a mermaid code block
     let currentCursorPosition = cm.getCursor('from').line
 
+    // we need to initialize the mermaid API here, since we can only get the
+    // CSS color variables by querying the computed style of created documents
+    // get theme colors and font
+    let mermaid_bg0   = getComputedStyle(document.querySelector(".CodeMirror")).getPropertyValue('--mermaid-bg0').replace(/\s+/g, '')
```
nathanlesage, line 44: Please transform these into `const`s

source/common/modules/markdown-editor/plugins/render-mermaid.js
```diff
@@ -27,8 +27,7 @@ const mermaid = require('mermaid')
   'use strict'
 
   // Initialise the mermaid API
-  // TODO: Theming!
-  mermaid.mermaidAPI.initialize({ startOnLoad: false, theme: 'dark'/*, theme: null */ })
+  // mermaid.mermaidAPI.initialize({ startOnLoad: false, theme: 'base'})
```
nathanlesage, line 30: You can remove this part completely, since you've effectively copied it to after you've extracted the necessary styles!

source/common/modules/markdown-editor/plugins/render-mermaid.js
```diff
@@ -39,6 +38,109 @@ const mermaid = require('mermaid')
     let codeblock = [] // Holds a mermaid code block
     let currentCursorPosition = cm.getCursor('from').line
 
+    // we need to initialize the mermaid API here, since we can only get the
+    // CSS color variables by querying the computed style of created documents
+    // get theme colors and font
+    let mermaid_bg0   = getComputedStyle(document.querySelector(".CodeMirror")).getPropertyValue('--mermaid-bg0').replace(/\s+/g, '')
+    let mermaid_bg1   = getComputedStyle(document.querySelector(".CodeMirror")).getPropertyValue('--mermaid-bg1').replace(/\s+/g, '')
```
nathanlesage, line 45: Additionally, you might want to run the .CodeMirror query select once, and save it to a variable. This might make computation a little bit better (I don't know whether Chrome caches this value, though …)

nathanlesage (changes requested):
Looks good! I can definitely see that the CSS API deserves some more love on the side of Mermaid, but as far as I can tell it looks good. The only problem might be that I still want to refactor the themes, but that's something you don't have to deal with, since I have to figure this out myself. I will then make sure to migrate these styles alongside :)

The only thing you might want to do is move the stuff from `theme-berlin/light/codemirror.less` to this file: https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/editor.less

This will ensure those styles are being loaded only when there's a main editor instance (that is: in the main window as well as QuickLook windows). This way we have those styles very close to where we need them.

nathanlesage:
Oh, and, btw, thanks for even including a pretty comprehensive test file! That's an incredibly helpful addition!

ThaddaeusWiedemer:
Ok, thanks for all the comments, I will update everything throughout this week.

As for refactoring the themes:

I think you can easily see the logic I used for the mermaid diagrams: Use the main theme color as the highlight color, switch the pie-chart colors around so the main theme color gradient comes first, use shades of the editor background color as the diagram background and use a grey tone for lines and text.

So, if you happen to change some colors in the themes while refactoring, e.g. dark mode background, you can easily adapt the mermaid colors as well.

nathanlesage:
> So, if you happen to change some colors in the themes while refactoring, e.g. dark mode background, you can easily adapt the mermaid colors as well.

Will do! Thank you!

ThaddaeusWiedemer:
Just updated everything as discussed

nathanlesage:
Hey, sorry for my super late response – the past two weeks were completely busy. There are now some merge conflicts, because I did some refactoring to the themes – sorry for that! Would you mind resolving them? Apparently I messed up a little bit and GitHub mentions they're too complex to be resolved with the diff editor. Afterwards, I can finally merge them!

nathanlesage:
@ThaddaeusWiedemer Any news on this?

nathanlesage:
Marking as stale due to inactivity.

Issue #1783, opened 2021-03-12
[ENHANCEMENT] ctrl-T toggle between task-list, checked, no-task-list


ThaddaeusWiedemer:
## Description
Currently, `ctrl-T` toggles a line between task-list and no task-list. It would be amazing if this could instead be a cycle task-list > checked task-list > no task-list. This is the case in OneNote for example and enables you to toggle tasks with keyboard input only.

## Caveats
Of course, removing the task-list would now require two key-presses, but in my experience with OneNote this is not a problem.

nathanlesage:
Sounds interesting, have to fiddle around with this a little bit, but certainly sounds interesting!

PR #2241, opened 2021-07-28
ctrl-T toggle between task-list, checked, no-task-list

ThaddaeusWiedemer:
This implements the enhancement I suggested a while ago in #1783. I saw it on the roadmap to Zettlr 2.0, figured it was rather easy to do myself, so here it is.

A related thing that keeps bugging me with task lists is how existing list items are treated. Currently, if the selection is a normal list item, it is changed from ` * list item` to ` - [ ] * list item`, rather than to ` - [ ] list item`, which I would find much more intuitive. On the other hand, the current behavior has the benefit that after removing the task, the line goes back to being a normal list item.

I still think it would be better to replace any existing bullet (or numbering) with the list item and just reverting to a normal line if the task list is removed. I could add that behavior to this PR unless the current treatment is the preferred workflow. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ The Custom CSS can now be edited directly in the assets dialog where you can als
 - Add a tray to the system notification area, off by default. To activate, see Preferences → Advanced → "Leave app running in the notification area" (or "Show app in the notification area" when using MacOS).
 - Fixed a bug that would mark some quotation marks as misspelled.
 - Fix the visibility problems under night mode mentioned in issue #1845
+- `Cmd/Ctrl+T` now cycles between normal line/task list/checked task list to allow for ticking task list boxes using only the keyboard. Provided by Thaddäus Wiedemer (https://github.com/ThaddaeusWiedemer).
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
The reason I didn't yet implement it is because the whole logic is a little bit inflexible with regard to this (also the block items and inline items stuff). I'm not happy with it but also haven't figured out a way we could make _all_ Markdown shortcuts be more flexible … I'll have a look after next week, hopefully! (If not, as always: feel free to ping me _next_ weekend – please not this)

ThaddaeusWiedemer:
I understand. I'm happy to brainstorm ideas for a better solution together, although I'm not sure I can help much in terms of specific JS implementations, as my experience is rather limited there.

nathanlesage:
Hey! First of all, apologies for replying so late! Then, after thinking a lot about this, I'm now certain that this approach to the new logic is perfectly fine and we can go ahead.

With regard to the question:

> I still think it would be better to replace any existing bullet (or numbering) with the list item and just reverting to a normal line if the task list is removed.

I think that as well, so if you would like to, please implement it like this and then we can finally merge this PR. Again, apologies for responding so incredibly late!

Issue #2259, opened 2021-08-03
[Enhancement] Support attribute lists in fenced code block headers


vijoc:
### Description

Attribute lists in fenced code block headers are supported by [some markdown parsers](https://python-markdown.github.io/extensions/fenced_code_blocks/#attributes), and are required to use some features of e.g. pandoc [mermaid-filter](https://github.com/raghur/mermaid-filter#options). Currently, this does not appear to be supported in Zettlr, and using attributes or the curly brace syntax breaks syntax highlighting and mermaid rendering.

I'm happy to look into this in more detail and give a shot at implementing it if it's deemed useful.

### Proposed Changes

Enable syntax highlighting / mermaid rendering for code blocks where the mode is expressed in an attribute list. The attributes themselves can be ignored, but the language declaration should be supported with either the form ` ```{ #identifier .mermaid caption="Caption Text Here" }` or ` ```mermaid`. 

It would seem that there are at least two locations that would need to be updated: [the openRegex in `modes/multiplex.js`](https://github.com/Zettlr/Zettlr/blob/b74bcfd0c5287f2f00bc954a7448b4b20aca49a5/source/common/modules/markdown-editor/modes/multiplex.js#L391) and the [special case of mermaid rendering in `plugins/render-mermaid.js`](https://github.com/Zettlr/Zettlr/blob/24ea70311b92bbec6d5bb083862ad54e8093a403/source/common/modules/markdown-editor/plugins/render-mermaid.js#L50).

### Caveats

Parsing the language declaration from within the attribute list will complicate the parsing logic to some extent. 

### Do you Wish to Attempt Implementing this Yourself?

Yes

### Zettlr Version

Beta (if applicable)

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Operating System Version

Windows 10

nathanlesage:
This has been partially required by the R highlighting already, and I have that on my radar! It requires a few changes in the multiplexer but that should do it.

vijoc:
Glad to hear that this was on your radar already!

I imagine that in addition to the multiplexer the [Mermaid rendering plugin](https://github.com/Zettlr/Zettlr/blob/24ea70311b92bbec6d5bb083862ad54e8093a403/source/common/modules/markdown-editor/plugins/render-mermaid.js#L50) would also need to be updated, as in my understanding it currently only triggers on ` ```mermaid`.

Would a PR be welcome, or would you like to make the changes yourself?

[...]

PR #2278, opened 2021-08-05
Fenced code block attributes

vijoc:
## Description

This PR enables syntax highlighting and mode detection for fenced code blocks where the language is expressed in an attribute list, as discussed in #2259. For example, ` ``` { #identifier .js }` will be recognized as a JavaScript block.

## Changes

- Updated the generation of the regex used to match the highlight modes. 
- Extracted the generation logic into an utility function and added tests, as the function will likely also be useful when detecting Mermaid blocks in [Mermaid rendering plugin](https://github.com/Zettlr/Zettlr/blob/c9fbe1dac8de857b6a67e14420994cd93a9b2ac3/source/common/modules/markdown-editor/plugins/render-mermaid.js).

## Additional information

This PR does *not* touch the Mermaid rendering, and thus using attribute lists will currently still prevent the rendering of Mermaid diagrams.

Tested on: Windows 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ While in the past it was possible to use the common zoom shortcuts to only zoom 
 - Add a tray to the system notification area, off by default. To activate, see Preferences → Advanced → "Leave app running in the notification area" (or "Show app in the notification area" when using MacOS).
 - Fixed a bug that would mark some quotation marks as misspelled.
 - Fix the visibility problems under night mode mentioned in issue #1845
+- Enabled syntax higlighting for fenced code blocks that use attribute lists as per issue #2259
 
 ## Under the Hood
 
--- a/source/common/modules/markdown-editor/modes/multiplex.js
+++ b/source/common/modules/markdown-editor/modes/multiplex.js
@@ -221,6 +221,8 @@ const highlightingModes = {
 })(function (CodeMirror) {
   'use strict'
 
+  const generateRegexForHighlightMode = require('../util/generate-regex-for-highlight-mode')
+
   /*  This function is a copy of CodeMirror.multiplexingMode with a small modification made to
       the token function.  CodeMirror's multiplexing mode addon involves a brute force check of
       every internal mode object, and when these checks are regular expressions on long lines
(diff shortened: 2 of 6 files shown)


tobiasdiez (commented):
Thanks for your contribution! The changes look good to me.

As the only suggestion, I would ask you to migrate the new files to typescript.

vijoc:
Alright, I went ahead and migrated the new files to typescript.

tobiasdiez (approved):

nathanlesage:
A quick comment from me as well: I played around a little bit with the regexp and found this: https://regex101.com/r/HRsmm6/1

(It has some weird beginning instead of \s since there we have a multiline flag, so we can ignore that).

That regexp would improve in that it allows the identifier string to occur anywhere within the curly brackets. What do you think?

By the way, huge props for commenting that beast! I think I never understood another person's regexp faster!

vijoc:
> That regexp would improve in that it allows the identifier string to occur anywhere within the curly brackets. What do you think?

I implemented the regex with the assumption that the first "class" or dot-prefixed string is the language declaration, as this is what e.g. [python-markdown](https://python-markdown.github.io/extensions/fenced_code_blocks/#classes) requires. I don't see this requirement/behavior explicitly mentioned in [Pandoc documentation](https://pandoc.org/MANUAL.html#extension-fenced_code_attributes), but all the examples there follow that form. Additionally, there is this quote (emphasis mine):

> If the `fenced_code_attributes` extension is disabled, but input contains class attribute(s) for the code block, **the first class attribute will be printed after the opening fence as a bare word.**

I think the point is that you could (in theory) have classes that are valid language selectors, and thus the order in which the regexes are tested would affect the behavior. For example, if I were to declare my code block as ` ``` { .r .javascript }`, then it would either be recognized as `R` or `JavaScript`, depending on which was tested first.

The current regex in this PR -- [as demonstrated here](https://regex101.com/r/B3rWEL/1) -- will match the first dot-prefixed string after a curly brace, so you can use e.g. `#identifiers` and `key="value"` attributes before the language declaration as well.

I do think the current proposed regex could be simplified, but I'm happy with the behavior it produces.

> By the way, huge props for commenting that beast! I think I never understood another person's regexp faster!

Thanks! I wanted to take the time to properly comment it in order to understand it better myself as well.

nathanlesage:
@vijoc That is some decent amount of research. I must confess I did not put that much research into my own regular expression; I just assumed it. But everything you've found seems to indicate "the first class will be taken as the language identifier if applicable", and it does make sense. Thus, I think we should leave it as you had done previously!

vijoc:
@nathanlesage Thanks!

I have a couple of PRs I could follow up with, including updating the Mermaid renderer to use the new regex. I also noticed some of the tests are broken on Windows, due to things like OS-dependent path separators. I could try to fix those.

nathanlesage:
@vijoc Yes please, feel free to fix whatever you find! I noticed myself that quite a lot of commands and stuff in the pipeline doesn't work on Windows, so I'm also adamant on fixing that as well.

PR #2296, opened 2021-08-10
Enable @lackadaisical/defaults-generator

Kangie:
To validate defaults file inputs and generate defaults file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   },
   "dependencies": {
     "@clr/icons": "^5.4.1",
+    "@lackadaisical/defaults-generator": "^0.4.2",
     "@types/sanitize-html": "^2.3.2",
     "@types/semver": "^7.3.8",
     "@zettlr/citr": "^1.2.2",
--- a/source/main/modules/export/index.ts
+++ b/source/main/modules/export/index.ts
@@ -13,11 +13,12 @@
  */
 
 // Modules
-import path from 'path'
-import { spawn } from 'child_process'
-import YAML from 'yaml'
 import { app } from 'electron'
 import { promises as fs } from 'fs'
+import { spawn } from 'child_process'
+import generateDefaults from '@lackadaisical/defaults-generator'
+import path from 'path'
+import YAML from 'yaml'
 
 // Utilities
 import isFile from '../../../common/util/is-file'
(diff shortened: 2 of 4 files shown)


Kangie:
I eventually worked out the webpack issue, and in the meantime you've completely changed how the exporter works. :P

The previous PR (when you were writing intermediate files) was stripping and parsing in-document metadata blocks to produce a consolidated set of defaults file options. As it stands now we'll just have to let pandoc parse the options and hope that the user was smart enough to handle them correctly (variables vs metadata, etc).

Now that this is working I've immediately run into a few issues that I'll need to address before this can be merged. 

1. The JSON Schema that the module uses does not like being passed input-files; I've evidently been too strict when restricting path-like formats!
2. The error thrown when content fails validation provides output to console but doesn't render or read very well. That should be improved.

I'll put some work into the JSON Schema so that this is actually *usable* and bump the PR, so this should finally be over and done with shortly.

nathanlesage:
> I eventually worked out the webpack issue, and in the meantime you've completely changed how the exporter works. :P

… sorry 🥺

***

Maybe to ease your mind so that you know that your work is not futile, let me explain to a certain degree my intentions behind how and why I rewrote the exporter:

I wanted Zettlr to fully hook into the pipeline as intended by Pandoc, so to offload any work that Pandoc can do there. This makes the possibility of bugs much less likely and splits up the work evenly between Pandoc and Zettlr. Pandoc itself already handles a lot of edge-cases and potential problems with defaults files and YAML frontmatter variable resolution, so we can safely offload that.

However, Zettlr still has some work to do which Pandoc doesn't do. First of all, project exports and specific settings users can take require us to do a little bit of defaults parsing, so what your code is invaluable for is to make sure we don't accidentally spit out a defaults file that will make Pandoc show us the finger. Basically your PR adds quite a lot of validation that my cute pseudo-code doesn't do, so if a user makes a mistake NOW, Pandoc will throw an error, afterwards it won't.

That being said, I can think of two very, very, very, VERY helpful additional positions where we can make use of your validator:

1. We could actually validate the defaults files as users write them to help mitigate errors before Pandoc complains. That is: Send the YAML contents to main, have your validator look over it and print any errors that might appear.
2. Secondly, we could even also validate the YAML frontmatters of the file, that is, check that i.e. the `title` is in fact just a title, that the `bibliography` contains a valid file, and so on.

I think that would be an immense help for people and this way we could fully utilize your work.

Specifically, I'm thinking about using this function:

https://github.com/lackadaisical-documentation/defaults-generator/blob/817b09e900cdcdf436dce5b3713579550332a880/src/index.ts#L169-L179

Instead of logging the errors to console (by the way that is something you shouldn't do, we need some way to configure that, i.e. by passing a DEBUG env-variable or something, don't log something just because! :D) we could actually return them and display them to the users.

So to answer to your two caveats in the above comment:

I think it would be great if your function would return actually two properties. One: The actual defaults file contents, and secondly, any potential errors as a list. If that list has length zero, we know everything is well, but if it does contain errors, we know that we should probably bail.

With regard to displaying the errors, I'd say we can be really dumb there and simply whenever the user changes some YAML stuff, simply have the full thing run through your function and check for those errors) :D

PR #2428, opened 2021-09-06
Better accessibility

Michael55555:
- [x] I documented all behaviour as far as I could do.
- [x] I have tested this extensively and paid extra attention to
- [x] I have made use of ESLint using the provided configuration from the
- [x] I have added an entry to the CHANGELOG.md.
- [x] I matched my code-style to the repository (as far as possible).
- [x] As far as JS-files are concerned, I made sure to copy (in case of new files)

## Description
Just a couple of UI improvements here. You can see the precise changes in the screenshots and commit messages below. ;)

## Changes
Screenshot from 2021-09-06 20-56-22
Screenshot from 2021-09-06 20-57-26
Screenshot from 2021-09-06 20-57-56
Screenshot from 2021-09-06 20-58-29

Tested on: Pop_OS! v20

### DIFF
--- a/source/win-main/App.vue
+++ b/source/win-main/App.vue
@@ -699,4 +699,23 @@ export default {
 
 <style lang="less">
 //
+ /* width */
+::-webkit-scrollbar {
+  width: 10px;
+}
+
+/* Track */
+::-webkit-scrollbar-track {
+  background: rgba(255, 255, 255, 0.05);
+}
+
+/* Handle */
+::-webkit-scrollbar-thumb {
+  background: rgba(255, 255, 255, 0.3);
+}
+
+/* Handle on hover */
+::-webkit-scrollbar-thumb:hover {
+  background: rgba(255, 255, 255, 0.4);
+}
 </style>
--- a/source/win-main/DocumentTabs.vue
+++ b/source/win-main/DocumentTabs.vue
@@ -316,6 +316,14 @@ export default {
 <style lang="less">
 @tabbar-height: 30px;
 
+body {
+  &.dark {
+    div #tab-container {
+      border-top: 1px solid rgb(30, 30, 30);
+    }
+  }
+}
+
 body div#tab-container {
   width: 100%;
   height: 30px;
@@ -385,10 +393,6 @@ body div#tab-container {
     }
 
     transition: 0.2s background-color ease;
-  }
-
-  &.dark {
-    border-top: 1px solid rgb(30, 30, 30) !important;
   }
 }
 
--- a/source/win-main/DocumentTabs.vue
+++ b/source/win-main/DocumentTabs.vue
@@ -320,7 +320,7 @@ body div#tab-container {
   width: 100%;
   height: 30px;
   background-color: rgb(215, 215, 215);
-  border-bottom: 1px solid grey;
+  border-top: 1px solid grey;
   display: flex;
   flex-direction: row;
   flex-wrap: nowrap;
@@ -341,7 +341,17 @@ body div#tab-container {
     overflow: hidden;
     padding-right: @tabbar-height; // Push the filename back
 
-    &:hover { background-color: rgb(200, 200, 210); }
+    &:hover { background-color: rgb(200, 200, 210);
+      .close {
+        display: block;
+      }
+    }
+
+    &.active {
+      .close {
+        display: block;
+      }
+    }
 
     .filename {
       line-height: 30px;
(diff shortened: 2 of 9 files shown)


source/win-main/App.vue
```diff
@@ -699,4 +699,23 @@ export default {
...
+  background: rgba(255, 255, 255, 0.05);
+}
+
+/* Handle */
+::-webkit-scrollbar-thumb {
+  background: rgba(255, 255, 255, 0.3);
+}
+
+/* Handle on hover */
+::-webkit-scrollbar-thumb:hover {
```
nathanlesage, line 718: If I see this file correctly, you are not at all namespacing these custom styles, right? This would (on macOS) lead to scrollbars that are not automatically hiding, and on Windows it might make them look weird.

Generally, we should refrain from customizing the scrollbars since most operating systems have their built-in styles. But if this is better, what we must do at least is make sure these **do not apply to macOS**, since there scrollbars generally automatically hide and the native scrollbars work in both dark and light mode. So for macOS this won't work.

source/win-main/DocumentTabs.vue
```diff
@@ -316,11 +316,19 @@ export default {
 <style lang="less">
 @tabbar-height: 30px;
 
+body {
+  &.dark {
+    div #tab-container {
+      border-top: 1px solid rgb(30, 30, 30);
```
nathanlesage, line 322: This can be merged into the section below by refactoring the `body div#tab-container` below to:

```less
body {
  div#tab-container {
    // some styles
  }

  &.dark div#tab-container {
    // some dark styles
  }
}
```

source/win-main/DocumentTabs.vue
```diff
@@ -316,11 +316,19 @@ export default {
...
+    }
+  }
+}
+
 body div#tab-container {
   width: 100%;
   height: 30px;
   background-color: rgb(215, 215, 215);
-  border-bottom: 1px solid grey;
+  border-top: 1px solid grey;
```
nathanlesage, line 331: Since the macOS and Windows styles don't have a border-top (at least I think so …?) we have to reset these in both the Windows and macOS styles so it doesn't break the design.

source/win-main/DocumentTabs.vue
```diff
@@ -341,7 +349,17 @@ body div#tab-container {
     overflow: hidden;
     padding-right: @tabbar-height; // Push the filename back
 
-    &:hover { background-color: rgb(200, 200, 210); }
+    &:hover { background-color: rgb(200, 200, 210);
```
nathanlesage, line 352: The `background-color` should go on the next line

source/win-main/DocumentTabs.vue
```diff
@@ -435,15 +454,16 @@ body.darwin {
 
   &.dark {
     div#tab-container {
-      border-bottom-color: rgb(11, 11, 11);
+      border-top-color: rgb(11, 11, 11);
```
nathanlesage, line 457: Again, we need to make sure that this doesn't break the Windows and macOS styles

source/win-main/DocumentTabs.vue
```diff
@@ -435,15 +454,16 @@ body.darwin {
...
 
       div[role="tab"] {
         color: rgb(233, 233, 233);
         background-color: rgb(22, 22, 22);
         border-color: rgb(22, 22, 22);
+        cursor: pointer;
```
nathanlesage, line 463: I'm pretty sure this shouldn't be done on any platform. The `pointer` cursor is (at least as far as I know, please correct me if I'm wrong) always intended for web-links, and never (!) for anything regarding the GUI, including the document tabs. Think about it: Is there a pointer-cursor on your browser's tabs?

source/win-main/DocumentTabs.vue
```diff
@@ -457,7 +477,7 @@ body.darwin {
 
 body.win32 {
   div#tab-container {
-    border-bottom: none;
+    border-top: none;
```
nathanlesage, line 480: Ah alright, here's the reset — but I don't see a corresponding reset on the macOS section, right?

source/win-main/DocumentTabs.vue
```diff
@@ -514,10 +534,21 @@ body.linux {
...
+        &.active {
+          background-color: rgb(50, 50, 50);
+        &.hover {
+          background-color: rgb(55, 55, 55);
+        }
+          }
+
+      &:not(:last-child) {
+        border-right: 1px solid rgb(30, 30, 30);
+      }
```
nathanlesage, line 550: As I mentioned in my comment in the Discussions section, we might get away with merging the Linux section into the generic body section and then simply resetting the unwanted styles for Windows and macOS.

nathanlesage (changes requested):
I still haven't gotten to actually testing this out, but I've found a few initial things that we might want to think about either way. I'll get back to you once I finally was able to test this out!

Michael55555:
This pull request doesn't consider the different OS styles, so I better close it 😅

PR #2465, opened 2021-09-13
remove `brew install pandoc-citeproc` from static tutorial md

afogel:
## Description
Addresses the issue that was opened in the wrong repo (! sorry!) - https://github.com/Zettlr/zettlr-docs/issues/114#issuecomment-918456119

## Changes
I removed outdated brew-install instructions in the tutorial across the three languages.

## Additional information

Tested on: Mac OSX (Intel)


nathanlesage:
Aaaaaaaaaaaah! I see, I haven't touched the tutorial files in like ages.

Actually, we can even completely remove the Pandoc installation guidelines everywhere since Zettlr ships with Zettlr :D 

Would you mind doing this or should I?

afogel:
I'm happy to do that -- so, to be clear, Pandoc is no longer a dependency at all? Or rather that it's part of the Zettlr bundle? 

If the latter, than all that needs to happen is to remove the installation references, right? Thanks!

nathanlesage:
Exactly. Pandoc is now part of Zettlr itself, so whenever you install Zettlr, Pandoc is there right with it, and Zettlr will use the internal, bundled Pandoc. So yes, we can completely remove the Pandoc installation references!

nathanlesage:
Perfect, thank you so much!

Issue #2365, opened 2021-08-29
[Bug] All BETA Versions File manager not working: new files are not shown until I quit and start again the program


vsanchez:
### Description

I have hesitated to create this Bug Report because a problem this serious, happening since the first beta, I found it hard to believe it was not posted before. But I cannot find it. The problem is the following: 

1. Create a new file under a workspace or directory. I normally use the default ID. 
2. The file opens and I can work it, etc. But I cannot see it in the file manager. I need to reboot the program in order for it to show. Same thing if renaming, the old name will still show, as if the file manager is never refreshing.
3. It happens with all Beta versions (1 to 3) in both Windows and Linux. 
4. It happens regardless of the configuration of the file manager. In general, the file manager is hard to navigate, sometimes not allowing changes in directories, etc.
5. I do use 3 workspaces, for what it's worth, and the files are in a dropbox synced directory. Both the windows machine and the Linux machine work on the same directory. 

### Reproducing

1. Try to create a file. The file will be created and can be modified and saved, but it will not show in the file manager until a reboot is done.
2. Other weirdness navigating the file manager, but not that clear.

### Zettlr Version

Beta (if applicable)

### Specify version

1, 2, 3

### Your Platform

- [X] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Kubuntu 21.04 latest KDE, Windows 10

nathanlesage:
It could be that the file cache is acting up. Would you mind quitting the app and then removing the complete `fsal`-directory from the application data directory? Then the app will recreate that cache and that might help. If not, it would be awesome if you could try to create a file and then share the logs (to see the logs, activate debug mode in the preferences and then use the "Open logs" item in the Develop-menu).

vsanchez:
Hi Nathan,

Interesting: clearing the fsal directory did not help neither in windows or Linux. The interesting part of the log (this is the Linux side):

[14:50:51] [Warning] [R] [Main Window] Was about to add descriptor /home/victor/escritos/ZKM_Cocina/Datos generales/20210829144953.md to the filetree, but the parent /home/victor/escritos/ZKM_Cocina/Datos generales was null! (index.js:2)
[14:50:51] [Error] [R] [Main Window] TypeError: Cannot set property 'dir' of null (index.js:2)

This has happened in two of the workspaces but in the first, the system works well. 

[...]

PR #2492, opened 2021-09-20
Fix for bug #2365 (Problem creating and deleting files when a workspace name is a substring of another)

vsanchez:
## Description

When two workspaces located in the same directory are such that the name corresponds to the initial substring of the other, i.e. ZKM and ZKM_Work,  inserting files or deleting files in the latter will fail to be shown in the file browser.

## Changes

A change was made such that initial substrings should math a propper sub-path.

## Additional information

Tested on:  Linux and Win64

### DIFF
--- a/source/win-main/store.ts
+++ b/source/win-main/store.ts
@@ -57,10 +57,7 @@ function findPathDescriptor (targetPath: string, tree: any, treatAsAttachment: b
       if (targetPath === descriptor.path) {
         // We have the correct element
         return descriptor
-        // I "fixed" eslint's complaint that both descriptor.path and path.sep must be explicit strings, but since I am new to JavaScript and 
-        // its relationship to node.js, why do I need to do this? Isn't path.sep defined in node.js as to return a string? Perhaps, when building the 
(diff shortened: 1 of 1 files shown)


nathanlesage:
Will try this out later today. There are a few coding issues, the lint should therefore fail and tell you (if you go into the details of the Lint task) what issues there are! But in general I think this should work.

nathanlesage:
Quick update: I just realized that what you're doing is absolutely correct. Why? Because the descriptor will be a directory at the time of comparison, so the following holds true:

* `dir.path` + `path.sep` will still yield a valid directory path
* If the targetPath really starts with the directory, it will for sure include this said delimiter. If the path is the same (without trailing slash), the previous `if`-condition would've already returned true.
* Since we do not normalize paths, on Windows the paths will contain "\" as the separator, so path.sep is a sensible choice here.

So once the linter problems are solved we can merge. Hint: To get rid of the "Both sides must be strings"-problem, simply wrap `path.sep` in `String()`, i.e.: `String(path.sep)`

PR #2526, opened 2021-09-29
Remove basic markdown colors from Berlin; close #2243

### LINKED ISSUE #2243: [Enhancement] Remove color from simple markdown in existing themes
### Description

__Edit:__ Remove colors for simple markdown from existing themes, and other conservative tweaks.

Original issue below:

---

### Description

With Zettlr 2.0 coming up, I'd like to bring up the idea of removing the old colors and permanent markdown display for simple markdown e.g. `_italic_ and **bold**` text. I know it's kind of a signature look to have those lines marked in the theme color, but the colors and the symbols make it look crowded. I find them to be interfering with my thought process too much.

image

So much even, that on texts where I used a lot of simple markdown, I tend to prefer to edit the notes on [Nextcloud Notes](https://apps.nextcloud.com/apps/notes) (compatible with Zettlr) where the color is the same as the rest of the text.

image

Additionally, it would be more zen to hide those markdown symbols like Zettr does with URLs.

### Proposed Changes

In stead of showing simple markdown symbols like `_italic_ and **bold**`, we can make it look more modern and easier to read by showing _italic_ and **bold** in the color of the rest of the text, and only show the markdown code when the cursor is over it, much like Zettr does with links.

### Caveats

You will lose some of the signature look, but a new 2.0 major version is the ideal moment to make a bigger change. Just like the "Search" is a big change. Although technically the change might not be so big, since the same concept is already used for links, which show differently when the cursor is on them.

I think it has a modern and peaceful look when the text displays normally until cursor is over it.

Redsandro:
## Description
Remove diverging colors from simple markdown in Berlin light and dark

## Changes
css only

Tested on: Linux Mint 20.2

### DIFF
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -8,13 +8,13 @@ body.dark {
   #editor .CodeMirror,
   .quicklook .body .CodeMirror,
   .dialog .CodeMirror {
-    color: var(--grey-2);
+    color: var(--grey-0);
 
     // First the code blocks
     .cm-comment,
     .cm-fenced-code,
     .cm-inline-math {
-      color: var(--grey-1);
+      color: var(--grey-0);
     }
 
     .cm-fenced-code {
@@ -123,7 +123,7 @@ body.dark {
     .cm-link,
     .cm-strong,
     .cm-em {
-      color: var(--c-primary-shade);
+      color: var(--grey-0);
     }
 
     .cm-spell-error {
@@ -246,7 +246,7 @@ body.dark {
 
     // Code background
     .cm-comment:not(.cm-formatting):not(.cm-fenced-code) {
-      background-color: var(--beige-5);
+      background-color: var(--grey-5);
     }
   }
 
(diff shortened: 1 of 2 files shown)


source/common/less/theme-berlin/dark.less
```diff
@@ -123,7 +123,7 @@ body.dark {
     .cm-link,
     .cm-strong,
     .cm-em {
-      color: var(--c-primary-shade);
+      color: var(--grey-0);
```
nathanlesage, line 126: If you want `var(--grey-0)` as the colors, you can completely remove these CSS statements, since they will inherit the font color from the top level (see line 11).

The same happens in light mode.

nathanlesage (commented):
Looks good. I just have a suggestion regarding cleaning up the CSS.

source/common/less/theme-berlin/dark.less
Redsandro: :+1:

nathanlesage:
Perrfect!

PR #2527, opened 2021-09-29
add rounded colors to backtick codeblock similar to inline code; closes #2374

### LINKED ISSUE #2374: [Enhancement] [2.0] Code block border-radius consistent with inline code
### Description

Make the "code block" look slightly more polished and consistent with "inline code" styling.

### Proposed Changes

Find a way to mark the first and last line of a code block with a class so that we can apply a `border-radius-top` and `border-radius-bottom` to the first and last line of the code editor. Then the `border-radius` of a code-block could be made consistent with an inline code-tag through CSS styling.

### Example

Currently code-block looks like this:
image

Mockup of 4px border-radius:

image

---

Backlinks: #2369

Redsandro:
## Description

Make code block look more in line with inline code tag styling with rounded corners. For this, the first and last line of a code block need to be identified and tagged with a class.

image 
image

## Changes

Two new classes are introduced and ready for custom CSS. Styling with rounded colors is included with Berlin light and dark theme.

## Additional information

You can see that originally, attempts were made in `codeblock-classes.js` to do some conditional parsing to make sure the tagging of code blocks is efficient and doesn't do unnecessary passes. However, **prior to my PR, indented code blocks were broken. I did not touch this code**, nor did I fix the problem because I can't follow what the idea was exactly. I think this is best left to the original contributor to fix.

image
(Indented code block prior error left unchanged)

Tested on: Linux Mint 20.2

### DIFF
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -112,6 +112,16 @@ body.dark {
       background-color: #333;
     }
 
+    .code-block-line-open {
+        border-top-left-radius: 4px;
+        border-top-right-radius: 4px;
+    }
+
+    .code-block-line-close {
+        border-bottom-left-radius: 4px;
+        border-bottom-right-radius: 4px;
+    }
+
     /* LISTS */
     .cm-formatting-list-ul,
     .cm-formatting-list-ol {
--- a/source/common/modules/markdown-editor/hooks/codeblock-classes.js
+++ b/source/common/modules/markdown-editor/hooks/codeblock-classes.js
@@ -26,6 +26,8 @@ function applyCodeblockClasses (cm) {
   let needsRefresh = false // Will be set to true if at least one line has been altered
   let isCodeBlock = false
   let codeblockClass = 'code-block-line'
+  let codeblockClassOpen = 'code-block-line-open'
+  let codeblockClassClose = 'code-block-line-close'
 
   // This matches a line that starts with at most three spaces, followed by at
   // least three backticks or tildes (fenced code block).
@@ -44,6 +46,14 @@ function applyCodeblockClasses (cm) {
     const line = info.text
     const wrapClass = (info.wrapClass !== undefined) ? String(info.wrapClass) : ''
     const isCurrentlyCode = wrapClass.includes(codeblockClass)
+
+    if (i > 0 && codeBlockRE.test(cm.lineInfo(i - 1).text)) {
+      cm.addLineClass(i, 'wrap', codeblockClassOpen)
+    }
+
+    if (i > 0 && i < cm.lineCount() - 1 && codeBlockRE.test(cm.lineInfo(i + 1).text)) {
+      cm.addLineClass(i, 'wrap', codeblockClassClose)
+    }
 
     // Second, check if we are NOT inside a fenced code block. If we're not, but
     // the line is indented by at least four spaces, we have an indented code
--- a/source/common/less/theme-berlin/light.less
+++ b/source/common/less/theme-berlin/light.less
@@ -197,6 +197,16 @@ body {
       background-color: #eee;
     }
 
+    .code-block-line-open {
+        border-top-left-radius: 4px;
+        border-top-right-radius: 4px;
+    }
+
+    .code-block-line-close {
+        border-bottom-left-radius: 4px;
+        border-bottom-right-radius: 4px;
+    }
+
     /* LISTS */
     .cm-formatting-list-ul,
     .cm-formatting-list-ol {


source/common/modules/markdown-editor/hooks/codeblock-classes.js
```diff
@@ -45,6 +47,14 @@ function applyCodeblockClasses (cm) {
     const wrapClass = (info.wrapClass !== undefined) ? String(info.wrapClass) : ''
     const isCurrentlyCode = wrapClass.includes(codeblockClass)
 
+    if (i > 0 && codeBlockRE.test(cm.lineInfo(i - 1).text)) {
```
nathanlesage, line 50: Potentially faster solution (especially useful for documents with a lot of code blocks): `cm.getLine(i - 1)`. This doesn't include all the metadata of the line which we don't need here. (and in the next if-condition)

source/common/less/theme-berlin/dark.less
```diff
@@ -112,6 +112,16 @@ body.dark {
       background-color: #333;
     }
 
+    .code-block-line-open {
```
nathanlesage, line 115: I suggest we call the classes `code-block-first-line` and `code-block-last-line`. The open/close metaphor works better for inline elements. At least that's my impression here.

source/common/modules/markdown-editor/hooks/codeblock-classes.js
nathanlesage: Additional idea: We can also leverage the variable `isCodeBlock`. If that variable switches from false to true, this means the current line is the first line of a code block. If the variable switches from true to false, this means the previous line is the last line of a code block. If the variable remains `true`, this indicates it's regular code block line.

This might make this piece of code many times faster because we can refrain from any additional regular expression checks.

nathanlesage (commented):
Looks perfect. Had two ideas of how to maybe make the code faster since this function can potentially be called several times a second. What do you think?

Redsandro:
@nathanlesage thank you for your recommendations. Unfortunately after rebasing to `develop` I'm getting build errors. I cannot undo my rebase. I shouldn't have rebased. I was trying to make sure to get that 'did a recent fetch' checkbox.

Redsandro:
This works. Now I've created a potential fix for the pre-existing indented code block rendering errors, but it probably needs some love because I'm trying to work with some previously written logic I don't entirely understand.

So if we can merge this, I can create a new PR on top of that for the indented code problems.

source/common/modules/markdown-editor/hooks/codeblock-classes.js
```diff
@@ -64,6 +79,7 @@ function applyCodeblockClasses (cm) {
     // codeblocks themselves should not be styled)
     if (codeBlockRE.test(line)) {
       isCodeBlock = !isCodeBlock
+      codeBlockLines.push(i)
```
nathanlesage, line 82: Interesting solution ... but if I understand it correctly, this array doesn't include all `codeBlockLines` but only those at a switch from codeblock to markdown, right? I was thinking that we can make use of the toggle of the `isCodeBlock`-variable above and just add a new variable "was previous not code block?" ... right?

Redsandro: > this array doesn't include all `codeBlockLines` but only those at a switch from codeblock to markdown, right?

Yes, only the lines that match the pre-existing `codeBlockRE`.

> I was thinking that we can make use of the toggle of the isCodeBlock-variable above and just add a new variable

Since we need to look behind and ahead, I think that would be overly complex. I mean, you may save 2 bytes of memory per code block, but you'll have to detect that a boolean that is true was false in the previous round so you need to store the boolean in a new boolean at the end of the looped block, but there are cases where a `continue` is used so you cannot rely on end of loop execution, but maybe you can if you wrote the code and know the logic, but for a random contributor it's surprisingly hard to navigate the small routine as it is. Especially since `isCodeBlock` starts as `false` so `isCodeBlock = !isCodeBlock` is ambiguous and should be written as `isCodeBlock = true` but maybe not because perhaps a toggle was intended in some unfinished logic for the indented code block.

Sorry, I'm rambling. This is my attempt at an elegant solution that touches the least amount of code. If you prefer a boolean approach, could you show how you'd do it? The patch is minor enough to ignore this and create a new PR, although this PR allows edits by maintainers too.

Redsandro:
You don't like my _"touch the least amount of code"_ approach. :sweat_smile: 
Want me to redo the whole routine including pre-existing code?

nathanlesage:
I think my main problem is that you use an array, it looks like an abuse of data structures :D But then I might be just overly protective about the code!

nathanlesage:
But, alas, it works, so well, I'm gonna merge it! But if you'd like to, yes, feel free to rewrite the routine!

Issue #2503, opened 2021-09-21
[Enhancement/Bug] Improve/Debug markdown link click behavior in editor


raphraph:
### Description

_I'm on Zettlr 2.0.0-beta4_

When inserting a file from the "Other files" section of the sidebar, Zettlr automatically creates a link.

When the file is an image, it becomes a block element. But when the file is not an image, for example a pdf file, it inserts a link with the filepath as the url.

When clicking on this link (pdf file) in Zettlr, the file is opened in a web browser as if the link was a hypertext link (therefore formatted such as `https://home/user/path/to/pdf/file.pdf`) which does not work.

Manually adding `file://` (on Linux...) in front of the filepath makes Zettlr behave normally and correctly open the pdf file in the default pdf viewer.

### Proposed Changes

In the `handleDragStart` function in [Sidebar.vue](https://github.com/Zettlr/Zettlr/blob/c9fbe1dac8de857b6a67e14420994cd93a9b2ac3/source/win-main/Sidebar.vue) there is the following line :
```
const data = FILETYPES_IMG.includes(ext) ? `![${basename}](${uri})` : `[${basename}](${uri})`
```
I think that a string can be added here (for non-image filetypes) in front of the uri in order to produce a correct link, but this is surely system dependent...

### Caveats

I don't know for sure

### Do you Wish to Attempt Implementing this Yourself?

Yes

### Zettlr Version

Beta (if applicable)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Ubuntu 21.04

### Additional Information

I selected yes at "Do you Wish to Attempt Implementing this Yourself?" but even if it's a small fix I will need help. So if someone is willing to guide me, please feel free :)

nathanlesage:
> I selected yes at "Do you Wish to Attempt Implementing this Yourself?" but even if it's a small fix I will need help. So if someone is willing to guide me, please feel free :)

Thank you very much!!

You're on the right track, but I think you should look here: https://github.com/Zettlr/Zettlr/blob/develop/source/common/util/make-valid-uri.js

This function is being executed whenever the user attempts to open a Markdown link, so we might just add it there. I just saw [here](https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/plugins/render-links.js#L148) that the editor itself uses the older function. Let me have a look.

[...]

PR #2539, opened 2021-10-01
Manage uri logic when opening markdown links in main process

raphraph:
## Description
Creates a new command for managing uri logic directly in the main process when trying to open markdown  (Issue #2503).

## Changes
Moved IPC call from `open-markdown-link.js` directly in `render-links.js`.
Created the `open-markdown-link.ts`command regrouping old `makeValidUri()` and `openMarkdownLink()` logic.

## Additional information
This is a work in progress.


raphraph:
Fixed by other commits

PR #2589, opened 2021-10-09
Indented code block round corners; closes #2583

### LINKED ISSUE #2583: [Bug] Indented code block not working
### Description

Indented code blocks are not rendering properly.

### Reproducing

Use four spaces

### Zettlr Version

Right off the develop branch

### Specify version

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Linux Mint 20.1 AKA Ubuntu 20.04

### Additional Information

Followup to #2374 #2527

Redsandro:
## Description

Apply back tick code block style to indented code blocks as well. See #2583

## Additional information

I couldn't figure out where the `cm-comment` class on all indented words are coming from. So for now I just reset the `background-color` for `cm-comment` inside code-blocks on light and dark Berlin. It appears these changes propagate to other themes.

### DIFF
--- a/scripts/test-gui/test-files/Rendering/code.md
+++ b/scripts/test-gui/test-files/Rendering/code.md
@@ -1,0 +1,62 @@
+# Code blocks and indented blocks
+
+### Fenced code
+
+This is a _fenced code block_. It starts with three tildes and optionally a language identifier to change code formatting colors.
+
+```javascript
+var js = 432;
+let a = 1;
+let b = 2;
+let c = 3;
+```
+
+Fenced code blocks create collapsible code (see the little triangles on the left):
+
+```js
+while (true) {
+    if (!false) {
+        console.log("booya");
+    }
+}
+```
+
+### Indented code
+
+This is _indented code_. No color coding, but monospaced. Annoyingly wrapped in inline-code classes by CodeMirror. Make sure those classes don't interfere with the block classes.
+
+    var indented = true
+    // Remove a space from the font of this line
+    // To test if the code block splits properly
+    // and the styled corners are intact.
+
+Testing `inline code`.
+
+    readText()
+    Te002 Te002 changeObj.text002-1 changeObj.text002-1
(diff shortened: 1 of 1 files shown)


Redsandro:
Please don't merge this until either:

- I had time to do a refactor; or
- 2.0 will be released

nathanlesage:
> I couldn't figure out where the cm-comment class on all indented words are coming from. So for now I just reset the background-color for cm-comment inside code-blocks on light and dark Berlin. It appears these changes propagate to other themes.

This class is coming from the underlying Markdown mode and I thiiiiink (not completely sure) that the gfm mode basically slaps `cm-comment` onto everything code-ish, so it's probably okay, especially since you only target comments *within* codeblock lines. So it should be fine; in case feel free to add another test-file to the test file folder so that we can visually check that it works!

> Please don't merge this until either:

I wanted to do one last sprint today and fix all the bugs so that I can release 2.0 ideally next Friday. Do you think you manage to do the refactor during the week?

Redsandro:
@nathanlesage it is done. It's harder than I initially thought because the fenced code block looks both ahead and behind, while indented code blocks are "dumb" and prestyled using that underlying Markdown mode you mentioned. So I retained the array. I could have used maps or sets so that it would ease your mind semantically, but arrays are simply faster, and sets are only syntactic sugar with a semantic license to be used more broadly.

That underlying markdown mode was really driving me crazy because it does half the detection on its own but can't be used for start and end detection AFAIK.  At a future time it might make sense to disable it specifically for indented code if possible, then we can remove all the "undo" css code.

What is gfm mode?

Anyway, test it if you like. I have found no issues, I believe I tested quite extensively, but I can always have missed a use case.

nathanlesage:
Just saw the code test file – looks amazing!

Just say as soon as I can merge, then we can ship this tomorrow :)

PR #2641, opened 2021-10-16
Disable indented code block fix; closes #2637

### LINKED ISSUE #2637: [Bug] Four spaces triggers a code block, impacting Hierarchical Lists
### Description

When you start a line with four spaces it triggers the start of a code block.

Ex:
```
This is normal text
 This is still normal text
    This line is a code block
```

This is unexpected behavior, and would be ok on its own, but it impacts hierarchical lists which get prefixed with four spaces for each tier:

```
- This is a list item
    - This is a list item after I hit tab
```

In Zettlr pre-2.0 this behavior was not present.

### Reproducing

- Start a line with four spaces.
- Observe that you are now in a code block

Alternatively:
- Start a list (- or *)
- Press Tab to indent one tier
- Observe that you are now in a code block

### Zettlr Version

Stable (most recent version)

### Specify version

2.0.0

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Big Sur 11.6

### Additional Information

_No response_

Redsandro:
Hotfix disables `indentedRE` fixes #2637

### DIFF
--- a/source/common/modules/markdown-editor/hooks/codeblock-classes.js
+++ b/source/common/modules/markdown-editor/hooks/codeblock-classes.js
@@ -88,7 +88,7 @@ function findCode (cm) {
       // If this is the first line and either already indented code or prepended by an empty line
       if (prevLine >= 0 && (codeBlockLines.includes(prevLine) || blankishRE.test(cm.getLine(prevLine)))) {
         // If this is not preformatted markdown (e.g. a list)
-        if (!cm.getLineTokens(lineNum).some(token => String(token.type).includes('formatting-list'))) {
+        if (cm.getLineTokens(lineNum).some(token => String(token.type).includes('formatting-list')) === false) {
           let probeLine = 1
 
           // Skip ahead to the end of the potential code block


Redsandro:
@nathanlesage I completely rewrote this part to enable both fenced and indented code blocks again. See the last commit in this PR for your consideration.

Redsandro:
The linter is (suddenly?) broken on my machine. Let me manually fix those.

Redsandro:
It is done.

Redsandro:
#@! accidentally closed.

Redsandro:
I'm used to a CI environment so initially I panicked yesterday and wanted to get rid of the issue ASAP. I've been working all night on this rewrite and it's probably close to as good as it gets. I hope it is pretty. :blush:

nathanlesage:
Looks good to me, the code makes sense and seems to do what we want it to!

With regard to this comment:

image

This won't be a problem for too long, since I plan to migrate to CodeMirror 6 in some time, so we can fully implement this whole yadda yadda ourselves.

The only potential performance improvement would be to check if a regex is easier to check for rather than `getLineTokens` as (as far as I undertand) that method is pretty heavy to compute, vis-a-vis a simple regex testing for `/\s*(\d\.|*|+|-)\s.+/`.

However, I'll first merge it so it's in the code base, and then have a look!

btw this code is the first one to explicitly use lodash, I've always refrained from using it because my brain somehow thinks that it's only for polyfills, which seems to be outdated knowledge :D 

(And yes, it does look pretty!)

nathanlesage:
> lodash is the swish army knife of missing functionality in javascript, originally an optimized fork of underscore, and is sometimes used as an example for ECMAScript to see what to implement in official ES next.

Precisely. I have always shied away out of ignorance, and also because I want to learn JavaScript quirks myself. Takes me an extra mile, but yeah, lodash is pretty much nice for all those small utility functions, so I'd argue it's good to use it. And I think the dependency won't go away soon because so many libraries depend on it, but you are right that it might make sense to pull in the most recent version of it via package.json, so feel free to do so!

And with regard not to pay too much attention to CM5 in view of the incoming CM6, yes, absolutely!

PR #2673, opened 2021-10-19
downgrade electron to 14.1.1 (2) fixes #2643

### LINKED ISSUE #2643: [Bug]  Mouse scroll broken in latest dev
### Description

I pulled dev to do a hotfix and noticed mouse scrolling is no longer working.

Previously, one tick of the mouse scroll would scroll one or more lines of text. Now, one tick of mouse scroll moves _one row of pixels_

### Reproducing

Fetch latest dev. I'm assuming this was introduced in the 2.0 release.

### Zettlr Version

Right off the develop branch

### Specify version

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Linux Mint 20.1 AKA Ubuntu 20.04

### Additional Information

I consider this bug serious since not being able to scroll is a highly visible and irritable bug. However at quick glance I do not see any commits that are related to mouse scrolling. So it's a bit puzzling.

Redsandro:
#2643 is high priority and we're all at a loss as to what causes the problem. The quickest fix is to downgrade from Electron 15 to Electron 14. Effectively this means Zettlr 2 goes from Electron 13 to 14 rather than from 13 to 15. So unless there are any other known problems with this version, this may be the best action to get this fixed for now.

Personally I consider the (merged) #2641 high priority too, so these two together can be a new Zettlr patch release.

It's probably never a good idea to upgrade a major version of crucial dependencies like Electron right after the last beta and before the final release. Consider locking major versions of dependencies by the time betas or release candidates are released for public testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "cross-env": "^7.0.3",
     "css-loader": "^6.4.0",
     "csso": "^4.2.0",
-    "electron": "^15.2.0",
+    "electron": "^14.1.1",
     "electron-builder": "22.13.1",
     "electron-devtools-installer": "^3.2.0",
     "electron-notarize": "^1.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -325,7 +325,7 @@
     xterm-addon-fit "^0.5.0"
     xterm-addon-search "^0.8.0"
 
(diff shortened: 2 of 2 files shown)


nathanlesage:
Closing as (fortunately) fixed!

PR #2687, opened 2021-10-20
add lodash to package.json; #2637

Redsandro:
This is a technicality. In case you want dependencies to be explicitly defined, even though they are already subdependencies.

@nathanlesage [commented](https://github.com/Zettlr/Zettlr/pull/2641#issuecomment-945740306):
> it might make sense to pull in the most recent version of it via package.json, so feel free to do so!

I'm not sure how this changes the compiled code since `lodash` is already `require`'able, but after compiling things still work the same so at least it doesn't change compiled code into a problem. :grin:

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "joplin-turndown-plugin-gfm": "^1.0.12",
     "katex": "^0.13.18",
     "ky": "^0.28.6",
+    "lodash": "^4.17.21",
     "luxon": "^2.0.2",
     "md5": "^2.3.0",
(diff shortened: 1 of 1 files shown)


Redsandro:
That sounds simple enough, and I know empirically it works, but the versions still puzzle me: There are three different older versions as subdependencies and the new one in package.json is the latest version. The idea is that every dependency can use their own version of a subdependency. Doesn't that mean that this new different version is appended \*somewhere\*?

Redsandro:
Thank you for explaining. I feel more at ease now. :+1:

PR #2688, opened 2021-10-20
auto focus query input on global search; close #2605

### LINKED ISSUE #2605: [Bug] Ctrl+Shift+F shows global search box, but does not give it focus
### Description

When I press "Ctrl+Shift+F", the global search shows in the sidebar, but it does not receive focus, so I have to manually click on the search box.

The [docs indicate](https://docs.zettlr.com/en/core/search/#searching-globally) that receiving focus is the intention, and would be much more useful/usable (and I think it worked this way for me on v1.x):

> The search input in the toolbar will receive focus

### Reproducing

* Press Ctrl+Shift+F
* Global search shows on the left
* The "Find ..." box is not focussed

### Zettlr Version

Beta (if applicable)

### Specify version

2.0.0-beta.5

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [X] ARM 64bit

### Operating System Version

Windows 10

### Additional Information

This was also the case on an earlier v2 nightly, though I didn't write down my version I'm afraid.
I haven't used a v2 which doesn't have this issue (only tried Windows 10 on arm64)

Redsandro:


nathanlesage:
Ah, found the error. You need an additional check:

```js
if (this.$refs['global-search'] !== undefined) {
  this.$nextTick(this.$refs['global-search'].focusQueryInput)
}
```

... and then define the function `focusQueryInput` in the GlobalSearch component's methods section as I mentioned:

```js
focusQueryInput: function () {
	this.$refs['query-input'].focus()
}
```

The reason you're getting the undefined error sometimes is because if the whole split view component is hidden (i.e. if neither the file manager NOR the global search are visible), the components are unmounted from the DOM, hence, the ref `query-input` will be undefined, and it takes two or more ticks until it is mounted. BUT there is already a call to `this.$refs['query-input'].focus()` in the `mounted`-function of the GlobalSearch component which will make sure the query input is focused once the element is mounted to the DOM.

Vue can definitely be very opiniated at times :D 

I can commit this if you like, or you implement it, how you like :)

PR #2812, opened 2021-11-14
Remove 'Uninstall' from ARP Entry

### LINKED ISSUE #2811: [Bug] Application name is "Uninstall Zettlr" instead of "Zettlr"
### Description

In Control Panel, application is referred to as "Uninstall Zettlr" instead of "Zettlr".

Usually, most applications don't have "Uninstall" shown in the name.

Screenshot taken from 13th of November 2021:
image

Screenshot taken from 10th of November 2021:
image

This might be a culprit from https://github.com/Zettlr/Zettlr/blob/develop/electron-builder.yml#L92 and affects package managers where wildcard is used for ARP entries.

### Reproducing

1) Boot VM with VHDX/ISO, or use Windows Sandbox.
2) Install latest release from https://github.com/Zettlr/Zettlr/releases/latest
3) Observe that the application name is "Uninstall Zettlr" instead of "Zettlr" in Control Panel.

### Zettlr Version

Stable (most recent version)

### Specify version

2.0.3

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 11 (daily builds)

### Additional Information

_No response_

Trenly:
## Description

This PR updates the name from "Uninstall Zettlr" in the windows Add/Remove Programs menu to be just "Zettlr"
This resolves #2811

## Changes

I made changes to the parameters passed into NSIS by updating electron-build.yml. There should be no breaking changes

## Additional information

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   Zettlr are bound to the inner workings of the app, we should always overwrite
   them if applicable (to ensure they are updated with the app), unlike the
   defaults which people should be able to modify with us messing with them
+- Updated the display name in the Add/Remove Programs entry for Windows
 
 # 2.0.3
 
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -89,7 +89,7 @@ nsis:
   perMachine: false
   allowElevation: true
   allowToChangeInstallationDirectory: true
-  uninstallDisplayName: Uninstall ${productName}
+  uninstallDisplayName: ${productName}
   installerHeader: "./resources/NSIS/NSISinstallerHeader.bmp"
   installerSidebar: "./resources/NSIS/NSISinstallerSidebar.bmp"
   license: ./LICENSE


nathanlesage:
I'm still a little bit confused: The entry does specify that this is the *Uninstaller*'s display name, right? Why does this variable get propagated to the actual program's display name? Or could it be that this might be necessary for older Windows, such as 7? I remember it was a common thing back then to have "Uninstall <app>" in the control panel, but since I switched to Mac I have become uncustomed to the customs of Windows …

The documentation is somehow not really explicit on the ramifications of this setting.

Trenly:
I think it really depends on the actual NSI script that runs. 

There are likely two components at play here. First is the actual executable to uninstall Zettlr, which should be named “Uninstall Zettlr”. The second is the registry key which is set in the Uninstall registry  for Zettlr. This is most likely where the confusion comes in. Although it is the “Uninstall” registry key, it actually is the entry for the Add/Remove Programs table, which shows the currently installed programs.  There are several values within the registry key, and it is only the “DisplayName” that needs to be updated here. 

Without beinf able to see the full NSI script, I am not able to say with 100% certainty exactly how the variable here is used. However, I’d be happy to help look at the actual NSI files to determine if this is actually the variable which should be changed

nathanlesage:
> Without beinf able to see the full NSI script, I am not able to say with 100% certainty exactly how the variable here is used. However, I’d be happy to help look at the actual NSI files to determine if this is actually the variable which should be changed

I actually tried to find the NSI script, but couldn't find it in the source of electron-builder, so I'm not sure where it is – but it should be downloaded from somewhere, so feel free to check the source of electron-builder yourself to see if you can find it! Otherwise, the natural thing to try would be to actually build the app and try to install it in various settings (both update and initial installation). If then everything looks good, we can continue to merge this PR!

Trenly:
@nathanlesage After some digging, I found this [documentation](https://github.com/electron-userland/electron-builder/blob/master/docs/generated/NsisOptions.md) which states that it is specifically the display name in control panel. A ctrl+f on `uninstallDisplayName` should take you right to it

nathanlesage:
@Trenly Thanks a bunch! Looks good; plus I think not adding the `${version}` template makes sense (as it looks cleaner in the control panel).

So I'll be merging this, but there are some merge conflicts, would you mind to fix them?

Trenly:
> @Trenly Thanks a bunch! Looks good; plus I think not adding the `${version}` template makes sense (as it looks cleaner in the control panel).
> 
> So I'll be merging this, but there are some merge conflicts, would you mind to fix them?

For some reason it isn't letting me push new commits onto this PR. I have rebased it on the latest develop, and tried checking out the head of the PR directly, but somehow one of the refs must be broken.  I'll try opening it as a new PR and seeing if that fixes the refs issue

PR #2827, opened 2021-11-17
Fix/2818/disable autoupdate

### LINKED ISSUE #2818: [Feature] Disable Update Check
### Checklist

### Feature Proposal

Disable the Update-check on startup. This is not always necessary, and providing the ability to turn off the check reduces the data-footprint generated by zettlr.

### Example Workflow

Start the app the first time, and let it check.
Go to settings, either general or advanced
unmark "update on startup" checkbox
Restart zettlr, without updatecheck

### Caveats

Disabling this obviously does not provide the most up to date translation

Yes

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

### Additional Information

_No response_

newhinton:
## Description

fixes #2818

## Changes

No Api changes-

## Additional information

Tested on: KDE 20.04


source/app/service-providers/translation-provider.ts
```diff
@@ -88,6 +88,12 @@ export default class TranslationProvider {
    * @return {Promise} Resolves if everything worked out, rejects otherwise.
    */
   async init (): Promise<void> {
+    const autoUpdate: boolean = global.config.get('autoUpdate')
+    if (!autoUpdate) {
```
nathanlesage, line 92: I'm unsure if that should be here, since users will associate "auto update" with the application updates themselves, and not necessarily translation updates, so we need to find a good solution here. Plus, maybe we should refactor the `init` so that `updateTranslations()` can be called if the setting is true. Otherwise this check here would make other potential init-actions impossible which might be necessary in the future.

source/win-preferences/schema/general.js
```diff
@@ -50,6 +50,11 @@ export default function () {
           type: 'checkbox',
           label: trans('dialog.preferences.avoid_new_tabs'),
           model: 'system.avoidNewTabs'
+        },
+        {
+          type: 'checkbox',
+          label: trans('dialog.preferences.autoUpdate'),
```
nathanlesage, line 56: As in the other PR: Please convert the translation string to snake_case.

nathanlesage (changes requested):
In general looks good. We just have the problem of how to make clear that this setting is more about "Don't connect to the internet" rather than "Don't check for updates". Maybe we need two separate settings for that. This would make it more clear.

source/app/service-providers/translation-provider.ts
newhinton: We could rename it to updateCheck/checkForUpdates, which probably is more reasonable. Also we could just wrap all the init code in that if

newhinton:
> In general looks good. We just have the problem of how to make clear that this setting is more about "Don't connect to the internet" rather than "Don't check for updates". Maybe we need two separate settings for that. This would make it more clear.

Granted, also disallowing loading image-resources from the web may be a good idea, but wouldn't this be a second pr?
I don't really know where to start preventing that, im not that well versed in electron. 
Also, do we need to expose the second setting via ui? Maybe we should just document that and leave it out.

That gives me an idea: currently the command line switches are rather stretched, maybe we should build a central system for that. Benefits would be that we can quickly control settings from outside, and we do not lose track about what flags do what because we can quickly look them up in the source (which does not mean we dont need to document them)
In general, flags should only control settings via config anyway, just like i did here: https://github.com/Zettlr/Zettlr/blob/cfd2f64ed69cf1fc8857864424098b92ca856600/source/main/modules/window-manager/index.ts#L119-L121

newhinton:
I have tested something here: https://github.com/Zettlr/Zettlr/pull/2882

nathanlesage:
> Granted, also disallowing loading image-resources from the web may be a good idea, but wouldn't this be a second pr?

I'm beginning to think that we shouldn't actually completely prevent internet access, that seems to reminiscent of an earlier web. Rather, people who don't want an app to connect to the outside would need to block internet access to that app via the OS. That is not our idea. So I'm thinking whether or not we should just leave it at having this "check for auto updates" toggle, and leave the rest alone. That wouldn't break anything, and people who want to be paranoid, can still do so by blocking connectivity to the app as a whole.

> That gives me an idea: currently the command line switches are rather stretched, maybe we should build a central system for that.

Not a bad idea. Feel free to explore this if you'd like

nathanlesage:
Heya! To revive this PR, I think we should do the following:

* Allow users to disable the update check at startup (enabled by default)
* Still, the app should always check for translation updates so that incomplete translations can be completed even after a release has been made
* If a user really wants to disable Zettlr's access to the internet, they can do so using third-party tools

That would mean: We can leave everything (albeit I would recommend to rename the config setting to `checkForUpdates` to make it more clear) and just remove the changes in the translation provider. Then I can merge it!

And, if you would like to, you can also remove the else-branch if there is no new update, since there is request #2973 and thus we could also close that one.

newhinton:
Isn't this a great opportunity for a startup-flag with the new cli-parser? this way we keep the ui clean, but can still disable it if wanted. (I dont think it is reasonable to expect the os to block access, at least on windows and linux this is a non trivial task)

That would be a seperate PR/Issue then, but i would create that one aswell when the cli one is merged.

nathanlesage:
@newhinton I wanted to suggest something like that. Feel free to implement that once we have the CLI thingy!

I'm going to link #2973 so we can close both at the same time, would you remove that simple else-thingy? :)

newhinton:
Sure!

newhinton:
Everything done!

nathanlesage:
Looks good except a merge conflict that apparently is too complex for GitHub :(

By the way, when did that setting merge into the general pane? Wasn't it in advanced before…?

nathanlesage:
@newhinton Very nice! :)

Yeah it would be great if we had the setting in the advanced tab, since iirc many other apps also have it there, right? (Plus, we don't want to encourage people to do that :D)

nathanlesage (approved):

nathanlesage:
Looks wonderful! Merging …

PR #2841, opened 2021-11-20
start minimized when --tray is in argv

### LINKED ISSUE #2840: [Feature] Start in tray
### Checklist

### Feature Proposal

Autostarting zettlr can provide quick access to your notes. Starting the app in tray (minimized) can be benefical because the app can be autostarted, and is ready to be used. 

### Example Workflow

Start the application with: `Zettlr --minimized/--tray` or `Zettlr.exe --minimized/--tray`
Tray icon shows up, but the main window not.

If the app is already started and the shortcut is clicked again, the main window should be shown, regardless of the option. 

This may be furher improved by assigning a global keyboard shortcut to open the app quickly.

### Caveats

May be unintuitive to use, especially when the app is already open/minimized

Yes

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

### Additional Information

I can only test on linux.

newhinton:
## Description
Fix #2840 

## Changes
Add a check that checks for --tray in argv and then minimizes the main window and creates a tray icon, regardless of "use tray" setting

## Additional information

Tested on: KDE 20.04

### DIFF
--- a/source/main/modules/window-manager/index.ts
+++ b/source/main/modules/window-manager/index.ts
@@ -116,10 +116,8 @@ export default class WindowManager extends EventEmitter {
       .then(() => {
         global.log.info('[Window Manager] Window Manager booted. Opening main window.')
         this.showMainWindow()
-        if (process.argv.includes('--tray')) {
-          global.config.set('system.leaveAppRunningOverride', true)
-        }
-        if (global.config.get('system.leaveAppRunningOverride')) {
+        if (process.argv.includes('--tray') || Boolean(global.config.get('system.startInTray'))) {
+          global.log.info('[Window Manager] Application should start in Tray. Main Window is minimized to tray on startup.')
           global.tray.add()
           this.getMainWindow()?.hide()
         }
--- a/source/main/modules/window-manager/index.ts
+++ b/source/main/modules/window-manager/index.ts
@@ -117,6 +117,7 @@ export default class WindowManager extends EventEmitter {
         global.log.info('[Window Manager] Window Manager booted. Opening main window.')
         this.showMainWindow()
         if (process.argv.includes('--tray')) {
+          global.config.set('system.leaveAppRunningOverride', true)
           global.tray.add()
           this.getMainWindow()?.hide()
         }
--- a/source/main/modules/window-manager/index.ts
+++ b/source/main/modules/window-manager/index.ts
@@ -116,7 +116,7 @@ export default class WindowManager extends EventEmitter {
       .then(() => {
         global.log.info('[Window Manager] Window Manager booted. Opening main window.')
         this.showMainWindow()
-        if (process.argv.includes('--tray') || Boolean(global.config.get('system.startInTray'))) {
+        if (global.config.get('system.startInTray')) {
           global.log.info('[Window Manager] Application should start in Tray. Main Window is minimized to tray on startup.')
           global.tray.add()
           this.getMainWindow()?.hide()
--- a/source/main/modules/window-manager/index.ts
+++ b/source/main/modules/window-manager/index.ts
@@ -118,6 +118,8 @@ export default class WindowManager extends EventEmitter {
         this.showMainWindow()
         if (process.argv.includes('--tray')) {
           global.config.set('system.leaveAppRunningOverride', true)
+        }
+        if (global.config.get('system.leaveAppRunningOverride')) {
           global.tray.add()
           this.getMainWindow()?.hide()
         }
--- a/source/main/modules/window-manager/index.ts
+++ b/source/main/modules/window-manager/index.ts
@@ -116,6 +116,10 @@ export default class WindowManager extends EventEmitter {
       .then(() => {
         global.log.info('[Window Manager] Window Manager booted. Opening main window.')
         this.showMainWindow()
+        if (process.argv.includes('--tray')) {
+          global.tray.add()
+          this.getMainWindow()?.hide()
+        }
       })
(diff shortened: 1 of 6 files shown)


nathanlesage:
Why do it using a CLI switch? I know a few programs allow to start minimized & at system boot, is that what you ultimately want?

nathanlesage:
> A setting may be confusing, because if it is not autostarted, the application would "not start" (because it would be in the tray)

Fair point! Didn't think about that. Let me take a look …

source/main/modules/window-manager/index.ts
```diff
@@ -116,6 +116,10 @@ export default class WindowManager extends EventEmitter {
       .then(() => {
         global.log.info('[Window Manager] Window Manager booted. Opening main window.')
         this.showMainWindow()
+        if (process.argv.includes('--tray')) {
+          global.tray.add()
```
nathanlesage, line 120: This looks as if it might break, since the Tray provider relies on the global configuration, and if the tray has been created and THEN the user activates the corresponding setting, it might break something. I would rather argue that we use the global config setting instead of circumventing that. (I.e. set it if it's not set, else only hide the main window)

Additionally, if we add a new CLI option, it must be documented in the README file.

nathanlesage (changes requested):
We have to think about the state management, and what happens in each case. Would you mind creating a small table with these? Just to be sure we don't accidentally break the state!

source/main/modules/window-manager/index.ts
cskeogh: Also if the config `system.leaveAppRunning` is false, --tray won't do anything. The user would expect --tray to do something.

newhinton: That is true!
Should we introdude a hidden state 'system.leaveAppRunningOverride' as to not modify the original setting? 

That way, we have an indepdendend way to enable this behaviour, and this would also allow to add a UI-toggle for starting in tray.

nathanlesage: A second, independent setting would be great. We can also conditionally enable/disable the `Show in tray` depending on the `start minimized` (or vice versa), so that the user sees that if you start minimized this requires the tray to be set.

newhinton:
|  |  | system.leaveAppRunning |  |
| -------------- | :--------- | :--------- | :--------- |
|  |  | true | false |
| system.leaveAppRunningOverride | true | Show Tray but no Window | Show Tray but no Window |
|  | false | Show Tray and Window | Show Tray and Window |

nathanlesage:
I just discovered something: If we put that into the global config, it's going to be persisted, meaning if you start with `--tray` once it will behave as if you've always set that. Hadn't we have this one instance where the main app object just checked if the setting was there and then just prevented opening the main window? I personally think it would be perfectly fine to have that logic also set the `system.leaveAppRunning` to `true` programmatically since a person who is able to start by passing that flag will also be able to find that setting and disable it if they don't want it (however I think a person who wants to start in tray would also very likely have that setting checked either way)

newhinton:
> I just discovered something: If we put that into the global config, it's going to be persisted, meaning if you start with `--tray` once it will behave as if you've always set that. Hadn't we have this one instance where the main app object just checked if the setting was there and then just prevented opening the main window? I personally think it would be perfectly fine to have that logic also set the `system.leaveAppRunning` to `true` programmatically since a person who is able to start by passing that flag will also be able to find that setting and disable it if they don't want it (however I think a person who wants to start in tray would also very likely have that setting checked either way)

hmm yeah, you are right that is probably a sideeffect. however, the cli switch should not actually persist settings unless explicitly designed to do so, so --tray should only do it once. That however, does not work with an override flag. 

There are two options: do not use system.Xoverride as to not persist it, and only do it on the fly (as i did in my first commits) or create a shadow system.

It could use a `override.system.leaveAppRunning` that is basically a map that does not get persisted. if it exists, then it is used instead of the original setting.

nathanlesage:
I still don't see why we couldn't simply then set the `tray` option in the preferences in case it is set to `false` if said CLI switch is present. This would alleviate a lot of potential side effects, keep the code clean and make this whole effort much simpler to implement.

newhinton:
Oh sure we can! 

I was just arguing that --tray should not persist anything. A ui-toggle is totally fine and probably way easier in the short term

nathanlesage:
Mh, then let's do that. The simpler the changes the better!

newhinton:
I just did that! I added it to the ui toggle and removed the cli-switch

nathanlesage:
Looks mergable now! :) Would you mind fixing the merge conflicts? (Should merge automatically after performing a rebase, since the file `advanced.js` has moved to `advanced.ts`, but no other changes necessary)

newhinton:
uhm yeah, i broke the branch. I will reopen another branch, i made a copy :facepalm:

nathanlesage:
Umm, there's still a merge conflict :S

nathanlesage (approved):

newhinton:
... i think this worked? :laughing:

nathanlesage:
Now it works!

newhinton:
Regarding the wording:

Starte Zettlr minimiert in der Benachrichtigungsleiste
Start Zettlr minimized in the notification tray


is there a repository where i could directly provide those strings?

nathanlesage:
Ah, I have now already added translations (very simple: "Start app minimized")

There's not a repository but a whole system, which I wrote, because I doubted that so many people would want to use text files directly, and I wanted to encourage as many people as possible to help: https://translate.zettlr.com

Issue #2577, opened 2021-10-06
[Bug] File List  is not updated when sorting by modification time


tscheims1:
### Description

Sort order in file list is not updated when the folder is sorted by modification time and files are modified.
(when the sort order is switched in the folder settings, (eg from ascending to desceinding) the sort order is updated.

the timestamp in the file list (eg. 1 min ago ) is updated properly.

### Reproducing

1.) open a folder that contains some files
2.) set option: "when sorting by time sort by modification time"
3.) set sort of folder to sort by time ascending
4.) modify some files. 
5.) the timestamp of the file (e.g. 1 min ago) of the file in the file browser list is updated properly. the order of the files remains.
![file-sorting-by-modification-time-not-refresh-properly](https://user-images.githubusercontent.com/5829119/136290664-0830d877-4e02-4c7f-b113-0a8f11828b1f.png)

### Zettlr Version

Beta (if applicable)

### Specify version

5

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Manjaro Linux, i3, packages up to date.

nathanlesage:
Closed in 6da4b24

newhinton:
Is this truly fixed? I build from develop (and made sure the change in 6da4b24 was present) and the sorting still does not reorder the list automatically after a change occured. (Tested in all three view-modes)

Is it supposed to update on change?

PR #2851, opened 2021-11-23
update filelist if a file has changed

newhinton:
## Description

When the Filesystem registers a change, only directory-changes trigger a sorting-action.
This PR adds that a filechange also triggers a sorting-action.

## Changes

Added else if-path for patchInFiletree for files

## Additional information

Tested on: Kubuntu 20.04


newhinton:
See #2577

source/win-main/store.ts
```diff
@@ -401,8 +401,12 @@ const config: StoreOptions<ZettlrState> = {
       // that the name did not change (because that would've resulted in a
       // removal and one addition) but rather something else, so we need to make
       // sure to simply re-sort it in case the sorting has changed.
+      // If we have a file, update the parent instead.
       if (ownDescriptor.type === 'directory') {
         Vue.set(ownDescriptor, 'children', sort(ownDescriptor.children, ownDescriptor.sorting))
+      } else if (ownDescriptor.type === 'file') {
+        const parentDescriptor = ownDescriptor.parent
```
nathanlesage, line 408: `parent` is null if it's a standalone file, so this would throw an error.

nathanlesage (changes requested):
Looks good to me except one comment!

source/win-main/store.ts
newhinton: i wrapped it in a nullcheck!

source/win-main/store.ts
```diff
@@ -401,8 +401,14 @@ const config: StoreOptions<ZettlrState> = {
       // that the name did not change (because that would've resulted in a
       // removal and one addition) but rather something else, so we need to make
       // sure to simply re-sort it in case the sorting has changed.
+      // If we have a file, update the parent instead.
       if (ownDescriptor.type === 'directory') {
         Vue.set(ownDescriptor, 'children', sort(ownDescriptor.children, ownDescriptor.sorting))
+      } else if (ownDescriptor.type === 'file') {
+        if (ownDescriptor.parent != null) {
```
nathanlesage, line 408: I would add that to the else if-condition (more style, but it makes the code a tad cleaner)

nathanlesage:
One final comment, then it's ready to merge!

source/win-main/store.ts
newhinton: i kept the constant, i think it makes it more readable than reusing ownDescriptor.children.parent everywhere

nathanlesage: I meant to add the `ownDescriptorparent != null` to the `else if (ownDescriptor.type === 'file')`-condition above.

nathanlesage:
The Vue 3 migration has introduced merge conflicts, it would be super if you could fix them! Should work easily.

newhinton:
I found the bug resulting in #2864. The fix is now in this PR aswell, should i make a second PR for this tiny change?

nathanlesage:
> I found the bug resulting in #2864. The fix is now in this PR aswell, should i make a second PR for this tiny change?

Nah, we can add this in here as well. Just saw it, seems easy, understandable, and good. As long as it's in its own commit, all good!

nathanlesage (approved):

nathanlesage:
And here we go!

PR #2862, opened 2021-11-26
Feature/2344/folder first

### LINKED ISSUE #2344: [Feature] Customize general Files-first/Folders-first sorting
### Discussed in https://github.com/Zettlr/Zettlr/discussions/2343

<div type='discussions-op-text'>

<sup>Originally posted by **krystian-r** August 22, 2021</sup>
Hi,

is there a way to change the sorting order for folders/files. I'd like to see directories first and then files (I use the "Combined" file manager mode)

thanks</div>

Basically this requires a setting to alter the concatenation of the file and directory arrays in the FSAL directory descriptors. Should be simple to implement.

newhinton:
## Description
fixes #2344

## Changes
added a checkbox and the corresponding config to sort folders above files.

Tested on: Kubuntu 20.04

### DIFF
--- a/source/main/modules/fsal/util/sort.ts
+++ b/source/main/modules/fsal/util/sort.ts
@@ -74,6 +74,7 @@ export default function (
   // Return sorted array files -> directories
   if (global.config.get('sortFoldersFirst')) {
     return ret.concat(d).concat(f)
+  } else {
+    return ret.concat(f).concat(d)
   }
-  return ret.concat(f).concat(d)
 }
--- a/source/main/modules/fsal/util/sort.ts
+++ b/source/main/modules/fsal/util/sort.ts
@@ -72,5 +72,8 @@ export default function (
   const ret: Array<MDFileDescriptor | DirDescriptor | CodeFileDescriptor> = []
 
   // Return sorted array files -> directories
+  if (global.config.get('sortFoldersFirst')) {
+    return ret.concat(d).concat(f)
+  }
   return ret.concat(f).concat(d)
 }
--- a/source/win-preferences/schema/general.js
+++ b/source/win-preferences/schema/general.js
@@ -98,7 +98,11 @@ export default function () {
             'ascii': trans('dialog.preferences.sorting_ascii')
           }
         }
-      ],
+      ], {
+        type: 'checkbox',
+        label: trans('dialog.preferences.sortFoldersFirst'),
+        model: 'sortFoldersFirst'
+      },
       [
         {
           type: 'radio',
--- a/source/win-preferences/schema/general.js
+++ b/source/win-preferences/schema/general.js
@@ -98,11 +98,14 @@ export default function () {
             'ascii': trans('dialog.preferences.sorting_ascii')
           }
         }
-      ], {
-        type: 'checkbox',
-        label: trans('dialog.preferences.sortFoldersFirst'),
-        model: 'sortFoldersFirst'
-      },
+      ],
+      [
+        {
+          type: 'checkbox',
+          label: trans('dialog.preferences.sortFoldersFirst'),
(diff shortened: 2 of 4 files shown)


newhinton:
It does not resort the list on a change in the settings. A manual refresh, a restart or a change in the sorting-order is required to apply the change at the moment

newhinton:
Also, maybe it does make sense to add a css-class to each last element of a type. So that we can add a small margin-bottom to create a little space between folders/files to make differentiation easier

nathanlesage:
> Also, maybe it does make sense to add a css-class to each last element of a type. 

Sounds like a good idea, but maybe then also on the first element. I'd suggest we address this in a separate PR after the Vue 3 migration which is almost finished. If you'd like to, please feel free to review the corresponding PR that'll be incoming probably on the weekend, to speed that process up. This is also what kind of prevents me from testing and merging your PRs right now … I'll see to getting to that in the coming days.

source/main/modules/fsal/util/sort.ts
```diff
@@ -72,5 +72,8 @@ export default function (
   const ret: Array<MDFileDescriptor | DirDescriptor | CodeFileDescriptor> = []
 
   // Return sorted array files -> directories
+  if (global.config.get('sortFoldersFirst')) {
+    return ret.concat(d).concat(f)
+  }
```
nathanlesage, line 77: I would make this an else here so it's clear that one of these results will happen; I think it makes it cleaner!

source/win-preferences/schema/general.js
```diff
@@ -99,6 +99,13 @@ export default function () {
           }
         }
       ],
+      [
+        {
+          type: 'checkbox',
+          label: trans('dialog.preferences.sortFoldersFirst'),
```
nathanlesage, line 105: To make the translation string consistent with the other translations, please convert this to snake_case.

nathanlesage (changes requested):
Looks pretty good. I got only two small comments, but then it's almost ready for merging.

One thing we might need to think about is that just changing this doesn't immediately apply. We need to find a way to re-sort all directories afterwards. So what we might need to do is add a listener for this config change in the FSAL that triggers a recursive re-sort so that the changes are applied immediately. Do you feel up to that task as well?

newhinton:
> So what we might need to do is add a listener for this config change in the FSAL that triggers a recursive re-sort so that the changes are applied immediately. Do you feel up to that task as well?

Wouldn't resorting also solve #2864 ? Albeit we probably have too keep track of expanded folders... i'll think about it

nathanlesage:
> Albeit we probably have too keep track of expanded folders... i'll think about it

That is another thing that would be really cool to have. Currently, Zettlr does not at all keep track of opened folders. But, rather a feat than a fix!

Shall we just merge this thing and then figure out a way to hot reload the directory tree in a separate PR?

If so, it would be great if you had a suggested wording for the English+German translations of the new translation strings we have in this PR, so I can add them to Translate!

nathanlesage (approved):

newhinton:
> If so, it would be great if you had a suggested wording for the English+German translations of the new translation strings we have in this PR, so I can add them to Translate!

dialog.preferences.sortFoldersFirst.en: `Show folders above notes`
dialog.preferences.sortFoldersFirst.de: `Zeige Ordner vor Notizen`

albeit i am not sure about "notes" is the right term, but "files" may be too generic or simply not friendly enough

nathanlesage:
> but "files" may be too generic or simply not friendly enough

I know what you mean, but since the wording is generally "files" and not "notes", this might be better. I will add that to the Translation service; thank you!

And also, welcome to the party with the first PR! ;D Now we just need this idea of reloading the file tree …

Issue #2874, opened 2021-12-01
[Feature] Link navigation without keyboard


zh217:
### Description

Currently the only way to navigate links (that I know of) is to hold down ctrl/command key and click. This is inconvenient when browsing through files.

### Proposed Changes

Have a method of navigating links without pressing any keys, suggestions:

* Contextual menu on right click
* Middle-click
* We have a preview popup on mouse-over the link, the popup can act as a further target so that click or double click on it opens the linked article

### Caveats

Contextual menu is certainly very safe. I am not sure if middle-clicks have any other functionality.

### Do you Wish to Attempt Implementing this Yourself?

Maybe

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Operating System Version

12.0.1 (21A559)

nathanlesage:
Context menu will work, middle clicking not since we need that to enable force-opening in a new tab (to align with browser-style behavior).

> We have a preview popup on mouse-over the link, the popup can act as a further target so that click or double click on it opens the linked article

Even more so, we might be able to add a button to follow a link.

zh217:
> Context menu will work, middle clicking not since we need that to enable force-opening in a new tab (to align with browser-style behavior).

Adding to context menu is fair enough, but I don't think the planned behavior aligns with that of browsers, since for browsers (and most applications) the behavior is that middle click and ctrl-click are the SAME (whatever that behavior is). My understanding is that ctrl-click always acts as a replacement for middle click for mouses without a middle button.

nathanlesage:
> the behavior is that middle click and ctrl-click are the SAME

That is true, but on webpages you normally don't edit text, so you don't have to differentiate between normal clicks (intent: edit) and Control-clicks (intent: follow the link). So in this case we must differentiate between Control-left-Click and Control-Middle-Click.

zh217:
> So in this case we must differentiate between Control-left-Click and Control-Middle-Click.

I agree, but this does not imply that middle-click without control should be a no-op.

nathanlesage:
> I agree, but this does not imply that middle-click without control should be a no-op.

Right, middle-clicks are currently no-ops. Didn't think about that. Ctrl+Middle should still be reserved for forcing a new tab, so when someone forgets to press Ctrl it'll open in the same tab (if the corresponding setting is set). However, in tandem with issue #1590 this should pose not that much of a problem.

leo-hstone:
[...]

PR #2875, opened 2021-12-01
Follow ZTN links on middle click

zh217:
Follow ZTN links on middle click, for https://github.com/Zettlr/Zettlr/issues/2874

## Description

Previously, ZTN links are followed when ctrl/cmd clicking on them. With this patch, the links are also followed with middle clicks, thus enabling keyboard-free navigation.

## Changes

Simply taking middle clicks into consideration when deciding whether to follow links

## Additional information

This is consistent with expected behaviour in browsers where ctrl/cmd clicks on links are equivalent to middle clicks (open link in new tab in the background in Chrome)

Tested on: MacOS 12, Windows 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed an error in the link filtering process that would throw an error when
   you attempted to remove internal links completely upon export
+- Follow zettelkasten links on middle clicks
 
 ## Under the Hood
 
--- a/source/common/modules/markdown-editor/index.js
+++ b/source/common/modules/markdown-editor/index.js
@@ -219,7 +219,7 @@ module.exports = class MarkdownEditor extends EventEmitter {
     })
 
     this._instance.on('mousedown', (cm, event) => {
-      // Open links on both Cmd and Ctrl clicks - otherwise stop handling event
+      // Open links on Cmd clicks, Ctrl clicks and middle clicks - otherwise stop handling event
       if (process.platform === 'darwin' && !event.metaKey && event.button !== 1) return true
       if (process.platform !== 'darwin' && !event.ctrlKey && event.button !== 1) return true
 
--- a/source/common/modules/markdown-editor/index.js
+++ b/source/common/modules/markdown-editor/index.js
@@ -220,8 +220,8 @@ module.exports = class MarkdownEditor extends EventEmitter {
 
     this._instance.on('mousedown', (cm, event) => {
       // Open links on both Cmd and Ctrl clicks - otherwise stop handling event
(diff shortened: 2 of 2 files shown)


nathanlesage:
I am not sure we should do that. The middle-click is reserved for force-opening something in a new tab (not yet implemented). Once implemented it will work like this:

1. Cmd/Ctrl-Clicking a link will open it, in the current tab if the "Avoid new tabs" setting is present
2. Cmd/Ctrl-Clicking a link using the middle mouse will always open it in a new tab

zh217:
How about:

* ctrl-click or middle-click opens link according to setting
* ctrl-middle-click opens link in new tab always (still, I think this combination is a little strange, but I guess people can get used to it)

nathanlesage:
> still, I think this combination is a little strange, but I guess people can get used to it

What do you think is strange with this? What would you do if you could remap everything at will?

(We possibly can't since people are already used to a certain way of behavior, and changing that might be very difficult)

zh217:
It is strange for me because only when my mouse does not have a middle button do I try to ctrl-click to see what happens, as for the applications I frequently use these two are usually the same.

BTW I tried to find some applications where the two are not the same

* VSCode: ctrl-click follows link (opens in browser), middle-click no-op, and on mouse-over link there is an immediate popup with a '"follow link" (cmd+click)' prompt, the "follow link" part is itself a link and can be clicked, is right above the cursor so that minimal further mouse movement is required. I consider this to be a very good solution since it is very fast after getting used to it and has the further benefit of discoverability. Also It is important to ensure functionality in the absence of middle button (e.g. most people do not know how to middle click on a trackpad).
* Microsoft Word: any click you do except right click opens the link in the browser. The only way to edit a link is to move the cursor using the arrow keys, or backspace/delete. This makes no sense at all, since (unlike in Zettlr) I don't see why anyone would want to follow links in Word.

nathanlesage:
Mh, with regard to your discovery in VSCode: Zettlr also has those tooltips, but they open after some delay. Would having them open immediately remedify that issue in your view?

zh217:
> Zettlr also has those tooltips, but they open after some delay. Would having them open immediately remedify that issue in your view?

I don't think those tooltips are clickable at the moment. If they were, then yes.

nathanlesage:
> I don't think those tooltips are clickable at the moment. If they were, then yes.

Shall we do that instead, then? This would make the change more "backward-compatible". What do you think?

zh217:
Yes, and it would support more devices too (Apple's magic mouse cannot trigger middle clicks either)

nathanlesage:
Okay, then I'm going to close this PR and add said button to the tooltips containing the note preview.

PR #2927, opened 2021-12-15
Feature/noid/yargs

newhinton:
## Description

This moves (nearly) all access to process.argv to a single class, which can be queried if a setting is set. This makes it easier to handle cli switches, since they are locally controlled and not strewn across the whole application.

It uses [yargs](http://yargs.js.org/) as a base.

## Changes
Moved (nearly) all access to process.argv to a single class.

## Additional information

Tested on: Kubuntu 20.04


newhinton:
Question: Does the cli-help need translations?

nathanlesage:
I like the idea, BUT I do not see the necessity to use a dedicated library for that. Zettlr is not a CLI program, so we will never implement any complex logic (for that what we want to do is implement custom protocols, not CLI switches). Thus, I cannot see the benefit of having a rather large library for the few CLI arguments we will ever support. Especially given the fact that Node does an already good job at parsing CLI arguments for us. So I suggest we remove the additional dependency and rather move the two lines of extracting arguments that have their own arguments (such as data dir) to the service provider and be done with it. We also don't need any help strings on the CLI, we will just document those switches here (and on the docs) and be done with it.

newhinton:
I think we could benefit from such a library (or even a more custom approach)
Currently Zettlr does not have many cli-arguments, but using an easy system like this could easily allow us to expand this. As an example: We could move the Developer Mode away from a toggle in the ui to a switch. (To me, it seems more reasonable this way)

Additionally, i thought about java/makefile - style argument parsing, like -dDEVELOPERMODE=True which allows arbitrary settings via the cli. Although i have to say that would be way into the area of really nice to have and not really important.

And i think it is a nice touch to have a cli tool, because this way we show that even if it is not the main goal, we care about the cli. Also it is way faster to use -h to get options, instead of using the git-page. 

I also found out that [yargs-parser](https://www.npmjs.com/package/yargs-parser) exists, which is half the size (130kb). Sure it is not small, but better than yargs and we can write our own helper output.

nathanlesage:
> We could move the Developer Mode away from a toggle in the ui to a switch. (To me, it seems more reasonable this way)

Not really. A GUI-program should do most stuff within the GUI. Plus, on macOS, CLI switches for GUI apps are an absolute hassle, so most CLI flags should be last-resorts.

In general, I don't think that just by using some random library indicates something, but rather that the actual flags (i.e. the ones we currently use + potential future ones) do so. So until we actually run into the NEED to have a library because the code becomes unmanageable, I don't really think we should add more packages.

My aversion to new packages is mainly driven by the maintenance stuff (there are VERY frequently new CVEs and the constant need for updates; which require me to manually bump the versions. I would rather keep the amount low) and the package size. So unless ABSOLUTELY necessary, I don't want to add new libraries, and I'm positive that at least for now we can do very well without an additional library. As soon as the CLI arguments become more, I will certainly be open about this, so please remember your research when that time arrives! :)

newhinton:
Should i still build a cli-parser with the current interface?
One that currently can ONLY parse the few commands we have (but including a help&version) and does not rely on any library?

newhinton:
Also on a sidenote: i found out why i wanted to also add a --tray flag instead of a ui switch:

When the user has that setting set, the normal application starter also opens in the tray if zettlr was not already started. This is likely not wanted behaviour. Should i open a new issue to discuss that?

nathanlesage:
Yes please! Your initiative to centralize that function was very good, so we should definitely go with that! I was only arguing that a full library might (as of now) be overkill!

nathanlesage:
> Also on a sidenote: i found out why i wanted to also add a --tray flag instead of a ui switch:
> 
> 
> 
> When the user has that setting set, the normal application starter also opens in the tray if zettlr was not already started. This is likely not wanted behaviour. Should i open a new issue to discuss that?

Please, because solely based on the description I cannot completely comprehend what the issue is.

nathanlesage:
Heya, so first sorry for the late response! I like what you did with the CLI provider, especially the version string.

Although since you NEED to import something from the CLI provider class in every place where you want something, here's the thing: I'm planning on refactoring all service providers to functional modules which contain the state and then expose a few functions that allow for the state to be read or set. And I think the CLI provider is the ideal place to start, since it neither has a boot function nor does it need to be shut down.

This way we could simply import the `enum` of the arguments and the `getArg`-function wherever we need it and don't really need to rely on the global.

But please say if you don't want to do another refactor since that PR is already a tad older -- then I can do that later on.

newhinton:
Could you give me an example on how that may look? While i do not have a "boot function", but the parsing of the version string and help rely on the fact that it only gets initialized once. (and quite early in the startup process, which allows us to exit the app very fast from the cli)

nathanlesage:
> help rely on the fact that it only gets initialized once. (and quite early in the startup process, which allows us to exit the app very fast from the cli)

Yes, that will also happen on a module-basis as well. It's kind of difficult to realize, but once one "accepts" that fact, it makes modules look much cleaner. But I understand the hesitation just too good.

With regard to the exiting: That makes me think that we should do any exiting maybe ONLY during the boot function inside `lifecycle.ts`? That would make it easier for us to keep all _wanted_ exits in one place so we are less prone to introduce bugs …?

***

With regard to how modules work, here a toy example:

```typescript
// This is a state that only gets initialized the very first time any other module imports this one
const myState = {
  someProperty: 0
}

// The same as above, but with `let` so that it is mutable
let someVar = 0

export default function modifyState (newValue: number): void {
  // Calling this function updates both the nested prop and the `let`-variable
  myState.someProperty = newValue
  someVar = newValue
}

export function readStateObject (): number {
  // Returns 0, unless `modifyState` has been called from any place with a different number
  return myState.someProperty
}
export function readStateNumber (): number {
  // Same as above
  return someVar
}
```

So basically, this pattern has this state as expressed by some variables you declare, and then the exports basically define an "API".

Instead of always referring to `global.myProvider.someMethod()` you first do `import someMethod from '@providers/myProvider'` and then just call `someMethod()`.

This is much better than the `global`-object (which I used because I didn't know it better back then), because it makes everything much more explicit: We can see from some `import someFunction from 'myProvider'` that some module uses a provider without having to search for `global.myProvider` AND we can ditch all that boilerplate code from the file `global.d.ts` perspectively. (But you don't have to do that, obviously!)

***

If you feel up to the task, I added an alias in 6e28e9eb:

Now, the line `import CliProvider from '../../../app/service-providers/cli-provider'` in your PR would become `import CliProvider from '@providers/cli-provider'`

I think that's convenient

newhinton:
@nathanlesage I did a thing. I am sure it does not belong in the provider folder anymore, but am not sure where to put it otherwise.


Also regarding your lifecycle question:

Generally i think it could help IF we have many exit condtions, but from my head i cound only four:
1. Single Instance
2. "Wanted"-Exits
3. --help-flag
4. --version-flag

Does that warrant it's own tiny module that handles exits? We can introduce one, but maybe we are overengineering that. But to be honest, i really dont have an opinion on either one

nathanlesage:
Hey, looks pretty decent! I'm also sure it is not a service provider somehow. I'm currently thinking that the module is rather providing some helper functions for the app, and they are generally required only in a few places. I would argue these functions should be mainly handled in the main.ts (if the app should quickly exit without spinning up all the internal magic). Plus, some of the things you address should be in other parts (for example, the window manager will become a service provider based on PR #3133 and should handle cache clearing itself).

But in general this looks good. However, I'm currently thinking to postpone this PR to after the "great migration" is done. Feel free to have a look at PR #3133 so you know what's happening. Basically then we can put the CLI helper module into the `app` directory and just request the modules where needed. So it's pretty finished, the only reason why I want to postpone this PR is that I'm afraid of the merge conflicts that happen when we throw both together — or what do you think? Maybe I'm overreacting because the migratory effort is so large!

nathanlesage:
So, it's merged! If any questions come up wrt the merge conflicts, just poke me!

newhinton:
@nathanlesage Hey! Sorry for the hiatus, but now i have updated this branch.

Should we change anything else?

newhinton:
@nathanlesage Um is it a good idea to use the stalebot on pullrequests?

nathanlesage:
> @nathanlesage Um is it a good idea to use the stalebot on pullrequests?

Yes it is, and even only for me to remember me that I need to finally take care of it!

source/app/service-providers/cli-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+ *                  It can be used to query arguments, as
+ *                  long as they are defined in the options object.
+ *
+ * END HEADER
+ */
+
+import { app } from 'electron'
+import path from 'path'
+
+export const DATA_DIR: string = 'data-dir'
```
nathanlesage, line 20: Since these are constant primitives, I think we can remove the type annotations, since these will be picked up easily by Typescript

source/app/service-providers/cli-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+export const DISABLE_HARDWARE_ACCELERATION: string = 'disable-hardware-acceleration'
+export const CLEAR_CACHE: string = 'clear-cache'
+export const LAUNCH_MINIMIZED: string = 'launch-minimized'
+
+export function getCLIArgument (key: string): any {
```
nathanlesage, line 25: Should we make the return type more specific in that it will return either a `string` or a `boolean` (for existing CLI arguments) or `undefined` if it doesn't exist?

source/app/service-providers/cli-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+      return process.argv.includes('--disable-hardware-acceleration')
+    }
+    case LAUNCH_MINIMIZED: {
+      return process.argv.includes('--launch-minimized') || process.argv.includes('-m')
+    }
+  }
+  return undefined
+}
+
+export function handleGeneralArguments (): void {
```
nathanlesage, line 54: This is a function that should only once be called by the `main.ts` entry point, correct? If so we should definitely state that, in comments. I just realized that we need Docstrings for the functions here

source/app/service-providers/cli-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  if (process.argv.includes('--version') || process.argv.includes('-v')) {
+    console.log(app.getName() + ' ' + app.getVersion())
+    process.exit()
+  }
+}
+
+function showHelp (): void {
+  console.log('This is the Zettlr Help!')
```
nathanlesage, line 67: Cute :D

source/app/service-providers/cli-provider.ts
```diff
@@ -0,0 +1,88 @@
...
+  console.log('Usage:')
+  console.log('-h | --help                            Show this help')
+  console.log('-v | --version                         Show the Version of Zettlr')
+  console.log('   | --clear-cache                     Removes all cached files')
+  console.log('   | --disable-hardware-acceleration   Disables Hardware Accelleration for systems that do not support it.')
+  console.log('   | --data-dir=FILEPATH               Set a custom directory for Zettlr\'s configuration files')
+  console.log('-m | --launch-minimized                Start Zettlr mimimized in the Systemtray')
+}
+
+export function getArgumentValue (key: string): string | undefined {
```
nathanlesage, line 78: How is this function different from `getCLIArgument`? Why do we need two functions?

nathanlesage (changes requested):
So, first of all sorry for taking so long to respond to the PR again. I think it looks pretty good. I have a few comments on that, but in general it works as advertised and therefore it's pretty good!

There are just two "meta" issues/questions that I have:

1. The linter is complaining, so it would be great if we could see whether the code is fine or if the linter is right to bark at us
2. Since that provider doesn't maintain its own state, wouldn't it potentially make more sense to add that as a module into the `common/modules` directory and then just import the functions whenever we need them? Would make much more semantical sense, wouldn't it?

source/app/service-providers/cli-provider.ts
newhinton: This basically makes the --key=value syntax available to everyone. It is used in the getCLIArgument-Function to get the '--data-dir' value. Currently this is the only usage, but i guess this will be useful if we dont integrate it in the other function.

If i remember correctly, this was a "leftover" from the original way zettlr got that value, and i just moved it there. Which reminds me, i have to fix the naming inside the function.

source/app/service-providers/cli-provider.ts
newhinton: I hope that is what you meant, i have added a comment explaining what it ~~does~~ should do

nathanlesage: Yes, now I see what is going on here. Please feel free to add DocBlocks to the other functions as well!

newhinton: I did that!

source/main.ts
```diff
@@ -46,11 +53,11 @@ if (process.platform === 'win32') {
 
 // Setting custom data dir for user configuration files.
 // Full path or relative path is OK. '~' does not work as expected.
-const dataDirFlag = process.argv.find(elem => elem.indexOf('--data-dir=') === 0)
+let dataDir = getCLIArgument(DATA_DIR)
 
-if (dataDirFlag !== undefined) {
+if (dataDir !== undefined && typeof dataDir === 'string') {
```
nathanlesage, line 58: Here the `typeof dataDir` check is sufficient!

nathanlesage (approved):
Looks good so far! One small beauty thing that doesn't impede functionality. Are we ready to merge? :)

PR #2940, opened 2021-12-17
Highlight most recently clicked search result (closes #2933)

### LINKED ISSUE #2933: [Enhancement] Highlight last clicked search result in global search sidebar
### Description

It would be nice if whenever one clicks on a search result in the global search sidebar, that entry would also get highlighted somehow. That would make navigating through search results easier, because you can see which result you last checked.

### Proposed Changes

Highlight the most recently clicked search result in the list of search results.

### Caveats

None that I'm aware of!

No (_actually, I'm interested in addressing this, and might work on this when I find time, but I don't want to commit myself now, so I selected "No"._)

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

_No response_

### Additional Information

_No response_

kyaso:
## Description
When a search result in the global search side bar is clicked, it will get highlighted, and stay highlighted until another search result is clicked. With the PR, I'm addressing #2933, which I opened.

## Changes
- Added two variables in `GlobalSearch.vue` to keep track of the index of the most recently click search result
- Added a conditional binding to the `class` of a result line
    - The most recently clicked result (aka the _active_ one) will have a distinct background color

## Additional information

- [x] Changelog entry todo

Tested on: Manjaro Linux


kyaso:
#2933

nathanlesage:
In principle already looks fine. There are a few optimizations we can do, and the variable names can be better (one index makes sense, two are confusing :D). Also, we can just toggle the class `active` (i.e. separate it from result-line), so that we don't run into the possibility of having to duplicate code.

Feel free to already tackle this, I'll do a better code review later!

kyaso:
Thanks for the feedback!

> Also, we can just toggle the class `active` [...]

I'm not sure if I fully understood this. Do you mean the `active` pseudo-class?

kyaso:
I hope it looks somewhat better now.

Looking forward to your feedback.

nathanlesage:
Heya, looks perfectly fine, but one question: It looks as if you only provided light mode styles, right? How does it look in dark mode? I guess we might need a different color then, don't we?

kyaso:
Indeed, that color would be applied to both light and dark mode styles. Here is a comparison:

light

dark

What do you think?

nathanlesage:
Looks good! But I'd nevertheless copy that color code to the dark mode settings, simply so that we have it (otherwise it is sometimes hard to figure out where some style has been defined, in case we want to style that differently in the future)

kyaso:
Gotcha! Will do that.

kyaso:
I also tweaked the colors a bit, as the highlight color seemed to be a bit to bright in darkmode, and a bit too dark in light mode. Now it looks like this:
light

dark

I am by no means an expert when it comes to color combinations and aesthetics, so this is the best I could come up with. Feel free to to give me your opinion on the colors :D

---

Also, I noticed that the rule for the _hover_ color in dark mode didn't get applied. So I changed the `span` to a `div` and now it works.

Previously:
hover_old

Now:
hover_new

(note: I'm actually hovering over the result, the screenshot tool just didn't capture my mouse pointer)

nathanlesage:
> I am by no means an expert when it comes to color combinations and aesthetics, so this is the best I could come up with. Feel free to to give me your opinion on the colors :D

Ah well, I'm no better either. Minimal consensus is: It's readable, and that applies, so all good :D Whoever has more knowledge can certainly adapt the colors and open their own PR!

PR #3044, opened 2022-01-15
Do not preventDefault on middle mouse click for closing tabs on Linux

### LINKED ISSUE #2663: [Bug] [Linux] Closing tab using middle-mouse button also performs unintentional paste action
### Description

When closing a tab using middle-mouse button, that tab is closed, but in the next tab, that gets focus after, whatever text you had highlighted (and thus is present in the respective system paste buffer) will get pasted at the current cursor location.

Also, when pressing the middle-mouse button within the text (to paste something), the clipboard content will be pasted _twice_.

Looks like the middle-mouse click is somehow triggering the event handler (or similar) twice?

---

(_for those that didn't know: In most Linux distributions it is possible to highlight some text, and then paste it somewhere else by pressing down the middle-mouse button (scroll wheel)_).

### Reproducing

Suppose there are two open tabs, `bar.md` and `foo.md`:

img1

Somewhere (doesn't have to be Zettlr), I have highlighted the text snippet "peanuts".

Now, when I close tab `foo.md` using the middle-mouse button, this is how `bar.md` will look afterwards:

img2

**Note**: When tab `foo.md` was the active one, and we close that using middle-mouse, the pasting in tab `bar.md` will also happen.

---

Regarding the double pasting: This is what happens, when we go to the next line and press the middle-mouse button _once_:

img3

### Zettlr Version

Stable (most recent version)

### Specify version

2.0.0

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Manjaro

### Additional Information

_No response_

Joelius300:
## Description
This PR adds a platform check to determine whether or not to call `event.preventDefault` when someone used the middle mouse button to close a tab. There was a fix in place to prevent a scroll action when using middle mouse on windows but this fix caused middle mouse click to malfunction on Linux.  
It also removes the call to `event.stopPropagation` which was added to resolve this very issue but didn't do anything.

Currently the platform is `!== 'linux'` which means that on mac, `preventDefault` will still be called (just like on Windows). This could also be changed to only check for Windows but for that someone needs to test how OSX behaves currently.

Closes #2663 

## Changes

This shouldn't be a breaking change unless someone out there used this handy-dandy feature to paste something and close a tab in the same action.. 🤔 Let's say no breaking changes :)

## Additional information

Comment of origin: https://github.com/Zettlr/Zettlr/issues/2663#issuecomment-1013676075

Tested on: 5.15.12-1-MANJARO (arch based Linux) with KDE desktop environment

It would be good if at least one Windows user, one OSX user and some other Linux users with different distros and DEs could test this before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Display inline math using monospaced font
 - Right-clicking citations doesn't select parts of it anymore
 - Inline-equations are now properly syntax highlighted
+- Fixed an error that caused a paste action when using middle mouse button to close a tab on Linux
 
 ## Under the Hood
 
--- a/source/win-main/DocumentTabs.vue
+++ b/source/win-main/DocumentTabs.vue
@@ -243,8 +243,10 @@ export default defineComponent({
       if (event.button === 1) {
         // It was a middle-click (auxiliary button), so we should instead close
         // the file.
-        event.preventDefault() // Otherwise, on Windows we'd have a middle-click-scroll
-        event.stopPropagation() // In response to #2663
+        if (process.platform !== 'linux') {
+          // Otherwise, on Windows we'd have a middle-click-scroll
+          event.preventDefault()
+        }
         this.handleClickClose(event, file)
       } else if (event.button === 0) {
         // It was a left-click. (We must check because otherwise we would also


kyaso:
Hey!

I tested your code, but unfortunately, on my machine (Linux manjaro 5.14.21-2-MANJARO; Gnome/i3) the issue is still present.

Joelius300:
@kyaso Thanks a lot but oh god.. Now it's even more than just OS level, now it's desktop environment level :tired_face:

Joelius300:
@kyaso Since you have Gnome, could you maybe check if the paste event is called for you when you middle click on the tab? You can do that by adding
```html
v-on:paste="pasteHandler($event)"
```
after line 18 in the `DocumentTabs.vue` file and
```typescript
pasteHandler: function (event: ClipboardEvent) {
  console.warn('Paste event fired')
  console.log(event)
},
```
after line 258 or somewhere in that block. Because for me, this doesn't trigger.

kyaso:
@Joelius300 Thanks for the tip! I tested it, and for me, the paste event also didn't trigger. But when I add the same "paster trigger test" to `MainEditor.vue` instead, I see paste events being triggered when I close a tab.

Could you check what happens on your machine when you do that?

Joelius300:
@kyaso Okay so I just did that (added the event to the outermost div in `MainEditor.vue` (after line 13) and it doesn't fire when I close a tab. It only does when I successfully paste text into the editor (by middle mouse clicking in the editor div, but not on any text span, that doesn't do anything).

Joelius300:
I just found something probably important. The paste event that's triggered in `MainEditor.vue` has nothing to do with the tab. It's the result of actually pasting contents into the editor, which only happens (for me) when `event.preventDefault` is called. So for some reason the way to prevent pasting contents for me is to _prevent calling `preventDefault`_..

Another thing I found is that using middle mouse button to paste text into the editor normally (again, only works when you DON'T click on a span but below), then the event has A SPAN as target and source element (that one WTF) and it also has `defaultPrevented` set `false`. However the paste event that's fired when closing a tab has `defaultPrevented` set to `true` (makes sense as `preventDefault` has to be called for it to even fire for me).

I'm not sure maybe I'm over-exaggerating these findings now but I'm very confused..

kyaso:
> You also did your tests using this PR as base or did you use the develop branch directly?

I used your `fix/middle-click-pasting` branch.

A very weird issue indeed 😅

nathanlesage:
Stupid idea, but if we un-focus the main editor before triggering the middle click, does that maybe help ...?

Joelius300:
Closing in favor of #3061 as this doesn't fix the issue on all desktop environments (namely GNOME).

PR #3045, opened 2022-01-15
Fix #2663

### LINKED ISSUE #2663: [Bug] [Linux] Closing tab using middle-mouse button also performs unintentional paste action

kyaso:
## Description

My proposal for fixing #2663. The idea was to react to middle-mouse clicks only when a _mouseup_ (instead of previously _mousedown_) event is detected. That way, an accidental paste into the new tab (that gets opened after the closed tab) is prevented.

## Changes

* Added a `mouseup` event in `DocumentTabs.vue`
* Added a handler `handleMiddleMouseClick()` for the `mouseup` event
* Moved code related to middle-mouse click from `handleClickFilename()` into the new handler
* Removed the previous attempt to fix #2663 (the `event.stopPropagation()`)

Tested on: Linux manjaro 5.14.21-2-MANJARO

## Additional info
It would be great if this could be tested by Windows, Mac, and other Linux users!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Display inline math using monospaced font
 - Right-clicking citations doesn't select parts of it anymore
 - Inline-equations are now properly syntax highlighted
+- Fixed a bug that caused unintentional pastes when closing a tab using the
+  middle mouse button on Linux
 
 ## Under the Hood
 
--- a/source/win-main/DocumentTabs.vue
+++ b/source/win-main/DocumentTabs.vue
@@ -15,6 +15,7 @@
       v-on:drag="handleDrag"
       v-on:dragend="handleDragEnd"
       v-on:contextmenu="handleContextMenu($event, file)"
+      v-on:mouseup="handleMiddleMouseClick($event, file)"
       v-on:mousedown="handleClickFilename($event, file)"
     >
       <span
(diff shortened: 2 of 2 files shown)


kyaso:
Related PR: #3044

kyaso:
Closed in favor of #3061.

PR #3048, opened 2022-01-16
Open in new tab button on preview tooltip Fixes #3047

### LINKED ISSUE #3047: Offer "Open in New Tab" also on the Preview Tooltip [Enhancement]
### Description

The preview tooltip has an "Open" button, that respects the current setting of "Avoid opening files in new tabs". While this preference (together with the new back/forward history) is very useful, sometimes it is desireable to open a file in a new tab. When opening files from the file manager, one already can use the context menu to achieve this via the context menu "Open in new Tab". But on the Preview Tooltip on an internal link there is only this one open button.

grafik

### Proposed Changes

I would suggest:
If the user has set the "Avoid Opening Files in new Tabs"... then (and only then!) offer an additional button behind the existing "Open" button, which allows to optionally open the link target in a new tab. 

This functionality would be similar what is possible in web browsers via context menu on links: one can choose.

grafik

### Caveats

Currently the tooltip is created on page-load. So, if user changes the "Avoid opening Files in New Tabs", the second button will only appear on freshly opened files, or after pressing F5 to reload.

Yes

### Zettlr Version

Right off the develop branch

### Your Platform
- [X] macOS

### Operating System Version

Monterey 12.1

### Additional Information

I have this already implemented this on a fork branch. Will send a PR ASAP.

derwok:
Details see #3047

  - [X] I documented all behaviour as far as I could do.
  - [X] I target the develop-branch, and *not* the master branch.
  - [X] I have tested this extensively and paid extra attention to
  - [X] I have made use of ESLint (yarn lint) using the provided configuration from the
  - [X] I have added an entry to the CHANGELOG.md.
  - [X] I matched my code-style to the repository (as far as possible).
  - [X] I do agree that my code will be published under the GNU GPL v3 license.
  - [X] As far as JS-files are concerned, I made sure to copy (in case of new files)
  - [X] I synced the latest commits to develop shortly before proposing

## Description
see #3047

## Changes
added creation of 2nd  button "Open in new Tab"

## Additional information
Tested on: MacOS


source/win-main/MainEditor.vue
```diff
@@ -572,7 +572,10 @@ export default defineComponent({
     mdEditor.on('zettelkasten-link', (linkContents) => {
       ipcRenderer.invoke('application', {
         command: 'force-open',
-        payload: linkContents
+        payload: {
+                linkContents: linkContents,
+                newTab: false // let open-file command decide based on preferences
```
nathanlesage, line 577: I would suggest to set this to `undefined`, which will perfectly work in the main process and furthermore express that we do not care whether or not this file will be opened in a new tab. Makes it easier to see.

source/main/commands/force-open.ts
```diff
@@ -26,12 +26,15 @@ export default class ForceOpen extends ZettlrCommand {
...
+    * Force-Opens a file, after click on internal link
     * @param {String} evt The event name
-    * @param  {Object} arg the parameters of the file to be deleted
-    * @return {Boolean} Whether the file was successfully deleted.
+    * @param  {Object} payload the parameters of the file to be opened
+    * @return {Boolean} Whether the file was successfully opened.
     */
-  async run (evt: string, arg: any): Promise<void> {
+  async run (evt: string, payload: any): Promise<void> {
+    const arg = payload.linkContents
```
nathanlesage, line 35: A potentially cool idea would be to use a spread operator here: `const { linkContents, newTab } = payload`

Also, this would make `arg` a tad more expressive; it was a stupid idea by me to name this variable like that 😄

source/common/modules/markdown-editor/hooks/note-preview.ts
```diff
@@ -101,9 +104,31 @@ function getPreviewElement (metadata: [string, string, number, number], linkCont
...
+                  linkContents: linkContents,
+                  newTab: true
+          }
+      })
+        .catch(err => console.error(err))
+    }
+
+    const openButtonNT = document.createElement('button')
+    openButtonNT.setAttribute('id', 'open-note-new-tab')
+    openButtonNT.textContent = 'Open in new Tab'
```
nathanlesage, line 126: There is a translation string for that, which we should use! (i.e. `trans('identifier')`)

source/common/modules/markdown-editor/hooks/note-preview.ts
```diff
@@ -101,9 +104,31 @@ function getPreviewElement (metadata: [string, string, number, number], linkCont
...
+    openButtonNT.addEventListener('click', openFuncNewTab)
+    openButtonNT.style.cssText += ';margin-left :10px;'
```
nathanlesage, line 128: Rather than using `cssText` please use the correct style itself:

`openButtonNT.style.marginLeft = '10px'`

source/common/modules/markdown-editor/hooks/note-preview.ts
```diff
@@ -92,7 +92,10 @@ function getPreviewElement (metadata: [string, string, number, number], linkCont
   const openFunc = function (): void {
     ipcRenderer.invoke('application', {
       command: 'force-open',
-      payload: linkContents
+      payload: {
+                linkContents: linkContents,
+                newTab: false   // let open-file command decide based on preferences
```
nathanlesage, line 97: I think the indentation here is off which the linter won't like …

nathanlesage (changes requested):
In general, a good initiative and a useful feature. There are a few things I'd change, but other than that it looks perfectly fine!

Also, thanks for fixing the old and wrong documentation on the force-open command 😄

derwok:
@nathanlesage thanks for the valuable & true review comments. Fixed them all (I think) and fixed the 'yarn lint' errors.

nathanlesage (approved):

nathanlesage:
Perfect, awesome!

Issue #3012, opened 2022-01-07
[Enhancement] Indicate in Table of Contents tab the section the cursor is in


jgsw:
### Description

The Table of Contents tab currently enables you to go to any section you want to by clicking on it, but it doesn't currently give a visual guide as to which section you are actually in. When working with a large document it's often helpful to be able to see at a glance which section you're in. Either emboldening or highlighting the name of the section the cursor's currently in on the Table of Contents tab would help with awareness of one's current location in the document. 

### Proposed Changes

Change the drawing of the TOC tab, so that the section the cursor's in is made more obvious -- either by highlighting it, or displaying it in bold.

### Caveats

This would require redrawing the TOC when the cursor goes from one section to another, but I don't think that this would be computationally expensive. 

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

kyaso:
I would also welcome this feature!

nathanlesage:
@kyaso This should be fairly easy to implement, if you like give it a shot!

All the information you need is already available to the MainSidebar component.

The ToC entries contain the starting line, so the cursor must be greater than or equal to the line of a ToC entry while being less than the next entry. The line where the cursor is at is stored in `this.$store.state.activeDocumentInfo.cursor.line`

kyaso:
> This should be fairly easy to implement, if you like give it a shot!

Indeed, I have thought about giving it a try, as I have plenty of time this weekend :smiley: 

Thanks for the hints, that should save me some time!

kyaso:
@nathanlesage I crafted something over at #3071.

Also, currently, only the _cursor_ position is relevant. However, I know of examples, where additionally the _scroll_ position is also considered (e.g. Google Docs). What do you think?

nathanlesage:
@kyaso Yes, definitely. For this we could add the viewport information to the document info. Then we could do that. The question would be which one makes more sense 🧐 idk, maybe discuss that on Discord...?

kyaso:
> maybe discuss that on Discord...?

Sure, feel free to start a thread!

nathanlesage:
Done!

PR #3071, opened 2022-01-21
Highlight ToC entry of section cursor is currently in

kyaso:
## Description

This PR implements the feature requested in #3012.

Depending on the current location of the cursor within a file, the corresponding section in the table-of-contents (ToC) tab of the sidebar is highlighted.

While the basic highlighting is working so far, there are still some minor tasks to be done:
- [ ] Adjust highlight color for light and dark modes
- [ ] Optimizations?

## Changes

Only `MainSidebar.vue` edited. Added a `v-bind:class` to the ToC entries that will conditionally set a highlight background, depending on the position of the cursor (determined by method `tocIsActive()`).

## Additional information


source/win-main/MainSidebar.vue
```diff
@@ -427,6 +428,25 @@ export default defineComponent({
...
+     * @param   {number}  tocIdx                Index of heading in ToC
+     */
+    tocIsActive: function (tocEntryLine: number, tocIdx: number) {
+      const cursorLine = this.$store.state.activeDocumentInfo.cursor.line
+      // Determine index of next heading
+      const nextTocIdx = Math.min(tocIdx + 1, this.tableOfContents.length - 1)
+      var nextTocEntryLine = this.tableOfContents[nextTocIdx].line
+      // When this is the last heading, use Infinity as the next "heading"
+      if (nextTocEntryLine === tocEntryLine) {
+        nextTocEntryLine = Infinity
```
nathanlesage, line 445: I would simply preset that var with Infinity and then overwrite this iff the current ToC entry is not the last one. Plus, I think the linter is set to only accept let and const, but not var!

source/win-main/MainSidebar.vue
```diff
@@ -18,6 +18,7 @@
         v-for="(entry, idx) of tableOfContents"
         v-bind:key="idx"
         class="toc-entry-container"
+        v-bind:class="{'toc-active': tocIsActive(entry.line, idx)}"
```
nathanlesage, line 21: Maybe rename to toc-item-active to make sure that not the full toc is active ...?

source/win-main/MainSidebar.vue
```diff
@@ -525,6 +546,11 @@ body {
         cursor: pointer;
         &:hover { text-decoration: underline; }
       }
+
+    }
+
+    div.toc-active {
+      background-color: rgb(200, 200, 200);
```
nathanlesage, line 553: Maybe instead of using the background color and put this in line with the rest of the application, apply a bold text weight and set the color to `var(--system-accent-color)`!

nathanlesage (commented):
Looks pretty good to me, I have some ideas to your questions :)

nathanlesage:
Note to readers: This PR is currently subject to discussion in #3012 and on [Discord](https://discord.gg/PcfS3DM9Xj). Please feel free to chime in!

source/win-main/MainSidebar.vue
kyaso: Done in 41d94f8!

source/win-main/MainSidebar.vue
kyaso: Done.

Btw, I renamed it to `toc-entry-active`, or is there a difference between ToC "item" and "entry"?

source/win-main/MainSidebar.vue
kyaso: Done.

Also, I moved the `toc-entry-active` (how it's called now) into to the `toc-entry-container` block. Before that, it was floating kinda on its own.

source/win-main/MainSidebar.vue
```diff
@@ -26,7 +26,12 @@
         <div class="toc-level">
           {{ entry.renderedLevel }}
         </div>
-        <div class="toc-entry" v-bind:data-line="entry.line" v-html="entry.text"></div>
+        <div
+          class="toc-entry"
```
nathanlesage, line 30: You can add that to the v-bind directive by using `toc-entry: true` -- this way there is just one class property!

nathanlesage (commented):
Saw another thing to improve :)

source/win-main/MainSidebar.vue
kyaso: Done!

nathanlesage:
Nice work thus far! :) Now the only thing remaining (at least based on what the discussion on Discord surfaced thus far) would be to potentially also highlight the viewport …? But potentially, this could also be done in a second PR?

It's your call: Either merge or wait :)

kyaso:
> Nice work thus far! :) Now the only thing remaining (at least based on what the discussion on Discord surfaced thus far) would be to potentially also highlight the viewport …? But potentially, this could also be done in a second PR?
>
> It's your call: Either merge or wait :)

I would also like the viewport to influence the highlight. But maybe we merge this PR first so that we at least have something to test in 2.1.3 (or earlier in the nightlies). And other users can try this out as well and see how they like it. Then, based on the feedback, we could create a follow-up PR. Ok?

Btw, I need to add a Changelog entry. Should I put this as a "new feature" then?

nathanlesage:
Sounds good! With regard to the changelog: I think a simple "The ToC now displays the currently active section" would suffice in this case. It's more an enhancement of the ToC-feature rather than a new feature :D (But these things generally can only be determined by using eye-balling)

kyaso:
> It's more an enhancement of the ToC-feature rather than a new feature :D

Right, this makes sense :D

nathanlesage:
Okay, I'm going to merge now, and add it manually :D

kyaso:
What about this: https://github.com/Zettlr/Zettlr/pull/3071#discussion_r790557160 ?

Shall I just revert the change and push?

nathanlesage:
Ouh, I completely overlooked that! But I just tested it out – I think it's relatively neat to have a little bit more spacing to the right. So idk, I'd say we keep that in until it annoys us in some other place – saves us the hassle. What do you think?

kyaso:
Sounds good!

PR #3081, opened 2022-01-24
Fix: Filemanager now shows character count or word count depending on preference setting, Fixes #3075

### LINKED ISSUE #3075: [Bug] Character counting error.
### Description

File manager counts words instead of character although I ticked "Count character instead of words".

image

### Reproducing

Ticked "Count character instead of words"?

### Zettlr Version

Stable (most recent version)

### Specify version

2.1.2

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 10

### Additional Information

_No response_

derwok:
## Description
see #3075

## Changes
Reactively checking the user preferences and then either counting word or characters

## Additional information
* I spotted, that the filemanager numbers in word count and char count differ from the currently active and open files meta info shofing in the toolbar. But this diff was already in the word count stats. So not introduced and not fixed with this PR.

* Tested on: MacOS 12.1

### DIFF
--- a/source/win-main/file-manager/file-item.vue
+++ b/source/win-main/file-manager/file-item.vue
@@ -53,7 +53,7 @@
         <div v-if="isDirectory">
           <span class="badge">{{ countDirs }}</span>
           <span class="badge">{{ countFiles }}</span>
-          <span class="badge">{{ countWords }}</span>
+          <span class="badge">{{ countWordsOrCharsOfDirectory }}</span>
         </div>
         <template v-else>
           <div v-if="hasTags">
@@ -89,7 +89,7 @@
               relation to a set writing target, if there is one.
             -->
             <span v-else-if="!hasWritingTarget" class="badge">
-              {{ formattedWordCount }}
+              {{ formattedWordCharCountOfFile }}
             </span>
             <span v-else class="badge">
               <svg
@@ -163,6 +163,9 @@ export default defineComponent({
     }
   },
   computed: {
+    shouldCountChars: function (): boolean {
+      return this.$store.state.config['editor.countChars']
+    },
     useH1: function (): boolean {
       return this.$store.state.config.fileNameDisplay.includes('heading')
     },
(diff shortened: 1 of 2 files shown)


derwok:
In the function `countWordsOrCharsOfDirectory`I introduced some ugly code redundancy... but I wanted to keep the expensive reactive preference check **outside** of the costly map/reduce/filter loop...

nathanlesage:
Looks good, and the code duplication could be moved into the mapper, but it should also work this way. I'll check this again tomorrow on desktop, but I don't spot any mistake right now. Thanks for opening the PR this quickly!

source/win-main/file-manager/file-item.vue
```diff
@@ -230,17 +233,24 @@ export default defineComponent({
...
-      const wordCount = this.obj.children
-        .filter((file: any) => file.type === 'file')
-        .map((file: any) => file.wordCount)
-        .reduce((prev: number, cur: number) => prev + cur, 0)
-
-      return trans('gui.words', localiseNumber(wordCount))
+      if (this.shouldCountChars) {
+        const charCount = this.obj.children
+          .filter((file: any) => file.type === 'file')
+          .map((file: any) => file.charCount)
```
nathanlesage, line 244: What you could do is run `map((file: any) => (this.shouldCountChars) ? file.charCount : file.wordCount)`. Then you won't have duplicate logic, and `shouldCountChars` will only be re-evaluated once it changes, so it's a simple lookup.

source/win-main/file-manager/file-item.vue
```diff
@@ -230,17 +233,24 @@ export default defineComponent({
       }
       return this.obj.children.filter((e: any) => [ 'file', 'code' ].includes(e.type)).length + ' ' + trans('system.files') || 0
     },
-    countWords: function () {
+    countWordsOrCharsOfDirectory: function () {
       if (this.obj.type !== 'directory') {
         return 0
```
nathanlesage, line 238: This is not your fault, but while we're at it, I think returning an empty string here is better … I don't know why I returned a 0!

nathanlesage (approved):
Should work perfectly fine, just two suggestions. But the code will run without them! :)

Oh, and btw, the file manager will show a different word/char count because it only counts what is saved to disk, while the toolbar counts the contents of the editor as they are!

source/win-main/file-manager/file-item.vue
derwok: Updated PR. Even did a curious quick performance time measurement check, on a 2,000 MD files "real live" workspace, I could not spot even an additional millisecond spent with the "in-loop" check... so: conviced. :-)

nathanlesage:
Looks good! Thanks for your work — merging.

PR #3085, opened 2022-01-25
Highlight neighboring ToC headings

kyaso:
## Description
This draft PR implements highlighting of neighboring ToC sections as the cursor moves through the file. Currently, a static highlighting scheme is employed, i.e., a fixed amount of sections is always highlighted, regardless of section sizes. The following screencast shows it in action:

https://user-images.githubusercontent.com/18646029/151075190-9fd72cc3-4e04-4d70-86c2-1637c9b35af5.mp4

## Changes

In `MainSidebar.vue`:
* The ToC index of the currently active (i.e., where the cursor is) section is now stored in the Vue data object
* Added a function `tocEntryContainerShade()` that returns an `rgba` string
    - The alpha value depends on the distance (between indices) to the active reading
    - The return value (rgba) is used as the background color for the `toc-entry-container` divs
    - The upper and lower three neighbors of the active section will be highlighted (see video above)

## Additional information / Open points
This is still WIP. Things that need to be discussed:
- [ ] Background colors (should be more subtle than in the video)
- [ ] How many neighboring sections to highlight
- [ ] Whether to keep the static highlighting OR adjust the alpha values of the headings depending on the cursor's distance to the currently highlighted sections
    - With the _static_ highlighting, you won't see anything change when moving through a very long section. In contrast, a _dynamic_ highlighting scheme would adjust the alpha values, and you would notice a shift as you approach the next section.

Also, in the near term, we have intended to base the highlighting on the current _viewport_.

### DIFF
--- a/source/win-main/MainSidebar.vue
+++ b/source/win-main/MainSidebar.vue
@@ -19,8 +19,7 @@
         v-bind:key="idx"
         class="toc-entry-container"
         v-bind:style="{
-          'margin-left': `${entry.level * 10}px`
-        }"
+          'margin-left': `${entry.level * 10}px`, 'background-color': tocEntryContainerShade(idx)}"
         v-on:click="($root as any).jtl(entry.line)"
       >
         <div class="toc-level">
@@ -162,7 +161,9 @@ export default defineComponent({
   data: function () {
     return {
       bibContents: undefined as undefined|any[],
-      relatedFiles: [] as RelatedFile[]
+      relatedFiles: [] as RelatedFile[],
+      // Contains the ToC index of the currently active section
+      activeTocEntryIdx: undefined as undefined|number
     }
   },
   computed: {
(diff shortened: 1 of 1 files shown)


derwok:
I like this new "Ui Candy". Really. 
But one thought (sorry, no Discord)... Not everything that is technically possible makes sense from the user's point of view.
So, when my cursor is e.g in chapter 1.1.1 - I understand why ch.1.1 and ch.1 is highlighted - because I'm in that hierarchy! But why is chapter 2 and chapter 3 highlighted? That I don't understand. Just because I'm "close" to it? A table of content is a hierarchy - and I think that sort of highlighting should support that semantics. My 2 cts.

kyaso:
@derwok Valid point! The initial idea behind the highlighting was to have a visual feedback on where you are within a document. That would be helpful once the viewport location is also considered because that way the "wave" would also move while you scroll through a document.

Right now (draft, so nothing is final yet! :)) it is indeed a bit too distracting, but one option could be to only activate the wave effect when you start _scrolling_. And not when you are editing (e.g. moving the cursor); the cursor location would only cause the current section to be highlighted (that has already been implemented in #3071).

@nathanlesage Your thoughts?

nathanlesage:
> Background colors (should be more subtle than in the video)

Agreed, but in general some very light gray should be the most suitable, since that's the prominent color of most of the UI across all three platforms, so I think your color scheme goes a lot into the right direction!

> How many neighboring sections to highlight

I would say it depends, but the amount of sections to highlight should have a relationship to the amount of sections inside the document – so more highlighted sections in documents with more sections, and less in smaller documents. I'd say at least 3 (the current visible one, the preceding one, and the following one).

> Whether to keep the static highlighting OR adjust the alpha values of the headings depending on the cursor's distance to the currently highlighted sections

I guess that a dynamic animation would be nicer, because this way you have more immediate feedback as you scroll!

***

Now the next set of issues raised by @derwok 

So the general discussion on Discord was basically the following: We wanted to add some more visual feedback to the ToC; one thing which universally made sense was to highlight the section where the cursor/primary selection is in, which we already implemented. Now another idea was to also reflect the current scroll position within the document inside the ToC. Code editors like VSCode use a minimap (see screenshot) where you can see what's in the document and where you currently are. While I think a full minimap might be too technical for a writing editor, the general idea to give some feedback where in a document the user is could be cool.

image

That's what we try to achieve.

> A table of content is a hierarchy - and I think that sort of highlighting should support that semantics.

Agreed, that's a valid point. My personal intuition was just that, if you just scroll while the ToC is open, you immediately realize what these highlights do: Show you where you are right now. But then, I'm used to those code minimaps, and I'm the one who had the idea, so I see the point that users might be confused by that. Do you think that after seeing the selection immediately change when you scroll a document, that this then makes it understandable for users, or do you see more difficulties in this regard? If so, do you think we can marry those semantic interpretation with such a scrollposition-highlight? Or do you think it might be better to not do it?

> but one option could be to only activate the wave effect when you start scrolling.

That actually might be a good compromise. Also, I think that we could make sure to always highlight the full width of the ToC. This makes sure that the wave "ignores" the hierarchy, and might make it easier to discern that this wave is not that much related to the ToC itself rather than to the document.

Issue #2029, opened 2021-05-28
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths


cskeogh:
## Description
Zettlr displays error:

```
WARNING in ./source/common/i18n.js
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing 
files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "C:/Users/craig/Zettlr/source/common/lang/ar-AR.json"
 * "C:/Users/craig/Zettlr/source/common/lang/ca-CA.json"
 * "C:/Users/craig/Zettlr/source/common/lang/cs-CZ.json"
 * and more ...
 @ ./source/common/modules/window-register/register-default-context.ts 4:15-36
 @ ./source/common/modules/window-register/index.ts 8:51-88
 @ ./source/win-main/index.ts 19:42-86

1 warning has detailed information that is not shown.
Use 'stats.errorDetails: true' resp. '--stats-error-details' to show it.

webpack 5.37.1 compiled with 1 warning in 12727 ms
```

## Reproducing
Start Zettlr. Error occurs on startup.

## Expected behaviour
Zettlr starts without error

## Platform
 - OS and version: Windows. Does not happen on Linux.
 - Zettlr Version: develop e358a054c7dc3e8edd431b21ea61f550dcff9e69

## Additional information
Introduced in 5061aca4041ca7ab2496401cd41553885bd2f8a3

nathanlesage:
Would you mind running that command using the `--stats-error-details` flag? Maybe that'll spit out more info. Other than that, I'm also not experiencing this issue on macOS, so it might have to do either with git (CRLF-issues?) or Webpack itself (maybe there's something on the web on this?)

cskeogh:
I added the following in `webpack.main.config.js`:

```javascript
module.exports = {
  stats: {
    errors: true,
    errorStack: true,
    errorDetails: true, // --display-error-details
  },
```

The error is:

[...]

PR #3087, opened 2022-01-26
Typed version of IPC

tobiasdiez:
## Description

Second try at a typed IPC, superseding https://github.com/Zettlr/Zettlr/pull/1048. The problem with that PR was that the IPC provider had to be a class implementing an interface. In this new version, we essentially don't change anything how IPC providers are defined.

With these changes everything in the IPC is now strongly typed: the caller as well as the definition of the provider. The only thing that didn't work was to properly check for the return type of the provider (so you can return something else than what is defined in the protocol). I tried several things, but fear that's a current limitation with typescript; will open an issue at the ts repo.

Inspiration came from https://github.com/antfu/webext-bridge#type-safe-protocols.
## Changes

- Create drop-in replacement of the `ipcMain` object from electron, with the only change that the `handle` method is now strongly typed.
- Create `ipcRenderer` object that acts as a proxy to create a more readable version to invoke ipc methods.
- Illustrate the process at the example of the tag provider. Only change necessary was to rename the commands to use pascal case, so that they fit the javascript convention of real names.

## Additional information

Cannot currently run it, because it fails with the following error:
```
WARNING in ./node_modules/electron/index.js
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "D:/Programming/Zettlr/node_modules/electron/dist/LICENSE"
 * "D:/Programming/Zettlr/node_modules/electron/dist/LICENSES.chromium.html"
 * "D:/Programming/Zettlr/node_modules/electron/dist/chrome_100_percent.pak"
 * and more ...
 @ ./source/common/ipc.ts 4:19-38
 @ ./node_modules/ts-loader/index.js??clonedRuleSet-8.use!./node_modules/vue-loader/dist/index.js??ruleSet[0]!./source/win-main/MainSidebar.vue?vue&type=script&lang=ts 24:14-36
```
any idea what's going on? Is it not possible to use imports from electron in the renderer? The same issue for the language files had been reported before in https://github.com/Zettlr/Zettlr/issues/2029, but that doesn't seem relevant.

Tested on: Win 11

### DIFF
--- a/source/app/ipc.ts
+++ b/source/app/ipc.ts
@@ -1,5 +1,5 @@
 import { TagProviderProtocol } from '@providers/tag-provider'
-import { IpcMainInvokeEvent, ipcMain as electronIpcMain, ipcRenderer as electronIpcRenderer, IpcRenderer } from 'electron'
+import { IpcMainInvokeEvent, ipcMain as electronIpcMain } from 'electron'
 
 /**
  * Defines all channels of the Ipc, and which commands are supported on this channel.
@@ -47,7 +47,7 @@ type PromisfiedAll<T> = {
 /**
  * Replaces all functions in all channels by functions that return a Promise.
  */
-type PromisfiedProtocol = {
+export type PromisfiedProtocol = {
   [K in keyof IpcProtocol]: PromisfiedAll<IpcProtocol[K]>
 }
 
@@ -117,22 +117,3 @@ class IpcMain {
   }
 }
 export const ipcMain = new IpcMain()
-
-export const ipcRenderer: PromisfiedProtocol & IpcRenderer = new Proxy({}, {
-  get (target: any, propChannel: PropertyKey, receiver: any): any {
-    const channel = propChannel.toString()
-    if (channel === 'invoke') {
-      // For backwards compatibility
-      return electronIpcRenderer.invoke
-    }
-
-    return new Proxy({}, {
-      get (target: any, propMessage: PropertyKey, receiver: any): any {
-        const message = propMessage.toString()
-        return async function (...args: any[]) {
-          return await electronIpcRenderer.invoke(channel, { command: message, payload: args })
-        }
-      }
-    })
-  }
-})
--- a/source/app/ipcRenderer.ts
+++ b/source/app/ipcRenderer.ts
@@ -1,0 +1,25 @@
+import { IpcRenderer } from 'electron'
+import { PromisfiedProtocol } from './ipc'
+
+export const ipcRenderer: PromisfiedProtocol & IpcRenderer = new Proxy({}, {
+  get (target: any, propChannel: PropertyKey, receiver: any): any {
+    const electronIpcRenderer = (window as any).ipc as Electron.IpcRenderer
+
+    const channel = propChannel.toString()
+    // For backwards compatibility
+    if (channel === 'invoke') {
+      return electronIpcRenderer.invoke
+    } else if (channel === 'on') {
+      return electronIpcRenderer.on
+    }
+
+    return new Proxy({}, {
+      get (target: any, propMessage: PropertyKey, receiver: any): any {
+        const message = propMessage.toString()
+        return async function (...args: any[]) {
+          return await electronIpcRenderer.invoke(channel, { command: message, payload: args })
+        }
+      }
+    })
+  }
+})
--- a/source/app/service-providers/tag-provider.ts
+++ b/source/app/service-providers/tag-provider.ts
@@ -17,7 +17,7 @@ import path from 'path'
 import { app } from 'electron'
 import broadcastIpcMessage from '@common/util/broadcast-ipc-message'
 import { ColouredTag, TagDatabase } from '@dts/common/tag-provider'
-import { ipcMain } from '@common/ipc'
+import { ipcMain } from '../ipc'
 
 interface InternalTagRecord {
   text: string
--- a/source/app/service-providers/tag-provider.ts
+++ b/source/app/service-providers/tag-provider.ts
@@ -14,14 +14,22 @@
 
 import fs from 'fs'
 import path from 'path'
-import { app, ipcMain } from 'electron'
+import { app } from 'electron'
 import broadcastIpcMessage from '@common/util/broadcast-ipc-message'
 import { ColouredTag, TagDatabase } from '@dts/common/tag-provider'
+import { ipcMain } from '@common/ipc'
 
 interface InternalTagRecord {
   text: string
   files: string[]
   className: string
+}
+
+export interface TagProviderProtocol {
+  getTagsDatabase: () => TagDatabase
+  setColouredTags: (tags: ColouredTag[]) => void
+  getColouredTags: () => ColouredTag[]
+  recommendMatchingFiles: (tags: string[]) => { [key: string]: string[] }
 }
 
 /**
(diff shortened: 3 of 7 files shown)


nathanlesage:
Thanks for this next iteration of the IPC types! It is a lot to digest, and I need time to understand the type magic in there. Potentially, that is also a cause of the problem that Vue loader experiences, maybe the types are too complex for the loader?

That being said, if we could somehow simplify that, this would be great. One thing I could think of would be to create an API interface that is weakly typed and simply includes explicit type checks and returns those. Another idea: What if we were to simply extend the ipcMain and ipcRenderer types that ship with Electron to just provide function overloads per each channel? That would be more boilerplate, but potentially more explicit and simpler to implement? Something like this:

```typescript
// If anything below is not correct syntax that's because I didn't write that with a linter ...
declare module ipcMainProtocol extends ipcMain {
  handle('some-channel', async (event: IpcEvent, message: { command: UnionType, payload: keyof UnionType }) => ReturnTypes)
}
```

Okay, there's really a lot wrong with that snippet, but I hope you get the idea …

Basically: The more we can spot at compile time the better. And something quick and dirty might also work. But in the meantime I'll dig into your type magic and try to understand that to give a more educated opinion!

tobiasdiez:
> It is a lot to digest, and I need time to understand the type magic in there. 

I know its quite a bit of black magic in there. Essentially, what these types are doing is to construct the right listener type information from the protocol interface. Of course, one could hard code this, but I think its neat to let typescript compute it from the interface. Makes it very simple to add new functions or even new channels. 

If you really want to dive into these types, I would suggest you copy the file to https://www.typescriptlang.org/play?# and then use the neat `\\ ^?` syntax there to show the typing.

> Potentially, that is also a cause of the problem that Vue loader experiences, maybe the types are too complex for the loader

I hope not. The loader should essentially call typescript, which then strips all the types since they are not needed for the runtime. It's also complaining about the electron import not the types...

> What if we were to simply extend the ipcMain and ipcRenderer types that ship with Electron to just provide function overloads per each channel?

I tried that first, but you cannot overwrite existing type information only extend them. That's why I went for a simple wrapper with typing information.

nathanlesage:
> I tried that first, but you cannot overwrite existing type information only extend them. That's why I went for a simple wrapper with typing information.

Really? Not even by re-declaring the stuff? Ouh …

> May I suggest that you try to convert one of the other simple providers to this new interface. Then you will directly see how VS is detecting the right typing information and even supports you writing something like if command === 'what comes here?'.

Yes, definitely, once we have figured out the problem with Vue :D 

Ah, I think I know what's happening: Since the process isolation is strict, ipcMain is only available when called from main processes, and ipcRenderer is only available when called from main processes. Therefore, by importing not just the types, but also the actual modules into the TS declaration file, you're creating incompatibilities since these modules are by that requested to also stay in the compiled code, which doesn't work.

nathanlesage:
Alright, I'm looking up the types now, and figuring out what to do. I think I already have two minor corrections to this:

1. The TagProvider interface should ideally be defined in the "@dts" folder, I try to separate types from the logic as much as possible. The reason is that we frequently have situations in which functions benefit from stronger typing without the need to import any function directly from the module which generates the types. This will be even more important once the main modules have been migrated to service providers, since the entire state is made up from composite types.
2. I think we should make the Promises a hard dependency. That is: Require everything that runs through the send and handle methods to be asynchronous (since that is essentially what a send->handle->response chain is). This would already remove the Proxification type, since we will know that everything in there is asynchronous. If there are any non-asynchronous functions that suddenly yield type errors, this would be the ideal time to refactor them.

tobiasdiez:
> Ah, I think I know what's happening: Since the process isolation is strict, ipcMain is only available when called from main processes, and ipcRenderer is only available when called from main renderer processes. 

Perfect! Splitting the ipc module in things for main and things for renderer indeed worked!

> The TagProvider interface should ideally be defined in the "@dts" folder, I try to separate types from the logic as much as possible. The reason is that we frequently have situations in which functions benefit from stronger typing without the need to import any function directly from the module which generates the types.

You know that type-only imports are completely erased during compilation time? Thus, it really doesn't matter where you declare your types. Generally speaking, you normally have a separate folder for types only to provide typing for external modules. I can move the interface of course, but I think its natural place is in the tag-provider file.

 > Require everything that runs through the send and handle methods to be asynchronous (since that is essentially what a send->handle->response chain is). This would already remove the Proxification type...

The thing is that some of the methods of the providers are not async. For example, all commands in the tag-provider finish immediately and thus don't return any promises. You are right of course that the ipc call will be async in the end, so the `invoke` result will be a promise. The Promisfied type models exactly this situation where your original methods are not async, but the ipc call is.

> The problem is that we need to have them run through the preload script, and the context isolation bridge will break with Proxies. That is also where my intuition to simply hard code everything comes from. Instead of mingling too much with the types (which I know you really enjoy) will just get us unnecessarily into problems.

The proxies are only there for syntactic sugar so that instead of the somewhat clumsy
```ts
await ipcRenderer.invoke('tag-provider', {
        command: 'recommend-matching-files',
        payload: this.activeFile.tags.map(tag => tag) // De-proxy
      })
```
you can simply write
```ts
await ipcRenderer.tagProvider.recommendMatchingFiles(
        this.activeFile.tags.map(tag => tag) // De-proxy
      )
```
And this seems to work, so I'm not sure what issues you are referring to with the preload script.

nathanlesage:
> You know that type-only imports are completely erased during compilation time? Thus, it really doesn't matter where you declare your types. Generally speaking, you normally have a separate folder for types only to provide typing for external modules. I can move the interface of course, but I think its natural place is in the tag-provider file.

Yes, you are right. However, I want to keep the files lean, and that is one really simple way of doing so. The FSAL, for example, is already quite huge (> 1000 LoC) even without the types. Just keeps it a tad tidier and also has a separation of concerns, so to speak.

> The Promisfied type models exactly this situation where your original methods are not async, but the ipc call is.

That's what I meant with that we should make this a strict requirement: Saves us the hassle of doing type magic.

> And this seems to work, so I'm not sure what issues you are referring to with the preload script.

Could be that I was too quick here again, apologies ;) 

But what about my last concern with the unified API?

tobiasdiez:
> That's what I meant with that we should make this a strict requirement: Saves us the hassle of doing type magic.

So that would then require to return promises in the `handle` methods. But that would be a runtime change, while the proxified type is only there for type checking... 

 > But what about my last concern with the unified API?

Sorry forgot to comment on this one. The last commit now also uses `window.ipc`, so no change there anylonger from the status quo. The new design also gives you a bit more flexibility, since in the new `ipcRenderer` one could replace the calls to the real electron ipc by something else if one is in a web environment.

nathanlesage:
That sounds good, yes. And regarding the runtime change: absolutely, but my guess would be that this could lead to cleaner code and better expectancy since we would know that an async return would be required, wouldn't it...?

tobiasdiez:
To be honest, I don't see the point of it. I think cluttering the code with `Promise.resolve` is less readable than having the `Promisfied` type in the background that you will never have to interact with it.

nathanlesage:
Hmhm, that's a point. I will think about this, but as always: I think the problem here is less your virtue in writing concise code in TS, but rather my lack of knowledge/ignorance!

tobiasdiez:
@nathanlesage how would you like to go forward with this PR?

nathanlesage:
Ahh, I totally forgot about this PR :S I'm quite busy until the beginning of June, but then I'll take a few days just for Zettlr, so if we can keep it sitting here until then, I could have another look. In general I think your approach is very valuable; but I'm just a slowpoke recently!

Issue #2245, opened 2021-07-30
[Bug] [2.0.0 beta] Recent documents not working anymore


kyaso:
### Description

When clicking on "File" -> "Recent documents", nothing is shown.

In 1.8.9 and older, any file that you opened either (I believe) through internal links, search results, or manually would show in the list.

Now, in the beta, it stays empty.

### Reproducing

1. Open a few files through, e.g., internal links
2. Open the "Recent documents" menu -> should show "Empty"

### Zettlr Version

Beta (if applicable)

### Specify version

2.0.0

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Manjaro

### Additional Information

Might be a bit unrelated to this issue, but still worth mentioning: in version 1.8.9, files you opened through clicking on their tab were not included (try by opening some tabs, then restarting Zettlr -> now when cycling through the tabs, the "Recent documents" menu will remain empty).

nathanlesage:
What I changed in between is that now Zettlr completely relies upon the recent document management of the operating system, i.e. it remains empty if the operating system does not support recent documents.

I just saw that this API apparently is only implemented for macOS and Windows. I can see to re-adding this for Linux operating systems.

nathanlesage:
Re-implemented the recent documents functionality. However, please make sure to test this since I introduced a few improvements in this regard and made it a little bit sleeker, but I cannot test it (since on macOS+Windows we're using the built-in OS support for recent documents and I don't have a Linux machine available).

nathanlesage:
@sith-on-mars Btw when you click any of the files – is something showing up in the logs, like a potential error or so? Would you mind sharing those?

sith-on-mars:
Of course not! Here you go:
```
[18:10:43]	[Application] Opening file [file path].md
[18:10:48]	[Config Provider] Writing configuration file to /home/mydesktop/.config/Zettlr/config.json...
[18:10:49]	[Application] Received a request to run command file-open, but it's not registered.
```

nathanlesage:
> Received a request to run command file-open, but it's not registered.

Whoops. Wrong command name 🙈

kyaso:
Somehow I completely forgot about this. I haven't used the recent documents feature in a long time, but when I tried it recently, clicking an item still doesn't do anything for me.

kyaso:
[...]

PR #3090, opened 2022-01-26
Add missing menu ID

kyaso:
## Description

Fix for #2245.

## Changes

Added missing ID field. ID contains file name.

## Additional information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
   count in preferences
 - When clicking on a heading in the ToC, the cursor is now set to that heading
   and the editor is focused immediately
+- Fix recent documents items not being clickable in Linux
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


kyaso:
Ah, before we merge that, I should also add a Changelog entry. I will do that today.

nathanlesage (changes requested):
Looks good, and since it's in main the click-handler should run, correct? If so, then go ahead!

But we shouldn't use the `path.basename`, because that won't be unique for two times the same filename. Two options: Either use the full path (should be fine since it's just a string, but better test it out, sometimes Electron breaks in the weirdest places) or the almost antique hash!

nathanlesage (approved):
Looks good! I presume this also works as expected when you click items, right? Then we can go ahead and merge!

kyaso:
> I presume this also works as expected when you click items, right?

When I tested it, clicking on a recent document opened that file as expected. Is that what you mean?

nathanlesage:
Yessssss!

PR #3099, opened 2022-01-29
Add support for custom latex commands and autocompletion

tobiasdiez:
## Additional information


nathanlesage:
Clever idea to leverage the snippets autocomplete for doing this! But this again prompts me to outsource the autocompletes to different plugins, since they all leverage the same API, but need different hooks especially on autocomplete insertion …

However, that being said, I think it looks almost perfect, I don't see all too many issues, but I'll let you finish up this PR before doing a proper review, except you want me to look that over?

tobiasdiez:
So, it's working for me now. Ready for review.

I agree that the auto-completion code could be refactored a bit. For example, extracting common functionality/structure to a `AutocompletionProvider` interface, and then having the current providers as instances of this interface instead of huge if-else-constructions.

tobiasdiez:
@nathanlesage do you still have remarks about the code? I've been using this branch for a couple of days now it works quite well.

source/app/service-providers/assets-provider.ts
```diff
@@ -77,6 +80,13 @@ export default class AssetsProvider {
         return await this.listSnippets()
       } else if (command === 'rename-snippet') {
         return await this.renameSnippet(payload.name, payload.newName)
+      } else if (command === 'getLatexCommands') {
+        try {
+          await fs.access(this._latexCommandsFile, fsConstants.F_OK)
```
nathanlesage, line 85: Up until now I've been using `fs.lstat` to check. Is there any benefit in using `access` rather than `lstat`? If so, we may want to switch the rest of the codebase gradually :D

source/app/service-providers/latex-commands.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface LatexCommand {
+  name: string
+  displayText: string
+  snippet: string
+  description?: string
+  definition?: string
+}
+
+function convert (command: {
```
nathanlesage, line 12: Do I read it correctly that this basically just makes sure that the commands are usable? If so, `convert` may be a tad misleading …?

source/common/modules/markdown-editor/hooks/autocomplete.ts
```diff
@@ -324,6 +327,21 @@ function shouldBeginAutocomplete (cm: CodeMirror.Editor, changeObj: any): keyof
     return 'snippets'
   }
 
+  // Can we begin autocompleting a latex command?
+  if (changeObj.text[0] === '\\') {
+    // Check if we are in an equation
+    const lines: LineInfo[] = []
+    for (let i = 0; i < cm.lineCount(); i++) {
```
nathanlesage, line 334: Are there any performance penalties when iterating over long documents in this regard?

source/common/modules/markdown-editor/hooks/autocomplete.ts
```diff
@@ -522,51 +540,9 @@ function hintFunction (cm: CodeMirror.Editor, opt: CodeMirror.ShowHintOptions):
...
-      // through them.
-      currentTabStops = getTabMarkers(cm, from, to)
-
-      // Now activate our special snippets keymap which will ensure the user can
-      // cycle through all placeholders which we have identified.
-      cm.addKeyMap(snippetsKeymap)
-
-      // Plus, move to the first tabstop already so the user can start immediately.
-      nextTabstop(cm)
+      insertSnippet(completion.text, cm)
```
nathanlesage, line 543: Maybe collapse into a single else if-statement using `['latexCommands', 'snippets'].includes(currentDatabase)`?

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -15,32 +15,29 @@
 import { getBlockMathRE, getInlineMathRenderRE } from '@common/regular-expressions'
 import CodeMirror from 'codemirror'
 import katex from 'katex'
-
+import { IpcRenderer } from 'electron'
 import 'katex/contrib/mhchem' // modify katex module
+import { LatexCommand } from '@providers/latex-commands'
+
+const ipcRenderer: IpcRenderer = (window as any).ipc
```
nathanlesage, line 22: These calls have been properly typed in 2.2.0, so we can get rid of a lot of the fuzz!

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -15,32 +15,29 @@
...
 
 const multilineMathRE = getBlockMathRE()
 const commands = (CodeMirror.commands as any)
-commands.markdownRenderMath = function (cm: CodeMirror.Editor) {
+commands.markdownRenderMath = async function (cm: CodeMirror.Editor) {
```
nathanlesage, line 26: Please no async commands. This way we have no way of catching errors since CM doesn't handle these. Instead, it's better to outsource async code to an external function that you could call from time to time and which updates a variable instead.

source/common/modules/markdown-editor/plugins/render-math.ts
```diff
@@ -77,7 +74,21 @@ commands.markdownRenderMath = function (cm: CodeMirror.Editor) {
...
+      .reduce<Record<string, string>>((map, command) => {
+      map['\\' + command.name] = command.definition ?? ''
+      return map
+    }, {})
+
+    // Render the equation
+    katex.render(myMarker.eq, mathSpan, {
+      throwOnError: false,
+      displayMode: myMarker.displayMode,
+      macros: macros
```
nathanlesage, line 90: Yes, looking at this code, you definitely should not be building this list here. Rather, you could hook to `window.ipc.on` and update some module-global macros variable whenever there is an update.

nathanlesage (changes requested):
Sorry for forgetting about the PR! The last weeks are heavy with course work and conference application deadlines … I've found a few things, but in general it looks good!

source/common/modules/markdown-editor/plugins/render-math.ts
tobiasdiez: Okay, but how do I then initialize the module variable (async calls on the module level are not yet supported, or are they?)

source/app/service-providers/latex-commands.ts
tobiasdiez: It's "converting" the json representation to the `LatexCommands` object (which happen to be almost the same thing). Do you have a suggestion for a better name?

source/app/service-providers/assets-provider.ts
tobiasdiez: Shouldn't matter much, at least after some quick googling. https://stackoverflow.com/questions/54514627/nodejs-fs-stat-or-fs-access-to-check-if-a-folder-exists has some discussion, and access seems to be preferred and is maybe a bit faster; but both should be fine.

source/common/modules/markdown-editor/hooks/autocomplete.ts
nathanlesage: > Maybe in the future it would make sense to have one markdown parser, that takes care of all parsing (probably using a state-dependent algorithm instead of regex) and keeps an up-to-date representation of the markdown code as a parsed object. Then all the editor commands and syntax highlighting etc could query this central information without the need to reparse the source code.

Sounds like a clever idea. Maybe we can bodge the CodeMirror 6 Lezer for that

source/common/modules/markdown-editor/hooks/autocomplete.ts
```diff
@@ -473,6 +473,10 @@ function hintFunction (cm: CodeMirror.Editor, opt: CodeMirror.ShowHintOptions):
     if (autocompleteStart === null) {
       throw new Error('Could not autocomplete: autocompleteStart was null')
     }
+    if (currentDatabase == null) {
+      // We don't know where to get completion data from, so we can't do anything.
```
nathanlesage, line 477: Here we should throw an error analogous to above because this really shouldn't happen and would indicate a bug

Issue #3123, opened 2022-01-31
[Bug] Auto-close brackets option not working


gerotakke:
### Description

In settings->editor->auto-close brackets (“Schließe Zeichenpaare automatisch” in the German localization that I'm using), the option doesn't actually do anything.
I'm pretty sure that I've tracked it down to the option just missing in `source/win-main/MainEditor.vue`.
I got it working by adding this line in L221:
`        autoCloseBrackets: this.$store.state.config['editor.autoCloseBrackets'],`
I also didn't quite figure out if `source/common/vue/CodeEditor.vue` is actually used; it seems to me like it isn't.

### Reproducing

Write a single bracket `(`, Zettlr will automatically add the closing bracket regardless of the value of the configuration option.

### Zettlr Version

Right off the develop branch

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [X] ARM 64bit

### Operating System Version

macOS Monterey 12.2

nathanlesage:
> I also didn't quite figure out if source/common/vue/CodeEditor.vue is actually used; it seems to me like it isn't.

Yes, that is the "smaller" code editor. I would argue that auto-closing brackets in these should be active always. And yes, your bug fix is indeed what is necessary. Feel free to open a PR, if you like to, or have me add that line!

PR #3125, opened 2022-01-31
#3123 fix auto-close brackets in main editor

gerotakke:
Issue #3123; fix auto-close brackets in main editor


nathanlesage:
Thanks! :)

PR #3133, opened 2022-02-01
[DEVELOP] Service Provider Migration

### LINKED ISSUE #3063: [DEVELOP] Migrate main modules to service providers
I'm opening this issue to have a communication channel with the rest of the community regarding a large refactoring I've begun.

Until now, there was this distinction between `modules` and `service providers`, where modules were roughly equivalent to simple Javascript modules (i.e. you have a folder with an index and in there resides the module) and service providers were modules which did not expose any access to other parts of the code. The only code which had direct access to the service providers is the lifecycle hook which boots them up and shuts them down. Access to the internal state of these modules was always facilitated using the `global` object (inside the main process) and the `broadcastIpcMessage` function (towards the renderer processes).

While that was a cool thing, the main reason I used the `global` object was both because module support in an early version of Node was not that good and simply because I honestly didn't understand how modules work.

Over the past two years, however, the whole setup of the app has changed quite considerably. That has had several reasons:

* First, I should name the efforts by the Electron team to mitigate security issues with Electron apps, among these deactivation of node integration, context isolation, and other features which were slowly included. This means that the renderers are now fully autonomous and we could – if we provide the correct APIs – just take some of the renderer process and display it as-is on a website.
* Then, @tobiasdiez has introduced TypeScript to the codebase, and, alongside with that, webpack, electron-forge and thus a host of new possibilities. Ever since his large PR was merged, the files are now combined, minimized and otherwise optimized.
* Furthermore, I lost my fear that something bad could happen in the main process. Or, rather: I now know how to keep code safe without locking everything up.

Additionally to these reasons, there are now new demands for the application: Progress is halting, mainly because code that I would like to implement doesn't have access to some of the codebase which is necessary to implement this simple. As a very palpable example, take the demand for a network visualization of the files: While we now do have a link provider, that link provider is currently unable to resolve the outlinks into correct path names. The functionality for that, however, is already there: the `findExact` method of the FSAL. However, since that is only accessible by the main Zettlr object, the link provider has no means to resolve those links.

In fact, the main Zettlr class is the last remnant from the early versions prior to 1.0 where I organized everything OOP, an hommage to my background with Laravel and PHP in general, where OOP is much more prevalent than in the JavaScript world.

Given that background, here's what I am currently trying to do:

1. Refactor the service providers from classes to functional components which maintain a module-wide state and expose functions which allow limited access to that state in a safe manner. This will not only remove our reliance on the global object, but additionally make typing much easier, since we can simply type the functions and thus remove a lot of boilerplate from the type files (which I already migrated into a new `types` directory)
2. Convert the FSAL, the DocumentManager, and the Importer and Exporter to service providers, that is: Functional modules residing on the top level of the app. This will then additionally end the distinction between service providers and modules: The main process will then only have service providers. The renderers will still have modules, but that is because there are multiple of them, and for example the Markdown editor is something that's needed in several places
3. After this has been done, one of my next goals would be if I can safely remove the `zettlr.ts` class file. I don't yet know how to do this the most proper way, but I'm sure after there is no more state which the class manages a few avenues will open up.

The benefits of this whole endeavor will be that the complexity of the main process is radically reduced and we can finally begin to implement a few other neat things such as the typing of IPC modules which Tobias has already begun working on. Furthermore, this will make it easier for us to slowly migrate the whole preload module towards a way that offers a simple API. This will in turn make it easy to – some day – take the main window code, move it to an external repository, and from there on build a web application (PWA) of Zettlr, which utilizes the code in two places: The desktop app and said web app.

***

In the end, I think this is simply the logical conclusion of going from "I have no idea what I'm doing" towards someone who actually knows what a desktop app needs and how to properly structure it. All in all, it's just another technical debt in the wall!

CC because you seem to be reasonbly active in making PRs and/or thus might be interested to know about this: @kyaso @derwok @Kangie @newhinton

nathanlesage:
## Description

This is the PR I'll continuisly keep updated in order to close #3063

## Changes

This PR has one job: To get rid of the `globals` and side-effects pertaining to these. In order to do these, the following changes have been made (checkboxes are ticked):

- [x] Create an `AppServiceContainer` which serves as a singleton class that can inject service providers into other service providers
- [x] Move all provider instantiations into the app service container which is launched and shut down just like the service providers before in the lifecycle functions
- [x] Inject providers on a need-be basis to the providers, that is: all receive an instance of the logger, the rest depending on their needs
- [x] Refactored the `WindowManager` into the `WindowProvider`
- [x] Fixed a race condition in which the window provider and the tray provider required each other simultaneously
- [x] Moved the loading of the language data into the translation provider, where it should've been for years now
- [x] Refactor the `DocumentManager` into the `DocumentProvider`
- [x] Refactor the `Zettlr` class into a `CommandProvider`
- [x] Completely remove all `globals` in the main process
- [x] Clean up the `global.d.ts` and remove any provider typings as possible, since the types are now propagated correctly via dependency injection
- [x] Clean the absurd dependency hell that has just unearthed in the FSAL
- [ ] Find a solution to providing the language data via the `globals`, since that is something I have not yet figured out without having every file require the translation provider (or is that feasible?)
- [x] The md2html converter depends on the citation provider, we need to inject that both in main and in the renderer

## Additional information

I did some quick testing and thus far the app starts without all too much hassle. However, I still don't trust the linter via ESLint vis-a-vis VSCode's linting vis-a-vis the typechecking during builds, since they give me mixed messages. Feel free to have additional looks at this.

Tested on: macOS 12 (Monterey), arm64 arch, but this PR should be platform independent

***

CC @tobiasdiez @kyaso — this PR is not yet done so you don't have to do any deep code review just now. Just feel free to keep an eye out on this PR, and feel free to suggest improvements if you see I'm running into the wrong direction somewhere.


source/app/service-providers/fsal/fsal-directory.ts
```diff
@@ -181,7 +190,21 @@ async function parseSettings (dir: DirDescriptor): Promise<void> {
  *
  * @return  {Promise<DirDescriptor>}           Resolves with the directory descriptor
  */
-export async function parse (currentPath: string, cache: FSALCache, parent: DirDescriptor|null): Promise<DirDescriptor> {
+export async function parse (
```
tobiasdiez, line 188: That's pretty heavy. Can some of these properties be reused (say in a `context` argument) among the methods? But even for a single method something like `parse("path", context: { currentPath, cache, ... })` might be more readable.

nathanlesage: That's what I'm currently doing. I figured out already a lot of different ways to remove this dependency propagation deep into the FSAL. I'm not yet done doing this, but I'm getting there. It turns out, removing the side effects on the globals opens up a lot of avenues for better code.

I want to avoid `context` objects for as long as possible, because that hides a lot of the dependencies again.

nathanlesage: > Then you only need to pass the dependencies once in the constructor, and not for every function call.

I thought about that, but I figured that, by using closures instead, we will end up having a little more wiggle room for potential customizations, since we can always pass those functions which we actually want instead of calling a function on these objects. But I mean we can definitely look at how it looks after we're done.

What would be cool is if you could have a look at the appServiceContainer if that's feasible as I did that right now, or if I should do something else?

nathanlesage:
Soooooooo, I think I'm getting somewhere with this PR. I'd say it's time for a first review, if you're not opposed! What is still missing is a lot of documentation, since I didn't bother to rewrite the documentation of functions that I've moved around. I definitely have to keep on this.

That being said, @tobiasdiez and @kyaso I would be very happy if you – at some point – can facilitate having a look over it.

Just so that it's faster for you here's what I changed:

* Luckily, many of the files that you will see have been "renamed without changes" or completely deleted, so these are all good!
* Moved the provider boots and shutdowns from `lifecycle.ts` to their own `app-service-container` that is literally doing the same. The lifecycle now just boots and shuts down that service container.
* The providers now extend `ProviderContract` and no longer an EventEmitter. Those providers that in fact needed an EventEmitter now have an `_emitter` property that does the same job.
* The service providers now hold a reference to whichever other providers they need, and they get that one via the service container during boot
* All service providers now uniformally feature boot and shutdown methods (including the new "service" providers FSAL, WindowManager and DocumentManager)
* Sometimes I was calling async code in the constructors. I strove to move that code into the asynchronous boot method wherever possible. Please add a comment where I forgot that.
* I moved all providers into a directory and renamed the file to `index.ts`. That has two reasons: first, some providers (especially the new ones) have a few more files since they would be too big otherwise. Plus, it looks nicer when referencing them as it removes the duplication of `provider` in the import statement (e.g. `@providers/appearance` instead of `@providers/appearance-provider`)
* Moved corresponding files from various places (`assets` in the service provider folder, `util` folders, etc.) closer towards the only service provider that's using them, if applicable
* Created a new `CommandProvider` that basically fulfills the `runCommand` functionality.
* The commands are the **only** instance that receives the complete service provider. The reason is that, since they form the middle layer between user space and service provider/system space, they should have an easier time accessing all providers. It felt as if it didn't make sense to just pass them single providers peace-meal
* Moved both the `importer` and `exporter` modules into the new `CommandProvider` since I realized that only commands accessed them. (Which makes sense since exports and imports should solely happen when the user requests so)
* CHANGES IN THE FSAL
    * The FSAL turned out to be quite a behemoth. There I had to do a lot more changes
    * First, I now provide the link provider, tags provider, and target provider to any methods that involve Markdown files, since they need access to these providers to report links/tags and retrieve their target. I think that this is fine. It's extending the function call params a little bit, but after a day looking at these it seems fine (and better than a `context` since it's quite easy to see what dependencies they have). There is definitely room to improve, but I feel we can also do this later, since I will have to get back to the complete FSAL again in the future to implement another abstraction layer on the file I/O system (replace calls to `fs` with an adapter so that we can implement other fancy shit such as WebDav, Dropbox, iCloud Drive, etc. synchronization and retrieve files from idk where without the consumers of the FSAL or even the FSAL itself noticing that)
    * Then, I got rid of the `isAttachment` function and reduced the overall boilerplate quite a lot. From now on, we have a function to check just for Markdown and Code files, and everything else will simply be loaded as an "other file". One thing I still have to do (but that's three lines of code) is to implement the filtering based on the attachment extensions in the sidebar, but I'll do that later.
    * Furthermore, I reduced a little bit of the dependency hell by using two closures, one for parsing a Markdown file, and one for sorting a directory. This way we can create those unnamed functions anew every time based on the most recent configuration parameters and simply pass them for the underlying code to call. I figured this is clever b/c this way we can even pass fully custom parser functions and/or sorter functions to the FSAL consumers in case we ever wanted to.
    * No function below the level of the FSAL entry point log anymore. Rather, I realized that I can just throw errors. (But I didn't test if I capture them all, soooooo … well, that's something I need to recruit people for)
* Also, many dependency issues with access to `global`s were reduced by simply providing a closure. That also has the benefit that the `md-to-html` helper can now be instantiated both in main and in the renderers without some weird dependency management depending on whether we can import the IPC renderer
* I moved a lot of code out of the `Zettlr` main class and into the respective service providers, which makes much more sense and makes the code cleaner. There is just a small skeleton left now, and then wen can get rid completely of the `main` folder!
* I finally figured out how to fully type the Window object. The internet was, again, wrong on something. Now `ipc`, `clipboard`, `path`, and `config` are fully typed.
* I also cleaned up the preloader while I was at it
* Also, no more globals in the renderer process, everything is now handled via the window object
* Furthermore, since we didn't need to provide ambient module declarations on the global object anymore, I finally got rid of that unholy ambient hell I created earlier. Now, all types are the classes itself (since these are singletons either way). The only ambient module declarations are those we *absolutely* need, that is: Declarations for file types, for the entry points of the BrowserWindow instances, and a few modules that don't ship their types

***

I figure that it's quite a lot, but it would be great to have two other people looking over it, because at some point a lot of the work was copy and paste, so I will have forgotten something.

Feel free to start the branch using `yarn test-gui`, it should run all good now. I already tested around quite a bit, but didn't stumble upon super bad issues, but feel free to move files around, create, delete, rename, etc. all that stuff in the testing environment.

Thanks already for looking over it, and I'll defo get you a beer when Covid is better! 🍻

source/app/service-providers/provider-contract.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+export default abstract class ProviderContract {
+  /**
+   * Service providers implement a boot method which must be called after
+   * instantiation of the singleton.
+   *
+   * @return  {Promise<void>}  boot must return a Promise.
+   */
+  public abstract boot (): Promise<void>
```
tobiasdiez, line 23: Maybe provide a default implementation for the boot and shutdown methods, since some of the providers don't need them and then you don't need to implement them.

source/app/service-providers/fsal/index.ts
```diff
@@ -140,6 +140,40 @@ export default class FSAL extends ProviderContract {
       this._logger.info('Clearing the FSAL cache ...')
       this.clearCache()
     }
+
+    // Start a timer to measure how long the roots take to load.
+    const start = Date.now()
```
tobiasdiez, line 146: Date.now is not exact enough for perf measurements. Something like https://nodejs.org/api/perf_hooks.html#performancenow is preferred (not sure though if it works in electron)

source/app/app-service-container.ts
```diff
@@ -0,0 +1,182 @@
...
+  private readonly _targetProvider: TargetProvider
+  private readonly _translationProvider: TranslationProvider
+  private readonly _trayProvider: TrayProvider
+  private readonly _updateProvider: UpdateProvider
+  private readonly _windowProvider: WindowProvider
+  private readonly _fsal: FSAL
+  private readonly _documentManager: DocumentManager
+
+  constructor () {
+    // NOTE: The order in which the providers are instantiated still matters!
```
tobiasdiez, line 64: Why is this?

source/app/service-providers/appearance/index.ts
```diff
@@ -13,39 +13,45 @@
...
   private _endHour: number
   private _endMin: number
+
+  private readonly _logger: LogProvider
+  private readonly _config: ConfigProvider
   /**
    * Create the instance on program start and initially load the settings.
    */
-  constructor () {
+  constructor (logger: LogProvider, config: ConfigProvider) {
```
tobiasdiez, line 43: Or shorter:
```suggestion
  constructor (private readonly _logger: LogProvider, private readonly _config: ConfigProvider) {
```
without the explicit class attributes `_logger` and `_config`.

source/app/service-providers/citeproc/index.ts
```diff
@@ -95,11 +99,22 @@ export default class CiteprocProvider {
...
+  private readonly _windows: WindowProvider
+
+  constructor (logger: LogProvider, config: ConfigProvider, notifications: NotificationProvider, windows: WindowProvider) {
+    super()
+    this._logger = logger
+    this._config = config
+    this._notifications = notifications
+    this._windows = windows
+
+    this._logger.verbose('Citeproc provider booting up ...')
```
tobiasdiez, line 114: What is the distinction between the constructor and the boot methods?

tobiasdiez (commented):
I really like it! Good job 🚀 

I kept following along and now skimmed over the changes. I couldn't really find anything of substance, only a few smaller remarks thus.

source/app/util/environment-check.ts
```diff
@@ -20,7 +20,7 @@ import isTraySupported from './is-tray-supported'
 import commandExists from 'command-exists'
 
 export default async function environmentCheck (): Promise<void> {
-  global.log.info('Performing environment check ...')
+  console.log('Performing environment check ...')
```
kyaso, line 23: Sorry for my ignorance, but why it is logged to the console in this file, instead of to the log provider?

source/common/modules/markdown-editor/index.ts
```diff
@@ -94,6 +93,8 @@ export default class MarkdownEditor extends EventEmitter {
     this._cmOptions = getCodeMirrorDefaultOptions(this)
     this._countChars = false
 
+    this._md2html = getConverter((window as any).getCitation)
```
kyaso, line 96: Would `window.getCitation` also work?

source/common/modules/markdown-editor/table-editor/table-editor.ts
```diff
@@ -137,6 +139,8 @@ export default class TableEditor {
     this._colAlignment = alignments
     this._edgeButtonSize = 30 // Size in pixels
 
+    this._md2html = getConverter((window as any).getCitation)
```
kyaso, line 142: (dito, in case my comment above is valid; otherwise, ignore this)

kyaso (commented):
Same as @tobiasdiez, I didn't find anything major so far!

Just some minor comments.

source/app/util/environment-check.ts
kyaso: Ah wait, it's because the environment check is run very early before the log provider is even initialized, right?

source/common/modules/markdown-editor/index.ts
nathanlesage: Hah, yes! That's the last API function I haven't yet typed! Thanks for finding it! 😍

source/app/util/environment-check.ts
nathanlesage: Yes but I agree it's not good style! But I may have an idea of how to solve that …

source/app/service-providers/provider-contract.ts
nathanlesage: Good idea

source/app/service-providers/fsal/index.ts
nathanlesage: It's not really a performance measure. It should rather serve as a rule of thumb indication which basically only I use during development to have a coarse measure of how well the FSAL does (so I see it in case I accidentally break something)

source/app/service-providers/appearance/index.ts
nathanlesage: Nice find, thanks!

source/app/service-providers/citeproc/index.ts
nathanlesage: In general, I thought of that like so: The constructors should make sure that the objects can be accessed, but any "impure" code (i.e. disk access etc.) is async and therefore should be part of the boot methods. This also has the benefit that most stuff that's running doesn't run into race conditions anymore.

source/app/app-service-container.ts
nathanlesage: Because there are still some race conditions, but I'll have an additional look now that I have the knowledge that the PR is mostly good to think about those race conditions.

source/app/service-providers/fsal/index.ts
nathanlesage: Has worked, so why not, we're now using the `performance` hook!

source/app/app-service-container.ts
nathanlesage: I've now completely removed any async code from the constructors. I still think we should instantiate the log provider and config provider first, since they can be queried first by other providers, but the rest should not really matter anymore.

source/app/util/environment-check.ts
nathanlesage: Amazing how easy it can be once you remove any unsafe code from the constructors!

nathanlesage:
Alright folks — thank you both for having a look at this PR! I've now incorporated everything and fixed two minor other things, and now we can test this out a little bit and if you can't spot any logical errors that I have introduced during this factoring, we can merge! :)

nathanlesage:
Had to revert one last commit, we need to instantiate the config synchronously before all other providers, because so many providers require it during instantiation … but now it works!

nathanlesage:
I just tested most of the functionality, and couldn't find any new bugs. How about you @kyaso @tobiasdiez ?

If you didn't notice anything, I'm going to merge!

kyaso:
I just tested it, and I saw that when creating a new file (by right-clicking in the file tree in the expanded sidebar mode), the default name in the input field was `(\d{14}).md`. Somehow the ID regex is not resolved?

Apart from that, looks good so far!

nathanlesage:
> Somehow the ID regex is not resolved?

Uhh, good catch! Gonna quickly look that over

nathanlesage:
Alright, I'm going to merge now — if something else comes up this is rather something minor I would say and we can fix this during beta :)

PR #3149, opened 2022-02-07
Remove title/tag optimization in global search

### LINKED ISSUE #3111: [Bug] Global search: When search term is in file name or as a tag inside the file, the search result entry in the sidebar is incomplete
### Description

When you do a full-text search for the term, say `honey`, files that either contain "honey" in their file name, or contain a tag "#honey" are found, but the search results don't list any lines of that file that contain "honey".

The only "result" that's shown is something in the form:

```
-1: <file-name>.md
```

### Reproducing

1. Create a file named, e.g. "honey-bee.md"
2. Write some text and make sure to add a couple of lines containing "honey"
3. Search globally for "honey" -> "honey-bee.md" should be found
4. Observe that the lines containing "honey" are missing
5. Rename the file to "bee.md" (i.e. remove "honey")
6. Search again for "honey" -> This time the lines are found

The above steps can be applied to a file containing a tag "#honey" as well (file name does not matter in this case). For example, simply add the tag "#honey" to the `bee.md` from above, and search again for "honey".

### Zettlr Version

Stable (most recent version)

### Specify version

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Linux manjaro 5.14.21-2

### Additional Information

The issue is also present in 2.1.2, and after searching a bit in past issues, I found this comment: https://github.com/Zettlr/Zettlr/issues/874#issuecomment-744025291

Is that maybe still the reason?

kyaso:
Closes #3111

## Description
Previously, the global search (aka full-text search) would stop searching through a file when the search term(s) occurred in the filename and/or tags. The corresponding search results in the sidebar would then only contain the filename, and any other occurrence of the search term(s) within that file wouldn't be shown.

With this PR, all files will be searched through, regardless of whether the title/tags matched or not.

## Changes

The `searchFile()` function in `search-file.ts` formerly would simply return a single special `SearchResult` object when the title/tag matched. Now, that special object is instead pushed to the front of the final search result array, and the other matches follow.

## Additional information
The search may take slightly longer in certain circumstances. I tested this on my system with ~900 notes and the search time was IMO still in the tolerable region.


nathanlesage:
Yes, looks good!

One thing, since you're at it: Would you mind improving the `-1` display in the global search as well? Something like `if line === -1 ""` or something like that, because people have complained about seeing that `-1` in front of the filename there, and maybe just not displaying that might solve it. (sorry that I'm dumping this on you, but … there was an opportunity! :D)

kyaso:
Sure!

nathanlesage:
Perfect thanks! I would like to merge but somehow the Changelog doesn't like that and I'm on mobile rn :(

kyaso:
> I would like to merge but somehow the Changelog doesn't like that

Fixed!

kyaso:
Wait, I found another issue.

kyaso:
Sooo, I created a test file named "negative.md" that just contained "foo bar". When searching for "negative" that file was _not_ found. I checked on the stable 2.1.3, and there it was found. Then I checked out to some commit prior to all the commits of this PR, and oddly, the file wasn't found.

I'm currently bisecting to see what happened.

nathanlesage:
Oh, alright -- easy, take your time! Thanks for making sure the PR works as expected :)

kyaso:
Found it :relieved:

source/app/service-providers/fsal/util/search-file.ts
```diff
@@ -167,7 +167,9 @@ export default function searchFile (fileObject: MDFileDescriptor|CodeFileDescrip
   // Now immediately check if all required terms have matched. If not, we can
   // disregard this file now and save some computing time below.
   if (termsMatched !== termsToSearch.length) {
-    return []
```
nathanlesage, line 166: Did you make sure that NOT-operators disqualify a file? We have to make sure that we don't return *too many* results! That was the reasoning behind the fact that I returned an empty array here, and I'm not sure returning `finalResults` has the effects you think …?

nathanlesage: Yes, I didn't completely remember the code, just wanted to make sure that didn't get lost :)

source/app/service-providers/fsal/util/search-file.ts
```diff
@@ -163,7 +167,9 @@ export default function searchFile (fileObject: MDFileDescriptor|CodeFileDescrip
...
+    // Make sure we return the finalResults array instead of an empty array, so
+    // we don't lose the file in case only its title has matched.
+    return finalResults
```
nathanlesage, line 172: You know what I just realized? Can't we just skip this condition in total? It was solely the shortcut, and this way if not all terms matched in the title, the file contents would still not be searched, or did I misread my own logic here …?

nathanlesage (commented):
Found something new 🙈

source/app/service-providers/fsal/util/search-file.ts
kyaso: I think it should be possible. If I understood it correctly, when this condition is met, we didn't find anything in the file anyway, so the following post-processing would be skipped.

However, I'd suggest we keep this as an optimization. Because we still have the `sort()` in line 182, and the `for` loop following it. Maybe my understanding of Javascript is lacking here, but these two would still consume some amount of processing power (even though not productive), right? Then having the shortcut, that is just a single comparison, might still serve as an optimization?

nathanlesage (approved):
Looks good!

Issue #3200, opened 2022-02-14
[Bug] Emphasis Rendering breaks Underscores in Inline Code


kyaso:
### Description

Assume we have the following code: `foo_bar_`. With the new emphasis rendering it will be rendered as follows:

![grafik](https://user-images.githubusercontent.com/18646029/153944941-cb3af5fd-a2d3-4840-bf64-b6fc1f9d8e07.png)

### Reproducing

1. Enable emphasis rendering in preferences
2. Write some inline code that contains two underscores, e.g.: `foo_bar_`
3. Click somewhere else so the rendering gets applied

-> The code part between the underscores is rendered italic

### Zettlr Version

Stable (most recent version)

### Specify version

2.2.1

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Linux manjaro 5.14.21-2

kyaso:
I also found that this happens to underscores which are inside links:

https://user-images.githubusercontent.com/18646029/153946806-e3f3966f-dd11-4c13-b2cb-4feffc60ce24.mp4

Notice how the link is not rendered when the emphasis is active; only when you click the italic part (so the "edit" mode is entered), and you _then_ click outside, the link is finally rendered.

nathanlesage:
Yeah, the new emphasis rendering has quite a lot of edge cases where it's difficult to decide what the plugin should do. Maybe a good idea could be to first collect a few of these cases and decide what should happen in each of these instances and then do one clean up PR to fix these. What do you think?

kyaso:
> Maybe a good idea could be to first collect a few of these cases and decide what should happen in each of these instances and then do one clean up PR to fix these. What do you think?

Sounds like a plan!

Redsandro:
> \[Let's] collect a few of these cases and decide what should happen

There are two problems here. 

**Problem one, the easy part.** The emphasis renderer is breaking the markdown convention of _requiring a space_ before and after the marked phrase. For example, the middle part should not be in italics:

> test_italics_test

But in Zettlr, it is:

> ![image](https://user-images.githubusercontent.com/1702193/154372309-65724d86-a5fa-4f20-8761-c72941b8727c.png)

Fixing this would fix some but not all problems. At least the reported problems, so arguably the most prevalent one. I will create a PR for this.

[...]

PR #3226, opened 2022-02-16
don't render emphasis without whitespace boundary; touches #3200

Redsandro:
## Description

The emphasis renderer is breaking the markdown convention of _requiring a space_ before and after the marked phrase. For example, the middle part should not be in italics:

> test_italics_test

But in Zettlr, it is:

> image

This PR fixes this.

## Changes

Add non-capturing whitespace lookahead and lookbehind.

## Additional information

Tested on: Linux Mint 20

### DIFF
--- a/source/common/modules/markdown-editor/plugins/render-emphasis.ts
+++ b/source/common/modules/markdown-editor/plugins/render-emphasis.ts
@@ -14,7 +14,7 @@
 
 import CodeMirror, { commands } from 'codemirror'
 
-const emphasisRE = /(?<![\\])(?<=\s)([*_]{1,3}|~{2})((?!\s)[^*_]+?(?<![\s\\]))(?:[*_]{1,3}|~{2})(?=\s)/g
+const emphasisRE = /(?<![\\])(?<=\s|^)([*_]{1,3}|~{2})((?!\s)[^*_]+?(?<![\s\\]))(?:[*_]{1,3}|~{2})(?=\s|$)/g
 
 /**
  * Declare the markdownRenderEmphasis command
--- a/source/common/modules/markdown-editor/plugins/render-emphasis.ts
+++ b/source/common/modules/markdown-editor/plugins/render-emphasis.ts
@@ -14,7 +14,7 @@
 
 import CodeMirror, { commands } from 'codemirror'
 
-const emphasisRE = /(?<![\\])([*_]{1,3}|~{2})((?!\s)[^*_]+?(?<![\s\\]))(?:[*_]{1,3}|~{2})/g
+const emphasisRE = /(?<![\\])(?<=\s)([*_]{1,3}|~{2})((?!\s)[^*_]+?(?<![\s\\]))(?:[*_]{1,3}|~{2})(?=\s)/g
 
 /**
  * Declare the markdownRenderEmphasis command


Redsandro:
## Wait, wait, do not merge!

I've got a problem with newlines not matching.

Redsandro:
## Okay, you can merge now.

I believe all lines are single lines, right? So `^$` should cover it. Amirite?

nathanlesage:
image

😄 😄 😄 

One thing, though: There are some instances where emphasis should be rendered even within a word. Rather than trying to monkey patch the plugin, it might be better to [implement the CommonMark specification](https://spec.commonmark.org/0.30/#emphasis-and-strong-emphasis) here.

Redsandro:
Shoot. You are right about those instances. I did not know that it is allowed to be used like that, even though sometimes I wished it did. :sweat_smile: 

After reading the CommonMark doc a couple of times, I can't figure out a regex that does conditional lookaround that works in all these cases. :unamused: Do you have an idea on how to tackle this? I love this feature and I'd hate to see this issue persist for many moons.

What I can say is this. The broken render behavior this PR fixes is frequently occurring, and the nested edge cases that won't render properly are rare ones. The CommonMark documentation itself mentions: _This is enough for most users, but [doesn't account for nested emphasis.]_

So while it might be better to implement the CM spec at some point indeed, I propose it might be good to merge this initially anyway. Consider that the current implementation _renders where it should not,_ and this PR _does not always render where it should._ The former is a bugged/broken implementation, and the latter is an incomplete implementation, which is arguably better than a bugged bugged one.

nathanlesage:
> I can't figure out a regex that does conditional lookaround that works in all these cases. 😒 Do you have an idea on how to tackle this?

Sometimes, ugly if/else conditions are inevitable (see, for example, how I handle braces in links in the [renderLinks plugin](https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/plugins/render-links.ts#L90-L114)) :/

> Consider that the current implementation renders where it should not, and this PR does not always render where it should. The former is a bugged/broken implementation, and the latter is an incomplete implementation, which is arguably better than a bugged bugged one.

Unfortunately, the latter is also seen as "buggy" behavior, since most people cannot know whether some behavior is known or buggy.

But I see that you do have a point. @kyaso @tobiasdiez what do you think? Merge it and next PR, or wait?

kyaso:
> Rather than trying to monkey patch the plugin, it might be better to [implement the CommonMark specification](https://spec.commonmark.org/0.30/#emphasis-and-strong-emphasis) here.

In the long run, we should try to achieve this, I agree.

nathanlesage:
Alright, then let me quickly merge this, but please, let's collectively not forget to get back at the emphasis rendering :D

Redsandro:
> let's collectively not forget to get back at the emphasis rendering :D

Would it help to move it to the [project](https://github.com/Zettlr/Zettlr/projects/7)/to_do?

By the way, the project may be more useful if you split **To do** into three columns, borrowing two columns from the [MoSCoW method](https://en.wikipedia.org/wiki/MoSCoW_method).

1. Backlog (to do 1/3)
2. Must have (to do 2/3)
3. Should have (to do 3/3)
4. In Progress (A manager's favorite, can be removed if not used)
5. Done

The part where the plugin broke rendering in certain use cases which forced people to turn the option off (i.e. what caused me to crawl out of my cave and do this PR) would have been a **must have**. The part that's now left could be a **should have** enhancement because it does not break existing functionality.

tobiasdiez:
I think it's already an improvement, so +1 also from my side (even though this is already merged, as I've been a bit late to the party).

One remark for the future: I think these markdown editors/highlighters would profit from some unit testing. For example, here one could extract https://github.com/Zettlr/Zettlr/pull/3226/files#diff-79f18fcae0aaa8f2fc9131fa896774539f5395c0f866f9b49bae2a427806b4c6R35-R39 into a new method (excepting the line and returning a list of matches), and then write tests for this method. In this way one could make sure that the existing functionality still works, as well as that the changes have indeed the desired effect.

nathanlesage:
> One remark for the future: I think these markdown editors/highlighters would profit from some unit testing.

This is what I'm gradually doing. Many of your suggestions plus some refactoring have helped that I can gradually create more and more unit tests; for example the ToC builder now has also its own unit test, etc.

Since there are so many different files that still lack tests, I thought it makes sense to only introduce them once I see those files. But I mean, everyone can add more unit tests.

With regard to some new projects: Agreed, while I was still completely alone in this, they seemed kind of overkill, but now that we are more and more people with write access to the repo, it makes sense in order to keep it transparent who is doing what and when which issue will be addressed.

The question is rather: What counts as a project — would it be `3.x`, or `2.3.x` or `2.3.1`, or something different altogether? I have not really an answer to this, I'm afraid. Especially since I tend to fix bugs/implement features the quickest when I feel I need them, too (not really democratic, I know)

Redsandro:
> [Initially] they seemed kind of overkill (...) I tend to fix bugs/implement features the quickest when I feel I need them

Don't we all? :wink: That still works fine. Close an issue, automation will move it to done.

A project can be helpful to keep track of things, sort through issues, even if you're alone, especially when you have a day job and may take long breaks from working on the code. So start small. But if you think the overhead outweighs the potential benefit, don't do it! I mentioned it only because I saw the project so I thought you were already actively using it!

In summary, my recommendation is to add a list or two to the existing project if it's already being used. Not to start using projects, because some people love 'em and some people hate 'em.
 
> The question is rather: What counts as a project — would it be `3.x`, or `2.3.x` or `2.3.1`, or something different altogether?

"Zettlr" counts as a project. Start small. Don't add a new one, just change the existing one. I would make one single permanent rolling versionless project. Where you can triage issues and features. Keep the things that are important or things you want to remember in dedicated (short)lists, and everything else in "backlog".

You could also add a list to the right for future versions. E.g.:

- Backlog
- Should have
- Must have
- Done
- Next (3.0)

tobiasdiez:
Sounds very good. I fully understand that it will take a while to cover most things with tests. The intend of my comment was also more to point out that its probably a good idea that people touching these files (in this case @Redsandro ;-)) should add tests themselves in their PR.

nathanlesage:
Agreed. I'll get to opening a new project cleanly, and then it would be great that everyone with write access to the repo (@tobiasdiez @kyaso — @Redsandro do you also want to take up a part of the responsibility? :D) would also help take a look at that project so that we can somehow manage it ourselves.

Also, what would really help is that people self-assign issues where they are like "Hey, I wanna do that!" because I think that also communicates a lot, and it's easier than projects (although I agree that we also should add those things which we self-assign to said project as well, just for the overview)

Redsandro:
> write access

You correct me often enough which is good but I'm not sure it would ease your mind to give me powers. :wink: 

I don't think I have a lot of time and I always end up having even less than I think, but If I can have permission to move cards and maybe labels, perhaps those moments that I do contribute can be more are helpful.

Redsandro:
> its probably a good idea that people touching these files (in this case @Redsandro ;-)) should add tests themselves in their PR.

@tobiasdiez it's definitely a good idea. Although for some reason I find tests notoriously difficult. One reason is that every project I'm involved with seems to use a different testing suite, making it a bit time consuming to learn how it works. The overhead compared to a 20 minute PR is unappealing.

It's often a lot easier to amend existing tests. So in this case, if @nathanlesage added a test when authoring this new plugin a few weeks ago, it would be easier for me to amend those tests. Also, it's easiest if no contributor needs to have prior knowledge about where to look. It's easy to mix different suites up.

So to that end, here are some random related thoughts:

1. Do we have a dedicated documentation that describes how to set up tests with Zettlr?
2. Can we link to that documentation in the default pull request comment?
3. From now on, when files are touched and tests added, can we comment in header of said file the `path/to/test`?
4. (And now, to reveal my lack of testing-knowledge: How do you write a test that is so visually telling? You can't just test matches on a text, because Zettlr does a lot of processing on top and every line is independent. That's why I initially made a newline mistake, because I tested "on the side".)

So for 2 above, I recommend the _new_ first rule of the obligatory checklist should be:

> 1. I have added or amended tests (link to test documentation)

nathanlesage:
> You correct me often enough which is good but I'm not sure it would ease your mind to give me powers. 😉
> 
> I don't think I have a lot of time and I always end up having even less than I think, but If I can have permission to move cards and maybe labels, perhaps those moments that I do contribute can be more are helpful.

Alright, no stress! I'll send out those invitations only if you want to (and I only offer that to people that have shown a genuine interest in Zettlr, so I trust those people, you included, to not do bad things :P)

***

With regard to your comments on tests: I do see your point, which is why I do not want to make tests mandatory. Right now Zettlr's setup is still extremely context-dependent, and especially with CodeMirror 5 that is still problematic since there is a lot of data-binding back and forth. Hopefully, once we have access to a proper AST with CodeMirror 6, many of these issues should remedify themselves. So for now I would suggest to include tests where possible but don't make that mandatory, simply because even if you were to spend the time to write tests, there is no guarantee as of now that there is no hidden dependency.

That being said, we should first focus on implementing this project somehow, and just take a look at each PR: If I or we see that it implements a pure function, I would say we can on a case-by-case basis demand a unit test, but not else. This also keeps the overhead simple since one can just copy a boilerplate skeleton from one of the existing tests and adapt it.

PR #3255, opened 2022-02-23
Update render-emphasis.ts

Redsandro:
Add backtick exception
https://github.com/Zettlr/Zettlr/commit/144ef799713d35502988f63ee507f9dfc18167ef#commitcomment-67307430

## Additional information


nathanlesage:
The problem with that is that this will not work if the code backticks don't border with the emphasis. This is something we have to check manually in the plugin, namely by copying over the comment check from the other plugins! I can do that real quick! :)

In the meantime (the mobile app unfortunately doesn't show commit comments so I can't read them) it would be nice if you could compile your list of test samples so that we can create a test file for the test environment. Feel free to do such a PR in the meantime! :)

Redsandro:
Like this? #3265

PR #3256, opened 2022-02-23
Enable smooth scrolling for tab container

### LINKED ISSUE #3243: [Enhancement] Add smooth scrolling effect to tab bar
### Description

I played a bit with the tab bar, and found that adding `scroll-behavior: smooth;` to the tabs div, gives the user a nice visual feedback when switching to a tab that is out of view.

Besides being a UI candy, the scroll animation also serves the purpose of indicating _where_ (to be more precise: in which direction) the target tab is located relative to the origin tab.

What are your thoughts?

### Proposed Changes

Add `scroll-behavior: smooth;` somewhere in this section:

https://github.com/Zettlr/Zettlr/blob/afa6ddfaf325e3e8219219a690ce62b6a65726f6/source/win-main/DocumentTabs.vue#L399-L409

### Caveats

Two potential issues:
* Are those kinds of animations welcome in Zettlr?
* When you have _many_ tabs open, and you switch between the first and the last tab, the scrolling effect could be annoying to some people. The file you switched to is opened instantly, while the scroll animation is only finished after one or two seconds.
    * But I believe that this scenario is rare
    * On the other hand, this could also be seen as a hint: _"Hey, you might have two many tabs open, maybe do some clean up?"_
        * Currently, with the static switching, you can't really tell that you have that many tabs open

Yes

### Zettlr Version

Stable (most recent version)

### Your Platform

- [X] Windows
- [X] macOS
- [X] Linux

### Operating System Version

_No response_

### Additional Information

_No response_

kyaso:
## Description
Closes #3243.

## Changes

Added the `scroll-behavior: smooth;` property to the tab-container.

Note: it can be disabled via custom.css:
```
body div#tab-container {
  scroll-behavior: auto;
}
```
I also added this information to the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@
 - Differentiate between inbound, outbound, and bidirectional links in the
   related files section
 - Fix related files display and link reporting
+- Added a smooth scrolling effect for the tab bar. If you don't like this
+  effect you can disable it with the following custom CSS:
+
+  ```
+  body div#tab-container {
+    scroll-behavior: auto;
+  }
+  ```
+
 - Added classes in order to conditionally style the related files section based
   on the type of relation, the classes are `tags` if the relation is based on
   keywords, `inbound` for incoming links, `outbound` for outgoing links, and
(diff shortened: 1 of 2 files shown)


nathanlesage (approved):
Looks good, although I would suggest to keep that change at the top (see my comment in the Changelog!)

nathanlesage:
Almost :D I've adapted it here: 0b438588

kyaso:
Great! :rocket:

PR #3265, opened 2022-02-24
Create render-emphasis.md

Redsandro:
It's a start.

## Description

List for emphasis tests per request

## Changes

Added .md file to test files

## Additional information

### DIFF
--- a/scripts/test-gui/test-files/Rendering/render-emphasis.md
+++ b/scripts/test-gui/test-files/Rendering/render-emphasis.md
@@ -1,0 +1,20 @@
+### Should match the following:  
+
+This is _italics._  
+This is _italics_.  
+_Italics_ at start.  
+_Only `match the _outer_ emphasis` in this sentence_  
+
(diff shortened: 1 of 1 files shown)


nathanlesage:
Thanks for this list! The complex matches are a great start, since these aren't yet handled correctly:

image

I would also propose to repeat these matches with asterisks only. The reason is that I found that asterisks are counted as word-characters by JavaScript regexes, other than underscores, so we should differentiate between underscores and asterisks.

nathanlesage:
Specifically, I think nested emphasis is still a problem

nathanlesage:
Why wasn't this merged …? Did I simply … forget it?

Redsandro:
I think we all forgot :laughing: although now I remember you suggested this PR, so I suggest you merge it. ;)

nathanlesage:
Your suggestion suggests that I suggestively forgot about my own words and will henceforth retreat in shame now. (after merging) 🙈

PR #3274, opened 2022-02-28
Added support for JabRef fileDirectory directive in BibTeX files

### LINKED ISSUE #619: [ENHANCEMENT] Support JabRef's fileDirectory feature
## Description
JabRef allows the user to choose a directory and then all references in the `file` field are relative to this directory instead of relative to the bib file.

## Proposed Changes
It would be nice if Zettlr could support this. Right now it is assumed that the linked attachments are relative to the bib file, so that none of the pdfs are found if a user uses the `fileDirectory` feature of JabRef.

## Additional Information
JabRef stores the file directory at the end of the file in the format
```
@Comment{jabref-meta: fileDirectory:\\..\\..\\..\\..\\Literature\\;}
@Comment{jabref-meta: fileDirectory-tdiez-TUD205196:H:\\My Documents\\Literature;}
```
The link can be relative (to the bib file) or an absolute file. Moreover, it is possible to store file directories per user (e.g. to support different file structures on different computers). I guess the later is not that important for the moment.

If you give me some pointers where to look, I can also try to implement it myself.

lpilz:
## Description
This PR adds parsing for `fileDirectory` comments in BibTeX files written by Jabref in order to correctly reference these files.

## Changes
I implemented a small parsing operation before the parsing of the entries to find the comments and extract the paths.

## Additional information
This is a first iteration, some additional features to discuss might be:

- more fine-grained control over which `fileDirectory`-paths get added (there are global and local ones, atm I just take the last)
- maybe a option in the settings to manually set a path if none is specified (potentially even overriding the `fileDirectory` paths)

Tested on:  Arch Linux  5.16.11-arch1-1

Closes #619

### DIFF
--- a/source/app/service-providers/citeproc/extract-bibtex-attachments.ts
+++ b/source/app/service-providers/citeproc/extract-bibtex-attachments.ts
@@ -14,6 +14,7 @@
 
 import path from 'path'
 import { bibtex } from 'astrocite'
+import { BracedComment } from 'astrocite-bibtex'
 import pdfSorter from '@common/util/sort-by-pdf'
 
 const AstrociteAST = bibtex.AST
@@ -37,14 +38,13 @@ export default function extractBibtexAttachments (
   let files = Object.create(null)
 
   // First we search for the jabref comments containing the files' root directories
-  for (let entry of ast.children) {
-    if (entry.kind !== 'BracedComment') {
-      continue
-    }
-
-    // The format of the value field is 'jabref-meta: fileDirectory*:<path>;'
-    if (entry.value.split(':')[1].trim().startsWith('fileDirectory')) {
-      baseDir = entry.value.split(':')[2].trim().replace(/;/g, '')
+  const comments = ast.children.filter(item => item.kind === 'BracedComment') as BracedComment[]
+  // The format of the value field is 'jabref-meta: fileDirectory*:<path>;'
+  const jabrefComments = comments.filter(item => item.value.startsWith('jabref-meta:'))
+  for (let entry of jabrefComments) {
+    const value = entry.value.split(':').map(e => e.trim())
+    if (value[1].startsWith('fileDirectory')) {
+      baseDir = value[2].replace(/;/g, '')
     }
   }
 
--- a/source/app/service-providers/citeproc/extract-bibtex-attachments.ts
+++ b/source/app/service-providers/citeproc/extract-bibtex-attachments.ts
@@ -36,6 +36,18 @@ export default function extractBibtexAttachments (
   // Return value will be a fast-access dictionary
   let files = Object.create(null)
 
+  // First we search for the jabref comments containing the files' root directories
+  for (let entry of ast.children) {
+    if (entry.kind !== 'BracedComment') {
+      continue
+    }
+
+    // The format of the value field is 'jabref-meta: fileDirectory*:<path>;'
+    if (entry.value.split(':')[1].trim().startsWith('fileDirectory')) {
+      baseDir = entry.value.split(':')[2].trim().replace(/;/g, '')
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


source/app/service-providers/citeproc/extract-bibtex-attachments.ts
```diff
@@ -36,6 +36,18 @@ export default function extractBibtexAttachments (
...
   let files = Object.create(null)
 
+  // First we search for the jabref comments containing the files' root directories
+  for (let entry of ast.children) {
+    if (entry.kind !== 'BracedComment') {
+      continue
+    }
+
+    // The format of the value field is 'jabref-meta: fileDirectory*:<path>;'
+    if (entry.value.split(':')[1].trim().startsWith('fileDirectory')) {
```
nathanlesage, line 46: Do all JabRef-comments start with `jabref-meta`? In that case, you can already reduce the complexity of the for loop by filtering:

```js
// Only retain comments
const comments = ast.children.filter(entry => entry.kind === 'BracedComment')
const jabrefComments = comments.filter(entry => entry.value.startsWith('jabref-meta:')

for (const entry of jabrefComments) {
  // ... your code
}
```

source/app/service-providers/citeproc/extract-bibtex-attachments.ts
```diff
@@ -36,6 +36,18 @@ export default function extractBibtexAttachments (
...
+      baseDir = entry.value.split(':')[2].trim().replace(/;/g, '')
```
nathanlesage, line 47: Possibly factor that out, i.e.:

```js
const value = entry.value.split(':').map(e => e.trim())

// value[1].startsWith('fileDirectory')
// value[2].replace(/;/g, '')
```

This way you can save some code.

nathanlesage (commented):
Looks very good thus far. I had just two minor ideas for improving the code. Then with regard to global and local fileDirectory paths: What exactly does global and local refer to? Local as in "per entry" versus global as in "per BibTeX file"?

lpilz:
Thanks for the quick review :) I implemented your suggestions and am testing them atm.

Concerning the JabRef global and local `fileDirectory` directives: 
As far as I understand the concept, there are 3 `fileDirectory` paths. The main `fileDirectory`, which can be set in the application (and is not saved in the `.bib` files) and then the general `fileDirectory` and user-specific `fileDirectory` which can be set per library (and are stored in the `.bib` files). When these are set, JabRef searches first the user-specific, then the general and finally the main `fileDirectory` for files which have relative links. This is supposed to enable e.g. accessing papers which are also shared with the whole research group vs. your personal papers.

As said - my approach is a first iteration, but one could maybe also think about doing the same kind of search for making relative paths absolute....

lpilz:
I'm actually having trouble implementing this solution - I'm a bit of a JS-noob. For some reason after filtering, the type changes and it complains about `Property 'value' does not exist on type 'Children'` `Property 'value' does not exist on type 'Entry'`.

Do you have any idea on what is happening there?

nathanlesage:
That is a typing error. I played around a little bit and you can solve this by type-casting.

First insert the following `import` statement below the astrocite-import:

```js
import { BracedComment } from 'astrocite-bibtex'
```

Then, you need to manually cast the type of the entries to `BracedComment`:

```ts
const comments = ast.children.filter(c => c.kind === 'BracedComment') as BracedComment[]
```

lpilz:
Thanks a lot, that worked nicely! Would have never figured that out by myself :D

nathanlesage:
Looks very clean now, I really like that! :)

I'll be adding a few minor things after merging for debugging purposes etc., but it looks great!

lpilz:
Thanks a lot for the quick response and your help! :)

PR #3321, opened 2022-03-14
disable auto translation update

bigbenzh:
## Description

Added a checkbox to disable Zettlr connecting to translation server on start, which gives Zettlr a fully offline mode.

## Changes

Similar to #2827 

## Additional information

Windows 10 Pro

### DIFF
--- a/source/app/service-providers/translations/index.ts
+++ b/source/app/service-providers/translations/index.ts
@@ -108,7 +108,7 @@ export default class TranslationProvider extends ProviderContract {
     // Get user's setting of translation updates
     const checkForTranslationUpdates: boolean = this._config.get('system.checkForTranslationUpdates')
     if (checkForTranslationUpdates) {
-      this.updateTranslations().catch(err => this._logger.error('[Translation Provider] Failed to update translations.'))
+      this.updateTranslations().catch(err => this._logger.error(`[Translation Provider] Failed to update translations: ${String(err.code)}`, err))
     } else {
       this._logger.info('[Translation Provider] User decide not to update translations.')
     }
--- a/source/app/service-providers/translations/index.ts
+++ b/source/app/service-providers/translations/index.ts
@@ -107,12 +107,22 @@ export default class TranslationProvider extends ProviderContract {
 
     // return file
     let response
-    try {
-      response = await got(TRANSLATION_API_URL, { method: 'GET' })
-    } catch (err: any) {
-      // Not critical.
-      this._logger.warning(`[Translation Provider] Could not update translations: ${String(err.code)}`, err)
-      return
+    const checkForTranslationUpdates: boolean = this._config.get('system.checkForTranslationUpdates')
+    if (checkForTranslationUpdates) {
+      try {
+        response = await got(TRANSLATION_API_URL, { method: 'GET' })
+      } catch (err: any) {
+        // Not critical.
+        this._logger.warning(`[Translation Provider] Could not update translations: ${String(err.code)}`, err)
+        return
+      }
+
+      // Alright, we only need the body
+      response = JSON.parse(response.body)
+      this._availableLanguages = response // Let's save the response
+    } else {
+      this._logger.info('[Translation Provider] Translation update cancelled due to user setting.')
+      this._availableLanguages = []
     }
 
     // Alright, we only need the body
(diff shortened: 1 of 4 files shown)


source/app/service-providers/translations/index.ts
```diff
@@ -107,18 +107,23 @@ export default class TranslationProvider extends ProviderContract {
 
     // return file
     let response
-    try {
-      response = await got(TRANSLATION_API_URL, { method: 'GET' })
-    } catch (err: any) {
-      // Not critical.
-      this._logger.warning(`[Translation Provider] Could not update translations: ${String(err.code)}`, err)
-      return
+    const checkForTranslationUpdates: boolean = this._config.get('system.checkForTranslationUpdates')
```
nathanlesage, line 110: This long `if`-statement makes this hard to read; is it possible to just check for `if (shouldNotCheck)` to return early here? As far as I can see, the rest of the code just pertains to downloading updates.

Also, it might make sense to factor that logic out into its own function, right? It looks like a remnant from back when I didn't properly organize my code …

nathanlesage (commented):
Looks all good, with one comment. Please follow up with what you think about it!

source/app/service-providers/translations/index.ts
bigbenzh: > This long `if`-statement makes this hard to read; is it possible to just check for `if (shouldNotCheck)` to return early here? As far as I can see, the rest of the code just pertains to downloading updates.
> 
> Also, it might make sense to factor that logic out into its own function, right? It looks like a remnant from back when I didn't properly organize my code …

I updated the logic and moved the downloading process to update(), and keep the check process in boot(). Is it better now?

source/app/service-providers/translations/index.ts
```diff
@@ -105,6 +105,20 @@ export default class TranslationProvider extends ProviderContract {
...
+    } else {
+      this._logger.info('[Translation Provider] User decide not to update translations.')
+    }
+  }
+
+  /**
+   * Get translation updates via translation server
+   * @return {Promise} Resolves if everything worked out, rejects otherwise.
+   */
+  async update (): Promise<void> {
```
nathanlesage, line 121: I think it might make sense to call this `updateTranslations` to make it explicit that what this function does is solely to download updated translations

source/app/service-providers/translations/index.ts
nathanlesage: Yes, wonderful this way!

source/app/service-providers/translations/index.ts
```diff
@@ -105,6 +105,20 @@ export default class TranslationProvider extends ProviderContract {
     global.i18nFallbackRawData = await fs.readFile(fallback.path, 'utf8')
     global.i18nFallback = JSON.parse(global.i18nFallbackRawData)
 
+    // Get user's setting of translation updates
+    const checkForTranslationUpdates: boolean = this._config.get('system.checkForTranslationUpdates')
+    if (checkForTranslationUpdates) {
+      await this.update()
```
nathanlesage, line 111: An additional thing I just noticed: Instead of awaiting this promise here, we could also write `this.update().catch(err => logTheError)`.

The reason is that this will speed up the boot process, because the whole app does not have to wait for translations to download. Any changes are only applied after a restart anyway. This could make the app snappier, especially when users have slow internet connections.

We could then simply log an error if something went wrong.

nathanlesage (commented):
Looks very good, I like that the functionality is now factored out into its own function. There are only two very minor things, but then this PR is good to go!

source/app/service-providers/translations/index.ts
bigbenzh: > An additional thing I just noticed: Instead of awaiting this promise here, we could also write `this.update().catch(err => logTheError)`.
> 
> The reason is that this will speed up the boot process, because the whole app does not have to wait for translations to download. Any changes are only applied after a restart anyway. This could make the app snappier, especially when users have slow internet connections.
> 
> We could then simply log an error if something went wrong.

Changed both. Is that look good to you?

nathanlesage (approved):
Looks great! I'll have a look over the code tomorrow, but I think that it's mergable as is :)

nathanlesage:
Checked, works, all fine! Merging …

PR #3384, opened 2022-04-10
Add copy filename/path/ID to tab context menus

### LINKED ISSUE #3069: Copy filename in editor pane context menu
### Checklist

### Feature Proposal

As a Zettlr user, I want to easily link my zettels together to form a zettelkasten. It would make things a little easier if I could copy the filename (minus extension) of a file (zettel) from the editing pane tab using the context menu.

### Example Workflow

As a zettelkasten user of Zettler, I want to quickly create links between relevant notes/files/zettels. I have a few of them open in the editing pane currently. I use the filename as the ID, which Zettlr already supports as default behavior. I want to quickly grab the ID of one of my existing zettels so I can put a link to it in another one I am editing. So I simply click the other file to get its context menu, I select "copy name" or something like that, and this copies the name of the file to my clipboard. I can then easily paste it into a link in another file.

### Caveats

The context menu currently has very few items, so I think there is no caveat against this addition.

No

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

### Additional Information

[](url
feature_request
)

kyaso:
## Description
This PR extends the document tabs context menus with the following items:
* Copy filename
* ~~Copy path~~ Copy absolute path
* Copy ID

The only thing that is missing is a translation for the ~~"Copy path"~~ "Copy absolute path" option. I'm not sure how to do that properly?

- [ ] Add translations for ~~"Copy path"~~ "Copy absolute path"

Closes #3069

## Changes
* Added the new entries in `tabs-context.ts`
* Added click handling for the new entries in `DocumentTabs.vue`


source/win-main/tabs-context.ts
```diff
@@ -34,6 +36,27 @@ export default function displayTabsContext (event: MouseEvent, callback: (clicke
...
+      type: 'separator'
+    },
+    {
+      label: trans('menu.copy_filename'),
+      id: 'copy-filename',
+      type: 'normal',
+      enabled: true
+    },
+    {
+      label: trans('menu.copy_path'),
```
nathanlesage, line 50: I would suggest to rename that to `menu.copy_abs_path` and give it an according ID, because there might be in the future the need to also enable relative path copying. Other than that, everything perfect!

nathanlesage (changes requested):
I see that I have so many notifications that this PR just sat unseen for twelve days ... apologies!

source/win-main/tabs-context.ts
kyaso: > I would suggest to rename that to menu.copy_abs_path and give it an according ID, because there might be in the future the need to also enable relative path copying.

Will do that! Could you possibly give me a hint on where I can define that ID?

nathanlesage: I have to manually add those to Zettlr translate -- it's not very automated as of now.

kyaso:
@nathanlesage I've fixed the trans string. You could take another look, and merge if everything is fine.

nathanlesage:
Sorry for being so tardy as of now, I have to fix a very badly behaving model right now for a conference presentation in two weeks. Merging now!

kyaso:
No worries!

Issue #3295, opened 2022-03-06
[Bug] Keyboard input lags


sebzeter:
### Description

Keyboard input is often lagging / delayed since v2.2.4.

### Reproducing

Open a file in Zettlr v2.2.4 then start typing.
Entered characters do not appear immediately on the screen.

### Zettlr Version

Stable (most recent version)

### Specify version

v2.2.4

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

openSUSE Leap 15.3 x86_64

### Additional Information

Downgrading to v2.2.3 solves the issue.

nathanlesage:
Ah, I was already puzzled and suspected lower power of my laptop for that. Likely a result of me unifying the renderer plugin stuff (specifically the function that checks whether rendering an element can be performed). I'll have a look, except @kyaso has some time on their hands?

kyaso:
> I'll have a look, except @kyaso has some time on their hands?

Go ahead! I also noticed a little bit of lag, especially on large files (1000+ words). If I find some free time today, I can also take a look.

nathanlesage:
Alright, then I'll leave it to you, I'm unfortunately still working 🙈

kyaso:
Ok, I spent some time figuring out where this issue originates from. First, I tried to find a reliable way of reproducing the issue, and I think the following steps consistently trigger the bug:

1. Type something
2. Immediately press `Ctrl+S` to save
3. Immediately type something again -> Text will appear delayed

I found that the issue first appeared in 27cf119ebcaa562daa4d8e1f1f66e238660feeb1 (the upgrade to Electron 17. Also note that this commit comes _after_ the renderer changes (f2fc705138ab95f982893e7bc0ba9792487140a0)).

The commit prior to that, 9e7f73fd385b8963defe09edfd6f777ba5c6b773, does not contain the bug.

@nathanlesage Could you verify whether the bug was indeed introduced with 27cf119ebcaa562daa4d8e1f1f66e238660feeb1?

nathanlesage:
@kyaso I can assure you it has nothing to do with Electron 17 (otherwise we would see this issue regardless of the state of Zettlr)

I can confirm that this only affects a few tabs.

[...]

PR #3392, opened 2022-04-13
fix: Simplify codeblock classes again

nathanlesage:
## Description

This PR attempts to solve #3295 by simplifying the codeblock class application.

## Changes

* Remove the specific checks for certain Markdown mode classes in order to distinguish list items from indented code block lines
* Indented codeblocks will now no longer have the background classes applied under the assumption that most people will probably want to use fenced code blocks either way (or is this assumption wrong? Feedback very welcome)

## Additional information

@kyaso I noticed that you complained that `startOperation` and `endOperation` mingled with the VIM mode, correct? The problem is that we have to have those operations. Otherwise the lag will be even significantly worse than before, because CodeMirror will force a refresh once classes are being changed via `addLineClass` and `removeLineClass`. This almost sounds like an interference between the vim mode and CodeMirror itself, noting that the vim mode has no maintainer and the code is several years old now.

Tested on: macOS Monterey 12.2.1, but really should be platform-agnostic I hope

### DIFF
--- a/source/common/modules/markdown-editor/hooks/codeblock-classes.ts
+++ b/source/common/modules/markdown-editor/hooks/codeblock-classes.ts
@@ -60,6 +60,9 @@ function findCode (cm: CodeMirror.Editor): void {
         // Increment the lineCount and apply the code start line to that line
         cm.addLineClass(++i, 'wrap', codeblockClass)
         cm.addLineClass(i, 'wrap', codeblockClassOpen)
+        // Remove a potential close class, in case we have just one line of
+        // code block; found by @kyaso
+        cm.removeLineClass(i, 'wrap', codeblockClassClose)
       }
     } else if (codeBlockRE.test(line) && inCodeBlock) {
       // End a codeblock: Remove any codeblock class
--- a/source/common/modules/markdown-editor/hooks/codeblock-classes.ts
+++ b/source/common/modules/markdown-editor/hooks/codeblock-classes.ts
@@ -12,13 +12,13 @@
  * END HEADER
  */
 import CodeMirror from 'codemirror'
-import { debounce, range } from 'lodash'
+// import { debounce, range } from 'lodash'
 const codeblockClass = 'code-block-line'
 const codeblockClassOpen = 'code-block-first-line'
 const codeblockClassClose = 'code-block-last-line'
 // The debounce timeout needs to be exactly the same as but no less than the
 // debounce timeout used in CodeMirror Markdown Mode.
-const findCodeDebounced = debounce(findCode, 400, { leading: true })
+// const findCodeDebounced = debounce(findCode, 400, { leading: true })
 
 /**
  * Hooks onto the cursorActivity, optionChange and keyHandled event to apply
@@ -29,8 +29,10 @@ const findCodeDebounced = debounce(findCode, 400, { leading: true })
  */
 export default function codeblockClassHook (cm: CodeMirror.Editor): void {
   cm.on('keyHandled', handleNewline)
-  cm.on('cursorActivity', findCodeDebounced)
-  cm.on('optionChange', findCodeDebounced)
+  // cm.on('cursorActivity', findCodeDebounced)
+  cm.on('cursorActivity', findCode)
+  // cm.on('optionChange', findCodeDebounced)
+  cm.on('optionChange', findCode)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


source/common/modules/markdown-editor/hooks/codeblock-classes.ts
```diff
@@ -29,8 +29,10 @@ const findCodeDebounced = debounce(findCode, 400, { leading: true })
  */
 export default function codeblockClassHook (cm: CodeMirror.Editor): void {
   cm.on('keyHandled', handleNewline)
-  cm.on('cursorActivity', findCodeDebounced)
-  cm.on('optionChange', findCodeDebounced)
+  // cm.on('cursorActivity', findCodeDebounced)
+  cm.on('cursorActivity', findCode)
+  // cm.on('optionChange', findCodeDebounced)
+  cm.on('optionChange', findCode)
```
kyaso, line 35: During testing the Vim mode issue, I found that removing the `optionChange` listener indeed made Vim mode work again. I didn't fully understand why we have to listen for the `optionChange` event in the first place?

Also, I believe it would be sufficient to just listen to the `change` event instead `cursorActivity`. Well, during _typing_ both would be triggered (if I'm not mistaken), but I don't see a reason why the `findCode` function should be fired when only _navigating_ with the cursor.

Finally, listening to the `swapDoc` event would make sure that the codeblocks are rendered when we open/switch to a different document. (Currently, the `optionChange` event covers listening to file open events, but when we remove that, `swapDoc` could take over that part)

I tested a bit using the `change` and `swapDoc` hooks, and so far I didn't see any issue with Vim mode (and the normal mode).

kyaso, line 35: During testing the Vim mode issue, I found that removing the `optionChange` listener indeed made Vim mode work again. I didn't fully understand why we have to listen for the `optionChange` event in the first place?

Also, I believe it would be sufficient to just listen to the `change` event instead `cursorActivity`. Well, during _typing_ both would be triggered (if I'm not mistaken), but I don't see a reason why the `findCode` function should be fired when only _navigating_ with the cursor.

Finally, listening to the `swapDoc` event would make sure that the codeblocks are rendered when we open/switch to a different document. (Currently, the `optionChange` event covers listening to file open events, but when we remove that, `swapDoc` could take over that part)

I tested a bit using the `change` and `swapDoc` hooks, and so far I didn't see any issue with Vim mode (and the normal mode).

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
```diff
@@ -29,8 +29,10 @@ const findCodeDebounced = debounce(findCode, 400, { leading: true })
  */
 export default function codeblockClassHook (cm: CodeMirror.Editor): void {
   cm.on('keyHandled', handleNewline)
```
kyaso, line 31: This hook shouldn't be needed anymore, right? If I remember correctly, it had something to do with the debouncing, and that you don't want to debounce the `Enter` key. But now that we go rid of the debouncing stuff, this hook seems to be redundant.

kyaso, line 31: This hook shouldn't be needed anymore, right? If I remember correctly, it had something to do with the debouncing, and that you don't want to debounce the `Enter` key. But now that we go rid of the debouncing stuff, this hook seems to be redundant.

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
```diff
@@ -57,93 +59,43 @@ function handleNewline (cm: CodeMirror.Editor, name: string): void {
...
+    if (!inCodeBlock) {
+      // Not a codeblock: Remove any class
+      cm.removeLineClass(i, 'wrap', codeblockClass)
+      cm.removeLineClass(i, 'wrap', codeblockClassOpen)
+      cm.removeLineClass(i, 'wrap', codeblockClassClose)
+      if (codeBlockRE.test(line)) {
+        // Begin a codeblock
+        inCodeBlock = true
+        // Increment the lineCount and apply the code start line to that line
+        cm.addLineClass(++i, 'wrap', codeblockClass)
```
kyaso, line 81: I noticed that when you add a newline to a codeblock that only had one line of code, the first code line still has the `codeblockClassClose` class. We could fix that by removing that class first:
```suggestion
        cm.removeLineClass(++i, 'wrap', codeblockClassClose)
        cm.addLineClass(i, 'wrap', codeblockClass)
```

kyaso, line 81: I noticed that when you add a newline to a codeblock that only had one line of code, the first code line still has the `codeblockClassClose` class. We could fix that by removing that class first:
```suggestion
        cm.removeLineClass(++i, 'wrap', codeblockClassClose)
        cm.addLineClass(i, 'wrap', codeblockClass)
```

kyaso (changes requested):
Looks good, and the lags are indeed gone! I left some comments regarding the Vim mode issue, and one styling fix suggestion.

nathanlesage:
The comments look all good, 90% forgot to remove commented code and go over the code again. But good to know that it gets rid of the lag! I'm going to apply your suggestions possibly tonight/tomorrow! Thanks!

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
```diff
@@ -12,13 +12,13 @@
  * END HEADER
  */
 import CodeMirror from 'codemirror'
-import { debounce, range } from 'lodash'
+// import { debounce, range } from 'lodash'
```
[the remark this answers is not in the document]
nathanlesage: And it's out!

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
nathanlesage: `optionChange` was indeed unnecessary, so removed that hook.

With regard to `cursorActivity` you are right. Changing as such!

Also, your suggestion with `swapDoc` is very good, added that event as well!

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
nathanlesage: Already gone!

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
nathanlesage: Implemented!

nathanlesage:
So, I think I got all your comments, feel free to have a look at it again, and if you don't find anything I think you can just merge it!

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
```diff
@@ -12,13 +12,9 @@
...
-import { debounce, range } from 'lodash'
 const codeblockClass = 'code-block-line'
 const codeblockClassOpen = 'code-block-first-line'
 const codeblockClassClose = 'code-block-last-line'
-// The debounce timeout needs to be exactly the same as but no less than the
-// debounce timeout used in CodeMirror Markdown Mode.
-const findCodeDebounced = debounce(findCode, 400, { leading: true })
 
 /**
  * Hooks onto the cursorActivity, optionChange and keyHandled event to apply
```
kyaso, line 20: Maybe we should update the description here?

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
```diff
@@ -28,22 +24,8 @@ const findCodeDebounced = debounce(findCode, 400, { leading: true })
...
- */
-function handleNewline (cm: CodeMirror.Editor, name: string): void {
-  if (name === 'Enter') {
-    findCode(cm)
-  }
+  cm.on('change', findCode)
+  cm.on('swapDoc', findCode)
 }
 
 /**
```
kyaso, line 31: dito

kyaso (commented):
Looks good!

Just two little comments, then we could merge :rocket:

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
nathanlesage: That code is already gone!

source/common/modules/markdown-editor/hooks/codeblock-classes.ts
nathanlesage: Same here -- also gone

nathanlesage:
Sorry for answering so late, GitHub didn't think a notification would be appropriate :D 

The two cases you mentioned are already gone (it was old code which I removed later on), so I'm going to merge now!

kyaso:
No worries!

I think there might have been a small misunderstanding. I was referring to the documentation comments above the two functions `codeblockClassHook()` and `findCode()`, as those seem not up-to-date anymore. But I guess, it's not that critical, so it can still be updated at some later time :D

nathanlesage:
Ah, then GitHub has messed the location of your comments up! Nevertheless, fixed that real quick! :D

PR #3422, opened 2022-04-28
Refactor Vuex Store

nathanlesage:
## Description

This PR is a first in a series of PRs of last changes required in order to implement split-views and multiple windows. I just realized that it would make sense to already open this PR as a draft PR to be transparent.

Overall, I will over the course of the next months open a series of PRs in order to prepare the main window components for the integration of said split views and multiple windows:

1. (This PR) Modularize the store and clean up the state management
2. Migrate the store to pinia, since apparently this has separate stores, and also un-proxied state ability (?) which is required for the complex documents of CodeMirror 5 and 6 to store them in there
3. Move the way too large state management logic from the MainEditor vue component into the store
4. Split up the main editor component into two components – one for Markdown files and one for code files. (This is only partially related, but since I aim to integrate a few more file formats, most notably PDF, into the app, this is a good way to have the modularity up and running when we have to refactor the state again in order to enable split views. This way we, or rather, I, don't do anything stupid that makes that harder than necessary)

## Changes

This specific PR has two broad aims:

1. Modularize the store into several files which are easier to maintain. I somehow completely forgot that with the large state which we have having everything in one file is ridiculous. As the PR is at the time of writing I already moved large chunks of the logic into separate files and it is much cleaner now
2. Clean up the state management itself. I saw that some properties of the state objects can be merged, others should potentially be split, and I aim to also move a lot of the localized states within the Vue components into the central state (sans the main editor, since its state management is too large for this PR), because at several points in the application, we are creating double work because two components rely on the same information

## Additional information

I'm not finished with it yet, and it will take a few more weeks due to intense workload. But @tobiasdiez and @kyaso feel free to look around already. But this PR is not yet review-ready.

Tested on: macOS Monterey


nathanlesage:
Turns out 90% of the bad state logic is actually the editor, which somewhat blocks complete clean up, so I'll have to do a final cleanup state after the editor's state has been merged.

Therefore, the PR is now ready for review.

nathanlesage:
@kyaso Do you have time over the weekend or so to do a code review on this?

kyaso:
Oh, shoot, it totally slipped my mind, sorry..!

I'll take a look this week or this weekend!

source/win-main/store/actions/filtree-update.ts
```diff
@@ -0,0 +1,109 @@
+/**
```
kyaso, line 1: Small typo in file name: Should be "filetree-update.ts", correct?

source/win-main/MainSidebar.vue
```diff
@@ -279,17 +279,20 @@ export default defineComponent({
     citationKeys: function (): string[] {
       return this.$store.state.citationKeys
     },
+    bibliography: function (): any {
+      return this.$store.state.bibliography
+    },
     referenceHTML: function (): string {
-      if (this.bibContents === undefined || this.bibContents[1].length === 0) {
```
kyaso, line 301: `bibContents` is then not used anymore in this file, right?

kyaso (changes requested):
Looks good! I hope I didn't miss anything major, but I only came across two things that we should address :D

source/win-main/MainSidebar.vue
nathanlesage: Good catch! Will update …

source/win-main/store/actions/filtree-update.ts
nathanlesage: Yup!

nathanlesage:
Thanks for the review! I've now addressed your two points, and also made sure to test it again once more to be sure that nothing broke. I'm going to merge now so that this change is also in tomorrow's nightly and we may catch some hidden bugs before the release of 2.3 :) 

Thank you for having a second look!

PR #3429, opened 2022-04-30
Fix related files latency

### LINKED ISSUE #3399: [Bug] Massive lags when checking for related files
### Description

When Zettlr is checking for related files in a file that actually has many related files (in my case >1000) it massively lags. (At least the guess is, that the related files functionality is causing this.)

### Reproducing

**Following the exact steps how to reproduce**
1. Make sure to have a folder with many files that relate to each other through tags or links. You can create one via script or [download an example folder (the link is valid for one week](https://share.anoxinon.de/f.php?h=1r9a5vibd7nuFzTM8ZpOr2). 
2. Open that folder as workspace in Zettlr
3. Create or open an file, make sure that the file relates to the others by putting the respective tag or link into the file. For my example this would be `#someTag` (case sensitive!)
4. Open the sidebar
5. Switch to the related files tab, if not already open
7. You will notice a lag immediately
8. Work in the file. 
9. Let it save automatically or save it manually
10. The related files list will empty, probably Zettlr is re-calculating it
11. Immediately type or do something again, after saving
12. It lags/freezes until the related files are shown again

**Following is how your Zettlr view should look like for reproducing.** 

* First image is showing Zettlr whit the related files list filled, no lag appears.

![Screenshot of Zettlr with a example workspace, an example file with `#someTag [...]` and the sidebar with related files tab open, showing a list of related files](https://user-images.githubusercontent.com/73606096/164289676-9bf63ea4-7aa6-49ed-85b9-b3454a194826.png)

* Second image is showing Zettlr whit the related files list empty, when probabaly re-calculating, this is when it freezes

![Screenshot of Zettlr with a example workspace, an example file with `#someTag [...]` and the sidebar with related files tab open, showing an empty list of related files](https://user-images.githubusercontent.com/73606096/164289622-fbe2fe26-0126-421b-a9be-9996fe0af79a.png)

### Zettlr Version

Stable (most recent version)

### Specify version

2.2.5

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

OpenSUSE Tumbleweed; Debian Bullseye (But only tested a few days ago there with version 2.2.4. However, all other info applies there as well and the Zettlr config is exactly the same)

### Additional Information

This issue was caused out of #3295

nathanlesage:
## Description

This PR attempts to fix #3399 by reducing the workload the DOM has to do with regard to repainting.

## Changes

Instead of vomiting the 2000+ files that may be related directly into the DOM of the sidebar, this PR makes use of the `RecycleScroller` which is already in use in the file list. The idea: While calculating the related files is very fast, displaying each on the DOM is very computation-heavy. This PR does not change the generation of related files, but only its display.

I tested this out with the quick and dirty file generation script from @Redsandro both with autosave off and autosave immediately. The lag was gone and while the related files still flicker a little bit, typing is as quick as it can possibly get.

To make sure, I also tested out the very same files on the older branch, and could confirm that there was suddenly some lag.

## Additional information

**I noticed a few other things we have to do later**

1. Disaggregate the sidebar. It's way too full. And iirc someone even called me out for this before 🙈 
2. Finally fix the sidebar in so far that the tab bar and the panel header always stay fixed and only the content scrolls.

But that is for another PR.

**Edit: What I must fix before merging is that the related file items can currently be of variable height (e.g. for longer file names). Recyclescroller doesn't like that. Ideas on how to do that design-wise welcome.**

Tested on:  macOS Monterey 12.3.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fixed a small visual glitch on Linux where the selected file was not
   highlighted in the file list in the thin or expanded file manager modes
 - Fixed heading ID links not working with, e.g., accents
+- Fixed heavy latency when working on a file with many related files
 
 ## Under the Hood
 
--- a/source/win-main/MainSidebar.vue
+++ b/source/win-main/MainSidebar.vue
@@ -609,19 +609,22 @@ body {
       padding: 10px;
 
       div.related-file {
+        // NOTE: The margin + height equal 42, which was the automatic height
+        // before we fixed it here. We have to fix it because the Recycle
+        // scroller requires completely fixed heights.
         margin-bottom: 10px;
         display: flex;
         align-items: center;
+        padding: 10px 5px;
+        height: 32px;
 
-        &:hover {
-          background-color: rgb(200, 200, 200);
-        }
+        &:hover { background-color: rgb(200, 200, 200); }
 
         span.filename {
           display: inline-block;
           font-size: 11px;
-          padding: 10px 5px;
           flex-grow: 8;
+          overflow: hidden;
         }
 
         span.icons {
--- a/source/win-main/MainSidebar.vue
+++ b/source/win-main/MainSidebar.vue
@@ -190,6 +190,15 @@ export default defineComponent({
     }
   },
   computed: {
+    /**
+     * The Vue Virtual Scroller component expects an array of objects which
+     * expose two properties: id and "props". The latter contains the actual
+     * object (i.e. the RelatedFile). We may want to merge this functionality
+     * into the RelatedFiles generation later on, but this is the safest way
+     * for now.
+     *
+     * @return  {{ id: number, props: RelatedFile }}  The data for the scroller
+     */
     scrollerRelatedFiles: function (): any {
       return this.relatedFiles.map((elem, idx) => {
         return { id: idx, props: elem }
--- a/source/win-main/MainSidebar.vue
+++ b/source/win-main/MainSidebar.vue
@@ -54,46 +54,53 @@
           {{ noRelatedFilesMessage }}
         </div>
         <div v-else>
-          <div
-            v-for="fileRecord, idx in relatedFiles"
-            v-bind:key="idx"
-            v-bind:class="{
-              'related-file': true,
-              'tags': fileRecord.tags.length > 0,
-              'inbound': fileRecord.link === 'inbound',
-              'outbound': fileRecord.link === 'outbound',
-              'bidirectional': fileRecord.link === 'bidirectional'
-            }"
+          <RecycleScroller
+            v-slot="{ item, index }"
+            v-bind:items="scrollerRelatedFiles"
+            v-bind:item-size="43"
+            v-bind:emit-update="true"
+            v-bind:page-mode="true"
           >
-            <span
-              class="filename"
-              draggable="true"
-              v-on:mousedown.stop="requestFile($event, fileRecord.path)"
-              v-on:dragstart="beginDragRelatedFile($event, fileRecord.path)"
-            >{{ getRelatedFileName(fileRecord.path) }}</span>
-            <span class="icons">
-              <clr-icon
-                v-if="fileRecord.tags.length > 0"
-                shape="tag"
-                title="This relation is based on tag similarity."
-              ></clr-icon>
-              <clr-icon
-                v-if="fileRecord.link === 'inbound'"
-                shape="arrow left"
-                title="This relation is based on a backlink."
-              ></clr-icon>
-              <clr-icon
-                v-else-if="fileRecord.link === 'outbound'"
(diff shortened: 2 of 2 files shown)


Redsandro:
> 1. Disaggregate the sidebar. It's way too full.

Wait, don't go killing too many darlings. This issue caused me to open the side bar for the first time (!) and turns out I like it! If you're thinking about grooming the sidebar to a considerable extent, I wouldn't mind being tagged in a discussion so I can weigh in. I definitely like the "related files" because I still have fights with "search" showing least relevant items first.

> the related file items can currently be of variable height (e.g. for longer file names). Recyclescroller doesn't like that. Ideas on how to do that design-wise welcome.

Can you provide screenshots, here or in a new issue? I don't see what the problem is, which probably means I'm not using the correct issue trigger.

image

nathanlesage:
> Wait, don't go killing too many darlings.

Hahaha, sorry I didn't mean to scare you :D With disaggregate I only mean to split up the sidebar components into their own files, not kill any of the features. It's barely manageable as it is now, so having all of the different tabs in their own respective components will make maintenance much better.

> don't really understand the logic of the order of the results since the january commit.

Basically: First bidirectional links, then incoming, then outgoing, then tags.

> Can you provide screenshots, here or in a new issue? I don't see what the problem is, which probably means I'm not using the correct issue trigger.

The reason is how the recycle scroller works. It needs to know how big the items will be in order to properly space it. You don't see the issue right now because I hacked together the (for macOS) correct height, so it's not visible. But I need to address this in this PR before merging so that we don't wake up to a bad surprise ;)

nathanlesage:
Now I got it to look like this:

image

Question is: Is this okay (because two lines will work just perfectly fine), or should we rather just force a single line (and then have it cut off at the end, like in the file list?)

kyaso:
> Question is: Is this okay (because two lines will work just perfectly fine), or should we rather just force a single line (and then have it cut off at the end, like in the file list?)

This looks fine to me.

kyaso (approved):
Looks good! I can confirm that the lags in the case of many related files are indeed gone.

nathanlesage:
Wonderful, thank you for the feedback!

Let's see that other users can confirm that nothing broke with Monday's Nightly, and if that works perfectly well, I'll make sure to release 2.2.6 sometime next week.

Redsandro:
> Question is: Is this okay (because two lines will work just perfectly fine), or should we rather just force a single line (and then have it cut off at the end, like in the file list?)

This is good enough, but if you really want to go for a triple A looking product, the spacing for 3 lines is too tight. Two lines are fine. Does Electron use webkit across the board? Perhaps you can use `line-clamp` with `ellipsis`:

```css
  overflow: hidden;
  text-overflow: ellipsis;
  display: -webkit-box;
  -webkit-line-clamp: 2;
```

nathanlesage:
@Redsandro CSS FINALLY HAS AN AUTOMATIC ELLIPSIS?!?!?!?! O___O

Gosh, I feel so out of the loop …

I don't know if webkit line clamp will work, but the ellipsis feature? Holy crap, I'm gonna add that asap

Redsandro:
lol well it is standardized for a while now but I'm happy you got excited for a moment. Sad to see - and a bit puzzled - that it is not working. Personally I'm more excited about the line-clamp but it still depends on webkit so electron seems like the perfect excuse to use it. But without the ellipsis? Hmm.. Perhaps it depends on some specific box model settings?

Check out this test in a webkit-based browser: https://css-tricks.com/almanac/properties/l/line-clamp/

nathanlesage:
AH-HAH!

The magic feature was `-webkit-box-orient: vertical;`. I didn't see that earlier. With that active, it works like a charm:

image

Redsandro:
> image

It works beautifully! Can it be combined with vertical-align: center or some inner-outer center so that they always line out to the center of the box, even with only one line?

Random example: https://codepen.io/anon/pen/htbxA

nathanlesage:
> Can it be combined with vertical-align: center

Not un-hacky, I'm afraid because for that we would need `display: flex` but that would overwrite the `webkit-box` property that is required for the line-clamp …

I mean we could do it like your example, but such things always look super hacky to me :/

Redsandro:
I understand that hacky is not the best feeling in the world, because it's a crime you'd have to carry with you. 

But you'd carry this weight for everyone else so that they can be happy. This time it's for a particularly good reason because correct alignment is super important. Especially for OCD people that like to take notes of their thoughts and observations like myself.

Example of bad and good optics:

image

Especially on-mouse-over:

image

I think your options are to either give the entries an automatic content-based variable height and deal with the recycle scroller, or do the hack. 

My opinion?

Do the hack. :shipit:

Issue #2987, opened 2022-01-03
Add a way for toolbar customization


nathanlesage:
Since we have had multiple requests (#2980; #2685) for new toolbar buttons in the past weeks after I cleaned up the toolbar, we need some mechanism to customize the toolbar. There are basically two ways of doing this:

1. The "cool" macOS way with drag and drop (hard to implement)
2. (For now) a simple array of checkboxes in the settings to programmatically activate/deactivate buttons to show up

If anyone wants to give that a shot before me, feel free, the three places to touch would be the config (to store the choices, using the current (!) array as defaults), the App.vue component in the main window (where you can instead of simply returning the array of toolbar buttons build that array programmatically depending on the config) and a new preferences schema in the preferences window offering a new tab with all these choices.

***

If you wish to get on this, please comment here so we know someone is working on this.

lanceli237:
Hi Nathan, this is Lance, hope you are doing well. We are a course group of six members from The University of Adelaide, trying to complete this issue and make a contribution. Thanks for your three hints above about the toolbar feature. We will try to kick off this issue from next week. If you come up with any more ideas, please don't hesitate to contact with us. Cheers

nathanlesage:
@lanceli237 Thank you for tackling this issue! Please feel free to reach out in case you have any questions! Please also feel free to share here a first concept as soon as you came up with one so we can have a look collectively if it's good or if there could be improvements made! :)

haoliudev:
@nathanlesage Hello Nathan,  as mentioned by @lanceli237, we are a group of students from the University of Adelaide and we love to learn from and contribute to the Zettlr project. 

We have added a new preference window [toolbar] to save the config control button display.
The code is here: https://github.com/EdgarTang/Zettlr/tree/customize-toolbar

Now We have two questions regarding toolbar customization. 

1. Is it a good idea to add a boolean "visible" attribute to button objects(which is set to true by default)? In this way, we can display/hide by ticking/unticking an array of checkboxes to change the "visible" values. To avoid affecting other buttons, the corresponding button will only be hidden when "visible" exists and is set as false.

2. When the toolbar config is changed, we noticed that the GUI must be refreshed for the settings to take effect. The problem is that we do not know how to call the method to refresh the GUI. Could you please give us some advice or explanation?


![2987](https://user-images.githubusercontent.com/85983213/165676827-74197cbf-891f-4785-9839-223f9ba64c62.gif)

nathanlesage:
[...]

Issue #2980, opened 2022-01-02
[Enhancement] Add toolbar buttons for back/forward navigation


kyaso:
### Description

I think it would be nice to also have dedicated toolbar buttons for navigating your file history, as an alternative for the already existing keyboard shortcuts.

### Proposed Changes

It could look like this (forgive my gimp skills):

![zettlr-back-forth](https://user-images.githubusercontent.com/18646029/147892042-e73e3b06-d0b6-4944-a0be-36aee0115afb.png)

Similar to a webbrowser, when you have reached either end of the history, the corresponding button should be greyed out.

### Caveats

I don't believe there are any.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (but an older version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Manjaro

nathanlesage:
> It could look like this (forgive my gimp skills):

No worries, you don't know how easy it is to imagine something with an image attached, no matter the quality.

However, since there are now multiple issues open with requests to new toolbar buttons, and I don't want to clutter everyone's toolbar, I think it might make sense to first implement a way to programmatically show and hide toolbar buttons. Something that macOS has (with this drag and drop thing) might be cool, although pure checkboxes in the settings might be a good middle ground.

curtcox:
I've been using Zettlr for 5 minutes. I'm evaluating it for personal use along with other tools. I'm looking for the equivalent of web browser forward and back buttons. @kyaso indicated that keyboard shortcuts are available, but I can't find them listed [in the documentation](https://docs.zettlr.com/en/reference/keyboard-shortcuts/).
The act of writing up the work-around that I did find (File >> Recent Documents) led me to find Previous File and Next File in the same menu. I suggest using "Recent Files" instead of "Recent Documents" for consistency with the rest of the menu.

nathanlesage:
Implemented in 2.3.0

Issue #2685, opened 2021-10-20
[Enhancement] Button for New File (as was in former releases)


AIBN60:
### Description

Ctrl N or several clicks will lead to a new file. Former releases had a button to create a new file. Would like to see this again. Button == Ctrl-N

### Proposed Changes

Adding Button with Ctrl-N as function

### Caveats

-

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Linux 5.4.0-89-generic #100~18.04.1-Ubuntu SMP Wed Sep 29 10:59:42 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Additional Information

-

efficacy:
As it stands in 2.0.1 the behaviour of the "New File" menu option and the Ctrl-N action are subtly different. 
- The menu item respects the context of the click, and creates a new file in the currently selected folder, immediately prompting for a name inline in the file manager. It is clear at all stages where the file will be created and what its name will be.
- The Ctrl-N action creates a "floating" file which does not really have a name or location until it is saved, at which point you have to deal with the system file chooser panel. This can be useful for strange situations such as creating a file outside the usual workspace, but adds needless extra complexity for regular use.

The old "New File" button behaved more like the right-click menu action, and this is what I would like to see reinstated.

AIBN60:
Funny, but both ways to create a file show exactly the same beahviour as you described it for the menu item. Nevertheless, there is one point that is slightly different from your description:

> The menu item respects the context of the click, and creates a new file in the currently selected folder, 
Yes!

> immediately prompting for a name inline in the file manager. 
No! (nor anywhere else)

Ctrl-N does exactly the same as the Menu item, which surprises me: I thought I experienced a different behaviour earlier. Maybe it was before 2.0.1

efficacy:
> > The menu item respects the context of the click, and creates a new file in the currently selected folder,
> Yes!
> > immediately prompting for a name inline in the file manager.
> No! (nor anywhere else)
> Ctrl-N does exactly the same as the Menu item, which surprises me: I thought I experienced a different behaviour earlier. Maybe it was before 2.0.1

I am using 2.0.1 and I still see the difference. This may be a settings thing. Do you have the "do not prompt" checkbox set in the advanced settings?
[...]

PR #3430, opened 2022-05-01
Allow customize toolbar

EdgarTang:
## Description
Allows users to customize toolbar buttons as [#2987](https://github.com/Zettlr/Zettlr/issues/2987)

## Changes
1. Add a boolean variable named `visible` to `ToolbarControl` model. Only when the `ToolbarControl.visible` is set to `false` will hide the display content connected to it.
2. Preference adds a `toolbar` tab to set the visible value related to ToolbarControl
3. Added a button for adding new files, from [#2685](https://github.com/Zettlr/Zettlr/issues/2685)
4. Added buttons for the previous file and the next file, from [#2980](https://github.com/Zettlr/Zettlr/issues/2980)
5. This is the list of new translation strings:
dialog.preferences.toolbar
dialog.preferences.toolbar.showOpenPreferencesButton
dialog.preferences.toolbar.showNewFileButton
dialog.preferences.toolbar.showPreviousFileButton
dialog.preferences.toolbar.showNextFileButton
dialog.preferences.toolbar.showToggleReadabilityButton
dialog.preferences.toolbar.showMarkdownCommentButton
dialog.preferences.toolbar.showMarkdownLinkButton
dialog.preferences.toolbar.showMarkdownImageButton
dialog.preferences.toolbar.showMarkdownMakeTaskListButton
dialog.preferences.toolbar.showInsertTableButton
dialog.preferences.toolbar.showInsertFootnoteButton
dialog.preferences.toolbar.showDocumentInfoText
dialog.preferences.toolbar.showPomodoroButton

Tested on: 
macOS 12.3.1 (Monterey)
Debian GNU/Linux 10 (buster)
Windows 10


nathanlesage:
Thank you for opening the PR! It is also very nice to see that you managed to fix three issues with just one PR. This is very nice, indeed.

I will see that I am able to do a first round of review later today. I already spotted some things where we might want to implement changes. But overall, the technical execution looks pretty good!

scripts/test-gui/test-config.example.yml
```diff
@@ -83,3 +83,13 @@ watchdog:
   activatePolling: false
   stabilityThreshold: 1000
 checkForBeta: true
+customizeToolbar:
```
nathanlesage, line 86: Do we need these settings in here? This file is primarily useful if some settings should – in testing mode – be different from the other defaults (for example we always activate the debug mode which is off by default).

I don't think those toolbar buttons need to be in here, so to make the PR leaner, it would be best to remove the overrides here.

source/app/service-providers/config/get-config-template.ts
```diff
@@ -227,6 +227,19 @@ export default function getConfigTemplate (): ConfigOptions {
       zoomBehavior: 'gui' // Used to determine what gets zoomed: The GUI or the editor
     },
     checkForBeta: false, // Should the user be notified of beta releases?
+    customizeToolbar: {
```
nathanlesage, line 230: I would not call this property group `customizeToolbar` because while this is the *intention of this PR*, it is not the function these settings perform. Rather something like `availableToolbarButtons` or `showToolbarButtons`. Feel free to think about a good, descriptive name yourselves!

source/common/vue/window/toolbar-controls/Button.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="toolbar-group">
+  <div v-if="control.visible !== false" class="toolbar-group">
```
nathanlesage, line 2: Please do not do this. Basically, this will mount the corresponding component but hide it completely, which wastes unnecessary resources, and might even give us performance/memory issues if we ever use these components in larger amounts.

Instead, what you want to do is have a look at the `WindowToolbar.vue` component (location: `source/common/vue/window`). In there you will see a `v-if="item.type === <some type>"` directive. This will basically only render the component whose type matches. I.e. if a toolbar control has type `button` it will basically mount a new `Button.vue`-component and then provide the settings in the `item` to that Button component instance.

What you want to also add in there is a check whether the `visible` property is false, i.e. in total, for example, you would end up with `v-if="item.type === 'button' && item.visible !== false"` — but for every possible toolbar component.

source/common/vue/window/toolbar-controls/Text.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div
+    v-if="control.visible !== false"
```
nathanlesage, line 3: See above

source/common/vue/window/toolbar-controls/Toggle.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <button
+    v-if="control.visible !== false"
```
nathanlesage, line 3: See above

source/win-main/App.vue
```diff
@@ -766,6 +804,9 @@ export default defineComponent({
             this.$closePopover()
           }
         })
+    },
+    customizeToolbarVisible: function (configName: string): boolean {
```
nathanlesage, line 808: Is this REALLY reactive? (I still can't believe it, but I do admit I did not yet test it out myself 🙈)

nathanlesage (changes requested):
In general, a very solid piece of engineering. There is one path you took which is not ideal, which I have pointed out in my review. But it basically just involves moving a few pieces of code from one file to another, so the basic logic is flawless, it's just that we can do it a tad better.

It already looks very decent!

But one thing: You may want to add your names to the `toolbar.ts` header since you came up with the file, not me :)

scripts/test-gui/test-config.example.yml
EdgarTang: You are right, my original intention was to do this PR without changing the current panel display. For example the new file and previous/next buttons should default to false. Apparently I completely forgot about this. I will remove the settings here.

source/app/service-providers/config/get-config-template.ts
EdgarTang: How about `displayToolbarButtons`?

source/common/vue/window/toolbar-controls/Button.vue
EdgarTang: Thanks for your guidance. My original approach was equivalent to loading and then hiding. This is really unnecessary. I will changed them as you said.

source/win-main/App.vue
EdgarTang: Yes, it worked well.
At first I tried to find out how to input variables into computed. The answer I got was that while I could do it with a closure method, e.g.

```
computed: {
	closure() {
		return function (a, b, c) {
			/** do something */
			return data
		}
	}
}
```

But based on the principle of Vue
https://vuejs.org/guide/essentials/computed.html
This is no different from writing directly in the method. So I just did it。

But this seems to give up the cached data, maybe it will cause the program to be slow? I didn't find this to be the case during testing. Do you have a good way to test.

EdgarTang: By the way, the method name `customizeToolbarVisible` is not good, too. As above, how about `getToolbarButtonDisplay`

source/win-main/App.vue
```diff
@@ -766,6 +804,9 @@ export default defineComponent({
             this.$closePopover()
           }
         })
+    },
+    getToolbarButtonDisplay: function (configName: string): boolean {
+      return this.$store.state.config['displayToolbarButtons.' + configName] === true
```
nathanlesage, line 809: Just for confirmation: You checked that, when someone changes one of the settings in the preferences the toolbar automatically updates, correct?

source/win-preferences/App.vue
```diff
@@ -171,6 +172,12 @@ export default defineComponent({
           controls: 'tab-advanced',
           id: 'tab-advanced-control',
           icon: 'tools'
+        },
+        {
+          label: trans('dialog.preferences.toolbar'),
```
nathanlesage, line 177: Before merging, it would be great if you could list all NEW translation strings that you have added during this PR as an edit to the first comment of the PR, just so that I can easily copy and paste them into Zettlr translate!

nathanlesage (approved):
In general it looks good, just two minor things, but the code itself looks clean and seems to work!

source/win-main/App.vue
EdgarTang: > 



> Just for confirmation: You checked that, when someone changes one of the settings in the preferences the toolbar automatically updates, correct?

Yes, when the user changes the setting of toolbar preferences, the toolbar automatically updates.

source/win-preferences/App.vue
EdgarTang: @nathanlesage Sorry I don't make sure which one is the first comment of the PR..... I have edited the list of translation strings into the change detail of this PR

nathanlesage: Don't worry, exactly that's where I meant :)

source/win-main/App.vue
nathanlesage: Perfect, thank you for confirming!

nathanlesage:
Thank you for the PR so far; I just cloned the branch and I will do a quick testing myself sometime today. In the meantime, there is one last thing to do:

Please add a changelog entry to the CHANGELOG.md file in the root of the repository. There is at the very top the correct section for 2.2.7. Just add two lines or so describing what you added. Since you're contributing a new feature, please prepend `**Feature**:` to the entry. Feel also free to add all GitHub handles of the people who contributed to it in the form `(thanks to @handle and @handle)` if you'd like to!

Then I can merge the PR!

EdgarTang:
> Please add a changelog entry to the CHANGELOG.md file in the root of the repository. There is at the very top the correct section for 2.2.7. Just add two lines or so describing what you added. Since you're contributing a new feature, please prepend `**Feature**:` to the entry. Feel also free to add all GitHub handles of the people who contributed to it in the form `(thanks to @handle and @handle)` if you'd like to!

I have updated the changelog.

Thanks for your help, it made me learn a lot during the first pr process of the open-source project.
I'm happy to have contributed to this project!

nathanlesage:
> Thanks for your help, it made me learn a lot during the first pr process of the open-source project.
> I'm happy to have contributed to this project!

That is nice to hear! :)

I just tested it out myself and it works pretty nice! One thing I figured, however, is that we might add the ability to hide two specific functions: The Pomodoro timer (this way people can hide that button if they don't use that at all) as well as the settings button (which one doesn't strictly speaking need).

Would you mind adding those two? Sorry that I only saw this now, but I think it should be easy! Then I can merge it :)

EdgarTang:
> I just tested it out myself and it works pretty nice! One thing I figured, however, is that we might add the ability to hide two specific functions: The Pomodoro timer (this way people can hide that button if they don't use that at all) as well as the settings button (which one doesn't strictly speaking need).
Of course yes!

nathanlesage:
> I guess no user want to hide the button after starting the timer.

I also think so, and if they do worst they have to do is reload or reopen the main window and it'll work again. I suppose one won't really change this setting either way. Plus if I remember correctly I added a safeguard precisely for that reason.

I think you only forgot to add those config settings to the config template or the interface, the linter just complained. As soon as that works again, we can merge!

EdgarTang:
> I think you only forgot to add those config settings to the config template or the interface, the linter just complained. As soon as that works again, we can merge!

I think I meet the problem. The workflow check is stopped because of an error on the `yarn install` step for lint check. 
I cloned my fork again and run the `yarn install --frozen-lockfile` and `yarn lint` to check. However, there is no error.
How can I run the workflow again?

nathanlesage:
> The error log shows the error code is 500 from the HTTP model.

Ahh, yeah, the Github Action servers do that sometimes. Does the linter run without errors locally for you? If so we can ignore the failed check!

EdgarTang:
> Ahh, yeah, the Github Action servers do that sometimes. Does the linter run without errors locally for you? If so we can ignore the failed check!

No error.
I always check `yarn lint` and `yarn test` have no errors before makingcommit.

nathanlesage:
Wonderful, and now the checks as well report no error. Merging …! Thank you once again!

Issue #3057, opened 2022-01-17
[Enhancement] Sort order of root directories with identical names


jothanne:
### Description

Currently two or more root directories with identical names and different parent folders are sorted in an unknown order (see screenshot). With a growing number of identically named root folders it gets difficult to locate the one to open.
![screenshot](https://user-images.githubusercontent.com/30828470/149772079-7567a7af-f0a1-428d-8c43-4fe1a463e66c.png)
 

### Proposed Changes

It would be very helpfull to have an option to choose the sort order or do it manually via drag & drop.

### Caveats

--

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Operating System Version

Windows 10

a1765159:
Hello, I am part of a group in a course at the University of Adelaide.
We are planning to have a go at working on this issue for our open-source contribution project.

PR #3481, opened 2022-05-18
Root Workspace Sort Dropdown (Alphabetical)

csjasonl:
## Description
Added dropdown box in the workspace directory list to sort root workspaces alphabetically (ascending/descending). 
Addresses enhancement #3057 which was concerned with same-name root directories of different parent folders not being
ordered alphabetically.

_Image of the dropdown_
image

image

image

## Changes

- Added .ts file 'source/win-main/file-manager/util/sort-directories.ts' containing the sort function.
- Modified .vue file 'source/win-main/file-manager/file-tree.vue' to incorporate GUI dropdown and function usage.
	small dropdown textbox is positioned next to the Workspaces header, and scales with the sidebar and window size.
- Added unit test file 'test/sort-directories.spec.ts' to test function use cases. Runs alongside other unit tests.

## Additional information
**- The development team is doing this as part of a university course, and would appreciate any feedback**
- Translations are required for the new phrases:
** gui.sort_by_option
** gui.alphabetical_asc_option
** gui.alphabetical_desc_option
	_placeholder text is available until translation files are updated_
- The default behaviour for ordering root directories is still available as the default "Sort By" option.
- Other ordering types (e.g. by access or creation date) can be more easily implemented in future.
- Layered parent same-name directories aren't shown, but they may still sorted (e.g. Directory A/X/Root & B/X/Root are shown as Root (X)).

**Tested on:** **Windows 10 Home 64-bit (10.0 Build 19044), Ubuntu 20.04.4 LTS, macOS BigSur 11.6**


kyaso:
> 1. [...] we won't lose that ability just because we forget to implement your sorter.

I have no particular opinion on point 1, but if the quoted part is a potential issue, I would agree with moving the sort to the configuration (@nathanlesage: is "configuration" in this context equivalent to the "store.state.fileTree"?)

> 2. Should this really be done using a dropdown inside the workspaces header?

I would also say that the preferences might be a more suitable place for that.

> 3. There is an Intl.compare function, which I think the file sorting algorithm of the FSAL utilizes. This way, we can make sure that even non-Western scripts are sorted appropriately.

I agree with this one.

> 4. Lastly, I just remembered that I think the roots are already sorted to some degree, so this functionality here would potentially interfere with that. We need to check that.

Unfortunately, I'm not too familiar with that part of the codebase. But yeah, I guess it might be worth investigating.

nathanlesage:
Yes, I meant to implement it in the config provider, or in the FSAL. One of those two places already implements some form of root sorting, I think it should be the FSAL.

Issue #3469, opened 2022-05-15
Slow typing in large documents


waterfallpit:
### Description

I noticed slow typing (a certain lag with which the characters appear in the document) for large documents (e.g., ~200 headings, ~10 000 words). After recording with the DevTools I saw that TOC and document info (word count) are updated for every keypress. This is especially noticable on slower computers or when running other somewhat CPU-demandings apps (say, videoconference apps).

### Proposed Changes

Updating TOC and document info after a delay via a timer would speed up the typing with only a short delay for TOC and the document info update.

I implemented this by setting a 400 ms timer for `updateTableOfContents` when fired due to the editor's `change` event. Setting such a timer for word count would speed thing up even more. However, if I remove committing `activeDocumentInfo` from `cursorActivity` event, then the cursor position is not updated correctly.

### Caveats

TOC and document info are updated with a short (<1 s) delay.

### Do you Wish to Attempt Implementing this Yourself?

Yes

### Zettlr Version

Right off the develop branch

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Operating System Version

Windows 10

nathanlesage:
I habe actually implemented such a delay once, but removed it because it didn't feel snappy anymore and it became hugely annoying.

What about instead of utilizing a timeout, trying an idle frame? This way updating would be instantaneous if your computer is fast enough but preference would still be given to typing?

waterfallpit:
It felt snappy to me, but that's probably because I never look at the TOC or the word count or the position indicator when I'm typing.
I don't know how to implement it by waiting for the idle period other than to rely on the timer. Do you have something specific in mind?

nathanlesage:
I was thinking about [this](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback).

waterfallpit:
Thank you for the suggestion. I rewrote the code and tested both versions (with timeout and with idle callback) while running prime95. Both feel snappier than without a callback. I'll test them some more on a slower computer and report back. My concern is that requestIdleCallback often executes the callback right after a keypress and then the following keypress could be slowed down, since the callback is still executing and the new keypress does not cancel the callback once that one is called.

waterfallpit:
I was testing it today and requestIdleCallback does make a positive difference in responsiveness.
[...]

PR #3491, opened 2022-05-22
Table of contents: delayed updated and caching

waterfallpit:
## Description
Table of contents and document info are now updated when the program is idle. Caching of the headings' texts is added to speed up the processing. Addressing the enhanchment in #3469.

## Changes
Chnages are made to MainEditor.vue and store.ts.

Tested on: Win 10


source/win-main/MainEditor.vue
```diff
@@ -553,16 +557,27 @@ export default defineComponent({
         isClean: activeDocument.cmDoc.isClean()
       })
 
-      this.$store.commit('updateTableOfContents', mdEditor.tableOfContents)
+      // The sidebar needs the correct table of contents, so signal the
+      // corresponding event to the renderer
+      // Do it when idle.
+      window.cancelIdleCallback(this.updateTableOfContentsWhenIdleID)
```
nathanlesage, line 563: Does this work with the timeout being undefined?

source/win-main/store.ts
```diff
@@ -230,6 +230,10 @@ export interface ZettlrState {
    * This variable stores search results from the global search
    */
   searchResults: SearchResultWrapper[]
+  /**
+   * Cached texts of the headings
+   */
+  headingCache: { [key: string]: string }
```
nathanlesage, line 236: Since JavaScript now has support for Maps, I think we should be using those. Also -- according to the internet (tm) -- Maps are faster: https://bretcameron.medium.com/how-javascript-maps-can-make-your-code-faster-90f56bf61d9d

nathanlesage: I'm using Maps frequently in Vue components, feel free to search for `new Map()` in the codebase

source/win-main/store.ts
```diff
@@ -265,12 +270,24 @@ function getConfig (): StoreOptions<ZettlrState> {
     },
     mutations: {
       updateTableOfContents: function (state, toc) {
+        // This function is called every time a document is changed
+        // (e.g., a letter is typed). Most of the time the headings
+        // stay the same. To avoid unnecessary computation, their
+        // parsed texts are cached:
+        // state.headingCache[unparsedHeadingText] = parsedHeadingText
+        const headingTextToHtml = function (headingText: string): string {
+          if (!state.headingCache[headingText]) {
```
nathanlesage, line 279: Here you can then use `if (!state.headingCache.has(headingText))`

Also, I'd suggest to refactor the for loop down below to something like this:

```js
for (const entry of toc) {
  if (!headingCache.has(entry.text) {
	headingCache.add(md2html(entry.text))
  }
  entry.text = headingCache.get(entry.text) as string
}
```

Makes it more concise

nathanlesage (changes requested):
Looks decent, and if that really speeds up typing we're almost good to go!

nathanlesage:
This PR is fortunately no longer necessary. Specifically, the Table of Content-generation in the new Codemirror update is (a) already delayed appropriately, and (b) thanks to the new syntax trees blazingly fast (with a moderately sized document with around 30 headings it takes 0.3 Milliseconds to generate)

PR #3502, opened 2022-05-24
Fix translate key conflict

EdgarTang:
## Description
Fix translate key conflict bug for this [PR](https://github.com/Zettlr/Zettlr/pull/3430) 
In `~/source/win-preferences/App.vue` line 177 use translate key as ``'dialog.preferences.toolbar'``, it is conflict with translate keys in file `~/source/win-preferences/schema/toolbar.ts` which like  `'dialog.preferences.toolbar.showOpenPreferencesButton'`.

## Changes
rename `'dialog.preferences.toolbar'` to `dialog.preferences.toolbar.customize`

### DIFF
--- a/source/win-preferences/App.vue
+++ b/source/win-preferences/App.vue
@@ -174,7 +174,7 @@ export default defineComponent({
           icon: 'tools'
         },
         {
-          label: trans('dialog.preferences.toolbar'),
+          label: trans('dialog.preferences.toolbar.customize'),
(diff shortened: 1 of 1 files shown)


nathanlesage:
Thank you for spotting this! I have to fix this in Translate, hopefully tonight!

nathanlesage:
I've now fixed the conflict on Zettlr Translate, thank you for finding this and pointing it out!

PR #3593, opened 2022-07-07
Implement split panes and multiple windows

### LINKED ISSUE #2123: [FEATURE] Multiple split tabs a.k.a Multiplexer
## Description

Current Zettlr is limited to one single pane. While multiple tabs can be open, only one shows on screen.
Split pane would help when working with literature (e.g. definitions), drafts, or coding while following a statistical analysis plan
Split pane could be done horizontally, vertically, and it could be nested

This would be a great feature for Zettlr 2.0 

## Example Workflow
Here is an exemple using obsidian.md, another notetaking app. 

image

This picture contains 4 different panes which can all be split again independently of each others. 
Each pane can be edited, closed, or pinned, or exported. That means each pane is fully independent.
Additionally, each pane can have a different fonts size (ctrl + ±) 
Finally, pane can be moved and re-arranged using drag&drop, they will automatically adapt to the location (further vertical split or side by side)

Pane layout can be saved so the structure remains when quitting and re-opening the app (currently needs a plugin)
## Caveats

Unsure how synchronization would work with two panes open on the same file.
Should there be a limit to the number of splits?

## Additional Information
Implementation is well done in [obsidian.md](https://obsidian.md/features#_)  https://help.obsidian.md/Panes/Pane+layout 
The picture taken is from my own files and may be a bit extreme, two independent panes would be great already

Maybe their github provides some insight on how to do this: [obsidian github](https://github.com/obsidianmd)

### LINKED ISSUE #2358: [Feature] Multiple Instances and "Projects"
### Checklist

### Feature Proposal

Hi at [the forum](https://forum.zettlr.com/discussion/99/how-to-start-2-instances-of-the-app) it is mentioned that Zettlr is currently envisioned to be a single instance app. I want to encourage rethinking this idea because of the following reasons:

- when using zettlr as a zettelkasten it can appear that you want multiple sessions open, one for business, one for private for example.
- another usecase is to use zettlr as an ad-hoc markdown editor. Here you may want to leave your current setup in place, but just open a separate zettlr for quick editing which you can then close again afterwards.

Currently I have to use typora or marktext or vscode instead of zettlkasten to do so.

### Why reusing a central zettlr session is not an option
- you may want to use the different instances on different virtual desktops, screens, activities
- you may want to have different window sizes and positions for the different instances
- you may want to have both instances on the same desktop next to each other

### Essentially needed additions for zettlr.
- In order to be able to reopen a specific session like business, a meta-workspace concept, lets called it "project" would be needed
- the workspaces, tag management and all other config could be shared among projects, only the folder selection and open files would differ
- alternatively each project could also have individual workspaces, however I wouldn't prefer this right now, as I like the feeling of one global zettlr which the sharing of workspaces would enable.

### Example Workflow

launching Zettlr a second time will open a new blank Zettlr

it would be nice to have the option to load an existing Zettlr "Project" by clicking on some button in a summary-user-interface

### Caveats

you may loose the convenience that when starting zettlr your last setup is restored,
however this should be the case if you have multiple projects, and it is ambiguous which project to load

if there is only one project, the current zettlr behaviour could be retained
if there are multiple it may be best to start with an empty zettlr and a list of loadable "projects"

No

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

### Additional Information

_No response_

### LINKED ISSUE #2480: Inconsistencies among the different ways to create a new file.
### Description

I have tested the 3 ways I'm aware of to create a new file. the behaviour of Zettler differs depending on the chosen way and this is quite onfusing.

1. using the menu bar file option and then selecting the type of file, i.e. : .md, .tex, .json, yaml:
- Zettlr asks to choose a type of file,
- Zettlr creates the file with the ID as name but does NOT prompt for a filename and the file is NOT saved on disk
- When using the "save" command after the file creation, the proposed directory where to save the file is NOT  the directory currently selected

2. using the shortcut Ctrl-N
- Zettlr creates a .md file (no choice of the type of file) with the ID as name but does NOT prompt for a filename and the file is NOT saved on disk
- When using the "save" command after the file creation, the proposed directory where to save the file is NOT  the directory currently selected

3. right-clicking from a selected directory (being a workspace or not)
- Zettlr prompts for a filename (_proposing the ID as name_) and when "enter-clicking" the proposed or the changed filename but Zettlr DOES save the file on disk in the selected directory. The created file is always of type .md

It would help to bring the following changes:
- always prompting for a filename independently of the way of creating a new file
- the proposed directory (in the file explorer) where to save the newly created file, should always be the currently selected directory
 

### Reproducing

Create a new file, resp. by: 
1. using the menu bar file option
2. using the shortcut Ctrl-N
3. right-clicking after having selected a directory

### Zettlr Version

Beta (if applicable)

### Specify version

2.0.0-beta.4

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Win 7

### Additional Information

_No response_

### LINKED ISSUE #3382: [Enhancement] – Please implement support for vertical editor vertical split view , so we can compare notes.
### Description

Hello,

I suspect that this will not be a very simple enhancement to implement, but it will be one that will be very useful to compare the text of two notes. Or a Fleeting note to a Permanent note.
Here I’m thinking to cover on 95 % use cases with minimal functionality implemented.
So, one could not edit the same files in different split panels, one file was on one panel or was on the other panel and only one of it was considered the active editing file, the one with the focus.
So, the user also could not edit the same file, in two different places ex: at the beginning and at the end, to simplify things.
Could have only vertical splits and not horizontal splits to simplify things even more.

I think that this compare of notes in different panels of a split view, is a very important feature.  

Thank you,

Best regards,
João

### Proposed Changes

To add the possibility to see at the same time two notes side by side on the editor.

### Caveats

Maybe difficult to implement.

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Linux Ubuntu 20.04

### Additional Information

_No response_

nathanlesage:
**Even though this PR is not yet ready, feel free to already review so we can fix things along the way! Every review is greatly appreciated, even if you only find a single typo!**

## Description

This PR implements the long-awaited multiple windows/split panes features. This is one of those PRs whose description is so much easier than its implementation:

1. Allows to open more than one main window
2. Allows to split the editor area into many smaller editors

## Changes

Quite a lot. I have summarized all my changes (insofar as they affect the API) in the CHANGELOG.md. In short, this is it:

* Documents are now represented as tree structures:
    * Top-level are windows, identified by a UUID4 identifier. These identifiers are shared between the document provider and the window provider
    * Within each window, there is a single DocumentTree.
    * The document tree consists of branches and leafs. Both are identified with UUID4 identifiers as well
    * Branches enable splitting into the horizontal and the vertical direction and hold information about the relative sizes of the child nodes
    * Leafs, then, consists of the previous contents of the document manager (main process) and main editor split (renderer process). They contain now so-called "Tab managers" that basically borrow all the functionality from the documents provider, because they are now managed decentralized
    * Each leaf has its own tab manager that manages its open files as well as the active file
    * A document tree can be serialized to JSON, which enables us to both persist it to a configuration file as well as send it to the renderer
* In the renderer, this has the following implications:
    * Similarly to the FSAL updates, I have implemented a delta patch algorithm that tries to apply as few updates as possible. This is mainly due to Vues reactivity, which will happily rerender everything under the sun as soon as the tiniest variable changes
    * This also means that, inside the renderer, we disentangle the data of the document tree (the open files and active file) from its structure (branches and leafs). This is necessary because otherwise Vue will always re-instantiate the editors, making them basically unusable
* The modified files are still managed centrally. This is because if you have a single file open in two or more leafs/windows, it needs to be counted as modified in all those places. The documents manager will take care of keeping track of what is modified, and even keep copies of all modifications in order to propagate those changes across renderers
* Creating/deleting leafs/editor panes
    * The creation of leafs or editor panes works by dragging and dropping document tabs.
    * This means you can still re-order document tabs, but if you drag those tabs completely out of the tabbar, you can drop them on editor instances. If you drop a document tab onto the edge of one of the open editors, this will initate a split (+move the corresponding file to the newly created leaf), if you drop it into the center of an editor, this means it will move the file into that leaf (without creating a new one)
    * The deletion of a leaf can be done by closing every file in the leaf. Additionally, you can right-click on the document tab bar and select to close the pane
* Two settings have changed: Now, by default Zettlr doesn't try to avoid new tabs, because they can be spread out easier. Also, it will now by default load remote changes. The old behavior can be restored.
* Lastly, there are no more unsaved in-memory files. I didn't have the patience to reimplement that on top of all the synchronization efforts that are going on to keep the states themselves in sync. Plus, users were confused by that either way.

## Additional information

This PR closes the following issues:

* Closes #2123
* Closes #3382
* Fixes #2480
* Closes #2358
* Please add any additional issues this PR touches

Tested on: macOS Monterey 12.3.1

## TODOs

* [x] Since we cannot access the individual editor instances, many things such as the toolbar buttons do not yet work. I have to think about a clever system of re-enabling that [solved: It's supported by Vue, but it looks super weird]
* [ ] A LOT of cleanup
* [ ] Extensive testing (I'm currently testing this beast in production to find all those things that don't yet work)
* [x] Currently document tabs spit out everything in the form `<window ID>:<leaf ID>:<absolute path>`. Was very clever, until I realized how absolute paths looked like on Windows after checking the PR on my Windows machine ...
* [x] Distraction free mode and readability mode need to be reenabled by integrating them into my solution for TODO number 1 above


nathanlesage:
Soooooo, I think I got everything/most things. I'm sure there are still some bugs lurking, but I've been using this PR in production for myself for two weeks and it's working surprisingly smooth.

I understand that reviews on this monstrosity will take extraordinarily long, so I don't judge anyone who doesn't want to do this over holidays :D

I'll be on vacation beginning Monday for two weeks, so I won't be looking in here very often, so all chill.

Nevertheless, it would be great to have a few reviewers here, maybe again @kyaso, and @tobiasdiez ? Maybe also @tarleb, although I would completely understand if y'all thought I'm stupid. If you do a review, I'm gonna owe you a nice evening in a pub with lots of beers! :)

nathanlesage:
I have already anticipated that probably no one has the time/skills/motivation to review this massive piece of PR, I'm just going to merge it into develop for today's nightly. I have used it for several weeks now and I could not find any major bugs or anything that would lead to data loss. Any more testing will then be done via the nightlies/the next betas.

Issue #2213, opened 2021-07-21
[Enhancement] Improve hyperlinks rendering in tables


raphraph:
## Description

In Zettlr tables adding an hyperlink when in **dark mode** renders a hard to read link (blue link on dark background).

Furthermore when the table is in table-editor mode it is not possible to directly click (or CTRL-Click) the link to open the link in browser.
When the table is in markdown edition mode there is a tooltip that appears when hovering the link and allows to directly click the link.

## Proposed Changes

- Define a different light link color for the dark-mode.
- Add a tooltip when hovering a link in a table when in table-editor mode.

## Caveats

I don't know.

## Additional Information

None

nathanlesage:
The old PR was very outdated. Also I just double checked, and at some point down the line, I appear to have made a change to the link rendering so that it now looks much better!

PR #3599, opened 2022-07-11
Improve hyperlinks rendering in tables

L20L021902:
## Description

This PR adds `cma` css class to `a` tag when table is rendered. It also removes `href` attribute from `a` tag and adds `title` attribute with links.

## Changes
- Changes are done in `table-editor.ts` file. All `a` tags selected then loop through the tags. If `a` tag does not contain `cma` class then add the class to it.
- The `link-tooltips.ts` search `cma` class and `title` to add tooltip on mouse hover. So, `href` removed and `title` added to `a` tags.

https://github.com/Zettlr/Zettlr/blob/f8d0ee7821a2607ecf6542185d2e5e45e8e3f1de/source/common/modules/markdown-editor/hooks/link-tooltips.ts#L38-L49

## Additional information

Closes https://github.com/Zettlr/Zettlr/issues/2213

Tested on: Windows 11

image

### DIFF
--- a/source/common/modules/markdown-editor/table-editor/table-editor.ts
+++ b/source/common/modules/markdown-editor/table-editor/table-editor.ts
@@ -322,6 +322,7 @@ export default class TableEditor {
         cell.addEventListener('blur', (event) => {
           this._onCellBlur(cell)
         })
+        this._updateUrl(cell)
       }
     }
 
@@ -344,6 +345,9 @@ export default class TableEditor {
     // Re-render the table element and save the textContent as data-source
     this._ast[row][col] = cell.textContent ?? ''
     cell.innerHTML = this._md2html(this._ast[row][col])
+
+    // update url with cma class
+    this._updateUrl(cell)
 
     // For a short amount of time, the table won't have any focused
     // elements, so we'll set a small timeout, after which we test
@@ -982,5 +986,22 @@ export default class TableEditor {
     }
 
     return caretPos
+  }
+
+  /**
+   * update A tags which does not contain cma class
+   * @param cell
+   */
+  _updateUrl (cell: HTMLTableCellElement): void {
+    for (const a of cell.getElementsByTagName('A')) {
+      if (!a.classList.contains('cma')) {
+        const url = a.getAttribute('href')
+        if (url !== null && url.trim() !== '') {
+          a.removeAttribute('href')
+          a.setAttribute('title', url)
+          a.classList.add('cma')
+        }
+      }
+    }
   }
 }


source/common/modules/markdown-editor/plugins/render-tables.ts
```diff
@@ -86,6 +86,16 @@ function writeTableToDocument (cm: CodeMirror.Editor, table: TableEditor): boole
   // because such a table header WILL NEVER be on the last line, plus
   // this way we can check for Setext headers without having to worry.
 
+  // render links in table
```
nathanlesage, line 89: I would suggest to do that in the TableEditor's element itself, since the editor just searches through all those elements already, and by performing this check only when a table cell has changed this makes it both more performant and more compartmentalized. The TableEditor requires further changes anyways.

nathanlesage (commented):
In general a good idea, but it would be great if you could check my suggestion if that works as well, as it's then less prone to break.

source/common/modules/markdown-editor/plugins/render-tables.ts
krmanik: I have implemented using suggestion.

source/common/modules/markdown-editor/table-editor/table-editor.ts
```diff
@@ -983,4 +987,22 @@ export default class TableEditor {
 
     return caretPos
   }
+
+  /**
+   * update A tags which does not contain cms class
+   * @param cell
+   */
+  _update_url (cell: HTMLElement): void {
```
nathanlesage, line 995:
```suggestion
  _updateUrl (cell: HTMLTableCellElement): void {
```

source/common/modules/markdown-editor/table-editor/table-editor.ts
```diff
@@ -983,4 +987,22 @@ export default class TableEditor {
...
+    let aTags = cell.getElementsByTagName('A')
+    for (let a of aTags) {
```
nathanlesage, line 997:
```suggestion
    for (const a of cell.getElementsByTagName('A')) {
```

source/common/modules/markdown-editor/table-editor/table-editor.ts
```diff
@@ -983,4 +987,22 @@ export default class TableEditor {
...
+      if (!a.classList.contains('cma')) {
+        let url = a.getAttribute('href')
```
nathanlesage, line 999:
```suggestion
        const url = a.getAttribute('href')
```

source/common/modules/markdown-editor/table-editor/table-editor.ts
```diff
@@ -983,4 +987,22 @@ export default class TableEditor {
...
+        a.removeAttribute('href')
+        if (url != null) {
+          a.setAttribute('title', url)
+          a.classList.add('cma')
+        }
```
nathanlesage, line 1004:
```suggestion
        if (url !== null && url.trim() !== '') {
          a.removeAttribute('href')
          a.setAttribute('title', url)
          a.classList.add('cma')
        }
```

nathanlesage (changes requested):
A few minor changes making the code more efficient & readable

krmanik:
> A few minor changes making the code more efficient & readable

I have updated it.

nathanlesage:
This PR is now probably already very outdated. The new TableEditor may improve things, however (if not, we can still get back to it!)

PR #3602, opened 2022-07-12
Update MainEditor.vue

duanxunxu:
## Description
We open Zettlr, using the file reference function, If two files exist: topic-subject.txt and topic-subject.md. The two files are of different types but have the same name, We use zettlr's own link recognition when quoting, but the problem is that the automatic generation of links provided by the program does not generate suffixes, If you do not add a suffix, only the .md file is recognized by default, We need to add our own suffixes to find the file we're really looking for. This results in automatic link generation of .txt files even if you select the generated link only recognizes the md file by default.
This is an image
This is an image
This is an image

## Resolution process
1.We first clone the entire file to the local and then configure the environment. When setting up the environment, if you report an error in the preparing native dependencies of the yarn run, use git bash to perform the yarn start, and then this step in the git bash can be successfully executed and may be stuck in the position of the main process code……We force the exit and then execute the yarn start with cmd to solve it.
This is an image
This is an image

2. We open the program, starting from the pop-up window referenced by the link, first find the mapping file of Chinese characters and then find it all the way according to the mapped variables, and there are two very important files that need our attention during the search process.
This is an image
This is an image
One is source\common\modules\markdown-editor\hooks\autocomplete.ts, most of the code in this file is related to code completion, such assetAutocompleteDatabase() for the preparation of various types of database data, hintFunction() for code completion (in the file index completion, This concludes by judging whether another name needs to be generated after the index based on the user's settings.
The other is source\win-mian\MianEditor.vue, This MainEditor.vue file is a vue file of the editor interface, and the various configurations in it are basically editor-related, and its role is: Remove separate openFiles from store, One of the methods is important: Vue Langue Features(Volar)>{} script>default>methods>updateFileDatabase，The method is to do an initialization of the file data and then pass it to the part that needs to be processed for processing. The project here uses a FileDatabase data structure to store all the file information there are three important fields are displayText, text, id, The text field is used when the code is completed, but observing the updateFileDatabase() method, we find that he has already removed the file suffix when the field is initialized at the beginning, which leads to the problem we encountered.
This is an image
	Our solution is to rewrite fname so that this field does not remove the file suffix when it is initialized. This will solve the problem we have.
This is an image
When we use the software again after the modification is completed, we can find that the file suffix appears in his automatically generated link, and we reference the file, which can be accurately identified, and will no longer be recognized as an md file by default.
This is an image
This is an image
This is an image

### DIFF
--- a/source/win-main/MainEditor.vue
+++ b/source/win-main/MainEditor.vue
@@ -685,8 +685,12 @@ export default defineComponent({
       const fileDatabase: any = {}
 
       for (const file of this.fsalFiles) {
-        const fname = file.name.substr(0, file.name.lastIndexOf('.'))
+        // const fname = file.name.substr(0, file.name.lastIndexOf('.'))
+
+        const fname = file.name
+
         let displayText = fname // Fallback: Only filename
+        
         if (this.useTitle && typeof file.frontmatter?.title === 'string') {
           // (Else) if there is a frontmatter, use that title
           displayText = file.frontmatter.title


nathanlesage:
While I like the simple approach, a better way would be the following: if the file autocomplete database ends up with duplicate entries, this is an indication that there are both an md and a txt file in there. This would then indicate that the extension is meaningful there, and then you can add the extension only to those duplicate entries to disambiguate the files. This would keep the no-file-extension default and add those solely for duplicate filenames.

nathanlesage:
Hej, as there unfortunately hasn't been much activity and because the whole system has now changed from the older CodeMirror 5 implementation to version 6, I'll close this PR. Please feel free to open a new PR addressing CodeMirror 6 if you have the time! Thanks!

Issue #1328, opened 2020-10-08
[ENHANCEMENT]: Scale images based on Pandoc classes.


Kangie:
## Description

Zettlr knows to hide Pandoc classes assigned to images: `{height=16px width=16px}`. 
It would be nice if Zettlr could parse these values and pass them to the renderer. 
## Proposed Changes

Zettlr should parse `height` and `width` values associated with images and pass this to the renderer.
## Caveats

Nil.
## Additional Information

Happy to investigate implementing this, just wanted to add an enhancement for it. :)

nathanlesage:
Sounds good!

nathanlesage:
Works by now

PR #3608, opened 2022-07-14
Scale images based on Pandoc classes

zzccppp:
## Description

This PR is for scale images in issue #1328. Documents imported with pandoc usually contain image size information, such as:
`{width="3.69in" height="1.93884in"}`
So just parse the information in braces and pass it to the corresponding attribute to control image size.
## Changes

- parse the parameters in the braces and pass them to image.style.maxWidth attribute

## Additional information

Tested on: macOS 12.4 Monterey Apple Silicon


source/common/modules/markdown-editor/plugins/render-images.ts
```diff
@@ -1,202 +1,227 @@
...
+      let altText = match[1] ?? '' // Everything inside the square brackets
+      let url = match[2] ?? '' // The URL
+      let title = match[3] ?? altText // An optional title in quotes after the image
+      let p4 = match[4] ?? ''
+
+      let widthAttribute = ''
+      let heightAttribute = ''
+      if (match[4] !== '') {
+        // Use regex to parse pandoc attributes
+        // like width="3.69in" height="1.93884in"
```
nathanlesage, line 70: Two questions: 1) are quotes required? If not one should account for that in below's regexp. 2) Regarding size attributes: Obviously, for preview-reasons we don't need to accurately represent inches etc., but what are the options for measures people have, and how should we maybe account for that? (50% width is easy to accomplish, but how much are 3.5 inches relative to the editor area?)

source/common/modules/markdown-editor/plugins/render-images.ts
```diff
@@ -1,202 +1,227 @@
...
+        let widthRE = /(width=)"([^"]+)"/g
+        let heightRE = /(height=)"([^"]+)"/g
```
nathanlesage, line 72:
```suggestion
        const widthRE = /(?:width=)"([^"]+)"/g
        const heightRE = /(?:height=)"([^"]+)"/g
```

This makes the first group non-capturing, i.e. below you can access them with `mat[1]` instead.

source/common/modules/markdown-editor/plugins/render-images.ts
```diff
@@ -1,202 +1,227 @@
...
+      caption.addEventListener('focusout', updateCaptionFunction)
+
+      // Retrieve the size constraints
+      const maxPreviewWidth = Number((cm as any).getOption('zettlr').imagePreviewWidth)
+      const maxPreviewHeight = Number((cm as any).getOption('zettlr').imagePreviewHeight)
+      let width = (!Number.isNaN(maxPreviewWidth)) ? `${maxPreviewWidth}%` : '100%'
+      let height = (!Number.isNaN(maxPreviewHeight) && maxPreviewHeight < 100) ? `${maxPreviewHeight}vh` : ''
+
+      // Use the pandoc attributes to control image size
+      if (widthAttribute !== '') {
```
nathanlesage, line 181: What would need to change if you were to preset those two attributes with `undefined` instead? That might make more semantic sense, and would probably alleviate, e.g., the needs for the `?? ''` operator above?

nathanlesage (changes requested):
In general looks good, however a few questions:

1. Unfortunately, GitHub displayed the file as if it had been removed and completely written from scratch. After turning off Whitespace change, it worked like a charm. Did you accidentally exchange, e.g., spaces with tabs, or LF with CRLF…?
2. There are two critical issues we have to think about: (a) the metrics question I've already posed, and (b) how to handle the settings' `maxPreviewWidth` and `maxPreviewHeight`?

source/common/modules/markdown-editor/plugins/render-images.ts
zzccppp: I try to use pandoc to import docx documents. The attribute generated by pandoc is in inches. Every available length unit in CSS images can be used here. For the absolute length units (cm mm in px), its size is fixed, does not change with the window, and corresponds to the same px (1in = 96px).
As for the maxPreviewWidth and maxPreviewHeight, if absolute length units are used, conversion is required. I don't know how to calculate it.

nathanlesage:
This has at some point in the past been implemented.

PR #3611, opened 2022-07-16
change Exporter tab to Citations

radifar:
## Description

This PR is to address #3610 

## Changes

I change the the `Citing tutorial with Zettlr` in 3 language.

## Additional information

For `de` and `fr` I change it according to `static/lang/de-DE.json` and `static/lang/fr-FR.json`. Please let me know if I'm mistaken in updating the tutorial.

Issue #3610, opened 2022-07-16
Incorrect tab name in Citing tutorial


radifar:
### Description

Hi, I'm using the newest Zettlr (2.3.0) and I found that the citing tutorial is using the incorrect name tab. It says that to load the CSL JSON I need to open the `Export` tab in the preferences, but the correct one is actually `Citations`.

I do not know if it is an artifact from the older version or something. Anyhow I'll do the PR for this later.

### Proposed Changes

Update the citing tutorial from `Export` to `Citations` in 3 languages (en, de, fr)

### Caveats

No

### Do you Wish to Attempt Implementing this Yourself?

Yes

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Fedora 36

PR #3653, opened 2022-08-01
Fix wrong tab name in citation tutorial

spcified:
## Description
The citation tutorial is referring to the `Export` tab within setting in order to load the reference database. This is wrong there is now way to specify the database from this tab.

## Changes
Fixed by proving the correct tab name `Citations`

## Additional information
Not required only changes within docs/tutorial


spcified:
issue #3610 is already addressed by PR #3611

PR #3776, opened 2022-09-16
Codemirror 6 Upgrade

nathanlesage:
## Description

Okay, I think it's time to make this PR a tad more visible. **~~This PR is not yet ready for review!~~ It's ready now!** You may, however, if you're curious, build it and run it in test mode (`yarn test-gui --clean`) to have a first look.

I must admit, I got really pissed with tons of pieces of the codebase and decided to clean up even completely unrelated parts (such as the FSAL). That explains the hefty amount of changed files (188 at the time of writing).

Was this a clever decision?

No. Not at all.

Was it a satisfying action?

You bet!

## Changes

The following ist is guaranteed to be incomplete, but I'll give it a shot:

1. Exchanged Codemirror 5 with Codemirror 6 (currently: Only the main editor, still todo: the smaller code editors)
2. While migrating the plugins from 5->6 gave most of them a facelift, thanks to, first, a MUCH better API, and second, a godsent AST
3. **Documents are now managed CENTRALLY in the DocumentsProvider**. Why does that matter? Because now there is **one** source of truth for all documents. This makes the shutdown logic much simpler, keeps every editor leaf in sync, and a ton of other benefits. Plus, it's fancy!
4. I finally decided to merge the Meta-descriptors in the FSAL with the "regular" ones. The distinction into meta and non-meta descriptors didn't make sense and is 100% on me. Now it's much better.
5. Disentangled the FSAL, Documents, and Target providers which previously were much tighter entangled. This had the problem of a thousand and one hard-to-track side-effects between the providers.
6. The FSAL now specifically has two utility functions that will return a descriptor for a random path on disk (as long as it's supported) and one that will return a (supported) file's contents. No more poking around in the guts of a provider.

## Additional information

Still TODO:

- [x] Move the CodeEditor to Codemirror 6
- [x] Re-implement the TableEditor
- [x] Allow for granular control of the renderers (currently hardcoded due to limitations in my imagination regarding the new config system)
- [x] Put the extensions into compartments for easy-ish reconfiguration of existing documents
- [x] Re-enable spell-checking
- [x] Implement a Markdown linter since that re-uses 99% of the code required for the new spell checker and because I can.
- [x] Write a few code-fold extensions to allow for better code folding, even where we're limited to the older `StreamParser`s
- [x] Better styling for the tooltips (which are now no longer sponsored by tippy but by the library itself)
- [x] Better styling for the autocompletes (strictly speaking also "only" tooltips)
- [x] Re-implement the emphasis-renderer because now we can do that properly (thanks to the AST)
- [x] More careful (secure) inline-HTML rendering (since we can do that now, thanks to the AST)
- [x] Re-implement iFrame rendering
- [x] When the user closes the last tab, empty out the editor
- [x] Add an exemption for the autocorrect to not replace --- with an em dash when it's the first line of the file (collides with YAML frontmatter)
- [x] Spellchecker currently takes sometimes more than a minute to check a word (due to the overhead of approx. 50-70ms per IPC call), I need to bundle multiple words together (basically: string array in, array of booleans out)
- [x] Implement `==.+==`-highlighting
- [x] Fix styles of emphasis/strong emphasis in and outside of lists

Tested on: macOS 12.6 (Silicon/ARM)


nathanlesage:
NOTE: I have just switched to this PR branch for my production setup. I will continue to test the new editor component. In a few days, I think it'll be good to go for a few more beta testers to try it out.

nathanlesage:
## Build Instructions for those who want to help out

Everything here assumes you're using a Terminal; use the Git Bash terminal on Windows after installing Git.

1. Make sure Node.js, Yarn, and Git are installed on your system (varies depending on your system)
    1. For macOS: [Install homebrew](https://brew.sh/)
        * `brew install node@16` (`brew install node` will install Node 18, which doesn't work)
        * `brew install yarn`
        * `brew install git`
    2. For Windows: Install Node 16, Yarn, and Git using their respective installers
    3. For Linux: Use your software repository (e.g. `yum` or `apt-get`) to install the three things (also ensure that Node is version 16: `node --version`)
2. Clone the repository to a folder in your system: `git clone https://github.com/Zettlr/Zettlr.git path-to-your-desired-folder`
3. Go into the directory: `cd path-to-your-desired-folder`
4. Switch to the correct branch: `git checkout codemirror-6`
5. Install dependencies: `yarn install --frozen-lockfile`
6. Launch the app in testing mode: `yarn test-gui --clean` (omit the `--clean` when you start the second time, except if something broke, `--clean` will just nuke the test directory and create it anew)
7. If you feel safe to use it in production: `yarn package` and then use the installer (on macOS it will already spit out an App-bundle, for Windows/Linux you can optionally also create a release package, but that may fail on Windows due to a lack of code signing certificates) — refer to the README.md-file in this repository for more commands/options

nathanlesage:
Two months in, I declare the PR successful and ready to merge! Have fun ;)

PR #3823, opened 2022-10-14
using deb-get as distribution mechanism

takov751:
## Description
In the readme I'm mentioning deb-get as a delivery and update mechanism for those, whom want to use deb to install software

## Changes
Just added few line to Readme to explain how to install deb-get bash script , install and keep updated the deb version without manually downloading anything later.
## Additional information

Tested on: Debian and all Ubuntu flavours

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,25 @@ With Zettlr, writing professional texts is easy and motivating: Whether you are 
 
 To install Zettlr, just [download the latest release](https://www.zettlr.com/download/) for your operating system! Currently supported are macOS, Windows, and most Linux distributions (via Debian- and Fedora-packages as well as AppImages).
 
+Debian and Ubuntu based distros:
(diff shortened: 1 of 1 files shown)


takov751:
https://github.com/wimpysworld/deb-get/issues/579
This PR was made as part of the mini-event at deb-get project to inform users of Zettlr that the deb version can be installed and updated simply.

nathanlesage:
Hey, sorry for the late response, I just overlooked your PR due to the influx of issues for the nightly! It's an interesting approach, but I don't want to overload the README file (it's already pretty large with all the development stuff in there).

I have a suggestion: Since I want the README to be more towards the developers and the documentation more towards end users, we could add a generic paragraph here in the README and direct people to the installation instructions in the documentation ("To see all available installation methods, see our documentation" or something like that) and add the `deb-get` mechanism there?

PR #3864, opened 2022-11-04
WIP: Replace nspell with nodehun

### LINKED ISSUE #899: Brazilian Portuguese
Could you add Brazilian Portuguese spellchecker?
And, perhaps, vim key-bindings?
Thx

### LINKED ISSUE #1539: Zettlr freezes when trying to activate pt-BR dictionary
## Description
Every time I try to activate the Brazilian Portuguese dictionary, Zettler freezes. CPU usage sparks to 100% or more.

## Reproducing
I tried installing the pt-BR dictionary from https://github.com/wooorm/dictionaries/tree/main/dictionaries/pt (this github repository is on the Zettler documentation). I just downloaded the dictionaries created the pt-BR folder inside the right directory and renamed the dictionaries to `pt-BR.dic` and `pt-BR.aff` but when trying to activate them on the Zettlr preferences makes the whole software to freeze and one of the CPUs to spark to a 100% usage.

## Expected behaviour
I expect the spell check to work for this language

## Platform

``` sh
$ cat /etc/lsb-release 
DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=20.2
DISTRIB_CODENAME=Nibia
DISTRIB_DESCRIPTION="Manjaro Linux"
```

 - Zettlr Version: I'm using Zettlr 1.8.3 from AppImage package

### LINKED ISSUE #3354: [Bug] Crash while installing Portuguese pt-PT dictionary is stalls the program forcing it to exit.
### Description

I have tried multiple times to install the pt-PT Portuguese dictionary and from the URL that you kindly  provide in the manual, the Portuguese language dictionary is your use case in the manual.

GitHub wooorm - dictionaries
https://github.com/wooorm/dictionaries/tree/main/dictionaries/pt-PT

It appears in the list of dictionaries in the configuration window, but when I select it in stalls the Zettlr program and forces me to exit the program. 

I tried also a different version of the dictionary directly from the pt-PT project at uminho from the maintainers of the dictionary and the behaviour is the same.

### Reproducing

Download the pt-PT dictionary from the github site:

GitHub wooorm - dictionaries
https://github.com/wooorm/dictionaries/tree/main/dictionaries/pt-PT

menu -> import dictionary

Create directory pt-PT
copy files:
pt-PT.dic
pt-PT.aff

then select the configurations window and the dictionary pt-PT appears correctly.
Select the dictionary and exit the window.

A window will appear forcing you to exit because the Zettlr program will stall. 

### Zettlr Version

Stable (most recent version)

### Specify version

2.2.4

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Ubuntu 20.04

### Additional Information

Thank you for all the work!

### LINKED ISSUE #3451: Italian dictionary loading problem
### Description

Is the problem with the Italian and Portuguese dictionaries resolved ? I saw that the [issue1022](https://github.com/Zettlr/Zettlr/issues/1022) was closed. Still, I cannot load Italian dict. Any chance will be fixed in the near future ? 

### Reproducing

Making the `it` directory and moving  `index.aiff` and  `index.dic `files inside. 

### Zettlr Version

Stable (most recent version)

### Specify version

_No response_

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

osx Catalina 

### Additional Information

_No response_

nathanlesage:
Hey all,

since we have a ton of problems with the spellchecker, as the current solution – `nspell` – fails at loading more complex dictionaries such as the Portuguese, Brazilian, or Italian ones, I just attempted to simply load hunspell directly. More specifically, I chose to add https://github.com/Wulf/nodehun as an alternative.

## The Good News: It works! The formerly problematic dictionaries load just fine!

While in development mode (e.g. via `yarn start`), everything works smooth and as it should be. Which is very nice! I tried it out with said, problematic dictionaries, and none complained!

## The bad news: I can't get it to bundle!

Now to the problem. I don't exactly know what the problem is, but I can't get a working bundle. Here's what I have tried:

* "As it is supposed to be": In this instance, it works fine in development mode, but as soon as I bundle it, Electron complains that "there is no module named 'nodehun'"
* Upon deeper digging, I found out that the `node_loader` is not picking the module up, because at no point is a `.node` file being imported. Changing the import from `nodehun` to `nodehun.node` obviously does not work.
* Importing `nodehun/build/Release/Nodehun.node` (the actual library) ensures that it gets bundled and that Electron will actually attempt to open the file. However, it turns out that that library isn't actually code-signed, so my Mac will complain. Plus, Typescript hates me because it "cant find the type declarations" (which are there, but only if you import 'nodehun')

To sum it up: I currently suspect that it's an interaction between the webpack assets relocator (since it will only move binary node files if they actually end in `.node`), and forge (which may expect libraries to-be-signed to be actually moved by the relocator).

I have now spent way too much time trying to get this to work, now I'm turning to whoever has an idea of how to fix this (ideally for all build processes, not just macOS).

## Related Issues

* Fixes #3354
* Fixes #3451
* Fixes #1539
* Fixes #899

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,12 @@ there.
   managed by the document provider
 - Completely removed all instances of `hash`; the FSAL cache now is being called
   with absolute file paths. Hashed paths are only used to determine the shard.
+- Exchanged `nspell` with `nodehun` so that we can load any Hunspell-compatible
+  dictionary file in Zettlr, including the previously too-complex ones such as
+  the Italian or Portuguese dictionaries
+- Fixed the build pipeline so that native Node modules are now code-signed and
+  work on macOS and Windows, meaning that `chokidar` should not need to fall
+  back to CPU intensive polling anymore
 
 # 2.3.0
 
--- a/forge.config.js
+++ b/forge.config.js
@@ -133,8 +133,10 @@ module.exports = {
     // notarization.
     osxNotarize: ('APPLE_ID' in process.env && 'APPLE_ID_PASS' in process.env)
       ? {
+          tool: 'notarytool',
           appleId: process.env.APPLE_ID,
-          appleIdPassword: process.env.APPLE_ID_PASS
+          appleIdPassword: process.env.APPLE_ID_PASS,
+          teamId: 'QS52BN8W68'
         }
       : false,
     extraResource: [
--- a/forge.config.js
+++ b/forge.config.js
@@ -104,7 +104,13 @@ module.exports = {
     appBundleId: 'com.zettlr.app',
     // This info.plist file contains file association for the app on macOS.
     extendInfo: './scripts/assets/info.plist',
-    asar: true,
+    asar: {
+      // We must add native node modules to this option. Doing so ensures that
+      // the modules will be code-signed. (They still end up in the final
+      // app.asar file, but they will be code-signed.) Code signing these dylibs
+      // is required on macOS for the Node process to properly load them.
+      unpack: '*.{node,dll}'
+    },
     darwinDarkModeSupport: 'true',
     // Electron-forge automatically adds the file extension based on OS
     icon: './resources/icons/icon',
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "mermaid": "9.0.1",
     "minimatch": "^5.0.0",
     "moment": "^2.29.2",
+    "nodehun": "^3.0.2",
     "nspell": "^2.1.5",
     "rimraf": "^3.0.2",
     "sanitize-filename": "^1.6.3",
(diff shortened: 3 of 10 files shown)


nathanlesage:
Fixed it. What an oddyssee.

richdouglasevans:
Fantastic work Hendrik! 👏 🎉 Thanks also for blogging about this, I learned something 🙇

PR #3968, opened 2022-12-01
update translation

Brli:
## Description

Update translation
## Additional information


nathanlesage:
Hej! Why are you updating the translations here? Because of the failing CI? (I'm just curious, because I didn't expect any translation-related PRs just yet)

Brli:
Because the current translator isn't native, and the tones differs.

nathanlesage:
Ah alright. Would it be okay for you to wait until PR #3970 has been merged? This would make it easier and streamline the efforts of updating the translations. Would that be okay for you? Merging should happen sometime before Dec 31st, if nothing goes wrong.

nathanlesage:
Quick update: The translations are now updated — you can find the new Taiwanese one here: https://github.com/Zettlr/Zettlr/blob/develop/static/lang/zh-TW.po

Please add your translations to that file and update that instead. If you want to use a program to do so rather than a text editor, I can recommend POEdit.

In the meantime, I'll be closing this PR, but please do not interpret this as a dismissal of you work, I am extremely grateful for the work you've done so far! I am looking forward for your next PR and that the Taiwanese translation receives an update!

PR #3970, opened 2022-12-03
Exchange current i18n system with `gettext`

nathanlesage:
## Description

This PR is currently work in progress and contains the required changes for exchanging the current i18n system for a more static one that makes use of the GNU `gettext` system for translating app strings.

## Changes

The following changes have already been made:

* [x] Replaced every invocation of `trans()` with an identifier (e.g., `dialog.preferences.title`) with their corresponding English translations (e.g., "Preferences")
* [x] Added `static/i18n.pot`, the central/main template file that contains every translation from the source code
* [x] Added a script `scripts/extract_trans.sh` that runs the command `xgettext` over the source to extract any invocations of the `trans()` function and merges them with the central template file, `static/i18n.pot`
* [x] Added a script `scripts/split-vue-sfc.js` that is ONLY to be used with the script `scripts/extract_trans.sh` and whose job it is to take every Vue file and extract the content of the script tags. This script stores the script tag contents into `*.ts` files in a temporary directory. That temporary directory gets deleted by the parent script, `extract_trans.sh`. This step is necessary, since `xgettext` is unable to extract the `script` tag contents itself.
* [x] Added the file `static/debug_i18n_strings.json`, which is for **debug purposes** only. It must be removed before actually merging this PR. This file is used to cross-reference the strings that are part of the new `trans()`-invocations with other translations (which are still keyed using the identifier system). This needs to be done after everything else is in place and I perform the final database dump over on Zettlr translate.

These changes still need to be made:

* [x] Migrate the existing `json` translation files to the new `*.po` file format
* [x] Rewrite the language file loader to use the new PO-file format instead of simple JSON
* [x] Adapt the `trans` functions in both the main and renderer processes to utilize the new format
* [x] Add a GitHub Actions workflow that automatically updates the central template every day/week/night with new translatable strings and propagates these to the existing `*.po`-files
* [x] Take a last database dump from Zettlr Translate and overwrite the corresponding `*.po`-files with the most up-to-date content
* [x] Remove the debug i18n file and the replace-trans script
* [x] Remove the old JSON files
* [ ] Merge this PR, then shut down Zettlr Translate

## Additional information

It would be great if you could take a look if I didn't overlook anything in my initial changes.

Tested on: Not yet tested


nathanlesage:
Update on this PR: I've now created a small script that takes the JSON files and creates PO files from them. So now all that is left to do is enable the trans() function to utilize the new PO files instead and clean up the repo.

nathanlesage:
Works like a charm now. I also managed to reduce the complexity of the translation providers. However, it would be great if someone could review this PR (obviously NOT the files where I just changed the `trans()` calls).

nathanlesage:
This PR is now ready to be merged! By merging this PR, this will switch the translation efforts of Zettlr from Zettlr Translate towards the `gettext` system.

PR #3988, opened 2022-12-16
Added test case and improved error message to be more descriptive.

MyGodItsFull0fStars:
## Description

Added code that will append a string to the test output if the word/char count is not equal to the expected values.
This appended string will state what the actual word/char count value is.
Also added a test case with an empty sentence on top of the test cases.

## Changes

Introduced to variables `wordCount` and `charCount` and used those variables with a ternary operator to append a string with the actual value if the value was not equal to the expected value.

Below is the current error message output, that "only" states what value was expected.

current_errormessage

In this PR, the string value now also displays the actual value for further information about what the `countWords(...)` function actually returns.

improved_errormessage

Also an edge test case was added, that test if the `countWords(...)` function correctly returns the expected values if the provided sentence is empty.

empty_sentence

## Additional information

Tested on: macOS Monterey 12.6.1

### DIFF
--- a/test/count-words.spec.ts
+++ b/test/count-words.spec.ts
@@ -18,6 +18,11 @@ import { strictEqual } from 'assert'
 
 const countWordsTesters = [
   {
+    input: '',
+    expectedWords: 0,
+    expectedChars: 0
+  },
+  {
     input: 'Lorem\n\n# Ipsum',
     expectedWords: 2,
     expectedChars: 12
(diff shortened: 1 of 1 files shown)


MyGodItsFull0fStars:
I did execute `yarn lint` before creating this PR. Also I configured my VSCode workspace to be aligned with this project.

nathanlesage:
Definitely more readable than the standard ouput of mocha, but I'm thinking that it may be more readable to extract that ternary out of the `it` call, and instead generating a default message, only appending it with the error part if the wordcount actually differs. Also, constants should use `const` :)

What are your thoughts?

Other than that, it looks good!

MyGodItsFull0fStars:
Thank you for the review and explanations. As I am still a noob in TypeScript, it is highly appreciated :)
I will change the code based on your suggestions.

MyGodItsFull0fStars:
I added two more variables `wordCountMessage`, `charCountMessage` that hold the strings for the mocha `it` functions. Two if-conditions will check if the expected and actual values for the `countWords(...)` function do not match, and in that case concatenate a string to the `wordCountMessage` or `charCountMessage` variables that states what the actual value is.

nathanlesage:
Alright, thanks! The final question I'm curious about is: Do we really need to output that during the test cases? Mocha has quite the extensive error output in case something goes wrong, so I'm not fully convinced we need these shenanigans ...?

MyGodItsFull0fStars:
Haha that's fair! Yeah, I did overlook the output at the end that states what the expected and actual values are (even with nice colouring). So I guess in the end it was more of a shenanigan and warm-up for myself.

Sorry for taking up your time and thanks again for your code review :)

nathanlesage:
Alright -- but the edge case with zero words/characters is definitely a valuable thing to have! Should we then revert the test itself back to its original form and retain the new test case that you've added so that we can merge this PR? :)

MyGodItsFull0fStars:
Okay, I will revert the test file back to the original form and will keep the edge case in this PR :)

nathanlesage:
Wonderful, thank you! Will merge now! :)

MyGodItsFull0fStars:
Thank you! :)

PR #4000, opened 2022-12-23
Add hash function tests

MyGodItsFull0fStars:
## Description

Added some basic tests for the `hash` function inside the `source/common/util` path.

## Changes

Simply added a test file that checks the hash function with some expected output. Nothing should be breaking because of it.

## Additional information

As far as I saw, the hash function is solely used inside of `fsal-cache`, and it surely will do it's expected job.
First I wanted to comment on it being a rather weak hash function, that is not even able to produce a value that always has the same length, but after seeing that it is used for the function `_determineShard(...)`, it surely will do its intended purpose very well.

Additionally, I added myself as the author of this file in the file header, I hope this is fine, if not I will remove it.

Macbook Pro M1 with MacOS Monterey 12.6.2


nathanlesage:
Exactly, the hash function was never intended to be strong, and for ensuring we only have 100 cache files at most it should do its job. If you have a good alternative for mapping an arbitrary amount of files onto a fixed set (such as 100), please feel free to suggest an improvement.

The PR itself looks perfectly fine, so merging!

MyGodItsFull0fStars:
Okay I will check if I can find a good alternative, but first I might look at how well distributed arbitrary strings are with this hash function. If they are well distributed and hence, hash collisions are omitted as good as possible, using another hash function would only make sense if it would affect the performance of the caching.

Thanks for merging the PR.

PR #4067, opened 2023-01-18
feat: Add LanguageTool integration

nathanlesage:
## Description

This PR adds a LanguageTool integration to Zettlr. It offers a similar API to the browser plugin, i.e.:

* Choose between the official LanguageTool.org provider and a custom server
* Select between the default and picky modes
* Provide username & API key to enable premium limits

> **NOTE: This PR is currently under embargo and WILL NOT be merged.** I am awaiting a reply from the LanguageTool team whether it is okay to provide their API by default, since we can expect several thousand people to switch this integration on at roughly the same time, and I don't want to brigade their servers by accident. If their reply is negative, I will have to tweak the settings of this PR before merging.

## Changes

* Add a new linter for LanguageTool
* Add a new command that performs the actual API request on behalf of the renderer
* Add corresponding settings in the config & the preferences dialog

Tested on: macOS 13.1 both with the official server and a custom installation. Not yet tested with a local server, but that should work as well.


nathanlesage:
The official response from the team at LanguageTool.org just arrived; we can go ahead. After the checks pass, I'm going to merge the feature.

PR #4128, opened 2023-02-04
Do not crash on empty window_state.yml

### LINKED ISSUE #4099: Crashes at start
### Description

Until this morning, Zettlr worked fine. When I start it now, I get the following error and Zettlr quits (without any GUI except the error has appeared).

```
TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at M.boot (/opt/Zettlr/resources/app.asar/.webpack/main/index.js:2:1741819)
    at async t.default._informativeBoot (/opt/Zettlr/resources/app.asar/.webpack/main/index.js:2:1501067)
    at async t.default.boot (/opt/Zettlr/resources/app.asar/.webpack/main/index.js:2:1498227)
    at async t.bootApplication (/opt/Zettlr/resources/app.asar/.webpack/main/index.js:2:1502716)
```

When I choose another `data-dir`, Zettlr starts without an error even if I re-add every directory which was open before, so it looks like an error in the configuration.

### Reproducing

Start Zettlr

### Zettlr Version

Nightly

### Specify version

3.0.0-beta.1 (Mon Jan 23rd, 2023 13:37:12)

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Ubuntu 22.04

### Additional Information

I guess this is not worth an in-depth analysis. I'll re-apply my settings and everything is fine. But I reported it because it might be helpful if it appears more often.

UFOMelkor:
## Description
Fixes #4099 

## Changes
Zettlr Crashes at start if `window_state.yml` exists but is empty. This happens because the stateContainer is initialized (because the file can be read) but when interpreting the file (`stateContainer.get()`), no content is available, so it returns null which cannot be handled by `Object.entries`.

## Additional information
The real error behind this is: Why is an empty `window_state.yml` written generally. But as the error seems to occur just at my PC, this might be too much to investigate. So, I guess this might be an acceptable workaround.

If you don't think so, or you have a better idea to fix this, feel free to close this PR.

Tested on: Ubuntu 22.04

### DIFF
--- a/source/common/modules/persistent-data-container/index.ts
+++ b/source/common/modules/persistent-data-container/index.ts
@@ -94,7 +94,7 @@ export default class PersistentDataContainer {
   public async isInitialized (): Promise<boolean> {
     try {
       await fs.access(this._filePath, FSConstants.R_OK | FSConstants.W_OK)
-      return (await this.get()) !== null
+      return (await fs.readFile(this._filePath, { encoding: 'utf-8' })).trim() !== ''
     } catch (err: any) {
       return false
     }
--- a/source/common/modules/persistent-data-container/index.ts
+++ b/source/common/modules/persistent-data-container/index.ts
@@ -94,7 +94,8 @@ export default class PersistentDataContainer {
   public async isInitialized (): Promise<boolean> {
     try {
       await fs.access(this._filePath, FSConstants.R_OK | FSConstants.W_OK)
-      return (await fs.readFile(this._filePath, { encoding: 'utf-8' })).trim() !== ''
+      const contents = await fs.readFile(this._filePath, 'utf-8')
+      return contents.trim() !== ''
     } catch (err: any) {
       return false
     }
--- a/source/common/modules/persistent-data-container/index.ts
+++ b/source/common/modules/persistent-data-container/index.ts
@@ -94,7 +94,7 @@ export default class PersistentDataContainer {
   public async isInitialized (): Promise<boolean> {
     try {
       await fs.access(this._filePath, FSConstants.R_OK | FSConstants.W_OK)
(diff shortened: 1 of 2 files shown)


nathanlesage:
I think your PR is spot on: The app should not crash on an empty settings file.

However, we can do even better, if you like to, and put this check into the `PersistentDataContainer`. You are right that an empty file should not be considered “initialized”, and hence, if the file is empty, the `isInitialized` function should return false.

Would you mind quickly moving this check? That would be great! Then we can merge this PR.

UFOMelkor:
:+1:

source/common/modules/persistent-data-container/index.ts
```diff
@@ -94,7 +94,7 @@ export default class PersistentDataContainer {
   public async isInitialized (): Promise<boolean> {
     try {
       await fs.access(this._filePath, FSConstants.R_OK | FSConstants.W_OK)
-      return true
+      return (await this.get()) !== null
```
nathanlesage, line 97: Sorry for my late response! I would advise against calling `get` here in this function to keep it clean. Rather, I would look at the actual file contents and see if it's an empty file (using `fs.readfile`). Anything else (invalid syntax) will throw an error anyways.

nathanlesage (commented):
Looks much better now!

nathanlesage:
Perfect, thank you! I changed your PR a tiny bit to make the empty file check a bit more readable and align the code style more with the rest of the code base.

Thank you again for your efforts!

Issue #4049, opened 2023-01-12
Beta 3.0.0 Fails to Open (Debian)


cyanitol:
### Description

When running application it fails to open and produces the following error:
`A JavaScript error occurred in the main process
Uncaught Exception:
Error: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /tmp/.mount_ZettlrNiLWyo/resources/app.asar.unpacked/.webpack/main/native_modules/build/Release/Nodehun.node)
    at process.func [as dlopen] (node:electron/js2c/asar_bundle:2:1822)
    at Module._extensions..node (node:internal/modules/cjs/loader:1226:18)
    at Object.func [as .node] (node:electron/js2c/asar_bundle:2:2049)
    at Module.load (node:internal/modules/cjs/loader:1011:32)
    at Module._load (node:internal/modules/cjs/loader:846:12)
    at f._load (node:electron/js2c/asar_bundle:2:13328)
    at Module.require (node:internal/modules/cjs/loader:1035:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at 19572 (/tmp/.mount_ZettlrNiLWyo/resources/app.asar/.webpack/main/index.js:2:1151389)
    at r (/tmp/.mount_ZettlrNiLWyo/resources/app.asar/.webpack/main/index.js:2:2148597)
`

### Reproducing

Attempt to run the program. AppImage and .deb both have the same issue.

### Zettlr Version

Stable (most recent version)

### Installation Method

Other (please specify below)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Debian 11.6

### Additional Information

Linux dddd 5.10.0-20-amd64 #1 SMP Debian 5.10.158-2 (2022-12-13) x86_64 GNU/LinuxArchitecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   39 bits physical, 48 bits virtual
CPU(s):                          4
On-line CPU(s) list:             0-3
Thread(s) per core:              2
Core(s) per socket:              2
Socket(s):                       1
NUMA node(s):                    1
Vendor ID:                       GenuineIntel
CPU family:                      6
Model:                           142
Model name:                      Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
[...]

PR #4155, opened 2023-02-17
run Linux builds on Ubuntu 20.04 machines

marado:
Builds made against Ubuntu 22.04's glibc won't work on environments with an older glibc, like Debian or Ubuntu 20.04's installations.

This patch changes the builder workflows to make GNU/Linux builds be created in an Ubuntu 20.04 environment, producing builds that work on both older and more recent environments.

Closes: #4049

I have  *not* added an entry to the CHANGELOG.md, as this introduces no change from the previous Zettlr version.
Tested on: Ubuntu 20.04

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
   # havoc because we don't push unstable code to master, BUT it looks bad.
   verify_version:
     name: 'Preflight: Verify version tag'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Clone repository (master branch)
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


nathanlesage:
This sounds like a reasonable course of action. Do you have any idea how long this fixing of the Ubuntu version might be necessary? I'm just thinking that once 20.04 isn't supported anymore we have to remember this so that we can update the runner OS again. What's your opinion on this?

marado:
Ubuntu 20.04's end of life is in April 2030. Unless something comes up, I see no reason to not try to support it at least that long...

nathanlesage:
Oh wow, I didn't know the LTS releases had *that* long of a support window. Alright, case settled.

Thank you again for researching the cause of this weird issue, and proposing a fix! Well appreciated.

PR #4156, opened 2023-02-17
Have more verbose (and useful) error messages

marado:
If a data container file is empty, we'll now have a more useful error message showing to the user.

### DIFF
--- a/source/app/service-providers/windows/index.ts
+++ b/source/app/service-providers/windows/index.ts
@@ -256,6 +256,7 @@ export default class WindowProvider extends ProviderContract {
       await this._stateContainer.init(Object.fromEntries(this._windowState))
     }
     const tmpObject = await this._stateContainer.get()
+    if (tmpObject === null) this._logger.error('[Window Manager] tmpObject is null');
     this._windowState = new Map(Object.entries(tmpObject))
     this._logger.info('[Window Manager] Window Manager started.')
   }
--- a/source/common/modules/persistent-data-container/index.ts
+++ b/source/common/modules/persistent-data-container/index.ts
@@ -139,6 +139,7 @@ export default class PersistentDataContainer {
     try {
       const content = await fs.readFile(this._filePath, { encoding: 'utf-8' })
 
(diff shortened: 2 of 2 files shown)


source/app/service-providers/windows/index.ts
```diff
@@ -256,6 +256,7 @@ export default class WindowProvider extends ProviderContract {
       await this._stateContainer.init(Object.fromEntries(this._windowState))
     }
     const tmpObject = await this._stateContainer.get()
+    if (tmpObject === null) this._logger.error('[Window Manager] tmpObject is null');
```
nathanlesage, line 259: This check should be unnecessary after I've merged the PR by @UFOMelkor, since there is no way right now that `get()` will return null.

source/common/modules/persistent-data-container/index.ts
```diff
@@ -139,6 +139,7 @@ export default class PersistentDataContainer {
     try {
       const content = await fs.readFile(this._filePath, { encoding: 'utf-8' })
 
+      if (content.trim().length === 0) throw new Error("Container at " + this._filePath + " was empty (content is null).");
```
nathanlesage, line 142: This could be a good addition to catch instances where the initializer found a perfectly normal file, and something caused the file to be emptied out in the meantime. Just two suggestions:

1. I tried to stop using shorthand `if`-statements across the entire codebase, and surround everything with `{` and `}` now
2. Instead of using string concatenation, please use template strings (with backticks and `${variable}`)

nathanlesage (changes requested):
This sounds like a good addition to ensure we actually know where an error originates. However, a few minutes ago I merged the PR by @UFOMelkor that moves the empty string check into the initializer. It would be great if you could check how the changes you propose here interact with that, just to be sure!

Other than that, it seems reasonable — see my other comments!

source/app/service-providers/windows/index.ts
nathanlesage: Also, I just saw that this change is also being obliterated by your other change to the `get()` accessor below.

marado:
Well, I should have checked other PRs before creating this one... I've just tested, and with #4128 my PR is no longer needed.

PR #4189, opened 2023-03-05
Use existing lastLeafId computed property in App.vue

kyaso:
Just noticed this while going through the code and thought of quickly fixing it :)
There is already a computed property "lastLeafId" implemented, but it was not used and accesses to the store were still direct.


nathanlesage:
Ah, good catch! Thank you! :)

PR #4197, opened 2023-03-11
refactor: Simplify link detection for file preview

kyaso:
## Description
This PR simplifies the detection of Zkn links for the file preview tooltip.

## Changes

Instead of manually searching for the link starts and ends in the line, we could make use of the **AST** to get the syntax node at the current position. If that node is of type `ZknLinkContent`, then we know that the cursor is hovering over a Zkn link.

## Additional information

Tested on: Linux Manjaro

### DIFF
--- a/source/common/modules/markdown-editor/tooltips/file-preview.ts
+++ b/source/common/modules/markdown-editor/tooltips/file-preview.ts
@@ -13,6 +13,7 @@
  * END HEADER
  */
 
+import { syntaxTree } from '@codemirror/language'
 import { EditorView, hoverTooltip, Tooltip } from '@codemirror/view'
 import { trans } from '@common/i18n-renderer'
 import formatDate from '@common/util/format-date'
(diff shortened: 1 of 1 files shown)


nathanlesage (approved):
Magnificient. This is a remnant from the time when the Zettelkasten links were not yet part of the syntax tree. I am quite happy with the much better solution you came up with.

Just one thing: Did you test this especially in nested settings? I have had a few weird instances where the resolution did only return some *parent* node. But this could've also been my inability to properly use the resolver back then 😂

kyaso:
To be fair, I got "inspired" by these lines in the editor module 😅: https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/index.ts#L320-L330

Could you maybe provide an example of such a nested setting (not sure if I understood that)?

nathanlesage:
Oh … okay, apparently then it was just a bug :D With nesting I basically just meant that sometimes, if a node occurs deep in the tree, I have had some issues resolving the correct node, but this looks fine. I just remembered that there was a bug fix a few weeks ago that fixed that in CodeMirror.

Let's just merge and see what happens :D

PR #4215, opened 2023-03-25
refactor: Make MainEditor single-document only

nathanlesage:
## Description

This PR refactors the main editor in such a way that, instead of one `MarkdownEditor` instance maintaining the entire barrage of documents in the `openFiles` array, the class now only handles a single file.

This has the following benefits:

1. The state management of the `MarkdownEditor` class is simpler: It does not have to cache the various documents, nor does it have to cache the scroll positions of the documents
2. The `MainEditor` Vue component is simpler, since it also doesn't have to implement too much of a switcheroo-logic to ensure the editor contains the correct document. This should make the type of bugs which @efficacy has described, impossible.
3. The entire state logic has now moved to `EditorPane`, i.e. it has been outsourced to the Vue library. This has two additional benefits:
    1. The documents will retain their scroll position, as they are merely hidden instead of completely dropped out of existence – all without any weird caching magic
    2. We now have the ability to add additional editors for various other purposes. Two specific things I would definitely like to add would be a PDF reader and an Image viewer. Further editors/viewers would be possible, but we should remember that Zettlr is not VSCode ;)

## Changes

* The document provider now emits the previously unused `FILE_SAVED` event, which the `MainEditor` now listens to
* Added documentation for previously undocumented variables of the `MarkdownEditor` class
* Removed the state and view cache from the `MarkdownEditor` class
* Remove any code required to swap documents
* Removed the weird extra-code for the `darkMode` setting which was a remnant from when I initially implemented the editor, since that is handled with the configuration itself by now
* The `MarkdownEditor` instances now expose a property `documentPath` that can be used to query what document they represent
* Exchange the single `MainEditor` component in the `EditorPane` component with a `template v-for`, utilizing `v-show` to show/hide the various editors programmatically, based on which one contains the `activeFile`
* Removed any code from the `MainEditor` component used to differentiate whether it should react to given commands (text insertion, moving around of sections, etc.), and instead this is done in the `EditorPane` component, which will simply pass an unreactive dummy to those instances that should not handle anything
* Refactored some code here and there to make it more readable

## Additional information

This should not include any behavioral changes, except that the state of the various documents is better retained even when they are not active and in the background. The biggest change users should spot are (hopefully) that the scroll positions are now precise.

Tested on: macOS Ventura (13.2.1); MBP 14in M2 Pro


nathanlesage:
Also see issue #4227 which affects this PR as well.

kyaso:
Hey! Sorry for the late response, I'm currently out of town, will have a look this weekend when I'm back!

nathanlesage:
All good, no pressure! Looking forward to your review! :)

nathanlesage:
Quick update: The file renaming bug is not present in this PR, I fixed a different bug just now.

nathanlesage:
@kyaso 👀

kyaso:
Sorry, was a bit busy the last few days, I'll finally have a look tomorrow!

nathanlesage:
Thanks for the notice, all good -- looking forward to what you think!

source/win-main/EditorPane.vue
```diff
@@ -11,18 +11,20 @@
...
+        v-bind:file="file"
+        v-bind:distraction-free="distractionFree"
+        v-bind:leaf-id="leafId"
+        v-bind:window-id="windowId"
+        v-bind:editor-commands="commandsForFile(file)"
+        v-on:global-search="$emit('globalSearch', $event)"
+      ></MainEditor>
+    </template>
+    <!-- Show a placeholder if there are no open files -->
+    <div v-if="openFiles.length === 0" class="empty-pane"></div>
```
kyaso, line 26: We could move the length check to a _computed_ prop with a descriptive name, which would then make the comment above obsolete.

source/win-main/MainEditor.vue
```diff
@@ -123,25 +123,29 @@ const props = defineProps({
...
     type: Boolean,
     required: true
+  },
+  file: {
+    type: Object as () => OpenDocument,
+    required: true
   }
 })
 
+// TODO: Update the activeFileDescriptor as soon as the appropriate event is
```
kyaso, line 133: Is this TODO something for the future or for this PR? If it is meant for this PR, we could add a todo item in the PR description to not forget this.

source/common/modules/markdown-editor/index.ts
```diff
@@ -190,65 +200,42 @@ export default class MarkdownEditor extends EventEmitter {
   constructor (
     anchorElement: Element|DocumentFragment|undefined,
     editorId: string,
```
kyaso, line 202: Should we rename this to `leafId` to not confuse it with `this.editorId`?

kyaso (commented):
While I had some initial troubles wrapping my head around the details of this PR, after a first review round it looks very good. Great job also on the details in the PR description, it definitely helped in understanding what has been changed!

I left some minor comments :)

kyaso:
Ah, one more thing: As the description says that this PR doesn't affect external behavior, we may consider changing the type from "feat" to "refactor".

What do you think?

source/win-main/EditorPane.vue
nathanlesage: Done.

source/win-main/MainEditor.vue
nathanlesage: That was mainly a TODO for myself, and I just checked – I finished that one already :D

source/common/modules/markdown-editor/index.ts
nathanlesage: Good catch. I also took the liberty to update the constructor docstring, since that has changed quite a bit :D

nathanlesage:
> Ah, one more thing: As the description says that this PR doesn't affect external behavior, we may consider changing the type from "feat" to "refactor".
> 
> What do you think?

More precisely, as I mention myself in the very first sentence of the description: "This PR refactors the main editor …" — so yes 🫠 You're absolutely right

kyaso:
> More precisely, as I mention myself in the very first sentence of the description: "This PR refactors the main editor …" — so yes

Haha, even I didn't notice that 😂

> While this does change the UX a little bit, I'm not too afraid of anything breaking, and honestly I think it's even better as it is now. But please, share any thoughts you may have.

I'm also ok with this.

> anything else, or can we merge?

LGTM, we can merge! 🚀

nathanlesage:
Wonderful, then let's go! Then I can finally publish the next beta :)

PR #4217, opened 2023-03-26
arch linux: upgraded from aur to a regular package

UweKrause:
For Arch Linux there are different package sources available, the "regular" and the "arch user repository" (AUR). If you must, you could compare it with Ubuntus "Personal Package Archives" (PPA). The Zettlr package was promoted from an AUR package to a "regular" one, therefore the reference to the "AUR" is no longer needed.

For any Arch user it is way more convenient to use regular packages over aur packages.

## Additional information


nathanlesage:
Thank you for fixing this! I've also adapted this on the main website.

Issue #4084, opened 2023-01-23
Proxy for LanguageTool


UFOMelkor:
### Description

I tried the latest beta to enjoy the LanguageTool support. At home, it works really nice (once I've configured it correctly), but at work the only response that is shown in the toolbar is *LanguageTool: [warning sign] (offline)*. 

I guess the reason – as the used credentials are the same – is the proxy, which is required at work. The system environment variables (http_proxy, https_proxy, no_proxy) are all set.

Another obstacle I stumbled upon is the configuration. I have an account for LanguageTool and I entered the credentials without setting the server. This results in a *404* error. It took a moment to understand that I needed to enter `https://api.languagetoolplus.com` as server. Perhaps it might be better to automatically use this as server if no server is given and credentials are used because languagetool.org seems to error otherwise.

### Proposed Changes

Add a configuration for a proxy or (better) read the system proxy and apply him.

### Caveats

I don't know any

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Beta (if applicable)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Ubuntu 22.04

nathanlesage:
> I guess the reason – as the used credentials are the same – is the proxy, which is required at work. The system environment variables (http_proxy, https_proxy, no_proxy) are all set.

I did not meddle with how Zettlr utilizes proxies; however, it may be that Electron, by default, does not utilize the system proxy. I just noticed that I accidentally forgot to log every error, which I fixed in the attached commit.

So in order to find the real reason for the “offline” notification, please wait for the next nightly and/or beta and then, once it fails, have a look into the logs. There should be an error logged that exactly describes what went wrong.

***

Now to your second point. I did not realize that LanguageTool is offering TWO servers, one at https://api.languagetool.org and one at https://api.languagetoolplus.org. The “official” server Zettlr uses is the first one, but if I understand you correctly, credentials only work with the second one. Could you please confirm this?

If so, then maybe the first server is just a legacy one, and the correct one we should be using is the second one. It would be a simple change, once we can confirm that we can indeed redirect everything to the plus-API.

Do you know anything about this? Otherwise, I'd double-check with them first. I really don't want to offer two official servers if possible, since this may confuse people.

UFOMelkor:
> please wait for the next nightly and/or beta and then, once it fails, have a look into the logs

:+1: 

---

[...]

PR #4233, opened 2023-04-01
Allow to configure a proxy using environment vars

UFOMelkor:
See the latest comment in [#4084](https://github.com/Zettlr/Zettlr/issues/4084#issuecomment-1492961421).

I do need to test it again at work, but if you have some spare time, you might provide some feedback on whether this implementation would be okay or you would like follow another approach.


source/global-agend.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'global-agent' {
```
nathanlesage, line 1: There's a typo in the file name

source/main.ts
```diff
@@ -47,6 +48,9 @@ if (!app.requestSingleInstanceLock()) {
 
 // If we reach this point, we are now booting the first instance of Zettlr.
 
+// Use a proxy if one has been configured
+bootstrap()
```
nathanlesage, line 52: If not necessary to do so before the `appReady` event, please move this call out of the main file and into the boot cycle. If this needs to be called first thing even before the boot cycle, please add a comment why.

nathanlesage (changes requested):
Thank you for this initiative! There are a few questions, however:

* Which types of requests does this affect? Only the ones made by got, anything that uses the net module, or some other subset?
+ Does this affect requests made by Electron (none node code) as well?
+ We should probably allow users to configure Proxies in the settings, similar to in other apps ("Use system settings", "use no proxy", "specify proxy")
+ Can we disable Proxies altogether if not needed? I.e., call bootstrap() programmatically, maybe even after the config provider has booted up? If this only needs to happen before the first request, we can easily query the config provider which doesn't make a request itself

UFOMelkor:
> * Which types of requests does this affect? Only the ones made by got, anything that uses the net module, or some other subset?
> * Does this affect requests made by Electron (none node code) as well?

According to the docs, it affects everything that uses node's `http.request`. But I'm not as deep enough into electron to know, but I would guess that this won't effect any requests made by electron. Are there specific requests I might test?

> * We should probably allow users to configure Proxies in the settings, similar to in other apps ("Use system settings", "use no proxy", "specify proxy")
> * Can we disable Proxies altogether if not needed? I.e., call bootstrap() programmatically, maybe even after the config provider has booted up? If this only needs to happen before the first request, we can easily query the config provider which doesn't make a request itself

I'll test a later calling point when next time at work. If it is possible to call it after the config provider has booted up, I would guess the way to go would be adding a service provider for applying proxy configuration after the config provider but before the update provider?

nathanlesage:
Thanks for the update! Regarding Electron requests: I was specifically thinking about images. These are automatically loaded once an image is added to the document. So one thing you could test is see if image loading in Markdown files after proxying are also using the proxy or if they bypass it.

Then, I think it would suffice to add a comment explaining what is affected just so that we know (especially since this is a side effect that we may wonder about in the future).

Regarding where and how: that's a good question. On the one hand it's environment setup so the most correct place would be the end checker, but then when it's user configurable it can only run after the config provider has booted up. I'm a little bit hesitant regarding another service provider just for one setting. The question is: could there be potentially more settings except the proxy configuration that may fall into this category of needing to run early but still only after the config? One thing I could think about is the language setting, which needs to run super early but is currently instantiated only after the config has loaded.

Maybe a dedicated function within the config provider that sets up envs from the config ...? Hmmm...

PR #4242, opened 2023-04-08
refactor: Move FSAL parser to utilize the Markdown AST Parser

nathanlesage:
## Description

This PR replaces the FSAL Markdown file parsing logic with our new Markdown AST. This reduces complexity and makes sure that what the file descriptors contain matches what the main editor sees.

## Changes

* Removed the manual parsing logic from the FSAL file parser
* Utilize the Markdown AST parser to extract the types of nodes that are important for collecting metadata on the files

## Additional information

**Breaking Changes**

- This PR now contains a breaking change: The character count without spaces is now gone. After refactoring, I took the liberty to remove that. The reasons are as follows:
    - I originally implemented that counter because – from a programmer's perspective – a plain character count is meaningful (which is why I included it), but when writing text, what should count is only the amount of non-whitespace characters (which is also the writing target for many teachers and journals)
    - However, since Markdown contains a large amount of non-significant whitespace, or, whitespace that has no bearing on the factual text content of a file (e.g. ending a line with two spaces to signify a single line break), a plain character counter has no real use while writing (academic) texts
    - Thus, **the character counter without spaces is now gone, and the regular character counter now excludes whitespace**
    - This behavior is now unified across the app, so there will only be a single character count, no matter where

There are a few TODOs left:

- [x] I realized that now with the new AST we are able to greatly reduce the complexity of the word counter just as well. So I'll have to go back and do that as well, since it's a no-brainer right now
- [x] Go over the entire code base again and eliminate any remnant of manual parsing of Markdown code
- [x] Remove the old legacy `md2ast` function; also check if there are any other MD parsers left in the code base

Tested on: macOS Ventura 13.3.0


nathanlesage:
@tobiasdiez I managed to finally do the last part of the migration to the new AST – would you mind reviewing this? That would be great!

@kyaso If you're bored after the Easter lunch tomorrow, you're invited too 😂

source/common/modules/markdown-editor/index.ts
```diff
@@ -710,11 +704,12 @@ export default class MarkdownEditor extends EventEmitter {
           const anchorLine = this._instance.state.doc.lineAt(sel.anchor)
           const headLine = this._instance.state.doc.lineAt(sel.head)
           const selContent = this._instance.state.sliceDoc(sel.from, sel.to)
+          const ast = markdownToAST(selContent)
```
tobiasdiez, line 707: It is relatively costly to parse the markdown content, so as an optimization you may want to move the words/char count calculation to a queue with a very short delay (if I start selecting everything in a document by using Ctrl + Line down, then I can easily get > 25% CPU on a relatively beefy machine).

tobiasdiez (approved):
That looks like a nice improvement! Good job.

source/common/modules/markdown-editor/index.ts
nathanlesage: Very good catch! However, since asynchronous callbacks in this case are always tricky to do and would increase the complexity again, I think I have had an idea:

In a previous commit, I have added the ability of `markdownToAST` to use a pre-parsed SyntaxTree, if you pass one as the second parameter. This completely obliterates the need for the AST parser to create a parse tree by itself. Since CodeMirror always keeps at least a partial parse tree (which will include at least the selections), what I have done now is to pass said partial tree to the function, which cut down the parse time by a factor of 5.

A second optimization is that I have now exchanged the "countChars" and "countWords" function with a novel function, "countAll", which has the char and word counter share the computationally heavy preprocessing (and, prospectively, will replace the individual functions), thus further cutting down the remaining time in half.

I have run a few performance test: The unoptimized code took about ~60ms per each keydown event (i.e. each move to the next line), the optimized code now cuts this down to ~12ms, of which the actual counting of the characters and words makes up approximately half to 1ms.

Do you think this suffices, or do you still think it may negatively impact users' computers?

Please feel free to have a look at the newly optimized code.

If you think that this still puts too much load onto computers (thinking especially of potato PCs, on which Zettlr should run smooth as well), I can reduce the update interval with a debouncer.

tobiasdiez: I think this should suffice. It's a non-blocking operation anyway, so it should be fine.

nathanlesage: Alright, thank you very much for your thoughts on this! In this case, I'll merge the PR now!

PR #4243, opened 2023-04-09
Update pt-BR.po

TheMemoir:
I reviewed and updated some text excerpts for Brazilian Portuguese.

## Description

I reviewed and updated some text excerpts for Brazilian Portuguese after I have read the full PO file for pt_br.

## Changes

Corrected and update terms which I observed were mistranslated.

## Additional information

### DIFF
--- a/static/lang/pt-BR.po
+++ b/static/lang/pt-BR.po
@@ -18,11 +18,11 @@ msgstr ""
 
 #: source/app/service-providers/commands/language-tool.ts:180
 msgid "Document too long"
-msgstr ""
+msgstr "Documento muito longo"
 
 #: source/app/service-providers/commands/language-tool.ts:185
 msgid "offline"
-msgstr ""
+msgstr "offline"
 
 #: source/app/service-providers/commands/dir-new.ts:36
 #: source/app/service-providers/commands/dir-new.ts:37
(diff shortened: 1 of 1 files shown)


static/lang/pt-BR.po
```diff
@@ -1383,7 +1383,7 @@ msgstr "Imagem não encontrada: %s"
 #: source/common/modules/markdown-editor/renderers/readability.ts:39
 #, javascript-format
 msgid "Readability mode (%s)"
-msgstr ""
+msgstr "Modo de legibilidade"
```
nathanlesage, line 1386: You are missing the `%s` placeholder here (for the actual algorithm used) — was this by accident or on purpose?

nathanlesage (commented):
Hey, thank you very much for your help in translating! This looks already pretty good, but it seems as if you've accidentally forgotten a template string (see my comment below)? If you could have another look at that, this would be great! Then we can merge!

Awesome work :)

static/lang/pt-BR.po
TheMemoir: Sorry. I forgot that. I'll correct it right away.

nathanlesage: The `%s` is very simple: If the translatable string contains these two characters – here, it's `Readability mode (%s)`, this means that the `%s` will be replaced with some other string while the app runs.

In this case, `%s` will be replaced with the readability algorithm that is being used. I.e., in the app, `Readability mode (%s)` will become `Readability mode (dale-chall)`.

You can put this placeholder wherever it makes sense. You could, in theory, put it in front of your translation if that makes grammatical sense. I think here you can just modify the translation as such: `Modo de legibilidade (%s)` (but I don't speak Portuguese, so think whether `Modo de legibilidade (dale-chall)` makes grammatical sense first!)

TheMemoir: I guess I got it right, now. Would please let me know? Tks.

nathanlesage:
This looks great! Thank you very much! :)

Merging …!

Issue #4322, opened 2023-05-12
Case-sensitive renaming of files in File Manager does not work


sensologica:
For example, if I have a file named `LOWERCASE.md` and I try to rename it to `lowercase.md` in the File Manager, the change is not saved. It appears that if the string of characters is the same (excluding capitalization), the file name change does not "stick".

nathanlesage:
Hah, finally a bug that has been fixed for good! :)

PR #4338, opened 2023-05-15
Allow case sensitive renaming

justjack2347:
## Description

This PR allows for the renaming of files with different capitalisation.

## Changes

Removed two instances of making the filenames lowercase when comparing (in file-rename.ts and fsal-directory.ts).

## Additional information

The check is still done by lower casing the file names before file creation since I believe that removing this could cause problems on case-insensitive file systems.

This pull request addresses issue #4322 

Tested on: Windows 11
University of Adelaide - Software Process Improvement
Additional Contributors:

- @josh-133
- @Mr-Verb

### DIFF
--- a/source/app/service-providers/fsal/fsal-directory.ts
+++ b/source/app/service-providers/fsal/fsal-directory.ts
@@ -411,7 +411,7 @@ export async function renameChild (
     throw new Error(`Cannot rename ${oldName}: Not found in ${dirObject.path}.`)
   }
 
-  const foundName = dirObject.children.find(child => child.name.toLowerCase() === newName.toLowerCase())
+  const foundName = dirObject.children.find(child => child.name === newName)
   if (foundName !== undefined) {
     throw new Error(`Directory ${newName} already exists!`)
   }
--- a/source/app/service-providers/fsal/fsal-directory.ts
+++ b/source/app/service-providers/fsal/fsal-directory.ts
@@ -391,7 +391,7 @@ export async function createFile (
  * @param   {string}         oldName        The directory's old name
  * @param   {string}         newName        The directory's new name
  * @param   {FSALCache}      cache          The FSAL cache object
- * @param   {boolean}        allowOverwrite Whether to allow overwriting existing files
+ * @param   {boolean}        forceOverwrite Whether to force overwriting of existing files when renaming to an existant filename
  *
  * @return  {Promise<DirDescriptor>}    Resolves with the new directory descriptor.
  */
@@ -402,7 +402,7 @@ export async function renameChild (
   parser: (file: MDFileDescriptor, content: string) => void,
   sorter: (arr: AnyDescriptor[], sortingType?: string) => AnyDescriptor[],
   cache: FSALCache,
-  allowOverwrite: boolean = false
+  forceOverwrite: boolean = false
 ): Promise<void> {
   // If old and new name are the same, no need to rename
   if (newName === oldName) {
@@ -420,7 +420,7 @@ export async function renameChild (
   }
 
   // Stops renaming if the new file will overwrite an old file and we don't want it to
-  if (newName.toLowerCase() !== oldName.toLowerCase() || !allowOverwrite) {
+  if (newName.toLowerCase() !== oldName.toLowerCase() || !forceOverwrite) {
     const foundName = dirObject.children.find(child => child.name.toLowerCase() === newName.toLowerCase())
     if (foundName !== undefined) {
       throw new Error(`Cannot rename ${oldName} to ${newName}: A file with the same name already exists!`)
--- a/source/app/service-providers/fsal/fsal-directory.ts
+++ b/source/app/service-providers/fsal/fsal-directory.ts
@@ -420,7 +420,7 @@ export async function renameChild (
   }
 
   // Stops renaming if the new file will overwrite an old file and we don't want it to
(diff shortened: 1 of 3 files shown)


nathanlesage:
Thanks for this PR, I appreciate your support.

The first change in the command it looks proper – case-sensitive renaming is possible, but when the app checks for duplicates, it does so case-insensitive to ensure cross-platform compatibility.

However, this fine-grained check is missing from the second file, the `renameChild` utility function. There, it fails to check if there is already a file case-insensitively, thus allowing case-sensitive renaming _but also overwriting of existing files_.

Would you mind adding the necessary checks for that as well? That would be awesome.

justjack2347:
Hi @nathanlesage, thank you for your comments. I have had a look and play around and completely understand your concern but cannot find an example of where we can overwrite existing files and so if you had an instance in mind, that would be great, but otherwise I will have more of a think. I have tried using the test-gui to rename a file to another filename with different capitalisation and it raises an override warning. The override warning is what we are looking for isn't it?

If the `renameChild` utility function renames the file/directory and then rebuilds the directory tree (from my understanding), and is called after the rename file or rename directory functions, would these functions include the protections for file overwriting. I am likely wrong here so please let me know.

Thanks again

nathanlesage:
I think the issue is that there are two functions responsible for renaming – the corresponding command as well as `renameChild`. Since `renameChild` may be called programmatically without the command running, we have to ensure that both functions employ the appropriate safeguards. I think the confusion stems from the fact that right now there is kind of a shared responsibility, and we may want to move all the logic to the `renameChild` function (i.e. that the command basically just executes the function).

What the `renameChild` function thus should check is:

0. Is the new name exactly the same as the old name? Then just silently return, as nothing has to change (and by not touching the file, we prevent an array of other event listeners to be active)
1. Is the new name equal to the old name regardless of capitalization? In that case, allow the rename.
2. Next, if it is not, is there any other file that already has the new name (regardless of capitalization)? If so, ask before overwriting.
3. Next, if the name is new and unique, just perform the renaming.

What the command should check is:

* Is the file modified?

This check needs to stay in the command as the `renameChild` function has no access to the service container and therefore cannot check if there are currently unsaved changes to it.

Does this make sense?

justjack2347:
Yes, it did make sense, thank you. So I have added some of the additional checks in. The only check that cannot be performed within the `renameChild` function is the overwriting check, since that requires access to the service container of the app and which it does not have access to. To counter this, I did add a parameter (which is set to false as default) called `allowOverwrite.` In the case where the new name is not equal to the old name case insensitively, it will check if the function is allowed to overwrite files (set by parameter). While it would be preferable to have the functionality for asking about an overwrite in the function, this way of implementation would mean that if this function is called, it will default to not automatically overwriting. 

It is worth noting that setting the `allowOverwrite` boolean isn't required for the current implementation since in the rename-file code, it runs the overwrite check and if that we allow the overwrite, it just deletes the file currently there.

Also, regarding moving the logic to the `renameChild` function, I think that most of the code in the rename-file code is dealing with the `appServiceContainer` and so I don't believe this can be easily transferred across.

Is there any concerns about this?

justjack2347:
Hi @nathanlesage, just wondering whether there were any thoughts about this? Is defaulting to safe behaviour (not overwriting) an acceptable approach to ensure using the function will not result in the files being overwritten unless the `allowOverwrite` is explicitly set to true?

nathanlesage:
Hey, I took a Short break from coding, but slowly getting back to it. I need to take another look at your new changes, then I'll reply more thoroughly again!

source/app/service-providers/fsal/fsal-directory.ts
```diff
@@ -399,8 +401,14 @@ export async function renameChild (
   newName: string,
   parser: (file: MDFileDescriptor, content: string) => void,
   sorter: (arr: AnyDescriptor[], sortingType?: string) => AnyDescriptor[],
-  cache: FSALCache
+  cache: FSALCache,
+  allowOverwrite: boolean = false
```
nathanlesage, line 405: I would suggest renaming this parameter to `forceOverwrite`. The reason is that, somehow (this may just be my personal preference, so feel free to comment on this suggestion by me), `allowOverwrite` sounds to me as if there was user choice involved, whereas `forceOverwrite` has a more programmatic sound to it.

Since we now do have the proper checks in the file renaming command, I don't think we ever need to ask the user to set this parameter. Instead, it makes sense to include this parameter purely for potential future use if we at some point allow Zettlr to programmatically rename files. Since it works properly in this implementation, I'd suggest we definitely keep it for future use, and if the parameter is never used, then, well, so be it.

nathanlesage (approved):
I think this all looks very solid. I had one comment, but that doesn't involve change if you disagree. Please either implement my suggestion, if you agree it makes sense, or simply comment that you disagree, and afterwards I will merge this PR!

Thanks already for your work here!

source/app/service-providers/fsal/fsal-directory.ts
justjack2347: I think that this change makes sense, change should be part of the pull request now.

nathanlesage: Wonderful, thank you!

nathanlesage:
Thank you so much for your work! I'll wait until the unit tests have finished, and then merge!

PR #4354, opened 2023-05-18
Upgrade Mermaid to 10.x.x

tz-lom:
## Description
Upgraded Mermaid to 10.1.0

## Changes
Mermaid dependency was upgraded
New version provides only asynchronous api, code calling it was adjusted using `.then` pattern

Tested on: 
Linux


nathanlesage:
Just saw that some time in between this PR and today, mermaid has been bumped to v10.

Issue #4340, opened 2023-05-15
Multiple bugs with comment function


noah-arroyo:
### Description

(Uncertain if this should be split into multiple issues.)

**Expected behavior:** 

1. A comment remains intact even if the user types a space after the comment's closing phrase (`-->`).
2. Comment-style rendering applies only to text inside the comment's bounds.
3. Text within a comment isn't tallied in the word count.

**Buggy behavior:** 

1. When the user types a space after a comment's closing phrase (if that comment does not begin a line), auto-correct turns `-->` into `→`, which disables the comment.
2. When a comment begins a line, comment-style rendering applies to all text in that line, including text occurring outside the comment. When this bug occurs, bug No. 1 will not occur in the same line.
3. Zettlr's word-count function counts the text inside the comment, as well as the comment's opening and closing phrases ("`<!--` this is a comment `-->` and this is text outside the comment" is counted as 13 words).

### Reproducing

(Already explained above.)

### Zettlr Version

Beta (if applicable)

### Specify version

3.0.0-beta.4

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 10

justjack2347:
@josh-133, @Mr-Verb and I will look to tackle some of this over the next week or so.

justjack2347:
Hi @noah-arroyo. Using release 3.0.0-beta.4 on a Windows 11 operating system, I am unfortunately unable to reproduce the first bug, however can reproduce both other bugs as noted. Hopefully either @josh-133 or @Mr-Verb can replicate your issue so we can find a solution. Cheers

justjack2347:
I have found an instance where the first bug can be reproduced. I was able to reproduce it when using the command "ctrl-shift-c" to create a comment around a selected piece of text. When doing this however, the comment styling bug does not occur. @noah-arroyo, was this sort of behaviour consistent with what you were experiencing or could you face all bugs at the same time?

Mr-Verb:
Did you mean this release: [Release v3.0.0-beta.4](https://github.com/Zettlr/Zettlr/releases/tag/v3.0.0-beta.4)? I downloaded the source code and ran yarn test-gui but cannot recreate the issue where the closing comment tag auto corrects to an arrow. I am on a windows 10 system running a WSL terminal. 

[...]

PR #4367, opened 2023-05-24
Change replacement of '--' to regex, to ignore autocorrect for <!--

justjack2347:
## Description

This changes the '--' replacement to a regex expression which does not allow '--' to autocorrect if the characters '<!' are in front of them. This should stop any problems with the term \
Changed the replacement string for '--' to a regex expression '/(?<!<!)--/'.

## Additional information

This is in regards to the issue #4340. Most changes recommended have been committed by @nathanlesage however I believe that this problem still persists. 
It is worth noting that the regex is potentially more confusing in the autocorrect tab of the preferences.

Additional contributors:
@josh-133 
@mr-verb

Tested on: Windows 11

### DIFF
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -195,7 +195,7 @@ export default function getConfigTemplate (): ConfigOptions {
           { key: '(r)', value: '®' },
           // Interpunctation
           { key: '...', value: '…' },
-          { key: '--', value: '–' },
+          { key: '/(?<!<!)--/', value: '–' },
           { key: '---', value: '—' }
         ]
       } // END autoCorrect options


nathanlesage:
As much as I like the idea, this won't work, unfortunately.

* The autocorrect values must be simple strings. This RegExp is not going to work, it will simply look for the literal string of characters you put in there.
* Furthermore, allowing regular expressions would open the door to potential DoS attacks and unnecessarily complicate the logic.
* The correct way is to look for "protected nodes", as you will see in the corresponding file that contains the plugin's code here: https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/commands/autocorrect.ts
* Lastly, as you will see there, I already added the last missing node that should not be autocorrected, inline `Comment`s. I did that after I closed issue #4340, when I noticed that I forgot to make that change.
* This being said, the bug fortunately does not happen anymore.

If you would like to improve the autocorrect, one thing that I would love to check out but don't have the time is if we still need the logic to check for YAML frontmatters inside the autocorrect function. Specifically, I am talking about this snippet:

https://github.com/Zettlr/Zettlr/blob/e920dee0741641c82707e04a30d9002ecb74d00a/source/common/modules/markdown-editor/commands/autocorrect.ts#L80-L84

I already added horizontal lines to the protected nodes, so I'm not even sure that this failsafe is necessary to detect horizontal rules. YAML frontmatters, on the other hand, are not yet part of the protectedNodes, but in general this approach should also work.

In other words, what you could do is try to get rid of the lines of code that I presented above, and instead add the correct nodes to the protected-check in the same file here:

https://github.com/Zettlr/Zettlr/blob/e920dee0741641c82707e04a30d9002ecb74d00a/source/common/modules/markdown-editor/commands/autocorrect.ts#L36-L45

I'm not 100% sure this would work, but someone needs to test this. If this works, this would clean up the function a lot. Additionally, this change would ensure that there are no autocorrections within the YAML frontmatter, which currently can happen (and therefore interfere with them).

justjack2347:
Thanks @nathanlesage, I apologise for not realising that basic strings were the only form for replacements, for some reason I was under the belief I had tested its usage. And thanks for suggesting an issue to work on, we will have a look into it!

Using the current version, I still believe that I am experiencing the autocorrect bug. I think that the inline 'Comment' node is only being created after the final --> is created. This means that when using "ctrl+shift+c" to add a comment, there are no problems since it has finished creating the 'Comment' node which is protected. However, if you manually type a comment, the opening \, the inline 'Comment' is not generated and so no autocorrects are blocked.

I hope this makes sense and I am not missing something obvious. Cheers

nathanlesage:
Ahh okay, yeah I see, the block comment exists as soon as you type the opening, and that was what I was thinking about and causing the confusion! Although in that case what people can do if they frequently do that is manually add a space in front of the `--` autocorrect to ensure this doesn't happen, so it's something the users can do if they do fancy! Nothing we should worry about.

And no need to apologize, I am more than happy about any help I can get with the app! :)

PR #4378, opened 2023-05-27
Fix display of title for files

### LINKED ISSUE #4363: FSAL (?) bug with titles from yaml
### Description

If `fsal/cache` is available, the title from the YAML frontmatter is not shown in the working directories, even if configured.

With beta-3 everything was working fine.

Referencing https://github.com/Zettlr/Zettlr/issues/3887#issuecomment-1517329762 which looks like the same issue.

### Reproducing

In the settings, choose to show the title of files if available (or however it is named in English).

Regardless whether a title is available in YAML-frontmatter or not, the file name is shown.

Stop Zettlr, remove the `fsal` directory in the config and start Zettlr again and everything is shown correctly until the next restart.

### Zettlr Version

Beta (if applicable)

### Specify version

3.0.0-beta.4 (but reproducible with current develop-branche)

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Ubuntu 22.04

### Additional Information

_No response_

UFOMelkor:
Resolves #4363

The current implementation of `safeAssign` [strips properties with `undefined`](https://github.com/Zettlr/Zettlr/blob/develop/source/common/util/safe-assign.ts#L60) as value from the reference. Hence, the properties from the `obj` are not passed.

As the [`yamlTitle` is initialized with `undefined`](https://github.com/Zettlr/Zettlr/blob/e920dee0741641c82707e04a30d9002ecb74d00a/source/app/service-providers/fsal/fsal-file.ts#L94) it is not copied from the cache.

### DIFF
--- a/source/common/util/safe-assign.ts
+++ b/source/common/util/safe-assign.ts
@@ -20,13 +20,12 @@
 /**
  * Overwrites properties on the referenceObject, if they also occur in obj.
  * @param {any} obj The new object to be safe assigned
- * @param {any} referenceObject The reference to use the props from (without props with undefined as value)
- * @param {any} keyReferenceObject The original reference to check whether a key occurs
+ * @param {any} referenceObject The reference to use the props from
  */
-function safeAssign (obj: any, referenceObject: any, keyReferenceObject: any): void {
+function safeAssign (obj: any, referenceObject: any): void {
   // Overwrite all given attributes (and leave the not given in place)
   // This will ensure sane defaults.
-  for (const prop in keyReferenceObject) {
+  for (const prop in referenceObject) {
     if (prop in obj) {
       // safeAssign updates even nested objects, which we'll
       // do here. The "else" is for primitives. "Why do you
@@ -37,7 +36,7 @@ function safeAssign (obj: any, referenceObject: any, keyReferenceObject: any): v
         !Array.isArray(referenceObject[prop]) &&
         referenceObject[prop] !== null) {
         // Update the sub-object
-        safeAssign(obj[prop], referenceObject[prop], keyReferenceObject[prop])
+        safeAssign(obj[prop], referenceObject[prop])
       } else {
         // Assign the primitive
         referenceObject[prop] = obj[prop]
@@ -58,8 +57,8 @@ function safeAssign (obj: any, referenceObject: any, keyReferenceObject: any): v
 export default function <A, B> (obj: A, reference: B): B {
   // Make sure to clone the reference object, so that users
   // do not have to worry about doing this themselves.
-  const clone = JSON.parse(JSON.stringify(reference))
+  const clone = structuredClone(reference)
   // After cloning, safely assign the object to the reference
-  safeAssign(obj, clone, reference)
+  safeAssign(obj, clone)
   return clone
 }
--- a/source/common/util/safe-assign.ts
+++ b/source/common/util/safe-assign.ts
@@ -20,12 +20,13 @@
 /**
  * Overwrites properties on the referenceObject, if they also occur in obj.
  * @param {any} obj The new object to be safe assigned
- * @param {any} referenceObject The reference to use the props from
+ * @param {any} referenceObject The reference to use the props from (without props with undefined as value)
+ * @param {any} keyReferenceObject The original reference to check whether a key occurs
  */
-function safeAssign (obj: any, referenceObject: any): void {
+function safeAssign (obj: any, referenceObject: any, keyReferenceObject: any): void {
   // Overwrite all given attributes (and leave the not given in place)
   // This will ensure sane defaults.
-  for (const prop in referenceObject) {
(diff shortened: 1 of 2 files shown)


Mela:
Hi,
I was told on the Discord chat, that the nightly build including your fix would remove the issue. 

It worked, but only after the first start after installing the nightly build. After the next start Zettlr again only displayed the IDs but not the title. 

Zettlr 3.0.0-beta.4.nightly (UUID: 94b6d7bf-8f90-490d-896e-0047dd29cec2)
System: MacOS/darwin 13.3.1 (architecture: x64)

Bildschirmfoto 2023-06-01 um 10 14 21

Regards,
Mela

nathanlesage:
@Mela This PR is still open, which means it is not yet part of the code base and hence not in the builds. As soon as this PR is merged, it will be included in the releases.

That being said, @UFOMelkor, I do not like having a third parameter passed in. Instead, would you please mind having a look if we can completely ditch any undefined and/or null checks? `safeAssign` is intended to use only the props of the `referenceObject` (to prevent rogue attributes being assigned), but back when I implemented the function, I did not consider that `null` or `undefined` would also be proper values. I think it should be safe to remove those checks to allow properties to be undefined/null, because we're still only looking at properties that are set on the reference object. (Which should be the same as what you've called a key reference object).

The only real improvement over that would be proper type checks, but I think that would be overkill.

UFOMelkor:
@nathanlesage 

I'm uncertain whether we are talking about the same thing. `null` is no problem at all. The difficulty is that JSON does not know the value undefined, and hence `JSON.parse(JSON.stringify(reference))` removes every undefined value.

So when we do
```js
const obj = {a: 'text'};
const ref = {a: undefined};
safeAssign(obj, ref);
```
it will call the (internal) method like `safeAssign(obj = {a: 'text'}, ref = {})`, iterate on the props of `ref` and do not assign `text` to `ref.a`. The only alternative, I imagine iterating on the keys of the original reference, would be iterating on `Object.keys(referenceObject).concat(Object.keys(obj))`. But this would lead to the following scenario (like the second input in `safe-assign.spec.ts`):

```js
const obj = {a: 'text', b: 'prop to be ignored because it does not exist in ref'}
const ref = {a: undefined}
safeAssign(obj, ref);
console.log(ref); // { a: 'text', b: 'prop to be ignored because it does not exist in ref'}
```

nathanlesage:
We are talking about the same issue indeed, but describe a different problem: the ref will never be a deserialised JSON string. Rather, it will have all properties populated during runtime, including some undefined. However, since we're operating on the keys, as long as the for-loop accesses set properties -- even if they are set to undefined -- this should work unproblematic, so I think your first example should work with the function, no...?

OH WAIT

Because I was afraid something might go wrong, I created a crappy JSON deep copy of the ref object. The safeAssign function should work with undefined properties of the ref, but I'm actually wrapping that function again. I don't know if we should do a proper deep clone of the ref object in the wrapper (that preserves potential undefined) or simply mutate the ref object directly (then it would be no longer functional code)?

Do you have an opinion? The only thing that safeAssign should do is only care for properties set on ref, other than that it should just copy anything over from the (cache) object.

UFOMelkor:
That's the charm of the third parameter. It's backwards compatible because it only affects the internal method, which is wrapped anyway.

Doing a deep clone of the object manually sounds like it could become more complex than it seems. Moreover, performance might be a problem.
Using the original object instead of a clone, on the other hand, might have side effects.
 
How about using [`structuredClone(reference)`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) instead of `JSON.parse(JSON.stringify(reference))`?

nathanlesage:
Oh, I didn't know about structuredClone! Yes, that sounds good (but without transferring anything, to actually create a clone without modifying the original object). That's definitely better than the third parameter!

UFOMelkor:
Tests are failing because they are running on Node.js 16, structuredClone is available since Node.js 17. Since we are using [Electron ^24.1.2 which upgrades Node.js to 18](https://www.electronjs.org/blog/electron-24-0), I guess it should be ok to enhance the Node.js version to 18 in all GitHub Workflow?

nathanlesage:
> I guess it should be ok to enhance the Node.js version to 18 in all GitHub Workflow?

Yes, it should be! I think we did not (yet) switch to Node 18 on the workflows, since there were some weird bugs a year ago. But I just updated the Unit test Node version to v18, so if you'll quickly rebase your branch to the current develop branch, the unit tests here should begin to run on v18 and thus no longer fail. Then we can merge! :)

In the meantime, I'll be updating Node for the rest of the workflows.

UFOMelkor:
Done :+1:

nathanlesage:
Wonderful, and unit tests work out! So let's merge! :)

PR #4379, opened 2023-05-27
Fixes the tag-extraction from files and searching by tags

### LINKED ISSUE #4366: Keywords not recognized in YAML frontmatter
### Description

```
—-
keywords: [foobar]
—-

Some Text
```
No variation of keywords are recognized: [foobar], foobar, - foobar. Changing keywords to tags doesn’t work either. Adding #foobar outside the frontmatter works.

### Reproducing

Copy minimal md above.

### Zettlr Version

Beta (if applicable)

### Specify version

3.0.0-beta 4

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 11

### Additional Information

_No response_

UFOMelkor:
Resolves #4366 

This pr resolves some errors regarding tags:

- Tags from within the yaml-body (`This is #a_tag`) were used along with their `#`. Hence, they were shown in the tag cloud with `#`.
- The properties `keywords` and `tags` were parsed in the fsal, but not assigned to tags and therefore ignored.
- The search for tags did not work. It assumed that the tags are already transformed to lower case, even if this was not the case.

### DIFF
--- a/source/app/service-providers/fsal/util/file-parser.ts
+++ b/source/app/service-providers/fsal/util/file-parser.ts
@@ -71,7 +71,7 @@ export default function getMarkdownFileParser (
     const ast = md2ast(content)
 
     const tags = extractASTNodes(ast, 'ZettelkastenTag') as ZettelkastenTag[]
-    file.tags = tags.map(tag => tag.value.substring(1))
+    file.tags = tags.map(tag => tag.value.toLowerCase())
 
     const links = extractASTNodes(ast, 'ZettelkastenLink') as ZettelkastenLink[]
     file.links = links.map(link => link.value)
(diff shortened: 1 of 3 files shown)


source/app/service-providers/fsal/util/file-parser.ts
```diff
@@ -71,7 +71,7 @@ export default function getMarkdownFileParser (
     const ast = md2ast(content)
 
     const tags = extractASTNodes(ast, 'ZettelkastenTag') as ZettelkastenTag[]
-    file.tags = tags.map(tag => tag.value)
```
nathanlesage, line 74: Before doing this, we would need to discuss if the `#` sign is important for the tags in the AST. If it is not, we should remove it in the AST parser already, not here.

Look here: https://github.com/Zettlr/Zettlr/blob/e920dee0741641c82707e04a30d9002ecb74d00a/source/common/modules/markdown-utils/markdown-ast.ts#L877

source/app/service-providers/fsal/util/file-parser.ts
```diff
@@ -115,6 +115,21 @@ export default function getMarkdownFileParser (
           file.yamlTitle = title
         }
       }
+
+      const tagsFromFrontmatter = []
+      if (file.frontmatter.tags) {
+        tagsFromFrontmatter.push(...file.frontmatter.tags)
```
nathanlesage, line 121: This assumes that `file.frontmatter.tags` is an array, but it can be a string just as well, so you should add a few more checks. Have a look at the old implementation that worked well:

https://github.com/Zettlr/Zettlr/blob/v2.3.0/source/app/service-providers/fsal/util/extract-tags.ts

source/app/service-providers/fsal/util/search-file.ts
```diff
@@ -72,7 +72,7 @@ export default function searchFile (fileObject: MDFileDescriptor|CodeFileDescrip
           // Break because only one match necessary
           break
         }
-      } else if (wd[0] === '#' && fileObject.type === 'file' && fileObject.tags.includes(wd.toLowerCase().substring(1))) {
+      } else if (wd[0] === '#' && fileObject.type === 'file' && fileObject.tags.map((each) => each.toLowerCase()).includes(wd.toLowerCase().substring(1))) {
```
nathanlesage, line 75: You should absolutely do the lowercasing in the file parser above, and NOT here! Always normalize the tags where they are at.

nathanlesage (changes requested):
I didn't realize that I had nuked the YAML tag detector when implementing the AST parser — I was always kind of like "… why do people say that YAML tags do not work anymore …?" 😂 

In other words, thanks for fixing this! There are only a few additional things to fix!

source/app/service-providers/fsal/util/file-parser.ts
nathanlesage: Just thought about it: We've always removed the `#`, so you can add the `substring(1)` (or `slice(1)`, which I've been using lately) to the AST parser itself which I linked.

UFOMelkor:
Should be all done :-)

source/common/modules/markdown-utils/markdown-ast.ts
```diff
@@ -863,7 +863,7 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
         name: 'ZknTag',
         from: node.from,
         to: node.to,
-        value: markdown.substring(node.from, node.to)
+        value: markdown.substring(node.from, node.to).slice(1)
```
nathanlesage, line 866: Why Not `node.from + 1`? :) Other than that looks great!

nathanlesage (approved):
Just a single thing, but other than that it's mergeable!

UFOMelkor:
Thank you for your fast reviews :+1:

nathanlesage:
Wonderful, looks great! And thanks for keeping my slightly annoyed comment on how YAML can be a hassle 😂 

Merging! Thanks for the awesome work!

nathanlesage:
@UFOMelkor Somehow the unit tests are now flaky, apparently something broke with this change. Would you please investigate?

See: https://github.com/Zettlr/Zettlr/actions/runs/5105904542/jobs/9177778190

PR #4381, opened 2023-05-28
Add support for kroki diagrams

UFOMelkor:
Currently, just a rudimentary WIP which needs more work, especially for edge cases.


dustdfg:
It doesn't seem to me a very good idea to use external site. Local tool is better IMO

nathanlesage:
Possibly, I'll set it on the agenda for the next Steering meeting on Jan 30!

ODAncona:
Awesome feature, I would have loved to have plantuml diagrams or mermaid embedded into Zettlr. Nevertheless, I believe it goes against the local philosophy of Zettlr. Would it be possible to have it running locally in a docker image and Zettlr using it that way ? The sameway plantuml extention works on vscode...

Thank you for your contribution

PR #4386, opened 2023-05-29
Run the check workflow on pull-requests

UFOMelkor:
Following [the documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request), adding `pull_request` should run the checks for every PR.


nathanlesage:
Interesting, that must've been a recent API change that I accidentally missed – the `push` thing was always working before for the PRs.

Thank you for digging into the issue and fixing this!

PR #4389, opened 2023-05-30
YAML Refactoring Protected Nodes

justjack2347:
## Description

This pull request looks to refactor the autocorrect code to make the replacement function easier to read. This is similar to what was discussed by @nathanlesage in pull request #4367. It makes '---' at the start of a file be parsed as a 'YAMLFrontmatterStart' node, which has now also been set to be a protected node.

It is worth noting that this was discussed in order to make the code simpler in the autocorrect function and include more in the protected nodes section. However, while we have been working to make more of the checks be whether the text is protected nodes, I am not sure whether the rest of the code has become more messy. Happy to discuss

## Changes

This has set the YAMLFrontmatterStart to be a protected node. The `posInProtectedNode` function has also been extended to check the position one position to the left of the cursor for the horizontal rule since for some reason, the horizontal rule nodes are not returned if the cursor is the final position. Then if you put the horizontal rule node as a protected node, if you are at the end of the '---' and click space or enter, it would not pick up that you are in a protected node and so will still autocorrect. Hence, there is a change that will test for these nodes in the position one place back from the current cursor location.

Then there is a change in the YAML parser where if it picks up that there are three dashes '---' at the start of the document, it will keep checking for the closing of the YAML frontmatter. However, my change is that if it cannot find the end, it will just assume that the first three dashes '---' are a frontmatter start. This does stop the parsing of the rest of the document however we might be able to run a partial parse over the rest of the document (I don't know how this works but I think @josh-133 and @mr-verb will have a look at this over the next few days). However, once the YAML frontmatter is closed, it reverts to being just a normal YAML frontmatter.

## Additional information

Additional Contributers:
@mr-verb
@josh-133 

Tested on: Windows 11


Mr-Verb:
After doing some experimentation, we found that the following need to be fixed:
Space after "---" changes node to not recognise YAML format on top line
Space after "---" anywhere else does not autocorrect to horizontal rule

@josh-133 and I will fix this now

Mr-Verb:
@josh-133 and I have worked on fixing the bugs identified by my first comment. 
Putting a space after the first '---' still deidentifies the YAMLFrontmatter
Putting '---' anywhere else in the document autocorrects to a horizontal rule. 
Below is a video outlining how YAMLFrontmatter works after the changes.
https://github.com/Zettlr/Zettlr/assets/50310102/f671d560-bc9e-4a2b-8fd8-1b9c43b07ec6

justjack2347:
@nathanlesage, I have heard from @Mr-Verb and @josh-133 that this pull request is at the stage for review. Keen if you have any feedback.

nathanlesage:
Currently on a business trip in the US, can look at it closer toward the end of June. I have the notifications on unread so I can slowly work through everything.

justjack2347:
No worries at all. Cheers!

source/common/modules/markdown-editor/commands/autocorrect.ts
```diff
@@ -35,12 +35,20 @@ const startChars = ' ([{-–—\n\r\t\v\f'
  */
 function posInProtectedNode (state: EditorState, pos: number): boolean {
   const node = syntaxTree(state).resolve(pos, 0)
+
+  // Having the cursor at the end of the HorizontalRule '---' is not considered inside the HorizontalRule node
+  // So we check the previous position to see if it is in a HorizontalRule node
+  const checkYAMLFrontmatterNode = syntaxTree(state).resolve(pos - 1, 0)
```
nathanlesage, line 41: The second resolve parameter indicates a "side" which, if I read the docs correctly, indicate into which direction it should look to find a node. It seems safer to me to use that parameter – if it does what you desire. Could you test this out? (Refer to the CM docs for a description of that parameter)

source/common/modules/markdown-editor/commands/autocorrect.ts
```diff
@@ -35,12 +35,20 @@ const startChars = ' ([{-–—\n\r\t\v\f'
...
+  if (checkYAMLFrontmatter) {
+    return checkYAMLFrontmatter
+  }
   return [
     'InlineCode', // `code`
     'Comment', 'CommentBlock', // <!-- comment -->
     'FencedCode', // Code block
     'CodeText', // Code block
-    'HorizontalRule'
+    'HorizontalRule' // ---
```
nathanlesage, line 51: (or `***`)

source/common/modules/markdown-editor/commands/autocorrect.ts
```diff
@@ -35,12 +35,20 @@ const startChars = ' ([{-–—\n\r\t\v\f'
  */
 function posInProtectedNode (state: EditorState, pos: number): boolean {
   const node = syntaxTree(state).resolve(pos, 0)
+
+  // Having the cursor at the end of the HorizontalRule '---' is not considered inside the HorizontalRule node
+  // So we check the previous position to see if it is in a HorizontalRule node
+  const checkYAMLFrontmatterNode = syntaxTree(state).resolve(pos - 1, 0)
+  const checkYAMLFrontmatter = [ 'YAMLFrontmatterStart', 'YAMLFrontmatterEnd', 'HorizontalRule' ].includes(checkYAMLFrontmatterNode.type.name)
+  if (checkYAMLFrontmatter) {
+    return checkYAMLFrontmatter
```
nathanlesage, line 44: If my above comment works (see the same function call on line 37, I think it should work if you replace the second parameter with `-1`, but again, please check the docs first), then you can remove this entire check and add those nodes below to the array of protected nodes, cleaning this up a little.

source/common/modules/markdown-editor/parser/frontmatter-parser.ts
```diff
@@ -16,6 +16,7 @@ import { type BlockParser } from '@lezer/markdown'
 import { StreamLanguage } from '@codemirror/language'
 import { yaml } from '@codemirror/legacy-modes/mode/yaml'
 import { partialParse } from './partial-parse'
+// import { log } from 'console'
```
nathanlesage, line 19: Please remove this line

source/common/modules/markdown-editor/parser/frontmatter-parser.ts
```diff
@@ -49,6 +50,9 @@ export const frontmatterParser: BlockParser = {
       // The parser has collected the full rest of the document. This means
       // the frontmatter never stopped. In order to maintain readability, we
       // simply abort parsing.
+      const node = ctx.elt('YAMLFrontmatterStart', 0, 3)
+      ctx.nextLine()
+      ctx.addElement(node)
```
nathanlesage, line 55: Please remove this. There is a reason an unclosed frontmatter is not parsed as a frontmatter. This will break compatibility with Pandoc and break expectations.

nathanlesage (changes requested):
I think this PR is slowly coming together, but there are a few things I need you to take a second look at – please see my comments for that!

nathanlesage:
> it isn't initially parsed as a HorizontalRule by Lezer

I saw this myself, and had a look at the CommonMark spec: It doesn't say that a horizontal line on the very first line shouldn't be parsed as an HR. I think this may just be an oversight, but indeed your feeling is right: This means we do have to persist the manual check for HRs at the very beginning.

nathanlesage:
Yeah, but it might be good to just add a comment to say why we have this manual check, so we don't forget it again and wonder why we don't do this via nodes! Also somehow GitHub doesn't show a proper diff anymore, so I think you need to do a quick rebase so that I can double check your code one last time before merging it!

justjack2347:
Umm, so I have tried to rebase but I have done something wrong.

justjack2347:
Have closed this in favour of a cleaned up pull request #4447

PR #4396, opened 2023-06-04
Update readme to reflect deprecation of --frozen-lockfile

mattstring:
...which is replaced with --immutable.

## Description
Update readme to reflect changes within yarn to replace --frozen-lockfile with --immutable. This is helpful for reducing output bloat when setting up the repo in a new workspace.

See https://classic.yarnpkg.com/lang/en/docs/cli/install/ and https://yarnpkg.com/cli/install for old and new flags respectively.

## Changes
Simply replace --frozen-lockfile flag in the setup instructions with --immutable.

## Additional information
There is also a new --immutable-cache flag;  I'm not sure if this is relevant or not for Zettlr.

Tested on: Windows 10 build 19045.2965

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,10 +116,10 @@ Then, simply clone the repository and install the dependencies on your local com
 ```bash
 $ git clone https://github.com/Zettlr/Zettlr.git
 $ cd Zettlr
-$ yarn install --frozen-lockfile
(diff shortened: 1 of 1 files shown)


nathanlesage:
Good catch, thank you very much!

Issue #4327, opened 2023-05-14
Numbers don't update correctly when reordering list items


noah-arroyo:
### Description

**Expected behavior:** When moving items up or down in an ordered list, the list content should move while the list numbers don't, maintaining a proper sequence. 

**Buggy behavior:** When moving items up or down in an ordered list, the list numbers follow the content, breaking the proper sequence of the list. 

### Reproducing

Write a list:

1- apple
2- banana
3- coconut
4- daikon

Navigate the caret to list item #2, press `Alt`+`↑` to move that item into #1 position. That yields the following output:

2- banana
1- apple
3- coconut
4- daikon

(I'm using "2-", "1-" etc. because if I use "2.", "1." etc. then github will render them sequentially rather than in my custom order, above.)

### Zettlr Version

Beta (if applicable)

### Specify version

3.0.0-beta.4

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 10

Kinack987:
I'm having a look into this issue now. Should hopefully have more details and potentially a PR up soon.

mattstring:
I've played around with this a bit and it isn't specific to moving lines with alt+up/alt+down, ordered lists item numbers don't update in general when altering them in some way (For example, if you have list 1 2 3 4 5 6, and you copy element 6 to clipboard then paste it above element 2, the list is now ordered 1 6 2 3 4 5) (The same behaviour as described above, but occurs in general when moving items around in existing lists).

Kinack987:
I've opened a pull request #4406 to fix this issue. However, for now, I have only addressed the case of moving lines with alt+up/alt+down. There is another pinned issue #937 open since 2020 which addresses the issue of fixing list numbers when deleting / copy-pasting items in an ordered list. As such, I'd suggest this issue could be safely closed if the above pull request is found to be sufficient.

noah-arroyo:
Ah gotcha, good catch. Guessing this requires someone else's weigh in -- I don't object to what you're describing.

nathanlesage:
As I've also written in the PR: There is this function that can automatically correct list markers for those lists that are touched by any of the selections within the editor.

[...]

Issue #937, opened 2020-06-16
[ENHANCEMENT] Keeping numbering consistent when deleting items from a numbered list


peczony:
## Description

There should be a way to keep numbering consistent when deleting items from a numbered list.
What I’m talking about is: suppose you have a list like

```
1. one
2. two
3. three
4. four
```

Then you delete point 2. You are left with

```
1. one
3. three
4. four
```

That’s not a problem if you export the content but when you e.g. copy Markdown, the inconsistency is still there. It would be better if there was some way to recalculate the numbering.

## Proposed Changes

I can think of two possible ways to solve this.

(1) When one presses enter to create a new list item, the numbering could be automatically recalculated, so instead of

```
1. one
3. three
4. four
5.
```

it would be

```
1. one
2. three
3. four
4.
```

Recalculating should also be done on indent change as it effectively creates new lists?

(2) A context menu item that would allow to recalculate numbering in the currently selected list

sebastienvercammen:
Side note: GitHub supports the following:

```
1. a
1. b
1. c
```

and it'll render as

```
1. a
2. b
3. c
```

as demonstrated:
1. a
1. b
1. c

A similar approach will simplify all numbering. I don't know how this affects other systems though (but if it does, perhaps it's time for this to become a standard?).

nathanlesage:
> A similar approach will simplify all numbering. I don't know how this affects other systems though (but if it does, perhaps it's time for this to become a standard?).

In Zettlr, you can also do lists like such:

```
#. One
#. Two
#. Three
```

And it'll automatically number; because this is a feature of Pandoc. Oh, and we might want to support that in the list indentation as well … let's remember this!

peczony:
Sure, I’m totally fine with `1. 1.` or `#. #.` syntax. On the other hand it would be convenient to quickly see what the resulting numbering would be without first exporting the content. Like some hotkey for quick non-editable preview maybe?

nathanlesage:
Of course, but there's no hotkey necessary. Just some improvements to the underlying algorithm, which is currently based on a CodeMirror plugin. We'd need to rewrite that thing. Possible, but quite some work!

peczony:
I mean maybe “show quick html preview instead of editable text” is less work than rewriting the plugin, with the same effect (and maybe even better for some use cases) :)

nathanlesage:
> I mean maybe “show quick html preview instead of editable text” is less work than rewriting the plugin

[...]

PR #4406, opened 2023-06-09
New Functionality Fixing List Numbers After Moving Lines

Kinack987:
## Description

Moving lines that are part of an ordered list using 'Alt-UpArrow'/'Alt-DownArrow' will now also correct the list numbers of any affected lists after the move. Fix for #4327.

## Changes

- Added new functions to lists.ts which will call correctListMarkers after moving a line up or down inside an ordered list using the 'Alt-UpArrow'/'Alt-DownArrow' shortcuts.
- Modified keymap.ts to call these custom functions in lists.ts rather than directly calling codemirror's moveLine functions. The modified functions will then call codemirror's moveLine functions.
- Updated CHANGELOG.md to contain an entry for these changes.

## Additional information

Changing keymap.ts I had to replace 'Alt-Up' with 'Alt-ArrowUp' for the relevant keypress to register properly.
It seems 'Alt-Up' may have been mistakenly used in an earlier revision.
I will open an issue explaining this for further reference.
As mentioned in a comment on #4327, list numbers also do not update correctly in other cases, such as when copy-pasting or when deleting a line that is part of an ordered list. This is actually addressed in an old pinned issue from 2020 #937 and is not something I'm able to fix right now. 

Tested on: Windows 11 v22H2


nathanlesage:
This looks good so far. I'm not entirely trusting my very own code here that corrects list markers, but your intuition is absolutely correct. I'll give this a quick try later and if it works, I'll quickly merge your PR. Thank you already for your work!

> Changing keymap.ts I had to replace 'Alt-Up' with 'Alt-ArrowUp' for the relevant keypress to register properly.

This sounds odd. Could be a change in CodeMirror, or indeed a simple mistake by me.

> As mentioned in a comment on https://github.com/Zettlr/Zettlr/issues/4327, list numbers also do not update correctly in other cases, such as when copy-pasting or when deleting a line that is part of an ordered list. This is actually addressed in an old pinned issue from 2020 https://github.com/Zettlr/Zettlr/issues/937 and is not something I'm able to fix right now.

Well, given that my function to correct list markers works, then all we would need to do is do what you've done here but with all other instances where things can get removed.

Actually, this gives me an idea: Wouldn't it be possible to write a `transactionExtender` that checks if the document has been changed, and, if so, does what you've implemented here? Since the function will only alter lists where there is a cursor, this should work with moving list items up and down, adding and deleting list items, and in other instances, all without messing up other lists in the document.

Would you be up for trying this out …? Here's what I have been thinking about: https://codemirror.net/docs/ref/#state.EditorState^transactionExtender

nathanlesage:
That was a long time open — merging, because it works and it fixes even two bug reports at the same time! I'll just quickly let the unit tests go through …

PR #4435, opened 2023-06-19
Translation/ru ru

krote5k:
full russian translate

### DIFF
--- a/static/lang/ru-RU.po
+++ b/static/lang/ru-RU.po
@@ -3,19 +3,19 @@
 # This file is distributed under the same license as the Zettlr package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 #
-#, fuzzy
 msgid ""
 msgstr ""
 "Project-Id-Version: Zettlr 2.3.0\n"
 "Report-Msgid-Bugs-To: https://github.com/Zettlr/Zettlr/issues\n"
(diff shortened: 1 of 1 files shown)


nathanlesage:
Thank you very much for your work in translating the application! Before we can merge, please resolve the conflicts that prevent me from doing so. Also, you have only committed the `*.mo`-file, not the `*.po`-file. Zettlr uses the latter and does not make use of the compiled versions.

krote5k:
Poedit seems to have mixed the strings and now the strings that were untranslated do not conflict with the translated ones.

nathanlesage:
There are somehow still merge conflicts, maybe the weekly translation update now conflicts with your changes …? (I know I have to get back to it at some point and fix the translation updater script as well)

nathanlesage:
There are somehow still merge conflicts. I do not know what is happening, but to speed things up: would you allow me to simply copy and paste your work directly into the dev branch? This may make things quicker.

krote5k:
> There are somehow still merge conflicts. I do not know what is happening, but to speed things up: would you allow me to simply copy and paste your work directly into the dev branch? This may make things quicker.

It will be the surest way.

nathanlesage:
Wonderful, thank you. It is now in the codebase. Thank you again for your work and help.

PR #4447, opened 2023-06-24
Added new protected nodes + comment for '---' autocorrect ignore

justjack2347:
## Description

This is a replacement for pull request #4389 which has cleaned up the commits and has fixed the diff of the file which showed the whole file is changed

## Changes

There are some additional checks for protected nodes and an additional comment regarding why we have kept the manual check for '---' and '...'

## Additional information

Tested on: Windows 11


nathanlesage:
Perfect, thank you for your work and for the investigation! I really appreciate your help!

PR #4508, opened 2023-07-22
feat: Add custom export commands

### LINKED ISSUE #1842: [ENHANCEMENT] Execute arbitrary command for export
## Description
An enhancement would be if one could execute arbitrary commands in a custom export function. Thus, it would be possible to use the editor's Markdown e.g. with a Docker command and produce all kinds of output with Zettlr.

## Proposed Changes
- Add a new icon to the list of export formats named something like "custom"
- Add a new section to the setting tab "Advanced" that relates to this icon.
- Add a new form field to this section that takes custom commands being executed on the text corpus when the custom icon is clicked.

## Caveats
Not imaginable yet. But could be that the custom pandoc command is no longer necessary with this enhancement.

## Additional Information
Main reason for this enhancement request is that we would like to use Zettlr more in the context of the single-source production we have developed in https://doi.org/10.15480/882.2902

nathanlesage:
## Description

This PR adds the ability to add custom commands to Zettlr. This lets users define certain commands that only receive the file or project folder path to export and can then handle a lot by themselves.

## Changes

Added a new configuration setting that allows users to add custom commands to Zettlr and giving them a proper display name. These commands can subsequently be selected in the single file export popover as well as the project settings and enable users to export their files and projects arbitrarily. This could, for example, be a shell script, a docker container, or anything really that processes them and does something to them.

**PLEASE NOTE** that in order to facilitate the new setting in the preferences, I had to heavily rework the ListControl as well. It is now fully typed, has more sanity checks and better support for, e.g., proper column labels. I took the liberty to type the entire preferences form control as well, so you may ignore those changes while reviewing.

## Additional information

Tested on: macOS Ventura 13.4.1. (c)

Closes #1842


nathanlesage:
Merging to have it test-able in today's nightly.

PR #4517, opened 2023-07-26
Search when clicking on frontmatter tags and keywords

### LINKED ISSUE #4498: Keywords YAML does not work in 3.0 beta 6
### Description

It seems Zettlr 3 beta 6 ignores all the keywords in the YAML keyword/tag list. They are no clickable anymore, and they don't result in the buildup of the related files tab.

### Reproducing

1- Create a file with a YAML front end.
2- Add keywords to the YAML front end.
3- Observe the behavior described above :-)

### Zettlr Version

Beta (if applicable)

### Specify version

3.0 beta 6

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 11

### Additional Information

If I move the keywords out of the YAML front end and write it with hashtag it works. The problem is that my 500+ notes all use YAML keywords :-(

UFOMelkor:
## Description

Triggers a change when a keyword or a tag in YAML Frontmatter is clicked.

## Changes

I could not get it working with simple using the information from the frontmatter-parser because the nodes were neither accessible nor hierarchical. I looked at the implementation of version 2.3 and stole the idea to use the YAML parser. It works so far, but it won't help for implementing the autocompletion for tags and keywords, where the same problem will occur.

## Additional information
Resolves #4498 

Tested on: Ubuntu 22.04

### DIFF
--- a/source/common/modules/markdown-editor/index.ts
+++ b/source/common/modules/markdown-editor/index.ts
@@ -83,7 +83,7 @@ import { addNewFootnote } from './commands/footnotes'
 import { copyAsHTML, pasteAsPlain } from './util/copy-paste-cut'
 import openMarkdownLink from './util/open-markdown-link'
 import { highlightRangesEffect } from './plugins/highlight-ranges'
-import { isAnyLiteralNodeContainingATag } from './util/yaml-tag-detection'
+import { isThisNodeAnyLiteralNodeInAYamlKeywordsOrTagsPair } from './util/yaml-tag-detection'
 
 import safeAssign from '@common/util/safe-assign'
 import { countAll } from '@common/util/counter'
@@ -387,7 +387,7 @@ export default class MarkdownEditor extends EventEmitter {
 
             const innerNodeAt = syntaxTree(view.state).resolveInner(pos, 0)
 
-            if (isAnyLiteralNodeContainingATag(innerNodeAt, view.state)) {
+            if (isThisNodeAnyLiteralNodeInAYamlKeywordsOrTagsPair(innerNodeAt, view.state)) {
               let tag = view.state.sliceDoc(innerNodeAt.from, innerNodeAt.to)
               if (innerNodeAt.type.name === 'QuotedLiteral') {
                 tag = tag.substring(1, tag.length-1)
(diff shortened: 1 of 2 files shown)


nathanlesage:
So, first of all sorry for answering this late. This still seems a bit hacky to me, but to be honest I don't yet have a good other way. Maybe wrapping up the entire Frontmatter in its own node…? Then we could check if the clicked position is somewhere in there, then just parse the entire chunk with YAML and retrieve the node at the given position (the YAML parser retains quite a lot of info, fortunately).

I don't want to include that in 3.0.0 stable yet, because I'm afraid there might be things breaking, but I feel you showed something very valuable here that should be included asap afterwards!

nathanlesage:
That looks like some serious spaghetti code, but I absolutely see that this happens with the parser 😅 But it is very nice to see that you could make fast progress! We only need to handle `keywords`, too, but I believe then with a bit of cleanup we should be good to go :)

nathanlesage:
Will do, I'll put it on my todo list for the weekend!

nathanlesage (changes requested):
Finally, I was able to review this thing!

First things first: I think that your code has a very high quality. It is very efficient, and tries to be as minimal as possible, which is a great feat especially in the context of Lezer parse trees, because these things have literally been built on heuristics and rules of thumb.

As you will see, none of my comments criticizes the functionality of what you're doing, they only indicate where some clarity is needed, especially when it comes to maintaining the functionality in a year or two from now.

Specifically, you have done a great job understanding the parse trees that Lezer will build for YAML code, and it speaks. However, it is crucial that you add your insights using comments; ideally into the YAML tag detection utility file as a somewhat longer comment below the imports and above the function declarations. The more you share about how these trees work and why, e.g., you have this weird childAfter-childAfter chain in there, the faster anyone (possibly me but also others) will understand what the function does. This way we can account for changes in how the parse trees look like in a year from now.

All in all, great stuff, it only suffers from a few small documentation issues that should be straight forward to fix!

UFOMelkor:
@nathanlesage 

I rebased for the current version. Is anything missing here, or is there anything I could do? :-)

PR #4519, opened 2023-07-27
Prototype autocompletion of tags and keyswords in YAML frontmatter

### LINKED ISSUE #4070: autocomplete for keywords in YAML frontmatter
### Checklist

### Feature Proposal

Autocomplete for keywords in YAML frontmatter would be great. It already works for tags and 

### Example Workflow

Write 
```
---
title: "Welcome to Zettlr!"
keywords:
  - general
...
#genius 
```

in `file1.md`.
Start writing 
```
---
title: "Welcome to Zettlr! 2"
keywords:
  - gen
 ...
```
in `file2.md` and getting autocomplete suggestion for `gen` to either `general` or `genius`. 
(As far as I understand, tags and keywords are treated the same way. If that is not the case, the autocomplete should only suggest `general`). 

### Caveats

I cannot think of any. 

No

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

### Additional Information

(I only tested this on linux with version 2.3.0 UUID: f77f0308-2ef0-4327-875a-0679c5d8c261)

UFOMelkor:
## Description

This PR provides autocompletion for tags and keywords in YAML frontmatter.

## Changes

It enhances the current autocompletion for tags so that it also applies to YAML frontmatter.
Currently, this is a minimal prototype. It does not handle tags and keywords that are enclosed in quotes, and does also not handle single tags and keywords. Hence, this might be split into a separate autocompletion.

## Additional information

Resolves #4070 

Tested on: Ubuntu 22.04

### DIFF
--- a/source/common/modules/markdown-editor/autocomplete/index.ts
+++ b/source/common/modules/markdown-editor/autocomplete/index.ts
@@ -57,7 +57,6 @@ export interface AutocompletePlugin {
 }
 
 const forbiddenTokens = [
-  //'YAMLFrontmatter',
   'YAMLFrontmatterStart',
   'YAMLFrontmatterEnd',
   'MathEquation'


nathanlesage:
I had imagined this much more hacky than what you did there 👀

Tbh I wanted to try to use Lezer's NodeProps to introduce autocomplete to YAML because it is (supposedly) more elegant and performant than manual checking, BUT for that we would need a CMv6 ready YAML parser, but nobody has done so so far :/ 

What is your opinion on this? Implement it like so so that we have autocomplete (we can duplicate the Tag autocomplete and adapt it only to YAML frontmatter, while making it use the same database, if you feel this will help) or wait until we can do it "the right way"?

nathanlesage:
Yeah, true. And if we compartmentalise it into its own autocomplete the maintenance costs should always be manageable. Good point.

nathanlesage:
@UFOMelkor Since I saw you've been updating the branches, have you noticed that Zettlr now uses the first-party YAML parser for frontmatters…?

nathanlesage:
Oh god, I think someone opened a related issue recently... let's hope it gets fixed soon.

nathanlesage:
The changes you are referring to have been made when I switched my makeshift frontmatter parser to the Leter one. This means the node names will have changed, meaning that now the same nodes that you are already familiar with from your other PR will be the correct ones.

nathanlesage:
Ah. I think I just generally disallowed autocomplete in YAML because, when I started to migrate to CodeMirror 6, there was no proper YAML mode, meaning that it was very difficult to autocomplete anything there at all. We may want to remove the node from the disallowed ones altogether. I don't currently see how any of the autocompletes may interfere with YAML, but I'm not entirely sure. Would you mind testing if any of the existing autocompletes would look weird? If so, just move that check to the specific autocompleter. This way, we can disable only certain autocompletes in the YAML frontmatter as needed.

nathanlesage:
Then so be it!

PR #4560, opened 2023-08-18
Lowering alpha to fix not-highlighting selection on code blocks

### LINKED ISSUE #4559: CSS class `cm-selectionBackground` is not visible when selecting code inside a `plain`, `text` or empty code block
### Description

The base theme `berlin`, from which the other take these CSS rules, does not highlight characters when selecting text inside an inline code word, like `this` or when using a multiline code block without language, like

```
this
```
which is the same as using `plain`, `text`, and probably others that apply the classes `cm-monospace cm-monospace` to the code lines.

### Reproducing

On any document, on dark or light modes, create an inline word or sentence using the backtick character like 
<pre>
`this` 
</pre>
or with triple backtick without language or using `plain` or `text` for a multiline code block, like
<pre>
```plain
this plain block
```
</pre>
or
<pre>
```text
this text block
```
</pre>
or
<pre>
```
this non-language code block
```
</pre>

and after creating any of these elements, select characters or words, and they won't be highlighted.

### Zettlr Version

Right off the develop branch

### Specify version

_No response_

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [X] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [X] ARM 64bit

### Operating System Version

Linux (Ubuntu 23.04) and macOS (Ventura 13)

### Additional Information

Screenshots:

not highlighting:
not highlighting

classes applying
claases applying

I will post a PR with CSS changes to the theme `berlin`. I understand I can apply my own CSS classes to Zettlr. However, the fix I was able to create won't work ok both dark and light themes simultaneously. This is my snippet for the light theme (which doesn't work on the dark theme)

```css
:root {
  --green-selection:       rgba(180, 240, 170, 0.6);
  --green-selection-dark:  rgba( 90, 170,  80, 0.6);
  --blue-selection:        rgba(200, 220, 240, 0.6);
  --blue-selection-dark:   rgba( 29,  55, 134, 0.6);
}

body .main-editor-wrapper .cm-editor .code-block-line-background,
body .main-editor-wrapper .cm-editor .cm-monospace {
  background-color: rgba(238, 238, 238, 0.5);
}
```  

Thank you.

guzmandrade:
Closes #4559 

- [X] I documented all behaviour within the code as far as I could.
- [X] This PR targets the develop branch and *not* the master branch.
- [X] I have tested my changes extensively and paid extra attention to potential cross-platform issues (e./g. Cmd/Ctrl/Super key bindings) 
- [X] I ran the `yarn lint` and `yarn test` commands successfully
- [X] I matched my code-style to the repository as far as possible
- [X] I agree that my code will be published under the GNUP GPL v3 license
- [X] In case I created new files, I added a header to them (see the existing files for examples)
- [X] Before opening this PR, I ran `git pull` a last time to prevent any merge issues

## Description

Lowering alpha to fix not-highlighting selection on code blocks and inline code described in #4559

## Changes

I lowered the alpha of some elements that prevent selected code in inline and multiline blocks.

## Additional information

Tested on: Linux (Ubuntu 23.04), and macOS Ventura 13

### DIFF
--- a/source/common/less/theme-berlin/colours.less
+++ b/source/common/less/theme-berlin/colours.less
@@ -2,11 +2,11 @@
 
 :root {
   // Selection colours
-  --green-selection:       rgba(180, 240, 170, 0.8);
-  --green-selection-dark:  rgba( 90, 170,  80, 0.8);
+  --green-selection:       rgba(180, 240, 170, 0.6);
+  --green-selection-dark:  rgba( 90, 170,  80, 0.6);
 
-  --blue-selection:        rgba(200, 220, 240, 0.8);
-  --blue-selection-dark:   rgba( 29,  55, 134, 0.8);
+  --blue-selection:        rgba(200, 220, 240, 0.6);
+  --blue-selection-dark:   rgba( 29,  55, 134, 0.6);
 
   --purple-selection:      rgba(216, 171, 241, 1);
   --purple-selection-dark: rgba(100,  30, 120, 1);
--- a/source/common/less/theme-berlin/colours.less
+++ b/source/common/less/theme-berlin/colours.less
@@ -2,6 +2,7 @@
 
 :root {
   // Selection colours
+  // Lowering alpha to 60% so it's easier to see
   --green-selection:       rgba(180, 240, 170, 0.6);
   --green-selection-dark:  rgba( 90, 170,  80, 0.6);
 
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -119,7 +119,7 @@ body.dark {
     // Style the code block line wrapper classes
     .code-block-line-background,
     .cm-monospace {
-      background-color: #333;
+      background-color: rgb(51 51 51 / 50%);
     }
 
     // NOTE: Our codeblock layer creates THREE blocks per codeblock:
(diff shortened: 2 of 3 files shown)


nathanlesage:
This is not the way to do this. The reason is that CodeMirror 6 features so-called *layers*.

Instead of playing around with the opacity (which is a bad idea anyways), CM6 offers to use several _layers_ that can stack various backgrounds on top of each other.

You will notice that the same happens already with code blocks, if you select entire sections of a codeblock, you will see that the current ordering is as follows:

1. Code character backgrounds
2. Selection layer
3. Code block background

In other words, what we need to do is to move the logic that applies the background color to inline-code away from the CSS and into the background layer logic (those can live on the same layer).

The correct file to start and amend to account for these changes is here: https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/plugins/codeblock-background.ts

(You should then also rename the file from `codeblock-background` to just `code-background`, and also remove the CSS declarations for the inline code blocks).

Do you think you can do this? If so, that would be great! Please open a new PR then :)

guzmandrade:
@nathanlesage Thank you for your prompt and detailed reply. Yes, I think I can try that and open a new PR.

PR #4564, opened 2023-08-19
fix: #4559 Background Layering Issues

### LINKED ISSUE #4559: CSS class `cm-selectionBackground` is not visible when selecting code inside a `plain`, `text` or empty code block

guzmandrade:
Closes #4559 

- [X] I documented all behaviour within the code as far as I could.
- [X] This PR targets the develop branch and *not* the master branch.
- [X] I have tested my changes extensively and paid extra attention to potential cross-platform issues (e./g. Cmd/Ctrl/Super key bindings) 
- [X] I ran the `yarn lint` and `yarn test` commands successfully
- [X] I matched my code-style to the repository as far as possible
- [X] I agree that my code will be published under the GNUP GPL v3 license
- [X] In case I created new files, I added a header to them (see the existing files for examples)
- [X] Before opening this PR, I ran `git pull` a last time to prevent any merge issues

## Description

Move the logic that applies the background color to inline-code away from the CSS and into the background layer logic.
 
## Changes

Following feedback from @nathanlesage in #4560 I opened a new PR with the following changes:

- Renamed `codeblock-background.ts` to  `code-background.ts`  and added a condition to consider these rules considering `InlineCode` in addition to `CodeText`
- Renamed CSS rules to match these changes. That is, changing the CSS class to `code-line-background` and removing the background color rule applying for `cm-monospace` for consistency with the inline and code styles.

## Additional information

Tested on: Linux (Ubuntu 23.04), and macOS Ventura 13

### DIFF
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -116,21 +116,20 @@ body.dark {
       color: var(--c-primary-shade);
     }
 
-    // Style the code block line wrapper classes
-    .code-block-line-background,
-    .cm-monospace {
+    // Style the code line wrapper classes
+    .code-line-background {
       background-color: #333;
     }
 
     // NOTE: Our codeblock layer creates THREE blocks per codeblock:
     // First line, rest of the block, and an empty block at the end.
     // Therefore we have to style elements 1, 4, 7, etc. and 2, 5, 8, etc.
-    .code-block-line-background:nth-child(3n+1) {
+    .code-line-background:nth-child(3n+1) {
         border-top-left-radius: 4px;
         border-top-right-radius: 4px;
     }
 
-    .code-block-line-background:nth-child(3n+2) {
+    .code-line-background:nth-child(3n+2) {
         border-bottom-left-radius: 4px;
         border-bottom-right-radius: 4px;
     }
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -129,7 +129,7 @@ body.dark {
         border-top-right-radius: 4px;
     }
 
-    .code-line-background:nth-child(3n+2) {
+    .code-line-background:nth-child(3n+2), .inline-code-background {
         border-bottom-left-radius: 4px;
         border-bottom-right-radius: 4px;
     }
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -117,7 +117,7 @@ body.dark {
     }
 
     // Style the code line wrapper classes
-    .code-line-background {
+    .code-line-background, .inline-code-background {
       background-color: #333;
     }
 
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -124,7 +124,7 @@ body.dark {
     // NOTE: Our codeblock layer creates THREE blocks per codeblock:
     // First line, rest of the block, and an empty block at the end.
     // Therefore we have to style elements 1, 4, 7, etc. and 2, 5, 8, etc.
-    .code-line-background:nth-child(3n+1) {
+    .code-line-background:nth-child(3n+1), .inline-code-background {
         border-top-left-radius: 4px;
         border-top-right-radius: 4px;
     }
--- a/source/common/less/theme-berlin/dark.less
+++ b/source/common/less/theme-berlin/dark.less
@@ -117,19 +117,19 @@ body.dark {
     }
 
     // Style the code line wrapper classes
-    .code-line-background, .inline-code-background {
+    .code-block-line-background, .inline-code-background {
       background-color: #333;
     }
 
     // NOTE: Our codeblock layer creates THREE blocks per codeblock:
     // First line, rest of the block, and an empty block at the end.
     // Therefore we have to style elements 1, 4, 7, etc. and 2, 5, 8, etc.
-    .code-line-background:nth-child(3n+1), .inline-code-background {
+    .code-block-line-background:nth-child(3n+1) {
         border-top-left-radius: 4px;
         border-top-right-radius: 4px;
     }
 
-    .code-line-background:nth-child(3n+2), .inline-code-background {
+    .code-block-line-background:nth-child(3n+2) {
         border-bottom-left-radius: 4px;
         border-bottom-right-radius: 4px;
     }
--- a/source/common/less/theme-berlin/light.less
+++ b/source/common/less/theme-berlin/light.less
@@ -97,7 +97,7 @@ body {
       background-color: #eee;
     }
 
-    .cm-monospace {
+    :not(.code.code-block-line) > .cm-monospace {
       border-radius: 2px;
       padding: 0 2px;
     }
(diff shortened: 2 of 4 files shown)


nathanlesage:
Thank you for the changes! On first look, it seems as if it's working as expected -- I'll be testing it later and confirm before then merging!

nathanlesage (changes requested):
I just tested it out. However, I noted that there are three issues:

1. For Inline code, the backticks are included which they should not be
2. For inline code, the `node.to + 1` results in the next character to be drawn in a background color as well, which is undesirable. The `+1` is necessary to ensure that the last line of a code *block* is being included in the entire draw.
3. The rounded borders (difficult to see, I agree) are bad in documents that include both inline code and code blocks, based on the explanations given in the CSS that explains the weird CSS rule above.

Therefore, we need to change a few things; I've tried to include suggestions where I think they're necessary:

1. Add a new class for the inline code that shares the same background color, but has rounded corners everywhere. Then the counting of the CSS classes (the `3n+1` thingy above) works, because the classes for the code blocks are separate from the inline code and thus are counted correctly so that every first, fourth, seventh, etc. element has only rounded borders at the top, every second, fifth, eigth, etc. element has only rounded borders at the bottom, and the rest no rounded borders at all
2. Add a few branches to draw those `RectangleMarker`s with both the appropriate ranges and class names for both types of element

**Here's what I would like you to do now**:

1. Look over my suggestions and think whether you understand them and agree with them. Otherwise, please comment and ask before proceeding
2. If this looks good, implement those changes, but please double-check and test those changes to ensure that I did not make an error during this code review
3. Ping me once you're ready!

guzmandrade:
@nathanlesage Thank you for your suggestions. I think I understand them and will test them out and let you know if I have further questions.

guzmandrade:
@nathanlesage  Your changes worked fine in general, as far as I can see. However, when testing this with your feedback in mind, I noticed that the CSS class `cm-monospace` is added to code blocks, which applies the following rule to lines inside a code block.

```
.cm-monospace {
  border-radius: 2px;
  padding: 0 2px;
}
```

which it is visible on the first line of each code block of this kind (plain, text, no language). 

Screenshot from 2023-08-19 19-55-30

To prevent this, and given that the issue seems to be on the lines with the structure `.code.code-block-line > .cm-monospace` (direct child), I added:

```
:not(.code.code-block-line) > .cm-monospace {
```

Please let me know if this works for you.

nathanlesage:
I saw that you modified the `cm-monospace`-class. What happens if we just entirely get rid of it? It could be a remnant from the old CodeMirror 5 code, in that case we shouldn't need it anymore.

I also saw that apparently the selector doesn't work only on the classes themselves, but rather the elements (so it counts all children, regardless of their class name, when using `:nth-child()`). See screenshot:

image


In other words: We apparently need yet another layer only for the inline code …

How to do that: Duplicate the entire `layer`-call, think of a proper class name for that layer (given that the code block background is called `cm-codeBackgroundLayer`, maybe `cm-inlineCodeBackgroundLayer`), and use the one layer only for the code blocks, and the other one only for the inline code.

Then, instead of exporting directly the codeBlocks thingy, do the following export and import it in the editor extension sets:

```typescript
export const backgroundLayers: Extension[] = [
    codeBackgroundLayer,
    inlineCodeBackgroundLayer
]
```

(Note that I changed back the original variable name, because it will then only apply to the code block backgrounds as before).

Sorry for the back and forth, I thought we could handle this without creating yet another layer, but the upshot is that I'm confident that this will be the very last change before it actually works as expected. If you can manage to implement this during this week, I'll merge this in just before 3.0.0 so that it's part of that release.

guzmandrade:
@nathanlesage Thank you. Ok, I will address these issues and update the PR.

guzmandrade:
@nathanlesage I added my understanding of your suggestions, with the following observations:

In `code-background.ts` I added a new layer `inlineCodeBackground` but by adding it as 

```javascript
export const inlineCodeBackground = layer({
```
the import line in `editor-extension-sets.ts` looks like 

```javascript
import { codeblockBackground, inlineCodeBackground } from './plugins/code-background'
``` 
So I think we don't need to export the layers as an array, as you suggested. Please let me know if this makes sense.

Regarding the `n-child` CSS rules. After adding the new layer, I'm not exactly sure how the `n-child` rules apply, since, as you point out, those count editor lines in general, nor just code block lines. For example, when watching at the final render of the code blocks in the code inspector, I see this structure:

```html
<div class="cm-line">...
<div class="code code-block-line cm-line">...
<div class="cm-line">...
<div class="code code-block-line cm-line">...
<div class="code code-block-line cm-line">...
```
and I think that would mean that the `background` classes, if are being added to the same elements as the `code-block-line` classes, might be counting everything and probably not applying to the intended elements.

nathanlesage:
Just a small update: I took a look into CodeMirror's code, and it seems to be easy to achieve the same code block background with just a single element if we, instead of using `forRange`, create the `RectangleMarker` ourselves:

image

I don't know if we really want to touch the code again, but it seems to be a possibility to reduce the complexity of the DOM. But let's maybe keep that for another day unless you want to compress the entire code down again.

guzmandrade:
@nathanlesage I updated this PR, and it's now exporting the background layers as an `Extension` array. Thank you for your patience and guidance. Do you think we should open a new issue (maybe an `enhancement`) with the additional change?

nathanlesage:
PR works, and it looks all decent – merging.

Regarding the other issue: I don't think we need a separate issue. If you want to give it a shot, please feel free to simply open a PR, otherwise the code works so we can leave it at that for now – I'm leaving this up to you!

PR #4601, opened 2023-09-02
fix: Handle Zkn link in Markdown to HTML conversion

### LINKED ISSUE #4600: Zkn link in heading becomes invisible in ToC
### Description

When a heading contains a Zkn link, the link is invisible in the ToC. See the following example:

grafik

I did some investigation and the cause seems to be that the [AST node to HTML conversion](https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-utils/markdown-to-html.ts#L279) doesn't consider the node type "ZettelkastenLink" so far. Hence an empty string is returned.

### Reproducing

Refer to the example image above.

### Zettlr Version

Stable (most recent version)

### Specify version

3.0

### Installation Method

None

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Manjaro

### Additional Information

_No response_

kyaso:
## Description

Closes #4600

This change makes Zkn links visible in the ToC again.

However, I'm not sure whether this might have any impact on export?

## Additional information

Tested on: Manjaro Linux

### DIFF
--- a/source/common/modules/markdown-utils/markdown-to-html.ts
+++ b/source/common/modules/markdown-utils/markdown-to-html.ts
@@ -364,6 +364,8 @@ function nodeToHTML (node: ASTNode|ASTNode[], getCitation: (citations: CiteItem[
     const close = tagInfo.selfClosing ? '' : `</${tagInfo.tagName}>`
     const body = tagInfo.selfClosing ? '' : nodeToHTML(node.children, getCitation)
     return `${open}${body}${close}`
+  } else if (node.type === 'ZettelkastenLink') {
+    return `[[${node.value}]]`
   }
 
   return ''


nathanlesage:
Oh God, this PR completely slipped my attention. Looks good, thanks!

nathanlesage:
Ah, and regarding your question: We can keep it like this for now; it ignores the export settings, yes, but those are anyhow mainly used for steering Pandoc's behavior; this md-to-HTML function is merely used as a crutch for quick-and-dirty exports right now, there are many other things that we would need to take into account. For example, I found out that due to me using that function to render the Changelogs, all the issue links that I add to the changelog get removed as well because tags don't count as text 🙄 

We'll have to improve that specific change anyhow in the future, but I think having it output is better than leaving it out completely, so I think it's absolutely good that you made this change.

PR #4652, opened 2023-09-13
Fix link preview for short notes

### LINKED ISSUE #4650: Link preview does not work for short notes
### Description

The link preview does not work when the target note is short (less than 10 lines and 300 characters, based on my reading of the code). I believe this is due to a bug in the FilePathFindMetaData (`'file-find-and-return-meta-data'`) command.

### Reproducing

1. Create a short note with less than 10 lines and less than 300 characters.

    Example:

    ```
    # Short Example

    This is a short example.
    ```

3. Create another note with a link to the first one.

    Example:

    ```
    Link to short note:[[short-note]]
    ```

4. Hover over the link. The preview does not display, and there is an error in the debug log reading: "Error received while running command: "Cannot read properties of undefined (reading 'length')"

### Zettlr Version

Stable (most recent version)

### Specify version

3.0.0

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Debian Testing

### Additional Information

I believe this is a bug in `source/app/service-providers/commands/file-find-and-return-meta-data.ts`, specifically in this code:

```ts
while (preview.length <= MAX_FILE_PREVIEW_LENGTH && i < 10) {
  const remainingChars = MAX_FILE_PREVIEW_LENGTH - preview.length
  if (lines[i].length <= remainingChars) {
    preview += lines[i] + '\n'
  } else {
    preview += lines[i].slice(0, remainingChars) + '…'
  }
  i++
}
```

If `preview.length` does not exceed the maximum length, then the code will try to read 10 lines, regardless of how many lines actually exist in the note. Therefore, if there are less than 10 lines, then `length[i]` will eventually become undefined, and reading the `length` property will lead to the error that appears in the log.

ObjectBoxPC:
## Description
Fixes #4650: Link previews do not work for short notes.

## Changes
* Limit the number of lines retrieved based on the actual number of lines in the note.
* Create a named constant for the maximum number of lines.

## Additional information
I was not sure if this belongs in the changelog; please let me know if I should add an entry.

Tested on: Linux (Debian Testing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   extensions
 - The generic text input context menu shows proper labels (#4655)
 - Improved the French translation
+- Fixed link previews for short notes
 
 ## Under the Hood
 
--- a/source/app/service-providers/commands/file-find-and-return-meta-data.ts
+++ b/source/app/service-providers/commands/file-find-and-return-meta-data.ts
@@ -17,6 +17,7 @@ import ZettlrCommand from './zettlr-command'
 import type { MDFileDescriptor } from '@dts/common/fsal'
 
 const MAX_FILE_PREVIEW_LENGTH = 300
+const MAX_FILE_PREVIEW_LINES = 10
 
 export default class FilePathFindMetaData extends ZettlrCommand {
   constructor (app: any) {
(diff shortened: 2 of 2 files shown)


nathanlesage:
> I was not sure if this belongs in the changelog; please let me know if I should add an entry.

Yes, that is definitely worthy of a Changelog entry. Thank you for opening the PR; I must've overlooked this when originally merging the PR last year. Thanks for fixing it!

ObjectBoxPC:
> Yes, that is definitely worthy of a Changelog entry.

Okay, I've added the entry to the changelog. Thanks for reviewing!

nathanlesage:
Wonderful, thanks!

PR #4665, opened 2023-09-17
fix: Make sure active tab is synced to config

kyaso:
## Description
I noticed that sometimes the last active tab was not opened on re-launch of Zettlr. After digging a bit I found that we may have a missing `syncToConfig()` for active tabs. This PR addresses that.

## Additional information

Tested on: Linux Manjaro


nathanlesage:
Oh … I was wondering what was causing that but I didn't think of investigating that 🌚 

Thank you! Looks good :)

PR #4669, opened 2023-09-18
fix: GraphView Label-Rendering and click open nodes in the last leaf

### LINKED ISSUE #4701: Clicking on graph nodes no longer opens the document
### Description

I have updated to version 3.0.1 and clicking on the nodes in graph view now no longer open the files when clicked. I have also observed issues with the linking as reported in https://github.com/Zettlr/Zettlr/issues/4700. The connections in graph view were not made until I changed the format to [[filename]] rather than (my text)([[filename]]) and restarted Zettlr.

### Reproducing

Updated to version 3.0.1 from version 2.3.0 and observed issue

### Zettlr Version

Stable (most recent version)

### Specify version

3.0.1

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Ubuntu 22.04

### Additional Information

_No response_

pseandersson:
## Description
Labels in GraphView are now rendering without strokes, to make them readable. Tracking focus of the last focused leaf and window to be able to open the file when clicking on the nodes in the GraphView.

Fixes #4701

## Changes
- **MainEditor** is now firing off a 'focus-leaf' event on the ipc-channel 'documents-provider', to notify the DocumentProvider of the last focused leaf.
- **DocumentProvider** keep track of the last focused editor. If no windowId nor leafId is provided in the openFile-call, it will try to fall back to the last focused editor otherwise resolve the last used main-window.
- In **GraphView** the labels node are set with a fill color and a stroke-width of 0, and enable the on-click code for vertices.

## Additional information
This is just a proposal for solving the issue.

4. I ran the `yarn lint` and `yarn test` commands successfully
  `yarn lint` returned a lint-issue not caused by me.
  `yarn test` failed due to mismatch of linux and windows paths.

Tested on: Windows 11


nathanlesage:
Thank you for these improvements. At first glance, they do look good. I'll have to take another look and test it out for myself, maybe I'll manage to slot this in during the week.

pseandersson:
@nathanlesage any chance that you'll have time to review it this week?

pseandersson:
This is the pull request I would like to be reviewed @nathanlesage @kyaso

nathanlesage (commented):
In general, the PR appears fine to me. However, I have one question: Do we really need the additional leaf map? I am not sure I understand what the function is, and whether we can't just simply use the existing windows array?

Furthermore: I personally believe that focus tracking is not strictly necessary. I implemented a lot of code to try to remember certain window states, but I feel that possibly the last focus-getter on the BrowserWindow object may be enough – we possibly don't have to move through all windows to find the file open, and rather open it in the last focused main window (remember that files can be opened multiple times, without too much overhead).

Thoughts?

pseandersson:
I added the leaf map so that I could have control over which leaf the selected node/document would be opened in, so that I could organize the leafs for different purposes. 

My other thought, was that if I had it already opened the file in the window, then I wanted it to reappear in its leaf. However, it felt a bit overkill to track whether the file was open in the window or not, so I can remove that part.

So now the commit has the following logic: 
'Which is my last main window?' -> 'Which was my last leaf for this file?' -> 'If not found, which was the last focused leaf?'

But perhaps should be simplified to
'Which was my last main window?' -> 'Which was my last focused leaf ?'

nathanlesage:
I agree. While it would definitely be "cleaner" to actually check if we *really* have to open the file again, I think that opening multiple copies of the file is actually quite a rare possibility: Users will open the graph view coming from some main window, and I am certain they are absolutely happy if files the click on simply open in that one, even if the file may already be open in another window. So simply taking the last focused main window and checking in its leafs is the simplest solution, and then only open the file in that window in the last focused leaf.

One thing I'd like you to try beforehand, however, is if you simply allow the imputation of windowId and leafId in case they haven't been provided without anything additional code. That would make this PR cleaner and also allow in general the arbitrary opening of files without having to specify a window/leaf. I think that makes sense from a code-perspective and to be honest I don't know why I didn't implement it right from the start in such a way.

pseandersson:
Yes I clarified the signature of the openFile, that windowId and leafId is not required to open a file, it will then fall back on the last focused leaf in the last focused MainWindow.

nathanlesage (changes requested):
I'm still a bit unsure about the amount of new code added. Here's what I think:

Why don't you keep two simple variables in the document manager – last windowId, and last leafId, which can be undefined. Then, whenever the user activates an editor, you update those two variables. This would reduce the complexity in the documents provider by a large margin.

Then, when it comes to opening a file without a provided window/leaf ID, you would query if those variables are not undefined and if the keys actually exist. If the leaf is gone by the time the open request comes, but the window exists, simply take the first leaf inside that window and open the file there (if the file is not yet open in said window). If neither the window ID nor the leaf ID are set or exist in the actually opened windows, simply take the last window in the array, and open it in the first leaf of that one.

I am extremely sure that this will cover 99% of the cases, but keep the amount of additional code very limited. Then, if we figure out that the ordering of windows and/or leafs is not satisfying with this very simple heuristic, we can still improve the complexity of that. But I really don't feel comfortable adding this much code – especially given the fact that the documents provider is already very large and I am thinking of slimming it down again.

Are you really convinced this very complex way is necessary, and if so, what prevents us from removing much of those code additions?

pseandersson:
I have been looking more into it. I can move the entire focus logic to a new class referenced from the WindowsProvider, so the document manager will not be touched at all, that makes more sense.

Then there need to be an additional call in DocumentTabs which call the focus-leaf when a filename is clicked, which is invoked now in the MainEditor. This calls could be gathered in EditorPane and then using callbacks, what would you prefer in this case?

The the GraphView could then utilize a custom event e.g. "open-file" on another channel like 'window-provider' which the WindowsProvider is listen on and assign the last focused window and leaf to it before it sends a 'open-file' on the 'documents-provider' channel.

That will cover 100% of the cases, and DocumentManager will not increase in size. What do you think?

nathanlesage:
I am afraid I don't quite understand what you mean, and it doesn't appear that you referenced what I was saying, which at its core implied to cover most, but not 100% of the cases. The problem I see is that the solution you are offering right now is over-engineered and doesn't make use of much of the information that is already available. And I am having some stomach aches due to this, because I really see something unmaintainable coming up. Additional IPC calls will make this rather worse than better.

pseandersson:
I thought the big issue for you was that too much content was added to DocumentManager, so I thought perhaps I should try to make it clean, and was trying to look for alternative solutions the the problem, therefore the branch out discussion. Regarding the complexity you mention, I agree, perhaps it was too complex/over-engineered, sometimes complexity may arise to lack of separation of concern. I was eager to see what my new idea would look like and you will find it here: https://github.com/pseandersson/Zettlr/tree/ExtensiveFocusTracking

I was curious to see how your proposal worked out, and there are some edge cases which may disturb the use, but perhaps it is good enough.

pseandersson:
@nathanlesage, I think I have updated the changes as you suggested, Please let me know if there are any other doubts you want me to resolve before merge.

nathanlesage:
@pseandersson Hej, sorry for being so tardy, I have been on vacation for a few days until now. I really like the changes, it makes it much simpler and thus easier to implement, thank you very much!

I only have one last request, and then I'll do some testing and then we're ready to merge:

Right now you are using a class to remember the last window/leaf IDs, but I would like to ask you to simply use a simple assigning of the object `{ leafID: undefined, windowID: undefined }`; I think with a simple type-hint in the class this should be sufficient and avoids an empty class where a simple state is sufficient. Other than that it looks good!

pseandersson:
@nathanlesage No worries, I will upload a new patchset tomorrow. I need to track the focus of the last main window in the WindowManager, otherwise it get weird if you have open up a blank window to be the target of opening files from the Graph. Also I will by default open in the same tab, and then still shift can open it in a new tab.

pseandersson:
I default the opening to the same tab, since if you navigate in a big graph, it could open a lot of tabs. Perhaps there should be an option for that behavior. Let me know.

Another thing, I tried to search for documentation where short-cuts are listed so I can add the shift-click behavior in graph view, will open it in a new tab?

nathanlesage:
Just a quick note: I am a bit embarrassed that I still did not get to your PR, I will tend to it next week, please accept my apologies for the very lengthy process.

nathanlesage:
So, I finally was able to test out the PR: Remembering the last focused leaf works perfectly well, so I am absolutely happy with that.

There is just one issue: When you open a file in said last-focused leaf, you are not respecting the setting that it should open in a *new* tab. Right now, regardless of your setting, any file you click in the graph view will replace whatever the current tab in the leaf is, which is not intended. Would you mind having a look at this again? That would be great. As I said, other than that, I like the PR so far!

pseandersson:
Sure, I thought the implemented variant would be preferred when you've got a big graph and would like to navigate between documents.

Do you want me to keep the shift-modifier, such that if you hold shift while opening a document from the graph, it will enforce opening the document in new tab?

nathanlesage:
> Do you want me to keep the shift-modifier, such that if you hold shift while opening a document from the graph, it will enforce opening the document in new tab?

The modifier for "force new tab" is `Alt`, not shift. So please implement the following:

* On a simple click, do what the config option says: If the user has selected "Avoid new tabs if possible", overwrite the active file if possible (i.e. not if it's modified). If the user has disabled that checkbox, ALWAYS open a new tab (if the file is not yet open, of course).
* If the user holds down `Alt`, always open in a new tab, regardless of setting.

pseandersson:
@nathanlesage changes have now been made accordingly to your feedback

nathanlesage (approved):
So, again a somewhat slow response, but I just tested out the changes, and the behavior appears to be as intended, so wonderful job on that!

I just found a single instance where you set the last leaf and window in the documents provider directly instead of using the function – there is no bug, I just added a minor suggestion.

But other than that, this PR is ready to go, and can be included in the upcoming v3.0.3!

Thank you for the great work and for implementing the changes, and bearing with me for my slow responses!

nathanlesage:
I just quickly committed the tiny refactor, because I think it's ridiculous to wait for you to click a button :D Thanks again for your work!

pseandersson:
Thanks! Great, looking forward to see it in 3.0.3! :)

PR #4707, opened 2023-09-30
fix: Directory deletion in file manager

### LINKED ISSUE #4656: Directory deletion in file manager does not delete it from file explorer of Windows
### Description

Im currently using Windows 11 latest version.
Bug: When i create a new directory in Zettlr and delete it, it is only deleted from the view of file manager of Zettlr.

Expected: Deleting a directory from Zettlr should also delete it from the File Explorer of Windows.
image
As shown in the above pic, the directory "Hello" is deleted from the file manager of Zettlr but it is not deleted from the system.

### Reproducing

I am a beginner and looking to get into open source contributions so I do not know how to fix this but am looking forward to fix this. I have no knowledge of JS and TS so I would love to team up with someone to fix this.

### Zettlr Version

Stable (most recent version)

### Specify version

_No response_

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 11

### Additional Information

_No response_

pseandersson:
## Description
Fixes #4656 "Directory deletion in file manager does not delete it from file explorer of Windows"

## Changes
- Replaced `fs.unlink` with `fs.rm` in `safeDelete` to support recursive removal of directories
- `fsal-directory::removeChild` now calls `pathExists` instead of `isFile` to make sure directories also are removed in removeChild
- `pathExists` wraps `fs.existSync` as the pattern with `isFile` etc.

## Additional information
Tried to write tests for it, but couldn't mock electron.shell.trashItem

Tested on: Windows 11


pseandersson:
Tested on MacOS Ventura

source/app/service-providers/fsal/util/safe-delete.ts
```diff
@@ -7,7 +7,7 @@ export async function safeDelete (absPath: string, deleteOnFail: boolean): Promi
   } catch (err: any) {
     if (deleteOnFail) {
       // If this function throws, there's really something off and we shouldn't recover.
-      await fs.unlink(absPath)
+      await fs.rm(absPath, { recursive: true, force: true })
```
nathanlesage, line 10: It might make sense to add a log entry here, because `force: true` will silently swallow errors

source/common/util/path-exists.ts
```diff
@@ -0,0 +1,24 @@
...
+import { type PathLike, existsSync } from 'fs'
+
+/**
+ * Check if a path exists.
+ *
+ * @param   {PathLike}  p  path to be tested
+ * @return  {boolean}      true if the path exists otherwise false
+ */
+export default function pathExists (p: PathLike): boolean {
+  return existsSync(p)
```
nathanlesage, line 23: Do not use any synchronous methods for file access. Please take inspiration from other files; usually we import `fs.promises` and `await` results.

nathanlesage (changes requested):
So, first of all sorry for the late review. Then, there is a bit of synchronous code in the PR which has to go – file operations should all be async. Otherwise, it looks good!

source/common/util/path-exists.ts
pseandersson: It says that fs.exists() is deprecated but not existsSync. I also checked the is-file and is-dir code and they are using lstatSync which is synchronous code. But I'm considering to upload a changeset with an async pathExists

```typescript
export default async function pathExists (p: PathLike): Promise<boolean> {
  let stat, stat2
  try {
    stat = await promises.lstat(p)
  } catch (err) {
    return false
  }
  return stat !== undefined
}
```

Is that what you've got in mind?

pseandersson: the is-file and is-dir is not just a search and replace. The will be awaits ending up in constructors of object so there are some thoughts needed there to get it right.

source/app/service-providers/fsal/util/safe-delete.ts
pseandersson: I added an error log message.

nathanlesage (approved):
Wonderful, thank you very much for the changes – now the PR is mergeable! Thank you for your support.

PR #4711, opened 2023-10-02
fix: Update the translate in zh-TW.po

seanmars:
Fixed the translate, make it more precise.

## Changes
- Chinese (Taiwan): `簡體中文` to `正體中文(台灣)`
- Chinese: `簡體中文` to `中文`

## Additional information


nathanlesage:
Thank you!

PR #4760, opened 2023-11-04
fix: Prevent onerous popups & cursor jumps

### LINKED ISSUE #4729: Cursor jumps to the start of the file
### Description

On the 3.0.2 version, I have the following issue: when editing my file, the text cursor jumps from time to time to the start of the file.

I couldn't measure the time, that varies. I couldn't identifiy a cause either. It just jumps. It's a little like when I reloaded the file on Bversion 2.XX (when I had other problems), the cursor was back to char 1 line 1. but it in this case, it doesn't seem to be a reload problem.

### Reproducing

I couldn't identify the causes of this behaviour.

### Zettlr Version

Stable (most recent version)

### Specify version

3.0.2

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

MS Windows 11.

### Additional Information

_No response_

### LINKED ISSUE #4732: Load remote changes pop-up on every save
### Description

Hello there! Problem's really simple actually: Since the new version 3.0.2 the below pop-up will show on EVERY manual save (autosave is off, "Always load remote changes to the current file" is unchecked). Is it supposed to pop up every time? Experienced it before, where it would show only about 1/10th of the time, with the same settings.
And the problem is: The little asterisk indicating changes next to the file name won't disappear after hitting the first pop-up option.

image

I'll be switching to 3.0.1 for now...

### Reproducing

Turn autosave off, uncheck "Always load remote changes to the current file". Change a file. Press Ctrl+S or save through the toolbar.
Select the first option of the pop-up window.

### Zettlr Version

Stable (most recent version)

### Specify version

3.0.2

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 10

### Additional Information

_No response_

nathanlesage:
## Description

This PR adds a (hopefully working as intended) minimal-invasive fix to issues #4729 and #4732. It adds an explicit check that *really* makes sure that a file is changed on disk before pestering the user. In other words: If the disk contents do not differ from whatever content has last been written to disk, Zettlr ignores that change entirely, i.e. no popups (if auto-load changes is set to off) and no cursor jumping (if auto-load changes is set to on).

## Changes

* Store a copy of a file's last saved disk contents in the Document descriptor within the documents provider.
* Compare that on every change event that is neither ignored nor has an appropriate modtime to the actual disk contents at that point to see if a document has actually changed.

## Additional information

Since I do not experience this bug, it would be amazing if some folks could test it out who were experencing these issues.

Tested on: not tested b/c I do not have a system that experiences this bug.

* Fixes #4729
* Fixes #4732

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,7 +126,7 @@ jobs:
     name: Windows (x86_64)
     env:
       npm_config_arch: x64
-    needs: [preflight, verify_version, i18n]
+    needs: [preflight, verify_version]
     runs-on: windows-latest
     steps:
       - name: Checkout branch ${{ github.ref_name }}
@@ -163,7 +163,7 @@ jobs:
   #   env:
   #     npm_config_arch: arm64
   #     TARGET_ARCH: arm64 # Seen at https://github.com/hydraulic-software/github-desktop/blob/conveyorize/.github/workflows/ci.yml
-  #   needs: [preflight, verify_version, i18n]
+  #   needs: [preflight, verify_version]
   #   runs-on: windows-latest
   #   steps:
   #     - name: Checkout branch ${{ github.ref_name }}
@@ -237,7 +237,7 @@ jobs:
     name: macOS (x86_64)
     env:
       npm_config_arch: x64
-    needs: [preflight, verify_version, i18n]
+    needs: [preflight, verify_version]
     runs-on: macos-latest
     steps:
       # Check out master for a regular release, or develop branch for a nightly
@@ -283,7 +283,7 @@ jobs:
     name: macOS (arm64)
     env:
       npm_config_arch: arm64
-    needs: [preflight, verify_version, i18n]
+    needs: [preflight, verify_version]
     runs-on: macos-latest
     steps:
       - name: Checkout branch ${{ github.ref_name }}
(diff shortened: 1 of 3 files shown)


kyaso:
This looks like a reasonable way to resolve the issue. I will test it out over the next few days and see if that fixes the issue.

kyaso:
Small update (with some hopefully good news): Yesterday, I have tested the version with the fix from this PR, and the pop-up did _not_ show up once! Today, I tested again the latest version for the whole day, and got the pop-up multiple times.

I'm going to test again the fixed version tomorrow.

nathanlesage:
Ah, amazing! I'll wait for your final check, but this does indeed sound very good :)

kyaso (approved):
Good news: Today, with the fixed version, again no pop-up. So, I guess, we are good to go! :)

nathanlesage:
Thank you very much for your support! :) Merging the PR now …!

PR #4848, opened 2023-12-17
Added Tamil Translation file -> ta-IN.po

TamilNeram:
## Additional information

### DIFF
--- a/static/lang/ta-IN.po
+++ b/static/lang/ta-IN.po
@@ -1,0 +1,2695 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR Hendrik Erz <info@zettlr.com>
+# This file is distributed under the same license as the Zettlr package.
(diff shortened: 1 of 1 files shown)


nathanlesage:
Impressive work, thank you!

I just double-checked, and found out that we possibly have to add the corresponding language code mappings accordingly to this file: https://github.com/Zettlr/Zettlr/blob/develop/source/common/util/map-lang-code.ts

Would you mind having a look if there is a correct country code already available? Other than that, it's great work and I really like that!

TamilNeram:
ta: trans('Tamil (India)') - This is the correct language code and country code. And the same is available in the list.

nathanlesage:
Ah, wonderful, thank you for double checking that! Then I'd say, it's time to merge :)

Again, thank you for the amount of work you've put into this PR! It is highly appreciated.

PR #4942, opened 2024-02-07
Update nodejs used in CI runs off of EOL version

alerque:
Electron got updated in v3.0.5, but the nodejs used in CI is still EOL and something we're trying to deprecate and get rid of. In building Arch Linux packages I was able to build it with Iron, so I don't think Hydrogen is a requirement any more.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,7 @@ defaults:
 
 # Global environment variables
 env:
-  NODE_VERSION: '18'
+  NODE_VERSION: '20'
 
 # Ensure only a single build workflow runs at any one time
 concurrency:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
   # havoc because we don't push unstable code to master, BUT it looks bad.
   verify_version:
     name: 'Preflight: Verify version tag'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Clone repository (master branch)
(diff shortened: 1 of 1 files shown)


nathanlesage:
What specifically is Iron and Hydrogen referring to? The Ubuntu distros have some weird animal names as codenames, correct?

Then, I see that you have upgraded the utilized node version to 20. However, Electron is (currently) still using Node 18, and in order to avoid any super wird and exotic bugs, I would like to keep the Node version in the CI pipelines as close to Electron's version as possible. Or do you have any concerns?

nathanlesage:
(And disregard the failing lint tests, since that is unrelated)

alerque:
NodeJS releases have code names too, mostly metals. Hydrogen is 18, Iron is 20. Non-LTS releases don't have code names I don't think.

https://nodejs.org/en/about/previous-releases

I see now looking it up that even Electron 28 is still specing Node v18 ... which is kind of shocking to me considering it is out of either "current" or "active" status. I guess however it is still listed as maintenance, not EOL like I thought.

We did get asked to try and deprecate `node-lts-hydrogen`. Since the bulid seems to work fine on `node-lts-iron` I think I'll leave it that way unless we get weird bug reports like you said. I can see why you might want to distribute Electron binary stuff with a matching NodeJS.

If you want to close this or put it in draft mode until the situation changes that's fine.

nathanlesage:
Quick and fortunate update: I just saw that Electron v29 has switched to Node 20 `lts/iron`, so we are good to merge this one and switch everything to Node v20!

Thanks again for making the changes and notifying me of it.

PR #4971, opened 2024-02-19
Added ability to specify target directory when exporting

jamescormack:
Hi,

Scratching an itch, I found that I wanted to export to latex to a specific folder containing my latex template assets. I didn't want to use the CWD which is what is set in the code as I wanted my markdown and output/template to be in a different folder. I made the attached changes and I'm submitting them back in case they are helpful.

Hope this is useful. I've started using Zettlr instead of Joplin and it's much better for academic writing - Thanks so much.

-James
  

## Description
With this change you can place the output file wherever you please independently of the current working directory. When exporting a single file the target directory can be specified in the YAML frontmatter as zettlr:pandoc_target_filename (similar to how the existing zettlr:pandoc_target_dir is implemented). When exporting a project the target directory can be specified in the .ztr_directory project json as targetDir.

## Changes
Changed Exporter.tx and Dir-Exporter.ts to incorporate target dir for exporter. Small change to FSAL to add target directory field.

## Additional information
TODO: Add a field to the project_settings:Files tab to specify the target directory. Currently I manually add 'targetDir' to the .ztr_directory json.

Tested on: MacOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ links as well.
   actual mouse position if the window's GUI was scaled absolutely (as per the
   preferences); now the context menu should always appear exactly where it
   should be
+- Added the ability to specify the target directory when exporting. This allows you to place the output file wherever you please independently of the current working directory. When exporting a single file the target directory can be specified in the YAML frontmatter as zettlr:pandoc_target_filename (similar to how the existing zettlr:pandoc_target_dir is implemented). When exporting a project the target directory can be specified in the .zsh_directory project json as targetDir.
 
 ## Under the Hood
 
--- a/source/app/service-providers/commands/dir-project-export.ts
+++ b/source/app/service-providers/commands/dir-project-export.ts
@@ -112,11 +112,17 @@ export default class DirProjectExport extends ZettlrCommand {
         template = config.templates.tex
       }
 
+      // Set target directory so that it is not always cwd. We check if the project .ztl_directory file has a 'targetDir' set and set targetDirectory to that value (or set targetDirectory = dir.cwd if it is not set)
+      let targetDir = dir.path
+      if (config.targetDir !== '') {
+        targetDir = config.targetDir
+      }
+
       try {
         const opt: ExporterOptions = {
           profile,
           sourceFiles: files,
-          targetDirectory: dir.path,
+          targetDirectory: targetDir,
           cwd: dir.path,
           defaultsOverride: {
             title: config.title,
--- a/source/app/service-providers/commands/export.ts
+++ b/source/app/service-providers/commands/export.ts
@@ -109,6 +109,16 @@ export default class Export extends ZettlrCommand {
         exporterOptions.cwd = fileDescriptor.frontmatter.zettlr.pandoc_working_dir
       }
 
+      // A user can override this default by providing in a frontmatter the
+      // key zettlr.pandoc_target_dir: /path/to/directory. This sets the destination for the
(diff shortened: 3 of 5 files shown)


nathanlesage:
First of all, thanks for your work!

Then, I have a question, because I'm a bit confused: The current version of the app already allows for a selection of the target directory in the exporter -- only the project-based export doesn't (yet). Is the latter the use-case which you wanted to address? Because the former is already addressed. I just want to make sure I fully comprehend the intention behind your PR.

jamescormack:
Hi Nathan, its probably me that's confused. You are right, my primary goal was to get the project based export to work as I'm going to be writing a large multi-chapter document. I did the per-file exporter too because I could not see how to do it otherwise. How do you specify the target dir on a per-file export?

jamescormack:
... I probably should be clearer as to what I want to achieve...

I have a folder structure like this:
--------
\.
\latex-template-folder
\images
\1_first_chapter.md
\2_second_chapter.md
\pandoc_template.tex
-----

I want to generate my latex file into the latex-template-folder as that contains all the style information so latex can then generate my document without having to move files around. My pandoc template stays on the top level as does my markdown and an images folder for pulling in assets.

Currently when doing an single file export in export.ts we are give a choice. We can set the target dir by selecting it from a dialog [ 'ask' ] , we can set target dir to the temp directory [ 'temp' ] or we can set the target directory to the current working directory ['cwd']. 

https://github.com/Zettlr/Zettlr/blob/7bc678982ebe64607c139ac982a8ad681100be63/source/app/service-providers/commands/export.ts#L112C1-L130C6

I don't want any of those options. I wanted the exporter to pick up the target directory from the metadata in my markdown file. If there is another way to do this please let me know. I couldn't find anything in the documentation or in my brief wade through the code.

I'm aware my solution might not be the cleanest but I wanted to do it with minimal code change. Happy if you want to propose a better way to do it - I'm sure there is.

regards, J

nathanlesage:
That is definitely what the Pandoc working directory is supposed to achieve. We should work from there. If I remember correctly I put together some documentation as well, so maybe have a read and see how we can work from there.

jamescormack:
Hi Nathan,

Looking at the code,  it looks to me like export.ts sets the exporterOptions.cwd from the pandoc_working_dir value (line 109) and then, if 'cwd' is set, sets the exporterOptions.targetDirectory to the same value.

https://github.com/Zettlr/Zettlr/blob/7bc678982ebe64607c139ac982a8ad681100be63/source/app/service-providers/commands/export.ts#L109C9-L109C83

This cwd specifies where the source files are located and where pandoc looks for source images etc, correct? If 'cwd' is set as an export option it also specifies the target directory which is how the output-file for pandoc is generated.

I want to override the output folder without changing where the source files are located. i.e. I want to chang targetDir so that when the default-exporter overrides the output-file in the defaults file, it can generate a filename that is NOT in the working directory.

https://github.com/Zettlr/Zettlr/blob/7bc678982ebe64607c139ac982a8ad681100be63/source/app/service-providers/commands/exporter/default-exporter.ts#L34

My understanding was that pandoc_working_dir was to specify the root folder of all input files/images etc for pandoc. Let me know if I've got it wrong. Meanwhile, I will do some further reading to see what I've misunderstood.

jamescormack:
... I'll also have a closer look at how cwd is used. Perhaps a simpler solution for me it just place the pandoc_template.tex into the latex-template folder and set the pandoc_working_dir to './latex-template'

Issue #5029, opened 2024-03-13
Update default number-sections settings according to new behaviour pandoc.


TeaDrinkingProgrammer:
### Description

Previously in Pandoc there was [an issue](https://github.com/jgm/pandoc/issues/5071) which meant that headings would show up as 0.1, 0.2 etc. when `number-sections` was set in the frontformatter. The issue has been [resolved](https://github.com/jgm/pandoc/commit/03bb426a4a65918bfd6fbd63d5bb32a657df5568), but the "fix" is still in the default frontformatter, so it now introduces a bug.

This
```yaml
number-sections: false
number-offset:
- 0
- 0
- 0
- 0
- 0
- 0
shift-heading-level-by: 1
```

Can now just be set to this:
```yaml
number-sections: false
```

### Reproducing

See mentioned issue

### Zettlr Version

Stable (most recent version)

### Installation Method

None

### Your Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

N.A

nathanlesage:
Three questions:

1. Do i understand correctly that we should remove two of the three properties that you removed in the second code block?
2. Can you provide example outputs for both settings?
3. Would you mind opening a PR with these changes?

TeaDrinkingProgrammer:
Hi there!
1. Yes, that's right
2. I can do that
3. I'll look into opening a PR-request some time soon.

nathanlesage:
Amazing, thank you!

TeaDrinkingProgrammer:
So I have done some more digging, and these are my findings.

On Zettlr X, XeLaTex X and Pandoc X, when you take the XeLaTeX PDF template and change `number-sections` from `false` to `true` and generate a PDF from this MD file:
```
---
title: "Test"
---
# Chapter 1
## Chapter 1.1
### Chapter 1.1.1
# Chapter 2
# Chapter 3
## Chapter 3.1
```
You get this:
![image](https://github.com/Zettlr/Zettlr/assets/20358521/3733d86f-7982-49cd-8af7-cd9bc80279b1)


When you change the base config and remove:
```
shift-heading-level-by: 1
```
And then change `number-sections` from `false` to `true`. This is the result:

![image](https://github.com/Zettlr/Zettlr/assets/20358521/a45d6cf7-8d52-466b-b3f9-ea0868726cc3)


[...]

PR #5073, opened 2024-03-31
fix: Removed shift-heading-level-by: 1 flag from defaults files. This was a workaround for Pandoc which is no longer needed.

TeaDrinkingProgrammer:
## Description
See title.

## Changes
On Zettlr 3.0.5, XeLaTex 3.141592653-2.6-0.999996 (MiKTeX 24.3) and Pandoc 3.1.3, when you take the XeLaTeX PDF template and change number-sections from false to true and generate a PDF from this MD file:
```
---
title: "Test"
---
# Chapter 1
## Chapter 1.1
### Chapter 1.1.1
# Chapter 2
# Chapter 3
## Chapter 3.1
```
You get this:
image

When you change the base config and remove:

shift-heading-level-by: 1

And then change number-sections from false to true. This is the result:
image

See #5029


TeaDrinkingProgrammer:
Do you want me to add this to the CHANGELOG.md?

nathanlesage:
@TeaDrinkingProgrammer If you wouldn't mind adding a section to the top that would be awesome. Feel free to use the following template (make sure to stick with hard-wrapped at 80 columns as the rest of the changelog; I don't have a line here to ensure that):

```markdown
## Changes to Pandoc Profiles

This update fixes a workaround that Zettlr had in place for a shortcoming of Pandoc regarding
the automatic numbering of headings. This workaround is no longer necessary, since Pandoc has
since fixed the issue. From this version onward, Zettlr will ship with changed default
profiles for various export formats. Since Zettlr never overwrites any data, the new defaults
will not be applied automatically. In case you want to use the new settings as well for your
own setup, you have the following options to adapt Zettlr after the update and ensure you use
the new defaults:

* In case you did not modify one of these profiles or are fine with resetting your changes,
  you can just delete them, which will prompt Zettlr to immediately recreate them – but now
  with the new settings
* In case you did modify them and want to keep your changes, rename the file to something
  else – Zettlr will then recreate the profile with the new default settings

You can also manually adapt them by replacing `TODO: INSERT OLD` with `TODO: INSERT NEW`.

The following profiles are affected by this change:

* TODO: List all files as they appear inside the assets manager of Zettlr
```

I leave it to you to find where to place that (trust your gut feeling regarding the importance, it's going to be fine). Also, feel free to adapt that suggestion as you find it useful for people who mostly *use* Zettlr (and not develop as me; I don't have the best view on how to formulate instructions for everyday users).

Thanks again for your work!

TeaDrinkingProgrammer:
Sorry for the radio silence. I normally rely on my email to get notified about updates from GitHub, but my account logged out, and I didn't notice until a quite recently. 

Thanks for the pointers on the changelog text, I have filled in the parts with `TODO` and made others a bit more straight to the point.

TeaDrinkingProgrammer:
@nathanlesage Anything I can do to push this foreward?

Issue #5046, opened 2024-03-19
Add a "Remove Link" option in the right-click context menu for links.


wideaperture:
### Description

Right-clicking a link in the editor window brings up a context menu with the options, "Open Link" and "Copy Link". It would be very useful if an option could be added to "Remove Link", which when clicked would remove the URL and associated markup from the markdown text, leaving only the text. This is a feature common to other apps with text-composing features. I would find it especially useful in cases where I've pasted in a quotation from a webpage that includes multiple links, as I currently have to manually edit the markdown for each one.

### Proposed Changes

See above.

### Caveats

None.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Nightly

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

Manjaro (Rolling)

obitwrite:
This would be really useful @wideaperture 

I've been puzzling with hyperlinks in Zettlr also. That's how I found this enhancement.

As this is added to the roadmap, I'd like to share an additional experience / use case, if you don't mind @nathanlesage 

Use case: Copying text from another app that contains a hyperlink
* Copy text from any other app containing a hyperlink.
* Paste in Zettlr
* Result: Zettlr creates a clickable link, however, the markdown []() is not added. Because it is clickable, to me, it creates the illusion that the proper markdown is present.

Discussion / 2 ideas:

* Idea 1: When the proper markdown is not present, the hyperlink perhaps should not be clickable, in order to avoid confusion. Why? Because when I want to apply additional markdown such as emphasis/italics to the pasted clickable url (that doesn't have the proper markdown), it adds an * to the end of the url. Also, when the hyperlink markdown is not present, emphasis/italics added to a full paragraph are not shown in Zettlr. So, when the proper hyperlink markdown is not present, it creates problems when a user wants to apply markdown emphasis to either the hyperlink or the paragraph that contains the hyperlink. (This is separate from the issue that Zettlr doesn't show bold or italics emphasis when aplied to hyperlinks) 

* Idea 2: Right now, the markdown suggestions don't say anything about hyperlinks that don't have the proper markdown. A suggestion that []() should be added, could guide users to apply the proper markdown.

PR #5126, opened 2024-04-30
"Remove Link" option in the right-click context menu for links

NMazzy:
## Description
#5046  Adds a "Remove Link" option in the right-click context menu for links, which when clicked would remove the URL and associated markup from the markdown text, leaving only the text.

## Changes
- Added a "menu.remove_link" field in the link right click context menu
- Created remove-markdown-link.ts file in markdown-editor/util

## Additional information

Tested on: 
- macOS Sonoma 14.1
- Ubuntu 22.04


nathanlesage:
Thank you! Looks good so far -- just one comment: instead of slicing the entire node, you can also just search for the various child nodes that it should have. If I remember correctly, the link title that you want is not represented by its own node, instead you would have to check for two CodeMark nodes, and slice from the end of the first to the start of the second. That makes it more resilient to edge cases and leverages the parser more. I think I already implemented that snippet in the markdown AST parser module. I think searching for `case 'Link'` might turn it up

MWBruce:
As suggested this has been implemented as suggested.

nathanlesage:
Looks good. I found one issue while testing this out, though: It doesn't work with angle brackets links, i.e.:

```md
This is some text with a plain link: <https://www.google.com>
```

Right-clicking that link will give you the option to remove a link, but it won't actually remove the angled brackets as I would expect it.

NMazzy:
> Looks good. I found one issue while testing this out, though: It doesn't work with angle brackets links, i.e.:
> 
> ```
> This is some text with a plain link: <https://www.google.com>
> ```
> 
> Right-clicking that link will give you the option to remove a link, but it won't actually remove the angled brackets as I would expect it.

I didn't even know this was a possible markdown link :)
This latest commit handles this case now.

source/common/modules/markdown-editor/util/remove-markdown-link.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const from = useNode.from
+  const to = useNode.to
+
+  var linkText: string
+  if (useNode.type.name === 'URL') {
+    // This is where the URL with <> formate is parsed
+    // We add +- 1 to remove the < > brackets from the text
+    linkText = view.state.sliceDoc(node.from+1, node.to-1)
```
nathanlesage, line 47: Can you be certain that the first and last characters will always be `<` or `>`, respectively? There may be other types of links that the parser understands that we don't yet account for.

source/common/modules/markdown-editor/util/remove-markdown-link.ts
```diff
@@ -0,0 +1,56 @@
...
+  } else {
+    // This is where the URL with []<> format is parsed
```
nathanlesage, line 49: You mean `[]()`, not `[]<>`, right? The latter isn't even a valid link type

source/common/modules/markdown-editor/util/remove-markdown-link.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export function removeMarkdownLink (node: SyntaxNode, view: EditorView): void {
+  // To deal with user being able to click on either part of the link (Text or URL)
+  // And the two different types of links []() and <>
+  // Check which node they clicked on before passing it
+  var useNode: SyntaxNode = node
+  if (node.type.name === 'URL' && node.parent != null) {
+    if (node.parent.type.name === 'Link') {
+      useNode = node.parent
+    } if (node.parent.type.name === 'Paragraph') {
```
nathanlesage, line 35: What exactly is the purpose of this specific check …? Did you verify what the syntax tree looks like for all types of links? The passed `node` should always be some type of link, correct …?

source/common/modules/markdown-editor/util/remove-markdown-link.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        removeMarkdownLink function
+ * CVM-Role:        Utility function
+ * Maintainer:      Hendrik Erz
```
nathanlesage, line 7: Add your name/GitHub handle/nickname here

source/common/modules/markdown-editor/util/remove-markdown-link.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Utility to remove a markdown link. It extracts the text from the markdown
+ * link and replaces the link with just the text.
+ *
+ * @param {EditorView} view - The editor view
+ * @param {SyntaxNode} node - The node containing the Link
+ */
+export function removeMarkdownLink (node: SyntaxNode, view: EditorView): void {
+  // To deal with user being able to click on either part of the link (Text or URL)
+  // And the two different types of links []() and <>
```
nathanlesage, line 29: Reminds me that there's a third type: Reference style: https://spec.commonmark.org/0.31.2/#link-reference-definitions

I am not exactly sure how Codemirror handles these, but with a bunch of console.logs that should be researchable

nathanlesage (changes requested):
We're getting there! Just a few more things!

source/common/modules/markdown-editor/util/remove-markdown-link.ts
NMazzy: No actually I found that there are 3 different cases of how a user can enter the right click context menu for a link:

For a [text](url) link:

1. They can right click on the [text] node which has type 'Link'
2. They can right click on the (url) node which has type 'Url'

For a <url> link:

3. They can right click on the <url> node which has type 'Url'

For cases 1. and 2. we extract the text from the 'Link' node, for case 3. we extract the text from the 'Url' node. I think the way which this was handled was a bit convoluted, I'll try simplifying this.

source/common/modules/markdown-editor/util/remove-markdown-link.ts
NMazzy: I found an existing function: shortenUrlVisually (/source/common/util/shorten-url-visually.ts) which handles what I am doing here, should I utilize this function instead?

source/common/modules/markdown-editor/util/remove-markdown-link.ts
NMazzy: I will look into this too.

NMazzy: image

At the moment the right click context menu doesn't appear when you right click on the [text][label] part of the reference style links. Is this something you want us to implement too?

MWBruce: It appears that a bug may be preventing the context menu from displaying for [Text 1][link1]. Additionally, the "remove link" option is currently visible for [link1]: https://example.com/. Nathan, do you have any suggestions on how we should handle this?

From my perspective, once the context menu bug for [Text 1][link1] is resolved, the "remove link" functionality should operate correctly and I think we should possibly raise an issue about this (which we will resolve next). However, the option to remove the link for [link1]: https://example.com/ will still be available. Could you please clarify how you would like us to proceed with this issue?

MWBruce: It appears that a bug due to the getURLForNode function call in the linkImageMenu may be preventing the context menu from displaying for [Text 1][link1] as it throws an error because of it not being able to find a URL child. Additionally, the "remove link" option is currently visible for [link1]: https://example.com/. Nathan, do you have any suggestions on how we should handle this?

From my perspective, once the context menu bug for [Text 1][link1] is resolved, the "remove link" functionality should operate correctly and I think we should possibly raise an issue about this (which we will resolve next). However, the option to remove the link for [link1]: https://example.com/ will still be available. Could you please clarify how you would like us to proceed with this issue?

source/common/modules/markdown-editor/util/remove-markdown-link.ts
nathanlesage: Please don't — `shortenUrlvisually` should only be used in contexts where what it spits out doesn't have to be a functioning url. As it says in the name: It's for visual purposes. Please just copy over whatever you need from that function here to prevent this from breaking.

source/common/modules/markdown-editor/util/remove-markdown-link.ts
nathanlesage: Ah, got it, thanks!

source/common/modules/markdown-editor/util/remove-markdown-link.ts
nathanlesage: I think those references … mh, it seems like we'd have to actually go searching in the syntax tree for the matching references before being able to decide whether to actually show a context menu. I would say this is out of scope for this specific issue, and should be tackled separately, especially since it needs to be fixed within a set of functionality, because iirc I never implemented reference links at any point …

MWBruce: Would you like us to report an issue for it?

NMazzy:
Hi @nathanlesage, I believe all your comments above, except the one regarding the reference style links have been implemented.

MWBruce:
@NMazzy I think it might be @kyaso that is reviewing PR's at the moment.

source/common/modules/markdown-editor/util/remove-markdown-link.ts
nathanlesage: Yes please do so!

MWBruce: Yes, done #5142

MWBruce:
@kyaso Further to a discord conversation I had with Nathan could you please advise if there is anything further required.

NMazzy:
> Sorry for the late reply!
> 
> I just tested your PR locally, and it seems to do its job, so everything fine code-wise from my side.
> 
> One thing we should add is a Changelog entry (see https://github.com/Zettlr/Zettlr/blob/develop/CHANGELOG.md). You can add it under the "GUI and Functionality" section.
> 
> Once this has been done, I guess, @nathanlesage, we can merge?

Hi, a changeling entry has now been added under the "GUI and Functionality" section.

nathanlesage:
Looks good for me, too, but the linter is complaining about an extra semicolon and about another issue (see the unit test) — once that's fixed we can merge. Thanks @kyaso for the code review!

NMazzy:
> Looks good for me, too, but the linter is complaining about an extra semicolon and about another issue (see the unit test) — once that's fixed we can merge. Thanks @kyaso for the code review!

Apologies should have checked this before hand, all should be fixed now.

nathanlesage:
All good, no worries! I'll wait for the unit tests to finish and then merge it in! Thanks again for the work!

PR #5128, opened 2024-04-30
Issue 4921 

imalexhu:
## Checklist
yarn test fails, but I can't really work out why

## Description
Changes the file preview so that the title display is now the file yamlTitle or firstHeading rather than fileName. This should make the file preview more descriptive.

## Changes
The file-find-and-return-meta data query now returns previewTitle in its 0th index rather than descriptor.name

Tested on: MacOs 14.3.1

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -262,7 +262,7 @@ jobs:
     env:
       npm_config_arch: x64
     needs: [preflight, verify_version]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       # Check out master for a regular release, or develop branch for a nightly
       - name: Checkout branch ${{ github.ref_name }}
@@ -304,7 +304,7 @@ jobs:
       CC: zig cc -target aarch64-linux-gnu
       CXX: zig c++ -target aarch64-linux-gnu
     needs: [preflight, verify_version]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout branch ${{ github.ref_name }}
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


nathanlesage:
I just ran the unit tests, let's see where it complains.

Then, this is a good suggestion, but there are settings that the user can check regarding what to view -- filename, first heading, or YAML title -- and you should respect them. These are simple Boolean toggles that you can access in commands using `this._app.config.get()` (or similar, but typescript should show you the way in VS Code).

The state machine for these is relatively simple, I've implemented them in a bunch of renderers, just search for `const useH1`.

If you could implement that, then it should be good to go!

imalexhu:
Thanks Nathan, Ill get onto this :)

imalexhu:
@nathanlesage Im still unsure why the tests are failing, but I have reimplemented the code so that it satisfies the condition outlined above. 

Thanks :)

imalexhu:
Looking at the config of zettlr 

Screenshot 2024-05-01 at 6 16 36 pm

Im assuming that the config "title+heading" is supposed to be "title|heading"? If that is the case, i can fix my code to render something different. 

Thanks :)

nathanlesage:
No no the config option is supposed to be one of three strings:

* `title`
* `heading`
* `title+heading`

The character in between doesn't matter so much, but that is why I do the "includes" check -- that will match both "heading" and "heading+title" and the same for the title property.

nathanlesage:
But that being said, as you implemented it should work as well. Just two final changes and then it's ready to merge:

1. I recommend not extending the class and instead using a simple function outside of the class but within the file. I'm planning to convert this whole class instantiation mess into simple callable functions and if the preview title generator is implemented as a simple function inside that file, it'll be easier to do the migration going forward
2. Convert the Changelog changes you added into a bullet item that simply states `fix: The file preview tooltip now respects the filename display settings`

imalexhu:
@nathanlesage Hello, sorry to ping. But I think I have redone this PR so that it fits what you describe. Please let me know how to progress at this stage

PR #5162, opened 2024-05-15
Preferences window design refinements

sensologica:
Made a number of small CSS tweaks to match the Preferences window implementation to the design spec.

### DIFF
--- a/source/common/vue/form/FormBuilder.vue
+++ b/source/common/vue/form/FormBuilder.vue
@@ -15,7 +15,10 @@
       </div>
       <fieldset>
         <!-- First, let's do some setup of the fieldset -->
-        <div class="form-header">
+        <div
+          class="form-header"
+          v-bind:class="{ 'space-between': fieldset.title === 'Application language' }"
+        >
           <!-- First the fieldset legend: Required -->
           <legend>
             {{ fieldset.title }}
@@ -323,10 +326,13 @@ function getModelValue (model: string): any {
 
     .form-header {
       display: flex;
-      justify-content: space-between;
       align-items: center;
-      column-gap: 10px;
+      gap: 24px;
       margin-bottom: 16px;
+
+      &.space-between {
+        justify-content: space-between;
+      }
 
       legend {
         font-size: 15px;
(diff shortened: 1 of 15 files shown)


source/win-preferences/App.vue
```diff
@@ -19,11 +19,11 @@
       <template #view1>
         <TextControl
           v-model="query"
-          v-bind:placeholder="'Find…'"
+          v-bind:placeholder="'Search'"
```
nathanlesage, line 22: Ah, this needs to be a variable so that it becomes translatable. How I did it in all Vue components is like so:

1. Define a variable in the script part of the file: `const queryPlaceholderLabel = trans('Search')`
2. Reference that like so: `v-bind:placeholder="queryPlaceholderLabel"`

Feel free to choose a different name -- and if there are already labels defined like this, group this new one together with them!

source/common/vue/form/elements/TextControl.vue
```diff
@@ -1,6 +1,7 @@
 <template>
-  <div v-bind:class="{ inline: inline === true, 'form-control': true }">
+  <div v-bind:class="{ inline: inline === true, 'form-control text-field': true }">
```
nathanlesage, line 2: I would separate the classes out; which makes it more verbose, yes, but also faster to parse that these are two classes

source/common/vue/form/elements/RadioControl.vue
```diff
@@ -80,22 +80,26 @@ function fieldID (key: string): string {
 body {
   .radio-group-container {
     break-inside: avoid;
-    margin: 10px 0;
+    // margin: 10px 0;
```
nathanlesage, line 83: Feel free to remove that line entirely if it isn't required!

source/common/vue/form/elements/CheckboxControl.vue
```diff
@@ -88,13 +88,13 @@ body {
...
     // user clicks far beyond the label string)
-    grid-template-columns: @input-size * 2 max-content;
+    grid-template-columns: @input-size * 1.6 max-content;
     grid-template-rows: 100%;
     grid-template-areas: "input label";
     align-items: center;
 
     &:not(.checkbox-outer-div-inline) {
-      margin: 10px 0;
+      // padding: 5px 0;
```
nathanlesage, line 97: Feel free to remove if no longer necessary

source/common/vue/form/elements/ButtonControl.vue
```diff
@@ -48,4 +48,17 @@ const fieldId = computed<string>(() => `form-button${props.name}`)
...
 <style lang="less">
+  body {
+    .form-control.inline {
+      margin-right: 24px;
+    }
+  }
+
+  body.darwin {
+    .form-control {
+      >button {
```
nathanlesage, line 59:
```suggestion
      > button {
```

source/common/vue/form/FormBuilder.vue
```diff
@@ -15,7 +15,10 @@
       </div>
       <fieldset>
         <!-- First, let's do some setup of the fieldset -->
-        <div class="form-header">
+        <div
+          class="form-header"
+          v-bind:class="{ 'space-between': fieldset.title === 'Application language' }"
```
nathanlesage, line 20: Is this really only necessary for a single field? This seems brittle, especially should we ever decide to change its name … feel free to explain what you want to achieve.

nathanlesage (approved):
Looks good, I'm going to trust you with the CSS changes :D

PR #5164, opened 2024-05-16
fr-FR.po: Mistranslation for "Save changes" and "Discard changes"

artav4:
## Additional information


nathanlesage:
Looks good! Could you remove the lines that say "fuzzy" directly above these messages to indicate that these translations now have been fixed? Otherwise this is perfect and can be merged. Thank you!

PR #5182, opened 2024-05-24
fix: search in folder autofocus

viktomas:
## Description
This MR fixes a bug when autofocus for the "search in folder" (aka. global search) panel stopped working

| before | after |
| -- | -- |
|  zettlr-focus-before | zettlr-focus-after |

I tested it locally

## Changes
The ref of the input component was incorrect.

## Additional information

The issue was introduced in https://github.com/Zettlr/Zettlr/commit/3ecba4659f0d8193f76cbd22f8aa1f3b0b5706ec

`yarn test` fails for me locally (immediately without running any tests)

```
Exception during run: TypeError: Cannot set property navigator of #<Object> which has only a getter
```

Tested on:  latest MacOS 14.5 (23F79)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ multiple queries.
 
 - **Change**: When searching for files in the filter field, only files and
   workspaces that match all queries entered will be displayed
+- **Fix**: Focus input field when search in folder (global search) is
+  triggered
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
The error you get seem to be something borked in your setup, but no worries, the pipeline correctly reports that all is good with your PR! Thanks, merging in

Issue #3552, opened 2022-06-17
Word count session target


nomad101:
### Description

It's great that it's possible to set an overall word count target for a file.

Although for daily writing, it can be very beneficial for productivity to also have a word count target for a _session_. For example, you may have an overall word count for a document to be 5000 words. And for a writing session, your target could be set to 500 words. 

### Proposed Changes

Add additional functionality to be able to set a session word count target. 
User would need input how many words required for their writing session. 
The target display could be both actual number of words written and % in relation to session word count target. 

### Caveats

None. Though would need to be displayed clearly. And not be confused with overall word count target for the doc. 

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Operating System Version

MX Linux

### Additional Information

Thank-your for considering this request.

wideaperture:
I like this idea. It also might be helpful to be able to set a word count target for a workspace folder in the same way one currently sets it for an individual notes. So, for example, if I'm working in a Zettlekasten and I have a "Permanent Notes" folder that I'd like to contribute 900 words to today there'd be a simple way to see this, even if the collective word count was spread across multiple notes in the workspace.

PR #5234, opened 2024-06-16
Word count implement

OOGZleo:
## Additional information


Wei-whey-way:
**Description**
We plan to fix the issue mentioned [issue 3552](https://github.com/Zettlr/Zettlr/issues/3552) on 64 bits windows 11, which involves a way to implement a daily word count session target goal

**Changes**
The attached image outlines the proposed features added. The top right item circled in image 1 shows the total word count of the document while the bottom circled item outlines the daily session word count as well as a method to set the session word count target on the right.

image

Image 2 creates a pop up when clicking on the 'set word count session target' component, where a user can set or update the targeted daily word count.

image

The test case includes a folder with brackets in its name alongside a template .tex test file in it, a screenshot of the working app is also included in the latest commit. Further example images are located in resources/screenshots.

**Additional information**
Tested on: 64 bits Windows 11

PR #5237, opened 2024-06-17
Update fr-FR.po

camillem:
Corrected some obviously incorrect fuzzies.
Not the best translations, but should at least create less confusion.

## Additional information


nathanlesage:
Sorry that it took some time – thanks for your efforts! Merging in …

PR #5243, opened 2024-06-23
feat: Rewrite TableEditor from Scratch

### LINKED ISSUE #1999: [ENHANCEMENT] Table Heading enforced
Hi, i've tried using a table without a header (by removing the |--|--| line). But it keeps getting added again. According to the documentation (at least how i understand it), the table header is optional.

https://docs.zettlr.com/en/core/tables/

"""Optionally, you can also add a heading row, which may also indicate the alignment using colons (left, right, or center):"""

## Description
remove the automatic addtion of the |--|--| line if it's not in the table

## Proposed Changes
heading line would only be optionally in bold writing

### LINKED ISSUE #3596: Large tables cause the view to scroll down upon update
### Description
Adding or removing rows or columns to a table can cause the view to scroll down the page.
Changing the theme produces the same effect.

My guess is that the table momentarily becomes raw text, which is shorter than a table row.
Then, the text expands back into a table and pushes the screen down with it

### Reproducing
Create a file with a (sufficiently large) table at the top and many lines of text at the bottom.
Add a row to the bottom of the table.

**or**: while the large table is in frame, change themes.

### Zettlr Version
2.3.0 (UUID: 56bb8ed2-45e8-4c39-a961-66a2589fd908)
running on win32 10.0.19043 (architecture: x64)

### Installation Method
From the Website

### Your Platform
- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture
- [x] ARM 64bit
- [ ] Intel 64bit

### Operating System Version
Windows 10

### Additional Information
No response

### LINKED ISSUE #4657: Horrible typing issues with a big table in a long document
### Description

I'm writing my thesis, it's therefore a long document (right now around 46k characters) and wanted to introduce a table. After reading the table editor instructions after a few failed tries (sorry, not a very intuitive feature), I created a table of around 8 rows and 4 columns. As soon as it was saved, and I wanted to start typing again, the typing was EXTREMELY slow and also did not type correctly - as it was catching every other typed letter.

### Reproducing

Create a long document (mine is around 45k characters) and a large table and start writing. Not sure if reproducible or just me.

### Zettlr Version

Stable (most recent version)

### Specify version

_No response_

### Installation Method

None

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 11

### Additional Information

_No response_

### LINKED ISSUE #5080: Data disappear from table when saving
### Description

Hello, 

Sorry if this has been discussed, but I didn't find any issue like this one.

When I save a file with a table (at least 2 rows and 2 columns), data disappear from the last edited cell, or sometimes it switches back to the previous version.

I haven't noticed any clue on why this is happening.

My configuration:
- Zettlr 3.0.5 (Windows)
- Theme: Berlin with dark mode
- "Accentuation du rendu" activated (sorry, I don't know how this option is called in English…)

It is very annoying since it makes the use of table unstable and unreliable.

Is it a documented issue?

Regards,

### Reproducing

As far as I know, you only need to create a table with at least 2 rows and 2 columns and add some data.

### Zettlr Version

Stable (most recent version)

### Specify version

3.0.5

### Installation Method

None

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 10

### Additional Information

_No response_

### LINKED ISSUE #5190: Moving the mouse toward the top left corner of the app causes Table Alignment hover menu to shows up.
### Description

When a note having a markdown file containing a table is opened and the Table Editor option enabled, hovering the mouse to the left of the "File manager" button (or to the left of "File button" with Native Window Appearance disabled) on the toolbar will cause the Table Alignment hover menu to appear when it should not.

### Reproducing

1. Open Preference > Editor and make sure Table Editor is enabled
2. Create an empty Markdown file
3. Create a Table

If Native Windows Appearance option is enabled, hovering the mouse to the left of "File manager" button on the toolbar will cause the Table Alignment hover menu to show up.

If Native Windows Appearance option is disabled, hovering the mouse to the left of "File" button on the toolbar will cause the Table Alignment hover menu to show up.

### Zettlr Version

Stable (most recent version)

### Specify version

3.1.1

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Fedora 39 (Xfce)

### Additional Information

It created many instances of all table hover items when the bug happen.

image

nathanlesage:
## Description

This PR provides a wholly rewritten table editor. The aim is to fix several data loss issues that are scattered around the repository right now. For that, the table editor now hooks much more into the novel CodeMirror 6 API and ensures that any changes -- inside and outside of tables -- is always retained in the main editor, and never in some additional state that needs to be synchronized. This addresses the main issue that many people had where it was intransparent that they had to perform a special action before changes were persisted to the actual document.

## Changes

* Completely remove the old TableEditor
* Provide a rewritten one that keeps all changes in the main view, and only provides a GUI on top of Markdown tables

## Additional information

Currently, this PR is in draft status, since I haven't been able to fully replicate the old editor's functionality yet.

Fixes the following issues:

* Fixes #4657
* Fixes #5190
* Fixes #5080
* Fixes #3596
* Fixes #1999

Tested on: macOS Sonoma 14.5


Liam-Twomey:
A similar project is the Obsidian advanced tables [plugin](https://github.com/tgrosinger/advanced-tables-obsidian), thought it could be a useful reference for this PR and wanted to bring it to your attention.

source/common/modules/markdown-editor/commands/tables.ts
```diff
@@ -0,0 +1,408 @@
...
+ */
+
+import { syntaxTree } from '@codemirror/language'
+import { ChangeSpec, EditorSelection, SelectionRange } from '@codemirror/state'
+import { EditorView } from '@codemirror/view'
+import { parseTableNode } from '../../markdown-utils/markdown-ast/parse-table-node'
+import { SyntaxNode } from '@lezer/common'
+
+/**
+ * Takes a tale node and the corresponding Markdown source and returns a list of
```
kyaso, line 24:
```suggestion
 * Takes a table node and the corresponding Markdown source and returns a list of
```

nathanlesage:
jfc, tables are hard.

I have by now fixed a few issues with empty cells and inserting whitespace at the start/end, but I totally forgot that I already included a custom Grid table parser, which the table parser in the MarkdownAST (so the second stage parser that takes a Lezer parse tree and emits an actual AST) now no longer supports. It's now more stable with regard to sloppy pipe tables, albeit it doesn't correctly parse pipe tables with "+" as a delimiter in the header row instead of the regular "|".

Also, the commands still haven't made progress. But anyhow, it is progressing, albeit *very* slow.

nathanlesage:
I'm making progress with the commands. What is still missing are the deletion commands (both rows and columns), the clear commands (same), and the auto-alignment transaction extender.

Quick TODO list for the table commands (reminder that this comprises the first development milestone, CC @sensologica so you have an overview over what needs to be done before I can get to the tab selection field as discussed in the appropriate issue):

- [ ] The commands currently insufficiently account for grid tables
- [ ] The column commands currently require the cursor to not sit somewhere in the whitespace around a table. This becomes an annoyance in tables with little content that are properly aligned (because the amount of whitespace padding the table exceeds the amount of characters at which positions the commands are triggered)
- [ ] The commands currently don't handle selection changes (especially swapping and adding should move the cursor to the swapped or added columns/rows)
- [ ] Both the Lezer and AST parser currently don't handle complex grid tables (which should be handled, even though I won't support them in the table editor)
- [ ] After implementing the auto-alignment, run some performance tests to see at which point we shouldn't update in real time anymore\*, and rather update on cell movements (also, make it disableable)
- [ ] Think of proper commands. I have now tried to stick with Tab and Enter for movement and Alt-Shift-Arrow keys for additions in all four directions. This works well, but we may want to discuss this.

\* *Auto alignment refers to the automatic re-alignment of a table in case of modifications; real-time here refers to an update every time the user inserts or deletes a character, whereas non-real time would involve updating only on cell movements, i.e., when the user presses Tab or Enter to move between cells*

Furthermore, we have to document the following:

* The table editor won't pre-render "simple" tables (though supported by Pandoc) or grid tables, because both can become very complex and especially grid tables can't be properly handled by the way the table editor is implemented. Simple tables haven't been supported in a long time and the Pandoc team itself mentioned we shouldn't bother to. Grid tables, on the other hand, are rarely used.
* Complex grid tables (with multiple lines per cell, colspans, and rowspans) aren't handled properly by both Lezer and AST parsers. We should at least attempt it so that at least Markdown→HTML conversion during copying works. But this is OOS for this PR. I'm happy with providing a bug-free table editor experience, I don't have the time to fix Markdown tables as such.

nathanlesage:
Update: Hit a soft milestone today: As of now, all table commands are implemented. *However*, there are still some issues. This merely solves the "blank page problem," the two primary issues that remain are the missing selection setting and that the commands still may work a bit iffy with grid tables.

I hope that another day of intensive coding will remedy this and we're done with the first milestone, after which we can get to other issues, before implementing the GUI for the table editor.

nathanlesage:
Some notes for the review later on (before I forget this). @kyaso You don't need to review just yet, since I want to fix a few things before. Here I wanted to only provide a few notes on what the code is supposed to do so that it is easier for you to dig through this code once I'm done:

**How the Commands are Supposed to Work**

I tried to design the commands as efficiently as possible, but they turned out to be already quite complex (without even yet fully supporting even simple grid tables!).

Each command calls `mapSelectionsWithTables`, which is essentially just a glorified `forEach` loop, that makes certain guarantees, and behavior:

* If `callback` is called, it will be for a valid table `SyntaxNode` that the AST parser could successfully turn into a `Table` node, and that contains at least one valid selection range.
* Within these callbacks, all commands receive all ranges that are somewhere contained within the table, and are responsible for iterating over the ranges if necessary. That means that, if you have multiple cursors in different cells, `Tab` will advance all these ranges to the corresponding next cells.
* I have defined quite a few utility functions that hopefully make the code simpler to read and ensure that the commands work consistently
* The callbacks can return `undefined` which gets automatically filtered.
* I use quite a lot of calls to `flat()`, since the way I implemented the code (I often try to follow a MapReduce procedure because I learned it at Uni and it seems very elegant for these purposes) often produces n-dimensional arrays. If you spot a way where we could simplify this that would be amazing.
* Most commands return `TransactionSpec`s from the callback, to preserve or set new selections, the call to `target.dispatch(...tr)` ensures that CodeMirror combines all of them into one transaction
* All commands return `false` when they didn't make a change (`changes.length === 0`) to ensure that overloaded commands get a chance to run if no selection was within a table

**Available keyboard commands**

To ensure we can properly test it, and also because I want these commands to also work when the user has disabled the TableEditor, I have assigned all commands a keyboard shortcut, sometimes overloading the existing ones (i.e., a shortcut will run a table command if in a table, or the default one if not). IMPORTANT: Many aren't yet set in stone; I tried to be reasonable, but I expect some to change before merging this PR. Here's the current, temporary keymap for table commands:

* <kbd>Tab</kbd>: Move to the next cell (<kbd>Shift</kbd> for previous cell)
* <kbd>Enter</kbd>: Move to the next row (<kbd>Shift</kbd> for previous row)
* <kbd>Alt-Shift-ArrowUp</kbd>: Add a row before the one where the cursor is in
* <kbd>Alt-ArrowUp</kbd>: Swap the current row with the previous one (accounting for delimiting rows where necessary)
* <kbd>Alt-Shift-ArrowDown</kbd>: Add a row after the one where the cursor is in
* <kbd>Alt-ArrowDown</kbd>: Swap the current row with the next one (accounting for delimiting rows where necessary)
* <kbd>Alt-Shift-ArrowRight</kbd>: Add a new column to the right of the one where the cursor currently is in
* <kbd>Alt-ArrowRight</kbd>: Swap the current column with the next one
* <kbd>Alt-Shift-ArrowLeft</kbd>: Add a new column to the left of the one where the cursor currently is in
* <kbd>Alt-ArrowLeft</kbd>: Swap the current column with the previous one
* <kbd>Shift-Mod-k</kbd>: Delete the current column
* <kbd>Mod-Backspace</kbd>: Clears out the current column (i.e., inserts whitespace)
* <kbd>Shift-Mod-Backspace</kbd>: Clears out the current row (i.e., inserts whitespace)
* <kbd>Alt-Shift-Mod-Backspace</kbd>: Clears out the entire table
* <kbd>Alt-c</kbd>: Aligns the current column center (by placing colons in the delim row)
* <kbd>Alt-l</kbd>: Aligns the current column left
* <kbd>Alt-r</kbd>: Aligns the current column right

… Jesus Christ, I just realized how massive the amount of changes already is. @sensologica But do you understand now why I believe I should finish this work before distracting myself with anything else? :D

kyaso:
@nathanlesage Thanks for the summary, this will be super helpful during the review!

Also, apologies for being inactive on this PR 🙁 I’ve been quite busy over the past few months and couldn’t find the time to work on it properly. But looks like you are in a flow state with the recent changes, so looking forward to reviewing it! :D

nathanlesage:
Don't worry, I have accepted that I'll just write the logic myself, and as long as you can give it a good review, then I'm absolutely happy and satisfied! 🙏🏼❤️

nathanlesage:
I think we're mostly set for testing the table editor! There's just a single thing that I haven't yet gotten to do: The drag and dropping of the table. But I feel as if that is something I can deliver during the actual beta phase. So for all intents and purposes, I believe the PR is ready for review @kyaso @sensologica 

Also @sensologica I feel that the padding for the tables is a tad too generous, I prefer more tight tables — any opinions?

telegram-cloud-photo-size-2-5433937381486096611-y

nathanlesage:
We have mainly two remaining "problems" that make the TableEditor somewhat inferior to raw code editing:

* Currently, selections aren't synchronized between the two views (because that somehow causes the subview to become unresponsive to mouse presses, idk)
* Some extensions don't work in the subview (mostly only those who shouldn't work there anyhow, but also a few useful ones such as smart quotes).

In an ideal world, I could just literally mirror everything, with the only change between the main and sub views being that the subview only shows the cell content. I'm not yet sure, however, how to do that. But that would be ideal.

nathanlesage:
Update: Did some digging. The problem with clicking again to move the selection is that the main editor view focuses itself on such an event, and if I pass everything just to the main view, this also includes selection pointer events.

In addition, there is something weird happening with the selections if I do it this straight-forward. I'll have to investigate further.

nathanlesage:
> When editing, the view would jump to the top of the document when switching cells with Tab or Enter and then jump to the cell being edited once I start typing.

This is likely caused by the way the subviews are mounted in the table. Specifically, I suspect (I am writing this mostly as a reminder to myself to check that), that, when you press tab, the selection in the underlying main view is forwarded (or retracted) to the next/previous cell. This then causes an update in the table, which will first remove the existing subview from the "old" cell, and then instantiate another subview. As soon as a CodeMirror instance loses focus, the main editor regains focus, and as all editors try to pull their own respective selection into view, it could be that in the small amount of time it takes to move the subview to the other cell, that the main view gains focus momentarily and manages to drag the scroll bar up, and the subview can't pull it back down. Maybe we can fix this by first instantiating the new subview, and only then removing the old one. But that's what I'll have to check!

***

Then, some additional observations from @wideaperture (thanks a ton!):

* [x] Upper controls cropped when table appears at the top of the editor pane
  * Likely an easy fix, as I did it with the other sides of the table (negative margin and padding)
* [x] Underlines appear intermittently when editing heading rows.
  * This is a simple CSS rule that should be quick to fix
* [x] Apparent drag/drop UI element doesn't do anything.
  * That's a very nice compliment to @sensologica that the element is dead-simple to understand! Yes, it shall drag and drop columns and rows… I just didn't implement that yet!
* [x] No way to opt out of using a heading row.
  * This is unfortunately a requirement of Markdown tables, and was also the case with the last table editor, too.
* [x] Deleting a row causes the table editor to degrade to raw markdown.
  * Good catch! This only happens in the top row (which is the only one my lazy ass didn't test), and should be mitigate-able by improving the corresponding command
* [x] "Insert new row below" does not appear to work in context menu.
  * Thanks for discovering, I'll look into this
* [ ] Column sorting does not appear to work.
  * Ah damn, that is something I forgot…

nathanlesage:
Another note: I just realized that the negative margins offset background layers, which makes it look bad, so I will have to find another way of doing that (see screenshot)

image

nathanlesage:
Aaand another observation: One doesn't even have to have a table cell active for the editor to exhibit jumping.

wideaperture:
One thing I noticed in the test install what’s that the block quotes in my documents ended up with weird indentation, where the first line of the quote appeared further to the left than the subsequent lines. Not sure if this might be some CSS from the table, editor bleeding over into the rest of the document.

nathanlesage:
That might be the additional padding I added to show the table buttons. Unfortunately, otherwise I can't make them appear; so we have the choice between weird style glitches OR a slightly indented table.

Or we could move the left side buttons to the right side. @sensologica thoughts?

nathanlesage:
This is how the table editor would look like with spacings and margins that work properly:

image

The big issue is that this indents it by a few pixels to the left (see the table below "alignment test table", note how indented it is.

@sensologica @wideaperture

kyaso:
@nathanlesage Just FYI, I'm already in the process of reviewing the PR bit by bit (whenever I find some time). But oh boy, is it massive 😱😂

nathanlesage:
Im sorry 😂🙈

But it's good to know that you're reviewing now -- I'll let the PR rest for now so that the code doesn't change -- take your time, the app has enough other construction sites to comb through 🚧

kyaso:
Oh, and just to clarify — my initial comment wasn’t about you pushing changes during my review. I just wanted to let you know I was on it and hadn’t lost track!

So, if you have anything to change, go ahead!

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * This syncAnnotation is used to tag all transactions originating from the main
+ * EditorView when dispatching them to the subview to ensure they don't get
+ * re-emitted.
+ */
+export const syncAnnotation = Annotation.define<boolean>()
+
+export function dispatchFromSubview (mainView: EditorView): (tr: Transaction, subview: EditorView) => void {
+  return (tr: Transaction, subview: EditorView) => {
```
sensologica, line 28: Maybe `tr` could be renamed to `t` or `tran`? Out of habit I read `tr` as HTML table row.

source/common/modules/markdown-utils/markdown-ast/parse-table-node.ts
```diff
@@ -12,149 +12,242 @@
...
 
-import { type SyntaxNode } from '@lezer/common'
-import type { Table, TableRow, TableCell } from '../markdown-ast'
-import { getWhitespaceBeforeNode } from './get-whitespace-before-node'
-import { parseChildren } from './parse-children'
+import type { SyntaxNode } from '@lezer/common'
+import type { Table, TableRow, TableCell, TextNode } from '../markdown-ast'
 import { genericTextNode } from './generic-text-node'
+import { parseChildren } from './parse-children'
+// import { logLezerTree } from './debug-tree-logger'
```
sensologica, line 26: If not needed, this import should be deleted.

source/common/modules/markdown-editor/table-editor/commands/tables.ts
```diff
@@ -0,0 +1,169 @@
...
+  
+      if (idx === undefined) {
+        return undefined
+      }
+  
+      const delimChar = ctx.tableAST.tableType === 'grid' ? '+' : delimLine.includes('|') ? '|' : '+'
+      const fillChar = ctx.tableAST.tableType === 'grid' ? '=' : '-'
+      const delimOffsets = getDelimiterLineCellOffsets(delimLine, delimChar)
+      const [ from, to ] = delimOffsets[idx]
+      if (which === 'left') {
```
sensologica, line 57: Maybe `which` could be renamed to `alignTo` to make these conditionals read better?

source/common/modules/markdown-editor/table-editor/context-menu.ts
sensologica, line 1: Are we sure we want to group context menu items by type of action instead of by target? See the attached screenshot for an alternative grouping.

Screenshot 2025-05-09 at 10 24 04

sensologica (commented):
Wow that is _a lot_ of code! Bravo, Hendrik!

The table editor looks very promising. Of course there are a few expected early-days rough corners here and there, but overall it's amazing how much better it works than the previous version.

I've left some comments, mostly just small opinions/suggestions. Feel free to ignore if you disagree. I will welcome any feedback on how to make my code reviews more useful in the future (what specifically you want me to look for, etc.).

Now my biggest gripes with the table editor are actually related to my design, now that I see it in-context. I think we can simplify it a lot. I am not sure the on-hover outline is needed. Also, the colors of the alternating rows are too contrasty and break up the uniformity of the table. And other things like the controls are too bulky and too close to the table border. But all this is on me. So please let me know once you are done on your end and I will go ahead and do a design cleanup pass to make sure the tables look as good as possible before we release.

sensologica:
Another example where clicking into a table causes jumping:

https://github.com/user-attachments/assets/1ce8ae14-3551-4c81-ae03-850365919d87

nathanlesage:
> Now my biggest gripes with the table editor are actually related to my design, now that I see it in-context. I think we can simplify it a lot.

Please, go ahead -- aside from some minor bugs that aren't related to the functionality itself, the table editor is mostly done so we don't do double work if you start giving it another pass! I'll check your comments soon!

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -344,14 +357,13 @@ export interface Table extends MDNode {
    */
   rows: TableRow[]
   /**
-   * A list of column alignments in the table. May be undefined; the default is
-   * for all columns to be left-aligned.
+   * A list of column alignments in the table.
    */
-  alignment?: Array<'left'|'center'|'right'>
+  alignment: Array<'left'|'center'|'right'>
```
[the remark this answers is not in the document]
nathanlesage: Usually, type hints will already provide you with only those strings, and the linter will underline any string which expects left, center, or right, but receives something else. An enum would require a bit more writing (e.g., `TableAlignment.Left` instead of simply `left`).

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
nathanlesage: I decided to stick with `tr`, as that is the standard variable name for transactions with CodeMirror. I agree that it introduces semantic ambiguity, but I don't feel that it's confusing, especially since you always have type annotations that clarify if its a TR element, or a transaction.

nathanlesage:
@sensologica I have now implemented all your (pretty good!) review comments.

Also, I have now removed a lot of the theming, to make the tables very simple and thus easier on the eyes.

Regarding the context menu: I have now implemented your suggestion, but I've seen that for Obsidian, for example, they've put some frequently-used items on the top level, and I think we should do the same. But I'd argue let's first see what items users actually use frequently.

nathanlesage:
Status update: I found and fixed one more bug, but there is one final annoying thing: I can't insert spaces at the end or beginning of table cells without the table editor widget swallowing them. But apart from that I can't find any new bugs. Once that is done, we're ready for a … lengthy beta phase.

nathanlesage:
Another update: It's almost good. I think we can soon merge this. There are a few remaining visual oddities, but it works almost flawlessly. Citation rendering support is now also there.

nathanlesage:
I think it's now finished. The only two remaining issues that don't work yet are sorting columns and dragging and dropping columns and rows. Apart from that, everything appears to work now. I'll let this rest a few more hours, and then merge so that it can be included in tomorrow's nightly for some rounds of alpha testing before we enter the beta phase.

PR #5257, opened 2024-07-01
5217-Copy link remove <> from links

### LINKED ISSUE #5217: Copy link does not remove <> from links
### Description

The right-click "copy link" feature does not strip out the markdown characters.

### Reproducing

With the new link format `<https://www.google.com/>`, when right-clicking on the link and selecting "copy link" it copies the link and includes the `<` at the start and the `>` at the end.

### Zettlr Version

Stable (most recent version)

### Specify version

3.1.1

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Debian Bookworm

### Additional Information

Code: https://github.com/Zettlr/Zettlr/blob/3c822acbb1387f84e79a9f4b339541ea55f71218/source/common/modules/markdown-editor/context-menu/link-image-menu.ts#L120

kus0023:
Fixed #5217 

## Description
Fixed the issue of URL, which is adding "<" and ">" in URL.

## Additional information
Any suggestions are welcomed. Thankyou.

Tested on: Windows 10 x64


kus0023:
- Yes, I have used JSStandard coding style. And I have given comments.

source/common/modules/markdown-editor/context-menu/link-image-menu.ts
```diff
@@ -124,7 +124,15 @@ export function linkImageMenu (view: EditorView, node: SyntaxNode, coords: { x:
 
   showPopupMenu(coords, isLink ? linkTpl : imgTpl, (clickedID) => {
     if (clickedID === 'menu.copy_link') {
-      navigator.clipboard.writeText(url).catch(err => console.error(err))
+      var sanitizeUrl = url;
+      if(sanitizeUrl.includes("<") || sanitizeUrl.includes(">")){
```
nathanlesage, line 128: We can do this more elegantly. I would suggest to use a regexp (eg `/^<.+>$/`, but double check it does what you want) and then `url.slice` or `url.substring`. Right now the if looks a bit vague (it tries to replace < even if there's only a > in there, and since both signs belong semantically together, we should also treat them as together, ie a url that starts with < but does not end with > should not happen).

I hope I was clear in this statement ...? If not please ask! Otherwise, thanks for the suggestion, it looks otherwise good and mergeable!

nathanlesage (changes requested):

nathanlesage:
Is it a mistake that you closed this pr...?

Issue #5138, opened 2024-05-05
Focus tag filter after clicking tag selector


chiphogg:
### Description

In previous versions of Zettlr (e.g., I had been on 2.3.x previously), I was able to click on the "Tags" button, and immediately start typing to filter for the tag.  Currently (I'm on 3.1.1), when I click, the note window still has focus, so if I start typing, the text gets added to the note.  I would need to make two clicks to start typing.

### Reproducing

1. Click the main note window to focus it.
2. Click the "Tags" button.
3. Start typing.

I would expect this text to go to the filter, as it used to, but it goes into the note instead.

### Zettlr Version

Stable (most recent version)

### Specify version

3.1.1

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Ubuntu 22.04

chiphogg:
Any update on reproducibility, and/or how hard the fix is?

This bug litters my various documents with meaningless little strings (i.e., the first few characters of my attempted tag search), in ways that are hard to find later.

nathanlesage:
Heya, sorry that issue fell right into my "I can't look at GitHub" phase -- I guess it should be an easy fix. So thanks for bumping I'd say

nathanlesage:
Yup, as I thought:

https://github.com/Zettlr/Zettlr/blob/develop/source/win-main/PopoverTags.vue

There on line 110 it needs to be the after mount hook, not before mount.

PR #5273, opened 2024-07-12
Update PopoverTags.vue

Barnamoyy:
## Description

Fixes issue #5138 where the tag filter didn't get focused after clicking once, It had to be clicked twice to register the typed text. This issue was observed on Linux platform.

## Changes

The code changes line 110 in file https://github.com/Zettlr/Zetttlr/blob/develop/source/win-main/PopoverTags.vue by replacing the onBeforeMount hook with onMounted hook.

## Additional information

None

Tested on: macOS

### DIFF
--- a/source/win-main/PopoverTags.vue
+++ b/source/win-main/PopoverTags.vue
@@ -52,7 +52,7 @@ import PopoverWrapper from './PopoverWrapper.vue'
 import TextControl from '@common/vue/form/elements/TextControl.vue'
 import TabBar, { type TabbarControl } from '@common/vue/TabBar.vue'
 import { trans } from '@common/i18n-renderer'
-import { ref, computed, onBeforeMount } from 'vue'
+import { ref, computed, onBeforeMount, onMounted } from 'vue'
 import { useConfigStore, useTagsStore } from 'source/pinia'
 
 const props = defineProps<{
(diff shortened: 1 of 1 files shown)


kyaso (changes requested):
Thanks for your contribution!
Just a small remark (see review). Also, would you mind adding an entry to our changelog (located in the project root)? You could add quick line mentioning your fix under the "GUI and Functionality" section!

chiphogg:
@Barnamoyy, it looks like maybe this PR has started to slip through the cracks.  I think it's very close to being ready to land.  Do you think you'll be able to finish it up any time soon?

chiphogg:
> Thanks for your contribution! Just a small remark (see review). Also, would you mind adding an entry to our changelog (located in the project root)? You could add quick line mentioning your fix under the "GUI and Functionality" section!

@kyaso, the original author appears to have abandoned this PR.  At the very least, we've had radio silence from them for multiple months, despite direct mentions.

This PR purports to fix my number one biggest nuisance regression in Zettlr.  :slightly_smiling_face:  Therefore, I created a new PR, #5444, which includes this PR's commit (so @Barnamoyy will still get credit), but also addresses your feedback so that we can get it landed.

WDYT about reviewing #5444 instead?  I didn't perform any tests, but I'll let you decide which ones you'd need to see to be confident that the change is correct.

nathanlesage:
Sounds good, I've let the unit tests run but it's basically perfect, will close this one and merge the other one as soon as the tests finish. Thanks!

PR #5314, opened 2024-08-05
Update ru-RU.po

ezhkov-ph:
Translated multiple entries for improved localization.


nathanlesage:
Hi, thanks for your help! There is a small error that the linter has found, please fix it, and then we can merge:

`Could not parse file: Error parsing PO data: Invalid key name "%s" at line 1753. This can be caused by an unescaped quote character in a msgid or msgstr value.`

ezhkov-ph:
@nathanlesage I did it 💪

nathanlesage:
Wonderful, thank you so much!

PR #5319, opened 2024-08-06
Update nl-NL.po

candelaere:
"Save changes" and "Discard changes" were translated to be identical, leading to a nonsensical prompt.

afbeelding

## Changes

Only translated two strings, did not properly build and test Zettlr.


nathanlesage:
Thanks a bunch!

PR #5384, opened 2024-09-19
fix: "Dancing" list items by normalizing line indentation

### LINKED ISSUE #4602: dancing list items
### Description

when clicken, marking text, or moving the cursor with the arrow keys and on every keypress, even in other lines,
the lines of the list items start to jump left and right.

see behavior here: https://youtu.be/EFQyN_0LcsI

### Reproducing

watch video above, or make a list like this and enjoy the show:

1. <item>:
	1.1. <item>:
	<text>

2. <item>:
2.1. <item>:
	- <item>
	 - <item>

2.2. <item>:
	- <item>r, 

2.3. <item>:
- <item>
- <item>
- <item>

### Zettlr Version

Stable (most recent version)

### Specify version

3.0.0

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Windows 11 Pro, Version: 22H2, Build: 22621.2215, Windows Feature Experience Pack 1000.22662.1000.0

### Additional Information

_No response_

nathanlesage:
## Description

This PR's aim is to finally solve the various bug reports complaining about "dancing" list items with users who use tabs instead of spaces for their Markdown indentation.

## Changes

The fix is basically to remove any tabs upon file loading and only re-add them upon saving the file. It sounds stupid, but I hope that it works, in which case it ain't stupid.

## Side Effects

There is a side effect that this PR will introduce, namely that indentation will automatically be normalized upon reading and writing any supported file with Zettlr, like line endings.

This means: If you have a mixed indentation in any file (i.e., two spaces and four spaces), the more prevalent indentation pattern will overwrite the lesser used pattern. This means that this PR has the ability to change your files simply by opening and saving them. **IN THEORY**, this should have no impact on people, but I wanted to mention it here.

## Additional information

For months now people have been complaining that their lists were expressing "dancing" behavior. The cause is simple to explain: Zettlr wraps long lines, and without any special treatment, indented lines would look like this:

```
- List item
    - An indented list item, which we will assume to
be wrapped at this point.
```

But we want it to look like this:

```
- List item
    - An indented list item, which we will assume to
      be wrapped at this point.
```

In order to do so, we have to continuously play around with the text-indent and first-line indent to match up the characters on the list item's lines. Since this also must work with variable-width fonts (since not all use one of the monospaced themes), we have to actually measure pixel-precise where these lines are at.

The issue with tab characters now is the way tab characters are supposed to work. Using tabulators assumes that you can take a page and split it up into invisible vertical columns. Inserting a tab character means that any text after it should always be moved to the beginning of the next available vertical column. This means, if you insert or delete some text from before the tab character, this will not change the position of the text following the tab character.

So, in our case: When we measure the positions of the two wrapped lines, the tabs will cause the lines to be at one of these vertical lines. If we now apply our calculations to the line and change the text-indent and first-line indent accordingly, the browser will immediately recalculate where the text should be according to the tab characters in the document. This will move the text, rendering our calculations wrong.

Since we re-calculate these indents on every single keystroke, the lines will move every time. If you observe this behavior, you will notice that it starts to cycle between a few positions back and forth as you type.

As I cannot influence how the browser's layouting algorithm recalculates tab sizes (again, tab characters are NOT a fixed width, they only have a fixed MINIMUM width and are thus variable in width), this PR circumvents this issue entirely by just removing all tab characters before displaying the document.

This is how tabs are now treated:

```
1. Load in a file that uses tab indentation
2. Zettlr replaces them with a space-based indentation
3. You modify the document, indenting and unindenting lines
4. You tell Zettlr to save your file
5. Zettlr replaces the space-based indentation with tabs
6. Write the file to disk.
```

## Things we definitely have to test

Since my time is limited, I decided to provide the basic implementation, but I'll have to leave testing to others:

- [ ] How does navigation work? Right now we probably can navigate the various spaces. Should we not do this and instead make the cursor movement "jump" across indentation units (i.e., four spaces at a time)?
- [ ] Does this thing even work in preserving tabs on disk but not displaying them?
- [ ] One thing I couldn't test is whether this now influences the way equality is being calculated to determine if we need to save a document (newlines are probably already accounted for …? Please check.)
- [ ] Also we definitely have to check what influence the tab size setting now has on the documents. I am considering if we should always use a fixed tab size in the editors of four spaces, and only apply the tab setting to the file saving process…? Comments requested.

Tested on: macOS Sequoia 15.0

Fixes (or should fix) the following issues:

- Fixes #4602

### DIFF
--- a/source/app/service-providers/documents/index.ts
+++ b/source/app/service-providers/documents/index.ts
@@ -36,6 +36,7 @@ import { markdownToAST } from '@common/modules/markdown-utils'
 import isFile from '@common/util/is-file'
 import { trans } from '@common/i18n-main'
 import type FSALWatchdog from '@providers/fsal/fsal-watchdog'
+import { changeFileIndentation } from './util/change-file-indentation'
 
 type DocumentWindows = Record<string, DocumentTree>
 
@@ -579,7 +580,14 @@ export default class DocumentManager extends ProviderContract {
       throw new Error(`Cannot load file ${filePath}`) // TODO: Proper error handling & state recovery!
     }
 
-    const content = await this._app.fsal.loadAnySupportedFile(filePath)
+    // Load in the file. This procedure here will already normalize the line
+    // endings as well as the file indentation.
+    const { editor } = this._app.config.get()
+    const content = changeFileIndentation(
+      await this._app.fsal.loadAnySupportedFile(filePath),
+      editor.indentWithTabs ? '\t' : ' ',
+      editor.indentWithTabs ? 1 : editor.indentUnit
+    )
 
     if (descriptor.type === 'code') {
       switch (descriptor.ext) {
@@ -1433,7 +1441,11 @@ current contents from the editor somewhere else, and restart the application.`
     // and BOMs. So here we will always use newlines. This should fix and in the
     // future prevent bugs like #4959
     const docLines = [...doc.document.iterLines()]
-    const content = docLines.join('\n')
+    const content = changeFileIndentation(
+      docLines.join('\n'),
+      doc.descriptor.indentChar,
+      doc.descriptor.indentSize
+    )
     doc.lastSavedVersion = doc.currentVersion
     doc.lastSavedContent = content
 
--- a/source/app/service-providers/documents/util/change-file-indentation.ts
+++ b/source/app/service-providers/documents/util/change-file-indentation.ts
@@ -35,7 +35,7 @@ export function changeFileIndentation (text: string, newIndent: '\t'|' ', newSiz
   for (let i = 0; i < lines.length; i++) {
     // Take the current indent in multiples of current indentation size in its
     // entirety (leaving out any trailing spaces)...
-    lines[i] = lines[i].replace(new RegExp(`^${currentIndent}{${currentSize}}*`), (match) => {
+    lines[i] = lines[i].replace(new RegExp(`^(?:${currentIndent}{${currentSize}})*`), (match) => {
       // And replace them with the same amount of newIndents times newSize.
       return newIndent.repeat(newSize * Math.floor(match.length / currentSize))
     })
--- a/source/app/service-providers/documents/util/change-file-indentation.ts
+++ b/source/app/service-providers/documents/util/change-file-indentation.ts
@@ -1,0 +1,45 @@
+import { detectFileIndentation } from "../../fsal/util/detect-file-indentation"
+import { extractLinefeed } from "../../fsal/util/extract-linefeed"
+import { normalizeLineEndings } from "../../fsal/util/normalize-line-endings"
+
+/**
+ * This function takes in a file and normalizes the file's line indentation to
+ * whichever is desired. NOTE: This function will implicitly also normalize the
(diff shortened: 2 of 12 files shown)


kyaso:
Hey @nathanlesage, FYI, I have this PR on my radar. Will try to review it over the next few days!

PS: Love the branch name 😆

nathanlesage:
Thanks! As I said, I guess what we need to do is a bit of testing whether this even works ... especially what it does to files with mixed indentation. And then the checks whether a document is actually modified might be off now. So no haste with this one, people had to deal with this for some time already and I'd rather have them deal with it a week more than having files break. I'll see to testing it myself while I'm on the road on Wednesday (but I can't promise it yet)!

And thanks for the branch name shoutout, I was curious if it would even be noticed 😂

nathanlesage:
Additional TODOs:

* Improve the cursor movement ("jump over" tab-sized spans of whitespace at the beginning of lines)
* Make the tab size setting restart (i.e. non dynamic during the app's run) to avoid issues with having to change tab sizes dynamically while several documents are open
* Ensure that the mechanisms work with reading files, writing files, and especially checking if remote changes have actually changed a file

kyaso:
I played a bit with the patch, noticed an interesting behavior. In the following, I provided some steps to reproduce it.

Zettlr config:
- Indent using 4 spaces
- Load remote changes enabled


(1) Outside of Zettlr (I'll use VSCode as an example), create a file with the following content:
```
- A
    - B
```
Make sure that B is indented using 4 _spaces_.

(2) Open the file in Zettlr -> Everything is ok, i.e. the file is correctly loaded and displayed.

(3) In VSCode, add another list item "C", indented using a _tab_:
```
- A
    - B
    - C
```
(note that I used spaces here, just to illustrate)

Save the file, and the change should appear in Zettlr. However, the C item is misaligned and will exhibit the "dancing" issue. (In the current upstream, both B and C will dance)

(3) In VSCode, add another list item "D", now with 4 _spaces_ indent:
```
- A
    - B
    - C
    - D
```

Save the changes, and the change will appear in Zettlr. "C" stays misaligned, but doesn't "dance" anymore.

(4) In VSCode, add another item "E" with 4 _spaces_ indent:
```
- A
    - B
    - C
    - D
    - E
```
Save, and again, the change will appear in Zettlr, "C" stays misaligned, does not dance.

Note, that at this point, there are 3 items with space indent (B, D, and E), and 1 item with tab indent (C).

(5) In VSCode, add another item "F", now with _tab_ indent:
```
- A
    - B
    - C
    - D
    - E
    - F
```
After saving, "F" will appear in Zettlr. Now, "C" is suddenly aligned to the other sub-items, and does not dance.

(6) In Zettlr, without doing any changes to the file, hit the save button. -> _All_ indents will be converted to _tabs_ (verify it using VSCode).

nathanlesage:
I believe this is because I simply forgot the normalization step after reloading changes from disk, you wanna bet? :D

nathanlesage:
Superseded by 04ec811c9

PR #5451, opened 2024-10-31
Update citing.md to clarify instructions

Onatcer:
## Description
I found that after installing Zettlr the Auto-Save Setting is turned off by default, which also means that the References Panel does not update automatically. It looks like it does only update on save, so the instructions for in citing.md did not work for me initially, because the reference of the tutorial did not show up in the sidebar. 

## Changes
I changed

> You will notice that the book is visible there — and as you continue to include references, this list will grow!

to

> After saving the file, you will notice that the book is visible there — and as you continue to include references, this list will grow!

to clarify that you need to save before the References show up. 

Tested on: macOS 14.1.1

### DIFF
--- a/static/tutorial/en/citing.md
+++ b/static/tutorial/en/citing.md
@@ -34,7 +34,7 @@ At the core of Zettlr, there is a mighty engine that is capable of untangling wh
 
 ## The References List 💻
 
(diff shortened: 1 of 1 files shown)


nathanlesage:
Looks good! Indeed, the sidebar lists only update on save since there's quite some work to be done behind the scenes which I didn't want to do on every input. Good catch!

Issue #5163, opened 2024-05-15
Default to Last-Used Export Even When Custom Command


dstark:
### Description

When using a defaults profile to export a document, Zettlr 3.1.1 (on Windows 11) defaults to that profile in the next export. When using a custom command, however, Zettlr defaults to a blank export. If this blank export is run (say, because the user inattentively doesn't select something else), Zettlr replaces the existing markdown file with a new markdown file with many (or all) of the markdown characters escaped. Defaulting to the same custom command would both provide an added convenience and avoid potentially overwriting an existing markdown file with a new file where the markdown is "broken."

![Screenshot 2024-05-15 074053](https://github.com/Zettlr/Zettlr/assets/205528/fd6fe685-6d4a-40f1-97fa-2635785c7956)
![Screenshot 2024-05-15 074133](https://github.com/Zettlr/Zettlr/assets/205528/96921778-cc0f-4102-9ff6-e13bfb219a59)
![Screenshot 2024-05-15 074145](https://github.com/Zettlr/Zettlr/assets/205528/29b9bbbf-f6b1-4ad8-bc3d-8936cffc4da7)

### Proposed Changes

The behavior to default to custom export commands should be the same as the behavior of currently defaulting to exports specified via defaults files.

### Caveats

I don't believe there are any, and there's the additional negative outcome of overwriting an existing markdown file that's avoided. Of course, if one wants to overwrite an existing file, one can create an export to do just that separately.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Operating System Version

Windows 10 and 11

### Additional Information

The existing UI would work great as shown in the screenshots above---provided that Zettlr remembers and defaults to the most recently run export (whether that's based on a defaults file or a custom command).

nathanlesage:
That very much sounds like a bug!

wideaperture:
Just seconding this ticket request. I've been using a custom export command frequently of late and having to re-select it each time — while not difficult — is nonetheless tedious. Thanks for all you do, @nathanlesage!

PR #5480, opened 2024-11-17
fix(exporter): Also consider custom commands when checking last profile

kyaso:
Fixes: #5163


nathanlesage:
I knew it was a very simple to fix bug that I just couldn't wrap my head around … thank you so much for getting to this!

kyaso:
Yeah, it also took me a bit to figure it out :D

PR #5501, opened 2024-11-26
feat: Add ImageViewer and PDFViewer

### LINKED ISSUE #3883: PDF Viewr [sic] Inside Zettlr
### Checklist

### Feature Proposal

Being able to open and view PDFs inside of Zettlr and see PDFs in a workspace (possibly with an PDF icon next to the name) would omit the need to use an external tool to view PDFs.

### Example Workflow

Making PDFs in the workspace visible with a PDF icon next to it.
Clicking on it will open the PDF in the editor space.

In a later version the PDF viewer function could also be used to quickly extract important information from a PDF and put it into a Markdown file, e.g. with copy/paste the text or taking a snippet of the PDF (either a text passage or an image).

### Caveats

Depending on the inner works of Zettlr, displaying a PDF inside the editor could be rather difficult to now implement as an additional feature.

If a suitable tool is available to display PDFs inside of Zettlr, this could mean an additional dependency if the tool is not already used.

No

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

### Additional Information

_No response_

nathanlesage:
## Description

This PR adds two features to Zettlr:

1. It allows displaying of various groups of files (images, PDFs, Office documents, data files) in the file manager in addition to the sidebar
2. It adds an `ImageViewer` which allows users to open images right next to their files, allowing, e.g., referencing plots and figures for papers in the text. It also adds a `PDFViewer` that allows to do the same, but with PDF files, leveraging Chrome's built-in PDF viewer.

## Changes

There is one notable change to the existing behavior: The default attachment extensions of Zettlr are now obsolete, as they have been grouped into one of the new "file groups" this PR introduces, so users can remove all file extensions from that array, as they won't have any effect by now. Users can still add their own file extensions to display in the sidebar as expected.

Most other places have only very minor changes to them, because I tried to make sure that the file manager and related components are already prepared to display other files, too. And, while implementing this PR, it turns out that except some type changes, most files didn't really had to be changed at all, which was nice to see.

## Additional information

* Fixes #3883

Tested on: macOS Sequoia 15.1.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,11 +90,20 @@ annotate your PDF files, please continue using your existing workflow.
 - **Feature**: Full TableEditor Rewrite. The new TableEditor keeps most
   functionality of the previous version, with the exception of more safeguards
   against data loss, and more ergonomic usage.
+- **Feature**: Image and PDF previews. Zettlr has now two dedicated viewers that
+  allow users to open common image types and PDF files right from within the app
+  for preview purposes (#5501).
 - **Feature**: Full-text (aka. global) search runs can now be cancelled via a
   dedicated button. You can now also trigger a new search while another search
   is already running.
 - **Feature**: Individual global search results can now be copied to the
   clipboard (#2070).
+- **Feature**: The file manager can now show and display other file types as
+  opposed to having those only in the sidebar. Images and PDF files can be
+  opened directly in Zettlr, while other files will be opened using the system
+  default application. You can use the appropriate section in the advanced
+  settings to customize this. By default, none of the new file types will be
+  displayed in the file manager (#5501).
 - **Change**: Snippets: The `$FILENAME` variable now does not contain the file
   extension anymore. Users who also want the extension should update their
   snippets to `$FILENAME$EXTENSION` (#4191).


kyaso (commented):
Overall it looks pretty good! I tested the image and PDF viewers, and noticed a few things we may want to consider (not necessarily in this PR, but maybe in the future).

**PDF viewer:**
- Ctrl+F to search is not working
    - Not sure if this is in our control though (as this is probably a limitation of Chromium?)
- PDF position is not remembered on restart of Zettlr
    - A small QOL feature that some users may want
    - Again, may be out of our control

**Image viewer:**
- A button to open the image externally (like we have for images in Markdown files) could be useful
- Allow zooming with scroll-wheel
- Zoom reset button

Maybe @sensologica has some inputs on the GUI aspects?

Further, I noticed the following:

- Right click on image/PDF tabs not working ➜ Is this intended? There could be users who want to pin image/PDF tabs
- ToC showing content of last opened _markdown file_ (applies to both PDFs and image)

**Edit**:
A screenshot of the current state of the image viewer for reference:
image

nathanlesage:
> Overall it looks pretty good! I tested the image and PDF viewers, and noticed a few things we may want to consider (not necessarily in this PR, but maybe in the future).
> 
> **PDF viewer:**
> 
>     * Ctrl+F to search is not working

This is probably my bad, because I may somewhere intercept this event and omit it because it's not a Markdown file. I'll have a look next week. (Feel free to dig before that :D)

>       
>       * Not sure if this is in our control though (as this is probably a limitation of Chromium?)

I think the Chrome PDF viewer should have a search, it would be weird if not.

> 
>     * PDF position is not remembered on restart of Zettlr
>       
>       * A small QOL feature that some users may want
>       * Again, may be out of our control

Yeah, that's pretty difficult. However, one plan I had was switch to pdf.js at some point, because I think that that one is just so much better than that weird Chrome thing. But it takes quite a ton of work to implement, so for now I just used the built-in Chrome PDF viewer. At least that one works.

> 
> 
> **Image viewer:**
> 
>     * A button to open the image externally (like we have for images in Markdown files) could be useful

Done

> 
>     * Allow zooming with scroll-wheel

Done. In addition, I also implemented drag-to-move (i.e., if the container overflows, dragging with your mouse makes the image move)

> 
>     * Zoom reset button

Missing a bit a good idea on how to do that (is it even necessary now that interaction is much simpler…?)

> 
> 
> Maybe @sensologica has some inputs on the GUI aspects?
> 
> Further, I noticed the following:
> 
>     * Right click on image/PDF tabs not working ➜ Is this intended? There could be users who want to pin image/PDF tabs

My bad, works now

> 
>     * ToC showing content of last opened _markdown file_ (applies to both PDFs and image)

This is something I would like some brainstorming for. Currently, PDF and Image files don't update the lastLeafId. They probably should, I guess…? Would you mind having a look if that would break anything (such as ToC generator)…? I imagine we probably only have to implement a few sanity checks that the current leaf's active file is a Markdown file instead of anything else.

kyaso:
> Missing a bit a good idea on how to do that (is it even necessary now that interaction is much simpler…?)

Isn't that simply setting `zoomLevel` back to 100, or am I missing something here?

kyaso:
> This is something I would like some brainstorming for. Currently, PDF and Image files don't update the lastLeafId. They probably should, I guess…? Would you mind having a look if that would break anything (such as ToC generator)…? I imagine we probably only have to implement a few sanity checks that the current leaf's active file is a Markdown file instead of anything else.

Hmmm, I just thought about something. Maybe we want to keep the current behavior, actually. Say, a user wants to view a Markdown and a PDF side by side in split view, but still wants to see the ToC of the Markdown file, even after clicking into the PDF document. The current behavior would allow that.

What do you think?

nathanlesage:
> Isn't that simply setting zoomLevel back to 100, or am I missing something here?

Exactly, but now that the user has so much more fine grained control the question is they even need that -- primarily prompted by the fact that I haven't found, like, a good place for such a button :D sounds like a job for @sensologica !

Also regarding your second comment, because what you describe was I think my unspoken assumption -- that, since images can't have a ToC, we can just leave the last opened file's ToC there. In this case, a very easy solution might be to just leave it like that and then see if users find this confusing? @professed What do you think, from a researcher's perspective?

sensologica:
Before you merge, let's please discuss this at the next meeting. I have some thoughts.

sensologica:
Here's my early proposal. Let's have a discussion.

zettlr_image-and-pdf-viewer_current

zettlr_image-and-pdf-viewer_proposal

This redesign eliminates unneeded options and the dropdown, thus putting most operations just 1 click away. (For now, don't focus on the visuals, this is just a wireframe so we can figure out the best layout and interactions first.)

Elements:

1. These buttons should operate in increments of 25% and lock to multiples of 25. So, in this example, clicking `-` should jump to 100%, and clicking `+` should jump to 125%.
2. The value here should reflect the current scale of the image and be rounded down/up to an integer for better legibility. This is also an input, allowing users to click into it and type in values.
3. Newly loaded images should always fit to screen by default. This option can also be thought of as a zoom "reset".
4. On horizontal (width > height) images, clicking this should increase the image height to fill the available space. On vertical (width < height) images, clicking this should increase the image width to fill the available space. Therefore, this one option eliminates the need for separate "Fit width" and "Fit height" controls.
5. This is for quickly zooming in to 100% and seeing the image at its maximum resolution.
6. The learnability for these controls is quick, so I think it's fine to leave them as they are. Although, I don't like that they look so cryptic as that causes some initial confusion. Maybe there is a better solution, I'll have to think about it.
7. Perhaps "Open in system viewer" is the more idiomatic name for this functionality that "Open externally"?

Other comments: 

* On a trackpad, pinching in zooms out and pinching out zooms in, which is counterintuitive and needs to be reversed.

nathanlesage:
Sounds good!

sensologica:
zettlr_image-and-pdf-viewer_proposal-a2_light-theme

zettlr_image-and-pdf-viewer_proposal-a2_dark-theme

Here is a more refined version.

1. The `-`/`+` buttons were replaced with icons for added clarity.
2. The wording was changed from "Fit screen" to "Fit to screen" for added clarity.
3. The elements are now not confined to boxes to make the toolbar feel more spacious and to improve legibility. On hover, an element's containing box becomes visible to suggest that the element is interactive.
4. The background-chooser icons were rounded to reinforce that they are also interactive.

At this point these look and feel good to me. If you don't have any comments, see the Figma file for implementation.

nathanlesage:
I've now improved the PR a bit again — however, I fear that most of the UI cannot really be properly implemented while we're still on the (at this point very old) macOS native style, or in general. I think it might actually be worthwhile to implement this without a perfect UI right now, but get to working on a new UI framework for Zettlr that departs again from the native ideal. At this point I think it should be a nice improvement for all platforms. Maybe after the table editor?

nathanlesage:
I've now cleaned the PR up and improved it here and there. I think in terms of the final GUI we should move that to a general UI overhaul, but functionality wise, this seems fine so far. Any other changes we can still implement where necessary.

I'll keep this a bit open now, but I plan to merge once we're in beta for the table editor.

nathanlesage:
Letting the checks finish, and then we're ready to roll with this thing.

PR #5524, opened 2024-12-19
Update Ukrainian translations

Fedik:
## Description

Update translation for Ukrainian language.

Side question: Is there a reason why the language code is `uk-UK`, instead of `uk-UA`?
I mean, in such format it is usualy `<Language Code>-<Country Code>`. And `uk-UK` it is like Ukrainian in United Kingdom :wink:  

## Additional information


nathanlesage:
> Side question: Is there a reason why the language code is uk-UK, instead of uk-UA?

That is a very good question, which I can't answer :D I just added a bunch of BCP-47 tags, and I am very certain that at some point my brain just stopped working properly. I'll fix that after merging your PR, thanks for letting me know!

And, of course, thanks for the translation work ;)

PR #5560, opened 2025-01-11
fix: FSAL Cache Issue

### LINKED ISSUE #5552: FSAL Cache Clearing Doesn't Work
### Description

I had a hunch already, but issue #5551 has confirmed that the menu item to clear the cache doesn't work as advertised.

### Reproducing

Use the menu item to clear the cache, and observe that it doesn't actually remove the cached files from the app data directory.

### Zettlr version

3.4.1

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [X] ARM 64bit (includes Apple Silicon)

### Operating System Version

Definitely macOS, I don't know about other OSs

### Additional Information

_No response_

nathanlesage:
## Description

This PR fixes the FSAL Caching procedure, which only happened partially before. This was due to the fact that (a) the FSAL provider has been booted very late in the chain, and (b) the asynchronous unlinking was happening while the rest of the party was continuing to boot.

Specifically, this led to the following chain of events:

* The user passes the `--clear-cache` command to the application (usually using the "Help" menu and restarting) or updates to a new version
* The FSAL begins its boot process
* The FSAL detects that it should clear its cache, and calls the clear method of the Cache
* The Cache immediately clears its in-memory cache and lists all cache files for deletion
* It starts calling unlinks on the various files, but using `fs.unlink`-callbacks and collects them in promises. However, it does not await the promises, and instead returns immediately.
* The FSAL boot process concludes.
* Other providers that need to access files start to boot
* Some cache files are deleted at this point
* Other cache files are being read because they are requested
* Those files are then being deleted
* (…and so on)

You see where the issue lies: Before the cache could be cleared in its entirety, other providers will request files, causing not-yet-deleted cache files to be read, meaning that parts of the cache are effectively not cleared.

## Changes

1. Fix the boot order. I am not 100% sure if this was completely necessary, but it made a lot of sense. The new boot order seems to work appropriately.
2. Promisify the cache clearing procedure to ensure that, once the FSAL boot function returns, the cache is guaranteed to be cleared.

## Additional information

* Fixes #5552

@kyaso It would be great to have someone else with a different setup double check this, but I assume it should just work. Nevertheless, I wanted to start pushing *every* fix of mine as a PR rather than messing with the develop branch directly.

Tested on: macOS Sequoia 15.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,14 @@
 
 ## GUI and Functionality
 
-(nothing here)
+- Fixed an issue that prevented the FSAL cache clearing from proceeding
+  appropriately.
 
 ## Under the Hood
 
 - Fixed the boot order of providers to ensure certain actions are taken before
   providers access each others (primary case: the FSAL needs to be booted asap)
+- Promisify the cache clearing procedure
 
 # 3.4.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ## Under the Hood
 
-(nothing here)
+- Fixed the boot order of providers to ensure certain actions are taken before
+  providers access each others (primary case: the FSAL needs to be booted asap)
 
 # 3.4.1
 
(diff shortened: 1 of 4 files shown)


kyaso (approved):
LGTM!

PR #5607, opened 2025-02-05
feat: Switch Windows Code Signing to Azure Trusted Signing

nathanlesage:
## Description

This PR switches the build pipeline from the expiring SSL.com code signing certificate to the new flow using the Azure Trusted Signing pipeline, using our Electron Builder CI workflow.

### Roadmap

In order to reduce the amount of potential nuisance for Windows users, the suggested roadmap for the switch will be as follows:

#### Step 1: Maintenance Release at the end of February

At the end of February, we should release a maintenance update for Zettlr including the most recent Electron binary to ensure it is as safe as possible, and prompt users to update. This will be the final release using the expiring certificate.

#### Step 2: Initial Release w/ New Certificate

Then, at the beginning of March, I want to release a new update that changes nothing except the certificate, and again prompt users to update. In case the new certificate has only basic initial trust, this means that users using a managed computer won't be able to update (one needs Admin access to override the Windows warning), so we have to rely on those users to "build up trust" with Microsoft.

#### Step 3: Resume Usual Workflow

Once the code signing works, and has been tested in the wild, resume the usual flow of updating the app once a few changes have built up.

> [!WARNING]
> This PR is a draft since it should only be merged **after the maintenance update** to reduce the potential amount of disruption. While my testing on a Windows 11 VM did not at all show the prompt that the binary is not trusted, I won't fully trust this, so I want to pull through with the change as initially planned.

## Changes

* Update to Electron Builder 26.0.1 to be able to use the new Azure Sign Options property
* Exchange the old `CSC` environment variables with new ones reflecting the usage of Azure
* Adapt a few breaking changes in the Electron Builder API

## Additional information

Some TODOs:

- [ ] @kyaso I would like you to test out the Linux build. Electron Builder introduced a change that can be remedied by moving the `StartupWMClass` Desktop file entry as indicated into the `entry` property, but I have not been able to check if that will still work. It would be great if you could check and see if anything broke on the Linux Debian installer.
- [x] @nathanlesage Add the corresponding authentication variables as repository secrets ahead of the switch
- [ ] @nathanlesage Test the macOS installers produced by this build.

Tested on: Windows 11 (non-activated VM); Access for the accounts tested via Azure CLI


kyaso:
> I would like you to test out the Linux build. Electron Builder introduced a change that can be remedied by moving the StartupWMClass Desktop file entry as indicated into the entry property, but I have not been able to check if that will still work. It would be great if you could check and see if anything broke on the Linux Debian installer.

Roger! I will do it once I have access to my linux machine again (from Feb. 17th).

nathanlesage:
Awesome, thanks!

kyaso:
I just downloaded the latest nightly from yesterday, and was able to install the deb on my Ubuntu 24.04 machine!

nathanlesage:
Very good, thanks! :)

PR #5631, opened 2025-02-17
UI text cleanup

sensologica:
## Description
* Improved consistency and clarity of UI text across various parts of the app.
* Fixed two small bugs.

## Changes
See commit messages.

## Additional information

Tested on: macOS Sequoia, Version 15.3.1


source/app/service-providers/menu/menu.win32.ts
```diff
@@ -439,15 +439,19 @@ export default function getMenu (
         },
         {
           id: 'menu.toggle_distraction_free',
-          label: trans('Distraction free mode'),
+          label: trans('Distraction-free mode'),
+          type: 'checkbox',
+          checked: false,
```
nathanlesage, line 444: These lines are new. I think I removed them some time ago, because it was difficult to sync the main window's state back to the main process and align the checkbox with the actual state in the window.

sensologica: There does seem to be some kind of conflict that causes the checkboxes to be wrong, depending on the sequence in which one turn on the Distraction-free and Typewriter modes. It's weird, sometimes I test it and everything works fine, and other times it doesn't. Do you know how to solve this issue? Or do we ignore this problem for the time being?

source/win-main/file-manager/util/file-item-context.ts
```diff
@@ -48,15 +48,15 @@ export function displayFileContext (event: MouseEvent, fileObject: MDFileDescrip
...
+      id: 'menu.duplicate_file',
       type: 'normal',
       enabled: true
     },
     {
-      label: trans('Duplicate file'),
-      id: 'menu.duplicate_file',
+      label: trans('Delete file'),
+      id: 'menu.delete_file',
+      accelerator: 'CmdOrCtrl+Backspace',
```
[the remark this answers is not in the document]
sensologica: In this file I just repositioned Duplicate and Delete, I didn't add any new lines. Do you want me to delete the accelerator line?

source/app/service-providers/menu/menu.win32.ts
nathanlesage: Ignore for now. I have a hunch how we can do it, but this requires deeper changes. The phenomenon you observe is that the checkboxes toggle whenever you click them, but if you use the keyboard shortcut, the menu items don't get notice of that, so they keep their old state.

Issue #5636, opened 2025-02-22
Not able to select document from pop-up when inserting internal links typing [[


BabakFarshchian:
### Description

Since the last update:
- When I write [[ a pop-up list of titles comes up.
- When I browse and hit enter the title is not inserted but I just get a new line in my document.

### Reproducing

- In an open document write [[.
- Choose a title you want to insert.
- Hit enter.

### Zettlr version

4.3.2

### Installation Method

Other (please specify below)

### Installation Method (Other)

Automatic update in Zettlr

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Windows 11

### Additional Information

There is also a related but that sometimes shows up: Writing [[ does not bring the pop-up menu up. I will see how to reproduce it and possibly make a separate issue if it does not exist.

BabakFarshchian:
I noticed that you can still select an item from the list through mouse click. It is only the enter key that does not work anymore.

nathanlesage:
The Tab key should also work -- similar to the other issue you've linked.

@sensologica It gets better, now you seem to be able to close two issues with the same PR 😎

Issue #5637, opened 2025-02-24
Citation auto-complete does not work in list items.


wideaperture:
### Description

When writing a list item, I can type `@` and filter my list of citations, per usual. However, when I hit `Enter`, instead of auto-completing the highlighted citation from the list, Zettlr instead creates the next list item.

For example, say I'm writing a list item and want to enter the citation `@westBuyNowHow2022`. In the example below, I begin typing the citation and, at the point that I've written `@westBuy`, the correct citation is highlighted in the auto-complete list…

```
- I am typing a list item and begin to enter this citation: @westBuy
```

…so I hit `Enter` and this is the result:

```
- I am typing a list item and begin to enter this citation: @westBuy
- 
```

…versus the expected result, which would be: 

```
- I am typing a list item and begin to enter this citation: @westBuyNowHow2022
```

…or, rendered:

```
- I am typing a list item and begin to enter this citation: West (2022)
```

### Reproducing

See above.

### Zettlr version

3.4.2-nightly from 2025-02-24

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Manjaro (Rolling)

### Additional Information

I use the appimage.

sensologica:
Sounds good, I'm on it 👍

wideaperture:
Just a quick addendum to say that this problem also occurs in block quotes. I.e.,

```
> I have the same issue when auto-completing a citation within a block quote. [@westBuy
>
```

Probably the same fix will extend to both issues, but wanted to make a note just in case.

Mela:
Issue appears for me not only when writing list items but with every citation.

nathanlesage:
… that took very long, apologies, but I finally had the head space to look over @sensologica's PR and I tested it, it seems to work now! So, Tab always worked, Enter now finally works, too.

PR #5646, opened 2025-02-27
Fix keymap conflicts that caused issues #5636 and #5637

sensologica:
## Description
Fix issues #5636 and #5637 where conflicting keymaps would cause pressing 'Enter' to jump to a new line instead of applying autocomplete suggestions.

## Changes
The fix involved configuring the @codemirror/lang-markdown package to NOT install its Markdown keymap because this keymap took precedence over Zettlr's custom keymaps causing undesired behavior.

Tested on: macOS Sequoia 15.3.1, Apple Silicon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Upcoming
 
+- Fix issues #5636 and #5637 where conflicting keymaps would cause pressing
+  'Enter' to jump to a new line instead of applying autocomplete suggestions. 
+
 ## GUI and Functionality
 
 - **Breaking Change**: Switched Windows Code Signing Certificate to the Azure
--- a/source/common/modules/markdown-editor/autocomplete/index.ts
+++ b/source/common/modules/markdown-editor/autocomplete/index.ts
@@ -19,11 +19,7 @@ import {
   type CompletionSource,
   type CompletionResult,
   autocompletion,
-  type CompletionContext,
-  startCompletion,
-  acceptCompletion,
-  closeCompletion,
-  moveCompletionSelection
+  type CompletionContext
 } from '@codemirror/autocomplete'
 import { type StateField } from '@codemirror/state'
 import { codeBlocks } from './code-blocks'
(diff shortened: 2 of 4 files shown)


sensologica:
@nathanlesage, I've just realized something. My fix disables two keymaps that were originally attached by the @codemirror/lang-markdown package: 

* Enter -> `insertNewlineContinueMarkup`
* Backspace -> `deleteMarkupBackward`

While we do NOT want the first command to run because we prefer to handle it ourselves, we actually don't have ANY overriding behavior for the second command and therefore the behavior of Backspace is slightly different in certain small scenarios as explained [here](https://github.com/codemirror/lang-markdown?tab=readme-ov-file#user-content-markdownkeymap).

I think that I actually prefer this new behavior but I can't say if that's the case for everybody else. Could you please take a look and let me know your thoughts? If necessary, I can import and place this command in our custom keymaps to restore the original behavior. Thank you!

nathanlesage:
The change you made is perfectly fine. But please add ALL commands from that keymap into our own one. This means to basically copy and paste the keymap verbatim, but move them so that the autocomplete Enter is ABOVE the newlineAndContinueMarkup command. (This way, the autocomplete command will first see "Is there an autocompletion to accept?" and if no, it will return false, prompting codemirror to try the next command bound to the Enter key, which is newlineAndContinueMarkup. The precedence is top-to-bottom, which is why I have added more specialized commands further to the top, and more general commands towards the bottom of the list.)

The rational I had for that is the following: I basically removed all keymaps, and instead manually added all commands and mapped them in the various keymap files I created in the CodeMirror directory. This way we can do three things:

1. Keep the functionality as intended by the provided keymaps, but locate them centrally in one place so that we can fix issues like the ones you are fixing here much more easily because we control the keymaps entirely.
2. Remap them where necessary (this is how I fixed the issues with the Italian deadkey backtick and the Swedish Å on Mac keyboard layouts).
3. (In the future): Allow users to assign custom shortcuts to each command.

I do see your suggestion, though, and I'm happy to discuss this. In that case, just so that we can keep track, please open an issue, describe what the command is doing, how removing it would make writing easier, and then we can also collect some user feedback.

I hope that helps :)

sensologica:
> I basically removed all keymaps, and instead manually added all commands and mapped them in the various keymap files I created in the CodeMirror directory.

Ahh, I was wondering how come most of the default keymaps were being manually reintroduced. This explains it! 👍

Issue #5571, opened 2025-01-17
CTRL+PagUp/PgDown to navigate among open md files


HerrAugust:
### Description

Hi,

Would it be possible to add the shortcuts CTRL+PagUp/PgDown to navigate among open md files? Like in Visual Studio Code. 

Thank you! 

### Proposed Changes

See description. 

### Caveats

None.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

None

### Your Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

HerrAugust:
I have implemented this feature myself.

nathanlesage:
Will be fixed in #6411.

PR #5657, opened 2025-03-08
[#5571] Custom shortcut to switch tab

HerrAugust:
## Description
Fix issue #5571 . Users can now select their favourite shortcut to open the next and previous active tab.

## Changes
GUI setting was update to allow selection of shortcut and 2 menu entries were updated to reflect the preference.
CTRL+PageUp/Down were chosen as in Visual Studio Code.

## Test
1. Open 3 documents
2. Open Preferences -> Editor, scroll to the end
3. Change the "Select next tab with" or "Select previous tab with" preferences
4. Close the Preferences window
5. Press the selected shortcut

Expected behavior: the next/previous tab is shown.

Tested on: Ubuntu 22.04 LTS


nathanlesage:
First of all, I appreciate the work that has flown into this. The code looks great and I don't see any major points I would disagree with. However, I think it is merited to think about this PR in the context of the demand for keyboard shortcut customization that has been open for quite some while. While this hasn't been realized yet, I do think that that would make implementing this feature easier, because I don't think adding a preference for a single shortcut, if we could also allow users to customize a broad range of keyboard shortcuts. CC @sensologica for a UX perspective

HerrAugust:
Hi, it is always a pleasure to enhance such useful and well maintained projects! If I understand that correctly, there is a broader feature to customise shortcuts, but it is not planned and probably will take a considerably long time to be merged. I would instead propose to take this MR for now, and then when and if the broader one will be merged, these few changes of this MR can be deleted and the new nicer version used. What do you people think?

BTW I will try to fix the issues of the pipeline during this weekend :)

nathanlesage:
Cc @sensologica @kyaso your take?

I would say we should stick with ONE shortcut and not add this way of allowing users to choose between two alternatives, because it will be an absolute pain to find and remove those settings again if we finally implement custom shortcuts. So I'd say we can add a shortcut for that, but I would remove the entire preferences code because that will become a PITA in the not so distant future.

HerrAugust:
Open PR https://github.com/Zettlr/Zettlr/pull/5715

PR #5470, opened 2024-11-08
adjust the clickAndSelect function

pradnyaakumbhar:
Fixes #5192
## Description
Update the clickAndSelect function to take into account multi-line elements and ensure correct position calculations.

## Changes
updated ClickAndSelect function in /markdown-editor//renderers/click-and-select.ts

    const rects = target.getClientRects();
    let fromPos = null;
    let toPos = null;

    //iterate over each line and find the first and last position
    for (const rect of rects) {
      const startPos = view.posAtCoords({ x: rect.left, y: rect.top });
      const endPos = view.posAtCoords({ x: rect.right, y: rect.bottom });

      if (startPos !== null && (fromPos === null || startPos < fromPos)) {
        fromPos = startPos;
      }
      if (endPos !== null && (toPos === null || endPos > toPos)) {
        toPos = endPos;
      }
    }

    // const { top, left, bottom, right } = target.getBoundingClientRect()
    // const fromPos = view.posAtCoords({ x: left, y: top })
    // const toPos = view.posAtCoords({ x: right, y: bottom })

PR #5682, opened 2025-03-19
fix: Tweak citation selection to be less greedy

### LINKED ISSUE #5192: Selecting a rendered citation works incorrectly when there is a line-wrap in the middle of the citation.
### Description

Typically, when you click on a rendered citation Zettlr reverts to the un-rendered view of the citation and highlights its contents. This is the expected behavior.

However, when a (soft) line wrap occurs in the middle of a rendered citation, clicking on it highlights not just the un-rendered citation text, but the text of the entire preceding paragraph. This is _not_ expected behavior.

See the following video for a demonstration of the bug:

https://github.com/Zettlr/Zettlr/assets/3809531/e81d0890-e3ed-4ee9-bf97-a8ed02a2b6f4

### Reproducing

Open a markdown document in Zettlr that includes a rendered in-text citation. Resize the window until a line wrap occurs in the middle of the citation text. Finally, click on the rendered citation.

### Zettlr Version

Nightly

### Specify version

3.1.1-nightly-20240603

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit

### Operating System Version

Manjaro (Rolling)

### Additional Information

I use the AppImage.

richdouglasevans:
Fixes #5192

## Description

When clicking on a thing -- such as a citation -- that has a (soft) line-wrap in the middle of the rendered text the current (unwanted) behaviour is that the entirety of all lines that the clicked-on thing spans are selected.

This fixes the behaviour so that *only* the clicked-on thing is selected, even when the clicked-on thing spans multiple lines due to a (soft) line-wrap.

## Before

Notice how text outside the citation is highlighted.

https://github.com/user-attachments/assets/6d77d04e-3601-4854-af6b-7da34a2f51ab

## After

Notice how only the text of the citation is highlighted.

https://github.com/user-attachments/assets/7104a5cb-373b-4008-a795-16c0db57c77f

## Additional information

The fix absolutely 100% belongs to @pradnyaakumbhar: I've just picked up their work from #5470 because it's been languishing in PR limbo. Solely in the interest of getting this over the line -- because it's been mildly annoying me when using Zettlr for my own work -- I've addressed the linting issues and tested it locally. Hopefully this is enought to get it over the line, all hail Pradnya 🙇‍♂️

> I'm not trying to jump Pradnya's train. I'm looking to contribute more to Zettlr and this was a `good first issue` that looked to be stalled.

Tested on: Windows 11 Home

### DIFF
--- a/source/common/modules/markdown-editor/renderers/click-and-select.ts
+++ b/source/common/modules/markdown-editor/renderers/click-and-select.ts
@@ -31,9 +31,34 @@ export default function clickAndSelect (view: EditorView): (event: MouseEvent) =
       return
     }
 
-    const { top, left, bottom, right } = target.getBoundingClientRect()
-    const fromPos = view.posAtCoords({ x: left, y: top })
-    const toPos = view.posAtCoords({ x: right, y: bottom })
+    // The thing we're clicking on may span multiple lines in the editor; for
+    // example, clicking on a citation that has a (soft) line-wrap in the middle
+    // of the rendered text.
+    //
+    // Grabbing the `getBoundingClientRect` is too coarse in such cases because
+    // the selection will span the *entirety* of both lines rather than just the
+    // text of the rendered citation (in this example).
+    //
+    // So, we need to find the *smallest* rectangle that applies in the list of
+    // all possible rectangles for the target; continuing with the citation
+    // example, clicking on a citation that has a (soft) line-wrap in the middle
+    // of the rendered text *must* only select (highlight) the rectangle of the
+    // citation text itself.
+    const rects = target.getClientRects()
+    let fromPos = null
+    let toPos = null
+
+    for (const rect of rects) {
+      const startPos = view.posAtCoords({ x: rect.left, y: rect.top })
+      const endPos = view.posAtCoords({ x: rect.right, y: rect.bottom })
+
+      if (startPos !== null && (fromPos === null || startPos < fromPos)) {
+        fromPos = startPos
+      }
+      if (endPos !== null && (toPos === null || endPos > toPos)) {
+        toPos = endPos
+      }
+    }
 
     if (fromPos === null || toPos === null) {
       return


nathanlesage:
Thanks, I just took a look as to why that PR was stuck in limbo, because I was for a moment very afraid of just having forgotten it 😂 🙈 

Anyways, in principle this looks good, although I'm currently thinking that it might be better if there was a way to just directly pass the `from` and `to`-positions to the function instead of this arguably hacky solution. Because then we'd have the already pre-computed from and to positions and didn't have to touch the DOM at all. I believe the way this function works is still a remnant from the era of CodeMirror 5, where all of this was not as easy as with CodeMirror 6.

source/common/modules/markdown-editor/renderers/click-and-select.ts
```diff
@@ -31,9 +31,34 @@ export default function clickAndSelect (view: EditorView): (event: MouseEvent) =
...
+    // So, we need to find the *smallest* rectangle that applies in the list of
+    // all possible rectangles for the target; continuing with the citation
+    // example, clicking on a citation that has a (soft) line-wrap in the middle
+    // of the rendered text *must* only select (highlight) the rectangle of the
+    // citation text itself.
+    const rects = target.getClientRects()
+    let fromPos = null
+    let toPos = null
+
+    for (const rect of rects) {
```
nathanlesage: Given how multiline spans work … wouldn't it be easier to check for the length of the client rects? And, if it's only one, set from and to as I did before, and, if it's at least two, take the first and last for the coordinates…?

nathanlesage (commented):
I think this one seems almost ready to merge, too — I just had one question!

source/common/modules/markdown-editor/renderers/click-and-select.ts
richdouglasevans: I'll check that now and report back today 👍

richdouglasevans: That seems to work; mind you, I've only tested it per the video below.

Here's the commit:

* 979b2b2c0954745630f6aa4afd29b87d6511e826

And here's a video demonstrating it in action.

https://github.com/user-attachments/assets/1a408985-2496-4b3f-858d-756ad563cc34

nathanlesage:
Smart solution! Let's merge this, it looks good! Thank you again!

PR #5685, opened 2025-03-20
fix: Spawn shell when starting test GUI on Windows

richdouglasevans:
## Description

When running the test GUI in Windows I encounter the following error:

```shell
$ yarn start
Starting app with old test environment ...
Starting Zettlr with custom configuration ...
node:internal/child_process:420
    throw new ErrnoException(err, 'spawn');
          ^

Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at spawn (node:child_process:762:9)
    at startApp (file:///C:/Users/44750/projects/zettlr/scripts/test-gui/index.mjs:142:16)
    at file:///C:/Users/44750/projects/zettlr/scripts/test-gui/index.mjs:53:3
    at ModuleJob.run (node:internal/modules/esm/module_job:263:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:540:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5) {
  errno: -4071,
  code: 'EINVAL',
  syscall: 'spawn'
}

Node.js v20.19.0
```

> Note that this error happens whether the test environment is new or old.

## Changes

If I change the `spawn` call so that it uses a shell -- [options](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options) -- when running on Windows then the test GUI runs just fine with no error. (Or at least, not that error.)

There are other places in the codebase that also explicitly request a shell when spawning: they _always_ request a shell rather than just when running on Windows. `{ shell: true }`

Thoughts? I need this fix in place to be able to run the test GUI on Windows. I can, of course, use WSL but hey, I thought I'd raise this speculatively and see what rolls out.

Tested on: Windows 11 Home, running Git-Bash.

### DIFF
--- a/scripts/test-gui/index.mjs
+++ b/scripts/test-gui/index.mjs
@@ -122,6 +122,7 @@ function startApp (argv = []) {
   const forgeArgs = [ 'start', '--', `--data-dir="${CONF_DIRECTORY}"`, ...argv ]
   // Spawn's options: Use the root as CWD and pipe the process's stdio to the parent process.
   const spawnOptions = {
+    shell: process.platform === "win32",
     cwd: path.join(__dirname, '../../'),
     stdio: [ process.stdin, process.stdout, process.stderr ]
   }


scripts/test-gui/index.mjs
```diff
@@ -122,6 +122,7 @@ function startApp (argv = []) {
   const forgeArgs = [ 'start', '--', `--data-dir="${CONF_DIRECTORY}"`, ...argv ]
   // Spawn's options: Use the root as CWD and pipe the process's stdio to the parent process.
   const spawnOptions = {
+    shell: process.platform === "win32",
```
richdouglasevans, line 125: This could also be hardocoded to `true` rather than just for Windows.

nathanlesage:
Interesting find — the other place you've likely found is the importer/exporter. I have noticed that if we run Pandoc without the `shell` argument, it doesn't work properly, that's why it uses the `shell` parameter. I didn't just add it to all `spawn` calls primarily because enabling shell means that there are some security implications since then we'd have variable expansion and stuff. Maybe the reason for why this is an option rather than just hard-coded in Node.js is that one can also just directly call a program; but if that needs to access paths we need the shell.

[This is the relevant documentation on Node.js](https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows). If you have some time feel free to try to understand the intricacies of providing `shell: true`. But I think, since this is an issue in the spawner for the test environment, we should probably be safe in just hard-coding a shell in there. Especially if that then solves the problem of launching on Windows!

@kyaso Do you have any thoughts on this issue?

scripts/test-gui/index.mjs
nathanlesage: Adding to the comment I just made on the PR itself: I can see in the preview that we pass the `data-dir` argument. It can be that it works on UNIX this way because we use an entirely different spawn call, but I may be wrong. Anyways, as I outlined, I don't think there are any insurmountable security implications with just providing a hard-coded `true` here!

Issue #5659, opened 2025-03-10
Automate common text transforms.


wideaperture:
### Description

There are a number of text transformations that amount to repetitive tasks commonly done by authors. To give one example, if I'm quoting a brief passage that contains quotation marks…

> During this time the  meanings of "liveness" and "presence" were indistinguishable from Western discourses of modernization.

…I'll need to convert the double-quotes used in the quotation to single-quotes in my paper, like so:

> Lisa Parks argues that "during this time the  meanings of 'liveness' and 'presence' were indistinguishable from Western discourses of modernization."

This sort of thing can be done by hand or with find-replace (if find-in-region were a thing in Zettlr; see #4040). But it's also the sort of yak shaving/repetitive task that cries out for automation, particularly when using a tool like Zettlr that contains other features of advanced text editors.

### Proposed Changes

Many programming text editors contain a menu or sub-menu from which users can select text transformations that are common in programming, such as changing tabs to spaces, toggling upper and lower case characters, straightening quotes, etc. For instance, BBEdit has a killer `Text` menu:

![Image](https://github.com/user-attachments/assets/caa059ea-7846-462f-8763-fa0e57431161)

I'm suggesting that Zettlr add a similar menu or sub-menu populated with text transformations that are common in academic or journalistic writing. These could include:

- Double quotes to single quotes. (See above.)
- Single quotes to double quotes.
- Italics to quotes (e.g., for changing `_Cultures in Orbit_` to `"Cultures in Orbit"` to satisfy a different style guide.)
- Quotes to italics.
- Join lines (a.k.a. unfill-region, see #4039)
- Sentence case to title case.
- Title case to sentence case.

…and possibly others.

Possible locations for these menu items could include the edit menu (or a sub-menu under the edit menu), the right-click context menu when text is selected in the editor, or a top-of-screen toolbar button. Some of the most commonly used transforms might also be given their own keyboard shortcuts.

### Caveats

None.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Nightly

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Manjaro (Rolling)

wideaperture:
I keep finding more transforms that would be useful.

- Text cut-and-pasted from PDFs sometimes has extra spaces in weird places, so "strip duplicate spaces" could be useful.
[...]

PR #5701, opened 2025-03-30
feat: Add transform to strip duplicate spaces

### LINKED ISSUE #5679: Text transform: Strip duplicate spaces.
### Description

This is a sub-issue of #5659.

When inserting quotations from a cited article, users will often need to cut and paste text from PDFs and other documents. Pasted text often needs to be reformatted in the process, which can be time consuming to do by hand. This feature automates one of these reformatting tasks.

Specifically, I propose a CodeMirror command that strips duplicate spaces from a selected region of text. In a not-inconsequential number of PDFs, line breaks in the original text shop up in spaces when copied and pasted into the editor, resulting in text that has extra spaces mid-sentence. So, a conversion that stripped out extra spaces — e.g., converting `technical things  have political qualities` to `technical things have political qualities` — would be quite useful.

This might also come in handy when cleaning text shared by the typewriter generation, who sometimes insert double-spaces between sentences, where only one is now needed.

### Proposed Changes

See #5659.

### Caveats

None.

No

### Zettlr Version

Nightly

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Manjaro (Rolling)

### Additional Information

_No response_

richdouglasevans:
Closes #5679

This adds one of the text transforms -- strip duplicate spaces -- from issue #5659.

I'm using this PR to both flush out the approach and learn any things about contributing to this codebase. If all goes well I'll implement the other text transforms secure in the knowledge that the approach has been signed-off by Hendrik (and others) and any "dumb" first-time contributor stuff is sorted.

## Description

This adds a new _Transform_ submenu to the editor which has exactly one transform in it (for now): strip duplicate spaces.

In the video I am:

* selecting text that I want to strip duplicate spaces from
* clicking the command from the aforementioned submenu

After running the command the duplicate spaces are stripped: that is, collapsed to a single space.

https://github.com/user-attachments/assets/4ac026f3-52f4-44cc-b608-ebb48bfaa3cc

## Additional information

Tested on: Windows 11 Home.

### DIFF
--- a/source/common/modules/markdown-editor/commands/transforms/transform-selected-text.ts
+++ b/source/common/modules/markdown-editor/commands/transforms/transform-selected-text.ts
@@ -1,0 +1,72 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        transformSelectedText function
+ * CVM-Role:        Utility function
+ * Maintainers:     Rich Douglas
+ * License:         GNU GPL v3
+ *
+ * Description:     A builder of `Commands` for transforming selected text.
+ *
+ * END HEADER
+ */
+
+import { ChangeSet, type StateCommand } from '@codemirror/state'
+
+/**
+ * Transform the supplied `text`.
+ *
+ * * strip duplicate spaces
+ * * trim leading and trailing whitespace
+ * * change the casing
+ * * etc.
+ *
+ * @param   {string}  text  The text to be transformed.
+ *
+ * @return  {string}        The transformed text.
+ */
+export type TransformText = (text: string) => string
+
+/**
+ * Return a `Command` to transform selected text.
+ *
+ * @param   {TransformText} transform The transformation to be applied to the
+ *                                    selected text.
+ *
+ * @return  {StateCommand}            A `Command` to transform selected text.
+ */
+export function transformSelectedText (transform: TransformText): StateCommand {
+  return ({ state, dispatch }) => {
+    if (state.selection.ranges.length === 0) {
+      // when nothing is selected there's nothing to do
+      return false
+    }
+
+    const changes = state.selection.ranges.reduce((change, { from, to }) => {
+      const text = state.sliceDoc(from, to)
+
+      const transformedText = transform(text)
+
+      if (transformedText === text) {
+        // when there's no change just return the existing change (if any)
+        return change
+      }
+
+      const nextChange = ChangeSet.of(
+        { from, to, insert: transformedText },
+        state.doc.length
+      )
+
+      return change.compose(nextChange)
+    }, ChangeSet.empty(state.doc.length))
+
+    if (changes.empty) {
+      // when there are no changes there's nothing to do
+      return false
+    }
+
+    dispatch(state.update({ changes }))
+    return true
+  }
+}
--- a/source/common/modules/markdown-editor/commands/transforms/strip-duplicate-spaces.ts
+++ b/source/common/modules/markdown-editor/commands/transforms/strip-duplicate-spaces.ts
@@ -1,0 +1,18 @@
+import { transformSelectedText } from './transform-selected-text'
+
+/**
+ * Strip duplicate spaces from selected text.
+ *
+ * Sequences of two or more spaces will be collapsed to a single space.
+ *
+ * Note that sequences of interleaved whitespace such as `\t \t \t ` will _not_
+ * lead to any collapsing: _only_ sequences of two or more _spaces_ will be
+ * collapsed to a single space.
+ *
+ * @param   {string}  text  The text to be transformed.
+ *
+ * @return  {string}        The text with all duplicate spaces stripped.
+ */
+export const stripDuplicateSpaces = transformSelectedText((text) => {
+  return text.replaceAll(/ {2,}/g, ' ')
+})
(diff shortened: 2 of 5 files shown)


source/common/modules/markdown-editor/commands/transforms/transform-selected-text.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Return a `Command` to transform selected text.
+ *
+ * @param   {TransformText} transform The transformation to be applied to the
+ *                                    selected text.
+ *
+ * @return  {StateCommand}            A `Command` to transform selected text.
+ */
+export function transformSelectedText (transform: TransformText): StateCommand {
```
richdouglasevans, line 39: This function encapsulates all the machinery of transforming selected text using the CodeMirror API.

Individual transforms are then plugged into this function -- via the `transform` parameter -- and don't need to duplicate logic and/or concern themselves with the CodeMirror API.

This is a `StateCommand` -- which just requires `state` and `dispatch` -- rather than a more general `Command` because it makes these transforms easier to (unit) test because we don't have to create a "full fat" `EditorView`.

* [StateCommand](https://codemirror.net/docs/ref/#state.StateCommand)

richdouglasevans:
I have another 7 transforms prepped and ready for PR-ing... I thought I'd get this one uop first because it demonstrates the approach and then, after making any changes from the reviewer, we can (hopefully) quickly get the rest of the transforms in.

nathanlesage:
I will need some more time to think about your style of coding because it is different from mine — not in a bad way, so this is not a criticism, but it's a bit more involved to think about this. A few things I already saw you may want to tackle in the meantime is:

1. I feel your unit tests are extremely repetitive. When you test for various inputs for a single function, for example, it is much simpler and easier to create an array of test cases at the top of the file, and implement the tester function once. This way the file is more organized, and it will be simple to just add additional test cases in the future.
2. I greatly appreciate your efforts to kickstart testing the codemirror related functionality. The main reason why we don't yet have such tests is literally just because I was too lazy to write those helper functions until now. What I would love to see is to have that a bit more generalized so that we can test more than just the transformation commands in the future. Please do not take this to completely over-engineer that; just start naming things "codemirror-test-utils" or something like that simply so that we know where to add additional functionality as the need arises and we start testing more things.
3. Please don't write longer strings by concatenating them, this looks horribly ugly 🙈 The 80 character column limit is not enforced in cases where the description just takes longer, that is fine. If you really do need multi-line strings (but not in the tests please), you can use backtick-strings.
4. I'll have to think if it's really adequate to have the transform commands implemented using a factory style, because it does add some complexity that is more difficult to disentangle in the future. My first intention was to literally just write command functions as-is (in the sense that they could be copied and pasted into different code bases and they work there, too). I do see the appeal of saving on some code, however, but I'm wondering what your thoughts are on my approach to the table editor commands (see PR #5243, specifically this folder: https://github.com/Zettlr/Zettlr/tree/table-editor/source/common/modules/markdown-editor/table-editor/commands)

I'll be back with a more thorough code review — also please feel free to remove the explanatory comments, because it makes focusing on the code more difficult than necessary. I appreciate the thoroughness, but it is easier to parse if you could transfer that information into your PR description and leave the files!

Thank you already for your help and work!

richdouglasevans:
> I feel your unit tests are extremely repetitive. When you test for various inputs for a single function, for example, it is much simpler and easier to create an array of test cases at the top of the file, and implement the tester function once. This way the file is more organized, and it will be simple to just add additional test cases in the future.

I have changed the test fixtures to conform to your preference.

* 9899d09cea50656a0c26a72318d8d6fd2375c3f7

> What I would love to see is to have that a bit more generalized so that we can test more than just the transformation commands in the future.

I have moved the supporting code into a folder per your sugggestion.

* 9128bda64435de3288e7f422ab8f0bd2e8872915

> Please don't write longer strings by concatenating them.

Updated to conform to your preference in 9899d09cea50656a0c26a72318d8d6fd2375c3f7

> remove the explanatory comments

I have deleted all of them.

> I'm wondering what your thoughts are on my approach to the table editor commands

Approach looks solid ✌️ In your case there are distinct differences between chunks of commands... while there is some repetition, it reads well enough.

I only refactored to the composition approach after writing the third text transform command because by then I noticed that I was repeating the same code in every command: in the specific case of these text transforms there's no difference to the CodeMirror algorithm piece... the composition allows the code that's distinct and relevant to just that transform to "pop." 

It's easy enough for me to un-compose it.

nathanlesage:
Sooooooo, after way too long I've been able to give it a Quick Look. And it has proven once more that sometimes letting something sit can change one's perspective on it. I am now much more comfortable reading your code and seeing the benefits of your approach. While I do believe that a bit of the code might be too verbose, that honestly boils down to personal preference, so I don't want you to change anything. I just gave it a skim and will want to try it out myself but I really believe that the code looks good to merge!

So thanks for holding tight during the past four months -- once I'm back at my desktop, I'll give it a quick run and then we can merge it.

I currently think that we can pull this through in two PRs: one that implements the base functionality plus examples (I.e., this PR as it is), and then a second one that simply implements all other transforms you have been discussing!

If you want to, you can already open that second PR (you can close the individual ones), because I'm pretty sure you won't have to change anything! And then we're good to go!

Thank you again for all of this exemplary work!

nathanlesage (approved):
Finally! Sorry that I wasn't able to check and review the PR before last week's conference, but I finally was able to sit down and take a look.

Here's how I'd like this to go: I'll merge your changes, so that they will be included in tomorrow's nightly and tell the users to test it out. Then any additional issues will resurface. A few things I noted (which can be fixed down the line):

* The `zapGremlins` might be turned into a RegExp with ranges instead of individual characters (makes it less verbose)
* The quote-fixing functions all assume en-US quotes, and would probably benefit from access to the quote config. I have a pretty extensive list of quotes [here](https://github.com/Zettlr/Zettlr/blob/develop/source/win-preferences/schema/autocorrect.ts). My personal hunch is that we should turn ALL types of magic quotes into straight quotes, but only turn straight single or double quotes to the configured type of quotes (do nothing if that option is set to straight quotes already)

But other than that I can literally see nothing else, so it should, as the pros say, just work™!

Thank you again!

PR #5706, opened 2025-03-31
feat: Cache ESLint to improve speed

richdouglasevans:
## Description

Running ESLint on the command line is slow.

This change enables ESLint caching which speeds up subsequent runs of the tool on the command line.

* [Caching](https://eslint.org/docs/latest/use/command-line-interface#caching)

Current timings with no caching.

```
real    1m32.620s
```

After change, with caching.

```
real    0m7.296s
```

I am using the ESLint VSCode extension for in-the-moment linting of files I'm working on. Still, I lilke to run the `yarn lint` and `yarn test` scripts before pushing just to ensure that I haven't b0rk3d anything.

Waiting near two minutes for the linting to complete is punishing. The caching is not a great solution -- waiting near 10 seconds is still pretty slow -- but it's a marked improvement on the performance without it.

Note that this doesn't have any impact on the GitHub Actions builds; they are still taking near two minutes to lint the codebase.

* [Recent build](https://github.com/Zettlr/Zettlr/actions/runs/14171139998/job/39695001535)

Addressing the GitHub Actions build times is a separate issue I think; this fix is all about improving (one small aspect of) the Developer Experience (DX) of working on the Zettlr codebase.

## Additional information

Tested on: Windows 11 Home.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,8 +27,9 @@ test-config.yml
 yarn-error.log
 yarn.log
 
-# Don't commit ESLint error reports
+# Don't commit ESLint stuff
 eslint_report.htm
+.eslintcache
 
 # Taken from https://www.gitignore.io/api/visualstudiocode
 
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "update:citation": "node scripts/update-citation.mjs",
     "test": "mocha",
     "test-gui": "cross-env NODE_ENV=develop node scripts/test-gui/index.mjs",
-    "lint:code": "eslint",
+    "lint:code": "eslint --cache",
     "lint:types": "vue-tsc --noEmit",
     "lint:po": "node scripts/lint-po.mjs",
     "lint": "yarn lint:code && yarn lint:types",


nathanlesage:
If I understand it correctly, this basically just tells ESLint to keep any results for files which haven't been changed since the last time the command was run, correct? In that case this would be an improvement, because yes, the linter is excruciatingly slow. So I'll take it.

With regard to linting performance in general, I'm happy for any improvement, because I agree that linting should not take so long (although 10s for such an insanely large codebase is pretty good I'd say)

richdouglasevans:
> If I understand it correctly, this basically just tells ESLint to keep any results for files which haven't been changed since the last time the command was run, correct?

Yes, that's right ✌️

HerrAugust:
Hello, on my Ubuntu machine it takes 5 minutes and that's so long...

nathanlesage:
Let's quickly merge this, too. One observation, though: During updating ESLint a few days ago, I noticed that apparently it became much faster than before, so there's movement on their side, too!

Issue #5703, opened 2025-03-30
New File Always Defaults to .md Instead of Selected Format (e.g., JSON)"


jyothisjoy123:
### Description

1. file -new-json
2. note the file name associated default to .md not json

When creating a new file in Zettlr, the expected behavior is that the file extension should match the selected format (e.g., JSON). However, regardless of the chosen format, the new file is always created with a .md extension. This forces users to manually rename the file or change the extension, impacting workflow efficiency.

<img width="673" alt="Image" src="https://github.com/user-attachments/assets/227d2f18-db44-4a93-ad67-350582e050f7" />

<img width="981" alt="Image" src="https://github.com/user-attachments/assets/f9f04993-9620-410e-bb35-0855b0dd1714" />

### Reproducing

1.Open Zettlr.
2.Click on "New File" or use the shortcut to create a new file.
3.Select a format such as .json or .yml.
4.Observe that the file is created with a .md extension instead of the selected format.

### Zettlr version

3.4.3

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

15.3.1

### Additional Information

The new file should be created with the selected extension (e.g., .json or .yml) instead of defaulting to .md.

jyothisjoy123:
Additionally, the app allows users to manually enter any extension (e.g., .java), even if it's not listed, without warning or validation.

richdouglasevans:
> [R]egardless of the chosen format, the new file is always created with a .md extension.

Yes, this is a configuration setting: `File > Preferences > Advanced`

* [Docs](https://docs.zettlr.com/en/reference/settings/#advanced)

The default pattern for new file names is `%id.md`... the `.md` exension that you're seeing is coming from that.

Even if this setting is changed -- for example, to `%id.json` -- it still won't match your expectation even if you chose to create a YAML file the generated filename would now have a `.json` extension.

The code to generate the file name shown is here:

* https://github.com/Zettlr/Zettlr/blob/develop/source/app/service-providers/commands/file-new.ts#L44

The `generateFilename` function could be changed to have an optional `extension` parameter; that value could come from the `args.type` in `file-new.ts`, defaulting to `md`.

Or, the `file-new` command could override the extension to use the one that's in `args.type`.

[...]

PR #5711, opened 2025-04-03
Fix: New File Always Defaults to .md Instead of Selected Format (e.g., JSON)

jyothisjoy123:
## Description

This PR resolves the issue where new files always defaulted to .md. Now users can create new files with .tex, .json, .yaml, or .md extensions. Unsupported extensions will throw an error in the logs.

## Changes

1.Added support for new file types (.tex, .json, .yaml, .md).
2.If an unsupported file extension is chosen, an error is logged.
3.No breaking changes to the existing file creation flow.

## Additional information

macOS 15.3.2
Node.js v22.14.0

Note: Since this is a bug fix, I wasn’t sure if an entry in CHANGELOG.md is required. Let me know if I should add one, and I’d be happy to update it!

### DIFF
--- a/source/app/service-providers/commands/file-new.ts
+++ b/source/app/service-providers/commands/file-new.ts
@@ -1,10 +1,22 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        FileNew command
+ * CVM-Role:        <none>
+ * Maintainer:      Hendrik Erz
+ * License:         GNU GPL v3
+ *
+ * Description:     This command creates a new file.
+ *
+ * END HEADER
+ */
 import ZettlrCommand from './zettlr-command'
 import { trans } from '@common/i18n-main'
 import path from 'path'
 import sanitize from 'sanitize-filename'
 import generateFilename from '@common/util/generate-filename'
 import { app } from 'electron'
-
 export default class FileNew extends ZettlrCommand {
   constructor (app: any) {
     super(app, ['file-new'])
@@ -17,6 +29,13 @@ export default class FileNew extends ZettlrCommand {
    * @return {void}     This function does not return anything.
    */
   async run (evt: string, arg: { leafId?: string, windowId?: string, name?: string, path?: string, type: 'md'|'yaml'|'json'|'tex' }): Promise<void> {
+    // A few notes on how this command works with respect to its input. As you
+    // can see, all parameters are optional and all which are missing will be
+    // inferred from context (otherwise the command will fail). The type
+    // defaults to Markdown, obviously, the path to the current directory and
+    // the name has the following function: If it is given, the user will not
+    // be asked for a filename, but if it's missing, a new name will be
+    // generated and the user is asked to confirm the name.
     const { newFileDontPrompt, newFileNamePattern } = this._app.config.get()
     const type = arg.type ?? 'md'
     const generatedName = generateFilename(newFileNamePattern, this._app.config.get().zkn.idGen, type)
@@ -24,6 +43,9 @@ export default class FileNew extends ZettlrCommand {
     const leafId = arg.leafId
 
     if (arg.windowId === undefined) {
+      // The caller didn't provide a window number. This can happen with, e.g.
+      // menu items. But since we want to fulfill the user's wish, we first try
+      // to fallback onto any main window and only fail otherwise.
       const firstMainWindow = this._app.windows.getFirstMainWindow()
       if (firstMainWindow !== undefined) {
         arg.windowId = this._app.windows.getMainWindowKey(firstMainWindow)
@@ -36,21 +58,34 @@ export default class FileNew extends ZettlrCommand {
       this._app.log.error('Cannot create new file: No window id provided')
       return
     }
+    // Assume there is no directory we could salvage, so choose the documents
+    // directory, but (by setting isFallbackDir to true) allow the user to
+    // change it.
 
     let dirpath = app.getPath('documents')
     const { openDirectory } = this._app.config.get()
     let isFallbackDir = true
     if (typeof arg.path === 'string' && await this._app.fsal.isDir(arg.path)) {
+      // Explicitly provided directory
       dirpath = arg.path
       isFallbackDir = false
     } else if (openDirectory !== null && await this._app.fsal.isDir(openDirectory)) {
+      // Explicitly provided directory
       dirpath = openDirectory
       isFallbackDir = false
     }
+    // Make sure we have a filename and have the user confirm this if applicable
+    // Also, if the user does not want to be prompted BUT we had to use the
+    // fallback directory, we should also prompt the user as otherwise it would
+    // be opaque to the user where the notes end up in.
 
     if (
+      // No name provided, but the user does not want an auto-generated name
       (arg.name === undefined && !newFileDontPrompt) ||
+      // The user does not wish to provide a filename, but we've had to revert
+      // to a fallback directory
       (newFileDontPrompt && isFallbackDir)
+      
     ) {
       let defaultFileName = generatedName
       // Add the extension to the filename based on the selected type
@@ -82,14 +117,18 @@ export default class FileNew extends ZettlrCommand {
       }
 
       arg.name = path.basename(chosenPath)
+      // The user may also have selected a different directory altogether. If
+      // that directory exists and is loaded by the FSAL, overwrite the dir.
       if (path.dirname(chosenPath) !== dirpath) {
         dirpath = path.dirname(chosenPath)
       }
     } else if (arg.name === undefined) {
+      // Just generate a name.
       arg.name = generatedName
     }
 
     try {
+      // Then, make sure the name is correct.
       let filename = sanitize(arg.name.trim(), { replacement: '-' })
       if (filename === '') {
         throw new Error('Could not create file: Filename was not valid')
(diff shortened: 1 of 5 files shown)


nathanlesage:
Thank you for opening this! I have two things I would like to clarify before doing a code review:

1. You seem to have removed and added things here and there, not necessarily with changing functionality. What is the reason? (I was especially confused in cases where you changed code comments)
2. In issue #5703 we have discussed adding another variable, `%ext`, but if I see it correctly, you have simply removed the filename extension from the pattern, correct? Why is this?

In general, it helps reviewing PRs if you only change what is actually necessary. Currently it's a bit hard to disentangle what you have changed code-wise due to the various other changes you introduced. But again, thank you so far, I'm sure we can turn this into a great PR in no time!

source/app/service-providers/commands/file-new.ts
```diff
@@ -72,24 +70,47 @@ export default class FileNew extends ZettlrCommand {
       dirpath = arg.path
       isFallbackDir = false
     } else if (openDirectory !== null && await this._app.fsal.isDir(openDirectory)) {
-      // Choose the openDirectory
+      // Explicitly provided directory
```
nathanlesage, line 73: Please undo this comment change

source/app/service-providers/commands/file-new.ts
```diff
@@ -72,24 +70,47 @@ export default class FileNew extends ZettlrCommand {
...
       // No name provided, but the user does not want an auto-generated name
       (arg.name === undefined && !newFileDontPrompt) ||
       // The user does not wish to provide a filename, but we've had to revert
       // to a fallback directory
       (newFileDontPrompt && isFallbackDir)
+      
     ) {
-      // The user wishes/needs to confirm the filename and/or the directory
-      const chosenPath = await this._app.windows.saveFile(path.join(dirpath, generatedName))
+      let defaultFileName = generatedName
```
nathanlesage, line 90: Can we not reuse the `generatedName` variable here?

source/app/service-providers/commands/file-new.ts
```diff
@@ -72,24 +70,47 @@ export default class FileNew extends ZettlrCommand {
...
+      // Add the extension to the filename based on the selected type
+      switch (type) {
+        case 'json':
+          if (!defaultFileName.endsWith('.json')) {
+            defaultFileName += '.json'  // Ensure the default extension for JSON files
+          }
+          break
+        case 'yaml':
+          if (!defaultFileName.endsWith('.yaml')) {
```
nathanlesage, line 99: I would add a check if the name ends with `.yml` which is also a valid extension for YAML files. We can still default to `.yaml` below, that's absolutely fine.

source/app/service-providers/commands/file-new.ts
```diff
@@ -113,30 +134,28 @@ export default class FileNew extends ZettlrCommand {
         throw new Error('Could not create file: Filename was not valid')
       }
 
-      if (!hasMdOrCodeExt(filename)) {
-        // There's no valid file extension given. We have to add one. By default
-        // we assume Markdown, but let ourselves be guided by the given type.
+      // Check if the file already has an extension, and don't add the default extension if it does
+      if (!path.extname(filename)) {
```
nathanlesage, line 138: Please undo the changes in the if/switch construction here.

First, `if(!path.extname(filename))` does not pass the linter, you'd need to explicitly compare it to a string w/o implicit boolean conversion. Second, this construction should've already worked. Or is there something problematic with how it works right now? Did I overlook something?

source/app/service-providers/commands/file-new.ts
```diff
@@ -113,30 +134,28 @@ export default class FileNew extends ZettlrCommand {
...
       }
 
       const absPath = path.join(dirpath, filename)
-
       // Check if there's already a file with this name in the directory
       // NOTE: There are case-sensitive file systems, but we'll disallow this
+
       if (await this._app.fsal.pathExists(absPath)) {
-        // Ask before overwriting
         if (!await this._app.windows.shouldOverwriteFile(filename)) {
```
nathanlesage, line 159: Please keep this comment

source/app/service-providers/commands/file-new.ts
```diff
@@ -146,17 +165,15 @@ export default class FileNew extends ZettlrCommand {
           await this._app.fsal.removeFile(absPath)
         }
       }
-
       // First create the file
       await this._app.fsal.writeTextFile(absPath, '')
-
-      // And directly thereafter, open the file
```
nathanlesage, line 153: Please keep this comment. (In general, please do not delete comments unless they become invalid due to any change)

source/common/util/generate-filename.ts
```diff
@@ -14,19 +14,25 @@
...
   let pattern = replaceStringVariables(filenamePattern)
+  
+
   pattern = pattern.replace(/%id/g, generateId(idGenPattern))
-  // In case a funny guy has removed the pattern from config.
-  if (pattern.trim().length === 0) {
-    pattern = uuid()
-  }
+  
+  ext = ext || 'md'
```
nathanlesage, line 32: `ext` is not optional, so this case is unnecessary. You can remove this.

source/common/util/generate-filename.ts
```diff
@@ -14,19 +14,25 @@
...
 
 /**
  * Generates a new filename based on the configured filename pattern.
+ * @param {string} filenamePattern - The pattern for generating filenames.
+ * @param {string} idGenPattern - The pattern for generating unique IDs.
+ * @param {string} ext - The file extension to be used (default: 'md').
  * @return {string} The new filename.
  */
-export default function generateFilename (filenamePattern: string, idGenPattern: string): string {
+export default function generateFilename (filenamePattern: string, idGenPattern: string, ext: string ): string {
```
nathanlesage, line 25: Suggestion to strongly type the extension:

```suggestion
export default function generateFilename (filenamePattern: string, idGenPattern: string, ext: 'md'|'json'|'yaml'|'tex' ): string {
```

source/common/util/generate-filename.ts
```diff
@@ -14,19 +14,25 @@
...
+  
+
   pattern = pattern.replace(/%id/g, generateId(idGenPattern))
-  // In case a funny guy has removed the pattern from config.
-  if (pattern.trim().length === 0) {
-    pattern = uuid()
-  }
+  
+  ext = ext || 'md'
+  pattern = pattern.replace(/%ext/g, ext)
```
nathanlesage, line 33: You never introduced this `%ext` variable, so this change won't do much. I suggest to use instead `uuid() + '.' + ext` construction. (Or maybe with template strings, that might look nicer).

source/common/util/generate-id.ts
```diff
@@ -21,6 +21,6 @@ import replaceStringVariables from './replace-string-variables'
  *
  * @return  {String}                            The final string after replacements.
  */
-export default function generateId (pattern: string = '%Y%M%D%h%m%s'): string {
+export default function generateId (pattern: string = '%Y%M%D%h%m') {
```
nathanlesage, line 24: Please undo this change, you removed the (for this change unaffected) `%s`-variable, and removed the return type.

nathanlesage (changes requested):
Sorry for taking so long with the review. I believe this PR is going in the right direction. Most of my comments are relatively simple. There are a few things I'm unsure about and would like you to clarify some of the changes you did — I am not too familiar with the exact logic in this command because I haven't looked at it in quite some time, so I'm happy to listen to your justifications!

Finally, there are now a few linter/test errors that would need to be fixed (see the check annotations).

jyothisjoy-a1911465:
Thank you @nathanlesage for the detailed evaluation. Working on it and will update the requested changes.

jyothisjoy123:
@nathanlesage I have reviewed your suggestions  and updated the file. I found that the core issue was in-fact from the config file, where the filenamepattern was set to `%id` and not `%id.%ext`.  Therefore, I undo the other changes and found that my fix passed all the test cases. `yarn test` showed 221 passed tests.

However, after resolving the merge conflict, I found that current `develop` branch does not run.  `yarn start` results in following error

```
node:events:496
      throw er; // Unhandled 'error' event
      ^

Error: spawn electron-forge ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn electron-forge',
  path: 'electron-forge',
  spawnargs: [
    'start',
    '--',
    '--data-dir="/Users/jyothisjoy/projects/Zettlr/resources/test-cfg"'
  ]
}

Node.js v22.14.0
```


How do you wish me to approach this situation? Do you want me to undo the "merge commit" and force push changes to this branch?

nathanlesage:
> How do you wish me to approach this situation? Do you want me to undo the "merge commit" and force push changes to this branch?

Umm… wouldn't it be possible to find the cause by diffing the develop branch with your code for any changes? One of them likely is the cause of this error, is it not? Feel free to ping me so that I can re-review the PR!

PR #5735, opened 2025-04-19
Add pause button to pomodoro timer.

### LINKED ISSUE #5667: Add pause button to pomodoro timer.
### Description

I've been using the pomodoro timer more lately and it's super useful. My one issue is that, occasionally, I'll experience a brief interruption — an important incoming message that has to be addressed, someone stopping by my desk, etc. — and there's no way to pause the pomodoro timer when this happens. I can only stop it and start the cycle of pomodoros all over again.

To give an example of where this is sub-optimal, if I've been going for almost two hours, and am interrupted briefly in the middle of my last work session left before my long break, it would be nice to be able to pause the timer for a moment, rather than my only choices being to let the timer run while I talk to a student or hit `Stop` and begin the entire cycle from the top, with four more work sessions to get through before a long break.

Obviously, it's easy enough to just use a non-integrated pomodoro timer that I can pause as needed, but that sort of defeats the purpose of the feature.

### Proposed Changes

See above.

### Caveats

Some people may like the "write or die" nature of having no pause button. But I think it makes sense for most of us.

No

### Zettlr Version

Nightly

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Manjaro (Rolling)

### Additional Information

_No response_

bjones4949:
## Description

Closes #5667 
This PR adds the option to pause and resume the pomodoro timer. It adds an additional button in the pomodoro popover.
## Changes

I added the button element to the PopoverPomodoro vue file as well as added a pause function and modifications to the the pomodoroTick function to enable the pause in App.vue. Style was added to maintain a spacing when both the pause/resume and stop buttons are displayed together. The start function also resets the paused to false so it doesn't retain the paused state when restarted. 

## Additional information

Tested on: Arch Linux

### DIFF
--- a/source/win-main/App.vue
+++ b/source/win-main/App.vue
@@ -111,9 +111,11 @@
     v-bind:target="pomodoroButton"
     v-bind:pomodoro="pomodoro"
     v-bind:sound-effects="SOUND_EFFECTS"
+    v-bind:is-paused="isPaused"
     v-on:close="showPomodoroPopover = false"
     v-on:config="setPomodoroConfig($event)"
     v-on:start="startPomodoro()"
+    v-on:pause="pausePomodoro()"
     v-on:stop="stopPomodoro()"
   ></PopoverPomodoro>
 </template>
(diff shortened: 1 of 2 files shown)


bjones4949:
Hi! Just a quick note — I accidentally closed and reopened this PR, so the workflow needs approval again. Thank you very much!

PR #5688, opened 2025-03-23
Update fr-FR.po

julesbouton:
## Description

Revise and complete fr-FR translations.

## Changes

I added many missing translations and updated a lot of outdated ones.
I further tried to make some translations more consistent.

PR #5738, opened 2025-04-24
fix typos in fr-FR.po

julesbouton:
# Description

Fix a few issues in fr-FR.po (relatively small update after the much bigger #5688)


nathanlesage:
Merci!

PR #5740, opened 2025-04-25
feat: Add Custom Shortcuts

### LINKED ISSUE #818: Custom keyboard shortcuts
I thought it could be handy to be able to add custom keyboard shortcuts for faster editing/writing. Cheers!

nathanlesage:
## Description

This PR adds support for custom keyboard shortcuts for the editor of Zettlr. This allows users to customize which keyboard shortcuts do what in the editor. Users can mitigate three specific issues with this:

1. The default keybindings are uncommon, and they would like to remap them to more convenient shortcuts to retain their muscle memory from other apps.
2. The default keybindings do not work on the user's keyboard layout. The custom shortcut input essentially records the new shortcut based on the platform specific keyboard layout, making it easier to find shortcuts that actually work.
3. The default keybindings are uncomfortable for their specific keyboard (hardware or layout; think ortholinear or Dvorak).

## Changes

This PR makes the following changes:

1. Add a new preference key for custom keyboard shortcuts in the config.
2. Remove specific keyboard shortcuts and replace them with dynamic keyboard assignments in the default keymap.
3. Add a new control to record shortcuts, utilizing the CodeMirror-specific libraries to mimic its behavior as closely as possible.
4. As a proof of concept, "dynamicize" two key bindings, for adding images and adding footnotes.

## Additional information

There are still a few todos and caveats:

1. [ ] The preference shortcuts are not yet passed into the keymap constructor.
2. [ ] The keymaps cannot be dynamically updated yet.
3. [ ] We need to reach a consensus as to which commands should be customizable. I don't believe we should customize, e.g., arrow key movement.

***

- fixes #818

Tested on: macOS Sequoia 15.4.1

### DIFF
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -16,6 +16,7 @@ import { app } from 'electron'
 import * as bcp47 from 'bcp-47'
 import { v4 as uuid4 } from 'uuid'
 import getLanguageFile from '@common/util/get-language-file'
+import { getPlatformSpecificDefaultKeybinding, type EditorKeyboardCommand } from '@common/modules/markdown-editor/keymaps/default-map'
 
 export type MarkdownTheme = 'berlin'|'frankfurt'|'bielefeld'|'karl-marx-stadt'|'bordeaux'
 
@@ -120,6 +121,7 @@ export interface ConfigOptions {
       replacements: Array<{ key: string, value: string }>
       matchWholeWords: boolean
     }
+    keyboardShortcuts: Record<EditorKeyboardCommand, string>
   }
   display: {
     theme: MarkdownTheme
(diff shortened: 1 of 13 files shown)


historical-theology:
I appreciate that you are taking your time to implement this important feature in a thoughtful way.

In the interim, I suggest that it could be useful to include the Markdown file in which the default Zettlr shortcuts are listed (/reference/[keyboard-]shortcuts.md, depending on language) _inside_ the application itself, perhaps as part of the "Zettlr Tutorial" bundle. 

In that way, no one would need to divert workflow attention out of Zettlr to the online docs just to remember a keybinding. Rather, by having the file be a part of Zettlr, it would be permanently searchable.

I made a print-out of the keyboard shortcuts for myself, so I do not need to enter a Web browser to learn and remember them, and it takes 4.5 US Letter pages. The total set of keyboard shortcuts for Zettlr is quite large. It is good that every action has a shortcut, and it will be better when every shortcut is customizable. Until then, I think that having an all-in-one checklist would be a boon to the user experience. 

Plus, _where_ the link/icon for a hypothetical default keybindings list is placed could be a _placeholder_ for the eventual customization interface -- even a testing zone, to see if folks like that primary link/icon location or would prefer somewhere else.

maxanthem:
Cheers from 2026 :)

Why not both highlighting existing shortcuts **and** letting the user customize them?

If the different layout mappings poses an issue, why not reusing existing known keys mapped from existing shortcuts? What I mean is that if at least CTRL, ALT etc work for existing shortcuts, it might possible to reuse them for other use-customized shortcuts?

But maybe I don't get it right.

PR #5754, opened 2025-05-05
feat: add emoji support to snippet insertion

### LINKED ISSUE #5746: Emojis on snippets
### Description

I'm a fan of using emojis on my notes kept on Zettlr, and I noticed today that if I add an emoji notation (i.e. `:books:`) on a snippet, when using that snippet on a note, the `:books:` text is not replaced by the expected 📚 emoji.

### Proposed Changes

I see two options:

- allow emojis to be added directly on the snippets editor
- parse the snippets whenever they are inserted to convert any emoji syntax to the actual emoji

### Caveats

It probably complicates either the snippets editor, or the parsing of snippets on the regular text editor, but hopefully not too much 🤞🏾 

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Fedora Linux 41

### Additional Information

_No response_

jujuliaa12:
## Description

Add emoji support in snippet insertion using :emoji_name: syntax.

## Changes

- Removed `node-emoji` dependency
- Replaced emoji rendering with `gemoji` lookup
- Applied emoji conversion in `applySnippet()`

## Additional information

Tested with `yarn dev` and manual :emojiTest snippet.
Tested on: Windows 10, Node.js 18, Yarn 4

Fixes #5746


jujuliaa12:
Thank you for the feedback! Here's a bit more detail for each change:

Removed node-emoji dependency
→ Because node-emoji is not actively maintained and lacks TypeScript support. It also caused build issues in the preload step, so we replaced it with a simpler solution.

Replaced emoji rendering with gemoji lookup
→ gemoji is already used elsewhere in the project and provides a complete emoji-to-name mapping. Using it avoids adding an extra dependency and improves consistency.

Applied emoji conversion in applySnippet()
→ The reason it's done during snippet insertion is to allow dynamic emoji rendering from shortcodes like :books:. This approach keeps snippets human-readable and editable in plain text while ensuring correct rendering at the moment of insertion.

PR #5755, opened 2025-05-06
Fixes #5660: Prevent image reloading on every keystroke using debounce

### LINKED ISSUE #5660: Images are reloading on input
I recently found (via a GIF, of all kinds) that images in the editor are constantly reloading due to a small bug. I don't fully know where the bug happens right now, but I know where to look for it. This issue serves as a reminder to myself.

freshinfrancis:
## Description
Fixes #5660

This PR resolves issue [#5660](https://github.com/Zettlr/Zettlr/issues/5660) by debouncing the DOM update logic for image rendering inside the markdown editor.

Previously, GIFs and other images would reload on every keystroke, resulting in distracting flickering or restart of animations. This fix:

- Introduces a `debounce` utility to limit update frequency.
- Applies debouncing to `updateDOM`.
- Adds a check to avoid resetting `img.src` if the source hasn't changed.

## Changes
Introduced a debounce.ts utility in markdown-editor/util/ to delay execution of frequent DOM updates.

Updated image rendering logic in the Markdown preview to use the debounced function, preventing constant reloading of GIFs/images during text input.

Added a conditional check to skip setting img.src if the URL hasn’t changed, avoiding unnecessary DOM writes.

No breaking changes to public APIs or components.

Created:
source/common/modules/markdown-editor/util/debounce.ts – A generic debounce utility to throttle function execution.

Modified:
source/common/modules/markdown-editor/renderers/render-images.ts – Applied debouncing to the image rendering logic in updateDOM.

Introduced logic to avoid unnecessary img.src re-assignments if the source URL hasn't changed.

image

## Additional information
This fix addresses [Issue #5660](https://github.com/Zettlr/Zettlr/issues/5660), where GIFs and images were reloading on every keystroke in the Markdown editor.

The change uses a custom debounce function with a 15ms delay to throttle image reloads and improve performance.

Added a conditional check to prevent reassigning img.src when the source hasn’t changed, further reducing unnecessary reloads.

Verified functionality by testing with GIFs and observing that reloads no longer occur during typing.

Tested on: Windows 10

https://github.com/user-attachments/assets/6718d18c-66ab-431f-b255-92e6d792d884

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-images.ts
+++ b/source/common/modules/markdown-editor/renderers/render-images.ts
@@ -23,6 +23,7 @@ import { trans } from '@common/i18n-renderer'
 import clickAndSelect from './click-and-select'
 import { pathDirname } from '@common/util/renderer-path-polyfill'
 import { syntaxTree } from '@codemirror/language'
+import debounce from 'source/common/modules/markdown-editor/util/debounce'
 
 // This variable holds a base64 encoded placeholder image.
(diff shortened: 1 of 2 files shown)


nathanlesage:
First of all, thank you for this first contribution of you! I really appreciate it. However, your solution is very over-engineered (why introduce debouncing? And then, why didn't you use the available utility function from underscore.js and instead copied and pasted presumably one of the implementations into the code base…? Other than that, it looks good. But please have a look at commit bc4d91d, which implements this using a minimal viable solution.

Issue #5783, opened 2025-05-31
Hide double square brackets in internal links


Lerpwave:
### Description

I would like to hide the double square brackets in internal links for better readability.

### Proposed Changes

This could be an option in the Markdown rendering preferences as a checkbox.
How it is now:

![Image](https://github.com/user-attachments/assets/26e95a2a-abae-40b9-a313-6abd1a5b1f2e)

What I would like to see:

![Image](https://github.com/user-attachments/assets/5e0fc639-d5f4-4c38-8726-d15bae2a8a85)

### Caveats

I can't think of any caveats.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Ubuntu 22.04.5 LTS

Linrui-Li:
Hi! I’ve submitted a pull request to address this issue. It uses AST nodes when available, and falls back to RegExp if not.
PR: [#5791 ]
Let me know if any adjustments are needed — happy to improve!

PR #5791, opened 2025-06-14
feat: hide double square brackets in internal links

Linrui-Li:
This pull request hides the double square brackets ([[ and ]]) used in internal links in the editor view, improving visual clarity.
If ZknLink AST nodes are available, the plugin hides the pipe (|) and internal target.
As a fallback, it uses a RegExp to hide brackets manually when no AST is available.
This resolves [#5783](https://github.com/Zettlr/Zettlr/issues/5783).
I'm happy to receive feedback and make any changes as needed.

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-links.ts
+++ b/source/common/modules/markdown-editor/renderers/render-links.ts
@@ -68,6 +68,18 @@ function hideLinkMarkers (view: EditorView): RangeSet<Decoration> {
     })
   }
 
+  // Fallback: Manually hide [[ and ]] using RegExp
+  const docText = view.state.doc.toString()
+  const bracketPattern = /\[\[(.*?)\]\]/g
+  let match
+  while ((match = bracketPattern.exec(docText)) !== null) {
+    const start = match.index
+    const end = match.index + match[0].length
+    ranges.push(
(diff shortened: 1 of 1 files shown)


source/common/modules/markdown-editor/renderers/render-links.ts
```diff
@@ -68,6 +68,18 @@ function hideLinkMarkers (view: EditorView): RangeSet<Decoration> {
     })
   }
 
+  // Fallback: Manually hide [[ and ]] using RegExp
```
nathanlesage, line 71: Why "Fallback"?

source/common/modules/markdown-editor/renderers/render-links.ts
```diff
@@ -68,6 +68,18 @@ function hideLinkMarkers (view: EditorView): RangeSet<Decoration> {
...
+  const docText = view.state.doc.toString()
+  const bracketPattern = /\[\[(.*?)\]\]/g
```
nathanlesage, line 73: I advise against regular expressions where not necessary. There are syntax nodes within the syntax tree for double brackets that you can target instead. The reason the links above are handled a bit awkwardly is simply due to the complex nature of Markdown inline and reference links.

PR #5818, opened 2025-07-29
fix: toSentenceCase no longer skips the first word

richdouglasevans:
## Description
Given a sentence where the first word was ALL_CAPS, the `toSentenceCase` function would transform the other, subsequent, words in the sentence to sentence case... but it'd essentially skip casing the first word.

Bug was spotted by @nathanlesage and [described here](https://github.com/Zettlr/Zettlr/issues/5659#issuecomment-3131843870).

## Changes
This fix adds a test case to flush out the bug and then fixes the bug in the implementation code.

## Demonstration

https://github.com/user-attachments/assets/16222c49-6273-4e24-88db-bca15583e2ec

## Additional information
Tested on: Windows 11 Home.


nathanlesage:
Looks good, thank you!

PR #5831, opened 2025-08-09
Update it-IT.po

Yattatux:
Update Italian translation.


nathanlesage:
Thank you for the work!

PR #5834, opened 2025-08-11
feat: Allow global search to be cancelled

kyaso:
## Description
This PR adds the ability to cancel/re-trigger an already in-flight global search run.

This feature was present in the past, but got removed at some point. While I'm not sure why the removal happened, I thought of reintroducing the feature, as it might be useful in certain cases where a user wants to cancel a long running search.

## Changes
- Added a new "Cancel" button next to the "Search" button
  - It is only enabled if a search is running
  - If clicked, it will result in the current search run being cancelled
- The search button is now also enabled during an already running search
  - If clicked, it will trigger a new search run based on the current value of the search field

Tested on: Ubuntu WSL, Windows 11

### DIFF
--- a/source/win-main/GlobalSearch.vue
+++ b/source/win-main/GlobalSearch.vue
@@ -29,7 +29,7 @@
         v-on:click="startSearch()"
       ></ButtonControl>
       <ButtonControl
-        label="Cancel"
+        v-bind:label="cancelButtonLabel"
         v-bind:inline="true"
         v-bind:disabled="!searchIsRunning"
         v-on:click="cancelSearch()"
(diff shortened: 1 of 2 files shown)


nathanlesage:
Im 90% sure it got accidentally removed -- Your implementation looks sounds so far! The only thing I'd do would be to denote that a feature instead of a simple fix since the start-new-search-while-old-one-runs sounds pretty much like a new feature. Also the new release is going to bump the minor version anyway! Other than that we can merge!

kyaso:
> The only thing I'd do would be to denote that a feature

I assume you mean in the Changelog, right? Sure, I'll do that later today!

source/win-main/GlobalSearch.vue
```diff
@@ -25,9 +25,15 @@
       <ButtonControl
         v-bind:label="searchButtonLabel"
         v-bind:inline="true"
-        v-bind:disabled="filesToSearch.length > 0"
+        v-bind:disabled="false"
         v-on:click="startSearch()"
       ></ButtonControl>
+      <ButtonControl
+        label="Cancel"
```
kyaso: TODO: Don't hardcode string

nathanlesage:
Exactly, yes! Feel free to merge on your own once you're ready! :)

Issue #4191, opened 2023-03-06
The $FILENAME variable should not return the extension


McGlear:
# Description

When using the `$FILENAME` variable within snippets, it returns not only the file's name, but also its extension (for which there is the additional `$EXTENSION` variable). There currently is no easy way to return `filename` instead of `filename.md` to use within snippets.

# Proposed Changes
`$FILENAME` should return the name _without_ the file extension. When the user needs the extension, they can easily use `$FILENAME$EXTENSION`.

# Caveats

I don't see any.

# Do you Wish to Attempt Implementing this Yourself?

No

# Zettlr Version

3.0.0-beta.2

# Your Platform

Windows 10

PR #5842, opened 2025-08-12
fix(snippets): Remove extension from filename variable

kyaso:
## Description
This a proposal to address #4191.

## Changes
- The file extension will be removed from the `FILENAME` variable
- Caveat: Users who _want_ the file extension as well, need to update their snippets accordingly -> Added a note regarding that in the Changelog

## Additional information

Tested on: WSL Ubuntu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@ so long to fix all of these issues at once!
 - **Feature**: Full-text (aka. global) search runs can now be cancelled via a
   dedicated button. You can now also trigger a new search while another search
   is already running.
+- Snippets: The `$FILENAME` variable now does not contain the file extension
+  anymore. Users who also want the extension should update their snippets to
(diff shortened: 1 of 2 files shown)


nathanlesage:
Looks good! Just one thing: to make this change more clear, I'd suggest we add this with a `**Change**:` (similar to the feature announcements) at the top of the changes list and, in addition, add a heading 2 section before the list of changes explicating this, as this reduces the likelihood of people missing this! But code wise it's all good, and the changelog entry itself is also clear :) 

Feel free to merge on your own then! And thank you!

kyaso:
@nathanlesage Regarding the heading 2 section, not sure if I got that. Do you mean to create a dedicated "Changes" section? For now, I've just added "Change:" in front of the entry (similar to other changes in previous releases).

nathanlesage:
No worries, I've now added that section! Merging now …!

Issue #2070, opened 2021-06-09
Copying search results to clipboard


break6700:
## Description

Right-clicking on a search result copies the text to clipboard.
## Example Workflow

User searches for term e.g. anechoic, finds the result they want and right clicks. This cops the text to clipboard and notifies the User with a pop-up. This can then be pasted into a larger document e.g. research paper.

Ideally this also copies the names of the file/ID of the file the text is from which serves as a backlink.

![image](https://user-images.githubusercontent.com/44082853/121337114-b9905800-c90b-11eb-8486-bdf31736591e.png)

## Caveats

Not that I can think of.

## Additional Information

Loving the new dev update! Thank you for making the writing of my dissertation a pleasure to do.

nathanlesage:
Hehe, first congrats on submitting ;)

Then, that's certainly an interesting idea, although I cannot promise it'll ship with 2.0. I'm currently very much in favour of simply closing off the stuff that's already in the pipeline and releasing 2.0!

nathanlesage:
Wow, I just saw that we fixed this issue after more than four years … holy hell, it's good to finally have some support @kyaso 😂 

By the way, @kyaso one note: Please next time include "Fixes" or "Closes" as a keyword in your PR description. Since we're performing squash and merge, we're only going to get a single merge commit, which overwrites your "fixes" keywords in the commit messages. We may also want to consider moving back to a rebase merge to keep the entire history, although I'm a bit afraid that this might blow up the commit history … but nothing of concern right now!

kyaso:
> Wow, I just saw that we fixed this issue after more than four years …

Haha, indeed!

> By the way, @kyaso one note: Please next time include "Fixes" or "Closes" as a keyword in your PR description. Since we're performing squash and merge, we're only going to get a single merge commit, which overwrites your "fixes" keywords in the commit messages. We may also want to consider moving back to a rebase merge to keep the entire history, although I'm a bit afraid that this might blow up the commit history … but nothing of concern right now!

Good to know, will keep that in mind from now on!

PR #5845, opened 2025-08-13
feat: Add copy option to global search result context menu

kyaso:
## Description
This PR addresses #2070.

## Changes
Added a "copy" menu item to the context menu of search results. When clicked, it will copy the corresponding line text to the clipboard.

## Additional information
Note that I _didn't_ implement the additional request by the issue submitter:
> Ideally this also copies the names of the file/ID of the file the text is from which serves as a backlink.

---

Tested on: WSL Ubuntu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,19 @@ In any case, we hope that the new TableEditor will finally fix the issues you
 experienced over the past years — and we would like to apologize that it took us
 so long to fix all of these issues at once!
 
+## Changes to the snippet `$FILENAME` variable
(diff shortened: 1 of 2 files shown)


nathanlesage:
Looks good! The async portion in the callback makes me wonder if we should update the context menu callback to also return a promise, but that's for another day :)

kyaso:
> The async portion in the callback makes me wonder if we should update the context menu callback to also return a promise, but that's for another day :)

Ahhh, I must have missed that 😅

Thanks for merging!

PR #5846, opened 2025-08-13
fix: Make sure cursor is moved behind insertion text

kyaso:
## Description
Problem: When inserting a ZKN ID, the cursor stays _in front_ of the inserted ID string. This disrupts the editing flow, as the user has to manually correct the cursor position before continuing to type.

This PR solves this by ensuring the cursor is always right behind the inserted ID.

## Changes
Modified the Markdown editor's custom `replaceSelection` method to also include a `selection` update (containing the new cursor location) in the transaction.

## Additional information
As far as I can tell, the `replaceSelection` method is currently only accessed when inserting IDs, or new tables. I have tested the new changes with both, and it seems to work as expected (also tested with and without having texts selected).

Tested on: WSL Ubuntu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ so long to fix all of these issues at once!
   dedicated button. You can now also trigger a new search while another search
   is already running.
 - Fixed WebP images not rendering from relative paths (#5181).
+- Fixed incorrect cursor position after inserting IDs (#5846)
 
 ## Under the Hood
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
I think this is a remnant from the pre-Codemirror 6-times. I just checked, and I think the `view.state.replaceSelection`-method does what we want to, but better.

In that case, it makes sense to completely remove that utility function and rely on the built-in method in general. Would you mind having a look?

kyaso:
Sure, I'll check it out!

kyaso:
@nathanlesage Let me know what you think of my latest changes!

nathanlesage:
Looks good! Let me merge real quick :)

PR #5847, opened 2025-08-14
[Feat] Clicking the diagnostic status toggles the lint panel.

benniekiss:
## Description

This PR allows clicking the diagnostics section of the statusbar to toggle the lint panel.

## Changes

Currently, clicking the diagnostics section only opens the lint panel.

## Additional information

While I was exploring the app, the current functionality was confusing to me. I did not notice the (x) close button at first, so my first instinct was to click the diagnostics again to close the panel. Since it did not work, I thought I would implement it since this seemed like a natural function.

macOS 15.6

### DIFF
--- a/source/common/modules/markdown-editor/statusbar/diagnostics.ts
+++ b/source/common/modules/markdown-editor/statusbar/diagnostics.ts
@@ -16,7 +16,9 @@ import { type EditorState } from '@codemirror/state'
 import { type EditorView } from '@codemirror/view'
 import { trans } from '@common/i18n-renderer'
 import { type StatusbarItem } from '.'
-import { openLintPanel, forEachDiagnostic } from '@codemirror/lint'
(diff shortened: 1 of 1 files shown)


nathanlesage:
This will likely lead to even more frustration than the current behavior since the internal state of the status bar is prone to get out of sync with the lint status.

I just checked the linter source code, and the commands (open panel and close panel) return false if they couldn't turn on or off the lint panel. So what I suggest is you check what these two commands return and use that to implement a toggle.

I recommend you check the source code of the Codemirror linter to see what works!

benniekiss:
I looked at the codemirror source, and I don't think `openLintPanel` returns anything but true, but correct me if I'm wrong. `closeLintPanel` does return false if the panel is open. Tested the new function and it is working for me!

nathanlesage:
Looks good! Thank you!

PR #5848, opened 2025-08-14
[Feat] Add toolbar popover for pandoc's fenced divs and bracketed spans

benniekiss:
## Description

This PR adds a popover toolbar item to insert Pandoc fenced divs or bracketed spans with the configured identifier, classes, and attributes. Two tools, highlight and underline, are added to the formatting toolbar related to bracketed spans to help illustrate the functionality. 

## Changes

A popover is added to the toolbar which provides three input forms to configure an identifier, classes, and attributes to assign to the fenced div or bracketed span. Fenced Divs or Bracketed Spans can be selected with the tab bar at the top of the popover, and they are inserted via the button at the bottom.

I chose the `drag-handle` icon since it is similar to the `:::` fence syntax, but I am open to any suggestions for a different one.

Two tools are added to the formatting toolbar to insert highlighting or underlines using bracketed spans according to pandoc syntax.

The popover click handling is refactored so that clicks on the toolbar itself are ignored. This allows the user to click the icon button icon again to close the popover, and it allows the user to drag the window and keep the popover open. Previously, clicking the popover icon would cause the popover to close and immediately reopen, which was counter intuitive to me. 

## Additional information

I was inspired by the Quarto editor's similar functionality. 

I was not certain about how to apply the formatting, so I used the existing `applyInlineMarkup` function, but I am open to refactoring the actual format handling.

A next step after this PR would be to implement rendering of fenced divs and bracketed spans. There is already a function to parse pandoc attributes, so the scaffolding is available.

Images:

Popover:
Screenshot 2025-08-14 at 11 43 22

Formatting Bar:
Screenshot 2025-08-14 at 11 43 33

Underlining:
Screenshot 2025-08-14 at 11 43 42

Tested on: 
macOS 15.6


nathanlesage:
This PR is quite large, but I think I'll focus on this one next. I'll give you some time to update it, though, first :D

benniekiss:
It really has! I'm happy to split it into two, the renderer/parser and then the UI components, but otherwise, I'll get it rebased today

benniekiss:
Should be pretty cleaned up at this point.

~~The only issue I've noticed is that in the print preview/md2html function, the closing `PandocDivMark` keeps being printed as a `Paragraph` node, and I cant track down why. The opening mark never shows, and the way the mark is registered is similar to `QuoteMarks`, so I would expect it not to appear.~~ It's a lezer parsing bug, but I've submitted a patch that should fix it!

Otherwise, this has been working well for me!

nathanlesage:
Wonderful! I've been struggling in the past week to maybe give this PR the attention it deserves, but given that I'm still full throttle in application cycle I think to make it faster splitting it up might help so that I can give better feedback, if that works for you?

I think having this support is nice in 4.0!

benniekiss:
That definitely works! I'll have this split into parsing/rendering and UI components sometime later today or tomorrow.

PR #5850, opened 2025-08-15
Fix for unit test errors caused by file path platform variances

sysopod:
## Description
This PR should solve unit test errors caused by expected file path values not accounting for win32 platforms.

## Changes
Changed extractBibTexAttachments, makeValidUri, and mergeEventsIntoTree unit tests so that when they are run on win32 platforms, the expected values are adjusted to match what can be expected on win32 platforms.

## Additional information
The makeValidUri unit test is impacted by the extra '/' inserted into Windows URI paths for local files by the makeValidUri function. The fix implemented in this change addresses that by removing the extra '/'. When that extra '/' is no longer being used to make Windows URI paths for local files work, this will be unnecessary and should be removed.

Tested on: Windows 10 Home, Version 10.0.19045 Build 19045

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@ the leading period of the extension, so do not write `$FILENAME.$EXTENSION`.)
 ## Under the Hood
 
 - Update Electron to version `37.2.5`.
+- Changed unit test path handling to be normalized across platforms, mainly to handle differences with win32 platforms
 
 # 3.6.0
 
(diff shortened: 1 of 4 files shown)


nathanlesage:
Hi, thank you for getting to work on this! I really appreciate it!

However, I think what we should aim for is to not modify the existing unit tests within the tests itself, but rather do it in the tests itself. There are two ways I could think about what we could do here:

Either, we add another flag that indicates whether a unit test should expectedly fail on windows/unix (if the functions are required to produce different paths on the operating systems). In this case, we could modify the tests insofar that it expects them to pass on one OS and expects them to fail on another OS. Then we can add additional tests that will work on windows but fail on Unix.

The other alternative is to adapt the functions themselves to produce the same paths on all OSs. However, given that the functions appear to work as expected, I'd assume that maybe the first alternative is what we need to go with.

Feel free to think about this and implement whatever alternative makes sense! No stress, however, because as of now everything still works :)

PR #5851, opened 2025-08-16
[Feat] Set the editor dark mode separately

benniekiss:
## Description

This PR allows a user to set the editor dark mode separately from the main UI.

## Changes

A radio button setting is added to the appearance settings pane which allows a user to set the editor theme mode to either `light`, `dark`, or `match`. If the mode is `match`, it will follow the main UI dark mode. 

A helper function was added to `dark-mode.ts` to get the right dark mode setting for the editor, and several calls to `darkMode()` and `darkModeEffect` were updated to use this helper.

In terms of CSS, I added a `background-color` setting to several of the preset themes, otherwise, the incorrect background was shown since they were inheriting a transparent background. Likewise, I moved the styling of the `cm-statusbar` and `cm-formatting-bar` so that they took on the styling of the main UI, not the editor. I think this gives a more consistent feel.

## Additional information

MS Word and LibreOffice both offer the option to have a dark or light page when setting dark mode, and Pages defaults to a white page, otherwise a user can chose the `Blank Black` template for a black page. Additionally, this functionality is important in terms of accessibility.

When I'm writing, I prefer a dark UI but a light page. I was looking into achieving this with only CSS, but there were several components that needed reconfigured, and I would have had to do it for every theme I wanted to use. I decided to implement it in the app.

Images:

Light Mode / Dark Editor:

Light Mode / Dark Editor

Dark Mode / Light Editor:

Dark Mode / Light Editor

Tested on:  macOS 15.6

### DIFF
--- a/source/common/modules/markdown-editor/theme/dark-mode.ts
+++ b/source/common/modules/markdown-editor/theme/dark-mode.ts
@@ -121,6 +121,25 @@ const darkModeSwitcher = EditorState.transactionExtender.of(transaction => {
 })
 
 /**
+ * This public helper function returns a boolean depending
+ * on whether the editor should use a light theme, dark theme,
+ * or match the main UI.
+ *
+ * @param   {boolean}   darkMode      The main UI dark mode setting
+ * @param   {string}    editorTheme   The theme of the editor. 'match' will
+ *                                    match the value of darkMode
+ *
+ * @return  {boolean}                  Whether to use a dark theme
+ */
+export function useDarkModeEditor (darkMode: boolean, editorTheme: 'match'|'dark'|'light' = 'match'): boolean {
+  return {
+    match: darkMode,
+    light: false,
+    dark: true,
+  }[editorTheme]
+}
+
+/**
  * An extension that enables a CodeMirror editor to quickly switch between light
  * and dark themes. Pass an initial configuration to the function, and continue
  * to configure the extension with the darkModeEffect.
--- a/source/common/modules/markdown-editor/theme/dark-mode.ts
+++ b/source/common/modules/markdown-editor/theme/dark-mode.ts
@@ -131,12 +131,15 @@ const darkModeSwitcher = EditorState.transactionExtender.of(transaction => {
  *
  * @return  {boolean}                  Whether to use a dark theme
  */
-export function useDarkModeEditor (darkMode: boolean, editorTheme: 'match'|'dark'|'light' = 'match'): boolean {
-  return {
-    match: darkMode,
-    light: false,
-    dark: true,
-  }[editorTheme]
+export function useDarkModeEditor (appDarkMode: boolean, editorTheme: 'match'|'dark'|'light' = 'match'): boolean {
+  switch(editorTheme) {
+    case 'match':
+      return appDarkMode
+    case 'light':
+      return false
+    case 'dark':
+      return true
+  }
 }
 
 /**
--- a/source/win-preferences/schema/appearance.ts
+++ b/source/win-preferences/schema/appearance.ts
@@ -69,6 +69,16 @@ export function getAppearanceFields (config: ConfigOptions): PreferencesFieldset
       title: trans('Editor Theme'),
       infoString: trans('Select a color and font theme for the editor.'),
       group: PreferencesGroups.Appearance,
+      titleField: {
+        type: 'select',
+        model: 'darkModeEditor',
+        inline: true,
+        options: {
+          match: trans('Automatic'),
+          light: trans('Light Theme'),
+          dark: trans('Dark Theme')
+        }
+      },
       help: undefined, // TODO
       fields: [
         { type: 'separator' },
--- a/source/win-preferences/schema/appearance.ts
+++ b/source/win-preferences/schema/appearance.ts
@@ -74,7 +74,7 @@ export function getAppearanceFields (config: ConfigOptions): PreferencesFieldset
         model: 'darkModeEditor',
         inline: true,
         options: {
-          match: trans('Automatic'),
+          match: trans('Follow App'),
           light: trans('Light Theme'),
           dark: trans('Dark Theme')
         }
--- a/source/app/service-providers/config/config-validation.ts
+++ b/source/app/service-providers/config/config-validation.ts
@@ -16,6 +16,7 @@ import { trans } from '@common/i18n-main'
 
 const RULES = {
   darkMode: 'required|boolean|default:false',
+  darkModeEditor: 'required|string|in:match,light,dark|default:match',
   autoDarkMode: 'required|string|in:off,system,schedule,auto|default:off',
   fileMeta: 'required|boolean|default:true',
   sorting: 'required|string|in:natural,ascii|default:natural',
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -60,6 +60,7 @@ export interface ConfigOptions {
   appLang: string
 
   darkMode: boolean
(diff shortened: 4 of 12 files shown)


nathanlesage:
Just as a heads-up: I haven't yet looked at this PR, because I am not yet fully convinced by it, because I personally believe this is ugly. But I know that there are many people out there who do want such a setting, which is why I'm actively working towards accepting this and being able to merge it. Apologies, I sometimes have this, it's not something you did wrong!

benniekiss:
Take your time! I'll be using this on my own end whether you accept it or not, so it'll stay fairly up-to-date whenever you get to it.

ariadesu:
Perspective from another user: I think both the mix-light/dark screenshots look better than both the light mode and the dark mode. So there are people who don't think it's ugly. But I think the UI of having a dark mode global toggle and then a radio box for light, dark, match editor is confusing and bad.

Wouldn't a better, more intuitive UI be to have twice as many themes, and to group half of them into a label "Light mode themes" and half into a "dark mode themes"? The app would automatically switch to the equivalent Light/Dark version of your selected theme, but you can also manually choose a conflicting one. There would be less room for confusion and the theme selector is already nice and visual and discovery-friendly.

benniekiss:
I'm not opposed to showing all the  themes, and they're already grouped in the code, but I don't think that would be helpful in how the themes are currently handled. 

there is still the need for a toggle because the dark theme can be configured to match the system, or it can be scheduled to change based on time, and I think its bad UX to make a user keep selecting their theme choice each time the dark mode changes.

ariadesu:
Configuring scheduled/heuristics switching can be a single check mark next to the theme or next to the main toggle for dark-mode chrome. That would still be less to parse (or it would at least give the content hierarchy).

Mockup:


If you don't like this, then maybe consider a new UI that doesn't present the option for the chrome theme as the "dark mode", but just another setting with the same importance as the content theme. It could be as simple as two drop downs or radio buttons, each with light, dark, system. The schedule UI would need a bit of rethinking with that approach.

benniekiss:
I like the checkbox and the mockup you provided! I'll see what implementing that looks like

benniekiss:
So I thought about it and looked into what changes would need to be made, and I won't be pursuing the all-themes-visible approach. It would require a refactor that I am not interested in pursuing at the moment.

I will stick with the radio toggle for now, but I am moving it under the editor theme section, not the dark mode section. I may also change the toggle to a drop-down menu.

benniekiss:
Updated screenshots:

Screenshot 2026-01-11 at 19 54 40

Screenshot 2026-01-11 at 19 55 15

nathanlesage:
I agree with @benniekiss here. The themes primarily deal with fonts and colors, so duplicating them not only makes the UI more bloated for little reason, but also confuses the coloring with the dark/light idea.

What I would propose is leave the theme selection as-is entirely, and instead move an editor theme display toggle to the App theme indicator. This way we collocate the dark/light switch of the app with the editor selection, and keep the themes separate.

One suggestion though: I remember many editors implement this as a dropdown select, where the three options are "Match application" (default), "Light", and "Dark"

benniekiss:
I'm not against putting the setting back up with the dark mode setting, but I think it makes more sense in the `Editor Theme` group. How does this look?

Screenshot 2026-01-12 at 06 52 42
Screenshot 2026-01-12 at 06 52 49

source/common/modules/markdown-editor/theme/dark-mode.ts
```diff
@@ -120,6 +120,25 @@ const darkModeSwitcher = EditorState.transactionExtender.of(transaction => {
...
+ *
+ * @return  {boolean}                  Whether to use a dark theme
+ */
+export function useDarkModeEditor (darkMode: boolean, editorTheme: 'match'|'dark'|'light' = 'match'): boolean {
+  return {
+    match: darkMode,
+    light: false,
+    dark: true,
+  }[editorTheme]
+}
```
nathanlesage: This looks a bit over-engineered. Can we increase the verbosity a little bit by actually just checking the `editorTheme` variable? Doesn't look as sleek, but makes understanding the code in the future easier.

source/common/modules/markdown-editor/theme/main-override.ts
```diff
@@ -33,26 +33,12 @@ export const mainOverride = EditorView.baseTheme({
...
-  '.cm-panels .cm-button': {
-    backgroundImage: 'none',
-    backgroundColor: 'inherit',
-    borderRadius: '6px',
-    fontSize: '13px'
-  },
-  '&light .cm-panels .cm-button': {
-    backgroundColor: 'white',
-    borderColor: '#aaa'
-  },
```
nathanlesage: What do we lose with removing these rules entirely?

source/win-preferences/schema/appearance.ts
```diff
@@ -69,6 +69,16 @@ export function getAppearanceFields (config: ConfigOptions): PreferencesFieldset
       title: trans('Editor Theme'),
       infoString: trans('Select a color and font theme for the editor.'),
       group: PreferencesGroups.Appearance,
+      titleField: {
+        type: 'select',
+        model: 'darkModeEditor',
+        inline: true,
+        options: {
+          match: trans('Automatic'),
```
nathanlesage: Maybe "Follow app"?

source/common/modules/markdown-editor/theme/dark-mode.ts
```diff
@@ -120,6 +120,25 @@ const darkModeSwitcher = EditorState.transactionExtender.of(transaction => {
...
+ * on whether the editor should use a light theme, dark theme,
+ * or match the main UI.
+ *
+ * @param   {boolean}   darkMode      The main UI dark mode setting
+ * @param   {string}    editorTheme   The theme of the editor. 'match' will
+ *                                    match the value of darkMode
+ *
+ * @return  {boolean}                  Whether to use a dark theme
+ */
+export function useDarkModeEditor (darkMode: boolean, editorTheme: 'match'|'dark'|'light' = 'match'): boolean {
```
nathanlesage: Maybe rename `darkMode` to `appDarkMode` or so to differentiate it better from the `editor` dark mode settings?

nathanlesage (changes requested):
Sooo, it took quite a while – apologies – but I think it looks sound so far. Just a few comments for understanding. Regarding the CSS changes, I'll just be pulling the PR sometime in the coming week and check it out. Any problems can also be spotted during the nightlies, I believe.

benniekiss:
If you're willing, I would actually prefer addressing #6229 first, as that PR would reduce a lot of the diff in this one, and make any CSS issues easier to solve

nathanlesage:
True, I forgot about that one -- I haven't looked at the many new PRs recently 👀

I'll give that one precedence then!

benniekiss:
I need to cleanup some merge issues and address review comments, so I'll ping once this is ready

benniekiss:
@nathanlesage if you wouldn't mind giving this a review, I think it's ready.

I haven't noticed any missing styling after resolving the merge conflicts, so fingers crossed none pop up

nathanlesage:
I was very motivated half an hour ago, but I think I will have to yet again postpone this PR of yours, but it's the next on my list. Time to finally break my hard stance against things I personally don't like too much :P 

Also, if you don't mind, could you take a look at the custom shortcuts implementation? I'll get to this hopefully tomorrow or over the weekend.

nathanlesage (approved):
Just took a look over this again, and it looks fine for me. Anything else you'd like to add or can we merge?

benniekiss:
this is good to merge for me!

nathanlesage:
Perfect, then off we go!

Issue #5518, opened 2024-12-14
Performance Impact when Resizing Sidebars & Splitting the Editor


nathanlesage:
### Description

I'm opening this issue as I have faced some performance impacts when resizing the sidebars (any of them) and splitting the editor, both as a reminder to myself and for transparency for you.

**Issue 1**: Lags during resizing

When resizing either the file manager or the sidebar with multiple (long-ish) documents open, during the resize operation you may notice lags. I have already tracked it down to very inefficient configuration updates. Since resizing the sidebars will update the corresponding config values in the main config, this will in turn tell all renderer processes to fetch the new configuration, too. Since that config is by large parts passed into all open editors, and they do a lot of update work, this will induce lag.

Mitigation: Move the sidebar size config into the document provider. This will then also mean that, having multiple windows open, all of them can remember their own sidebar sizes instead of sharing one global one. This may take some time since the document manager isn't yet prepared for such updating procedures.

**Issue 2**: Lags during editor splitting

When moving editor tabs to either side of the editor to create a split, this will induce a lag, as all editors in the affected leaf will have to be reconstituted. This is expected, as there's not really a way to retain most of the data, but it shouldn't be *this* laggy. Specifically, I believe that the similar performance degradation as with issue 1 are responsible, which is why I decided to add it here, too.

### Reproducing

**Issue 1**: Try resizing the file manager or sidebar. The lag will be more noticeable the more files are open.

**Issue 2**: Try splitting an editor. The more documents are open in the affected editor, the larger the lag will be.

### Zettlr version

3.3.1

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [X] ARM 64bit (includes Apple Silicon)

### Operating System Version

macOS Sequoia 15.2

### Additional Information

Do you experience similar lags in other parts of the app? Feel free to chime in below so that I can get a better view of what is going on. I won't be able to fix everything in this issue, but it may help knowing how well the app performs right now.

nathanlesage:
Did some digging, and I believe there are two things we want to do here:

1. Disentangle the big `configField` and instead move everything into the various plugins as they need it
2. The TableEditor is the biggest resource hog. Let's see if the rewrite will already improve things.

benniekiss:
[...]

PR #5857, opened 2025-08-19
Improve performance in large documents

benniekiss:
## Description

This PR implements an incremental word counter and fixes an event trigger for the citations sidebar tab.

## Changes

The `update` method of the `countField` statefield was refactored to incrementally count the words and characters of the document rather than recount the whole document on every change. It uses the codemirror [`iterChanges`](https://codemirror.net/docs/ref/#state.ChangeSet.iterChanges) method to only count the words/characters in the changed text.

There was a small error in the `ReferencesTab.vue` file which was causing the bibliography to be recomputed on every document modification. This caused really large delays in typing when editing large documents. The comment in the code says that it should be called on document save, not modification. So I corrected the event trigger. 

## Additional information

I initially brought this up [here](https://github.com/Zettlr/Zettlr/issues/5518#issuecomment-3198424344), but since the performance improvement was so great, I cleaned up the patches and decided to submit them. 

The machine I tested this on is an M2 Max macbook pro

Without this PR, I struggled to type in an 80,000 word document.

With this PR, I'm able to type without delay in a ~1,000,000 word document. However, I have to disable the markdown linter and spellcheckers. The initial startup does take a long time, maybe 1 to 3 minutes, but once the TOC has been loaded and the markdown rendered, it is easy to type. 

This does not address all of the lag issues, and you will likely still experience some problems, but this addresses a major cause of the editor's slow down and should help performance in general.

Tested on:  macOS 15.6


benniekiss:
Most recent commit improves selection lag for large documents. When first making a selection, the AST for the whole document is calculated. Making it so only the selection range is turned into an ast makes it much snappier.

benniekiss:
Recent commit implements linting for only the changed regions of a text. It can likely be refactored some, but it seems to be working quite well for me.

source/common/modules/markdown-editor/linters/md-lint.ts
```diff
@@ -74,10 +130,73 @@ function placeToOffset (place: Point|Position|undefined, source: Text): { from:
...
+    const composedChanges = value.composeDesc(transaction.changes.desc)
+
+    return composedChanges
+  }
+})
+
+function getSurroundingLinePosition (view: EditorView, from: number, to: number): { from: number, to: number } {
+  const totalLines = view.state.doc.lines
+  const fromLine = view.state.doc.lineAt(from).number - 5
+  const toLine = view.state.doc.lineAt(to).number + 5
```
benniekiss: the `-5` and `+5` were arbitrary decisions, so they could be expanded, reduced, or made dynamically

benniekiss:
With the changes to the TOC generation, performance should be usable. I think the TOC performance could still be improved some, but the performance improvement at this point is substantial.

What's left is probably some refactoring to move the logic into separate/reusable functions. I don't know where the best place to put these is, however.

 I also am working on converting the languagetool linter to operate on changed text, similar to the spellchecker.

benniekiss:
At this point, this PR could probably be split into different PRs, but the individual performance impact of each change is not really noticeable unless they are all tested together. 

I'm happy to separate the changes, but this one should be kept open for the time being so that the overall improvement can be tested.

nathanlesage:
Before proceeding we should definitely talk about some of the changes, as I am very skeptical of  especially how you've solved the linter issues. I am strongly in favor of having the linters run across the entire document, and move that to a worker process rather than risk missing important things like non-closed fences code blocks which happen quite often.

However, I won't be able to do a full code review before I have defended my thesis as I don't have time unfortunately...

benniekiss:
No rush whatsoever, and good luck on your thesis!

I agree with your point about the markdown context, and I hadn't considered unclosed divs, etc. Moving it to a background process would be preferred, but I don't have experience with that. I'll be looking at ways to more intelligently expand the markdown context in the meantime.

---

quick edit - but it may also be helpful to add a "Force Lint" button to the toolbar, or somewhere else, which would trigger a lint of the whole document.

benniekiss:
Cleaned up the commits to make the features separate per commit

benniekiss:
Closing because https://github.com/Zettlr/Zettlr/pull/5903 has made most of this unnecessary.

PR #5866, opened 2025-08-28
feat: Text transform to remove ligatures

### LINKED ISSUE #5820: Text transform to remove ligatures.
### Description

Some documents contain ligatures—characters that take the place of two or more connected letters. These look more polished in print, but when cutting and pasting from these texts, users will want the more typical sequence of characters. I.e, `official` rather than `ofﬁcial` (note the subtly connected `f` and `i` characters in the latter). Not all typefaces support ligatures, so these may interfere with exports. And when ligatures aren't removed in pasted text, they may also throw off Zettlr's spell checker, which is looking for the typical sequence of characters.

I suggest a transform that eliminates ligatures from selected text. For reference, here's a list of common ligatures from [this government website](https://lhncbc.nlm.nih.gov/LSG/Projects/lvg/current/docs/designDoc/UDF/unicode/NormOperations/splitLigatures.html). The columns represent, respectively:

- Unicode
- Mapped ASCII
- Character
- Unicode Name 

```
U+00C6	AE	Æ	LATIN CAPITAL LETTER AE
U+00E6	ae	æ	LATIN SMALL LETTER AE
U+0132	IJ	Ĳ	LATIN CAPITAL LETTER IJ
U+0133	ij	ĳ	LATIN SMALL LETTER IJ
U+0152	OE	Œ	LATIN CAPITAL LETTER OE
U+0153	oe	œ	LATIN SMALL LETTER OE
U+FB00	ff	ﬀ	LATIN SMALL LIGATURE FF
U+FB01	fi	ﬁ	LATIN SMALL LIGATURE FI
U+FB02	fl	ﬂ	LATIN SMALL LIGATURE FL
U+FB03	ffi	ﬃ	LATIN SMALL LIGATURE FFI
U+FB04	ffl	ﬄ	LATIN SMALL LIGATURE FFL
U+FB05	st	ﬅ	LATIN SMALL LIGATURE LONG S T
U+FB06	st	ﬆ	LATIN SMALL LIGATURE ST
```

Many more are possible, per [this more comprehensive listing](https://en.wikipedia.org/wiki/Ligature_(writing)#Ligatures_in_Unicode_(Latin_alphabets)) on Wikipedia. But a transform for common cases would be super helpful in its own right.

### Proposed Changes

See above.

### Caveats

Interestingly, whether ligatures show up in copy-pasted text is sometimes dependent on the application from which text is copied. I had an offending PDF (which I can share on request) and the ligatures in the document showed up in pastes from one PDF viewer, but not from the same PDF when I cut and pasted text from two other PDF applications.

No

### Zettlr Version

None

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Manjaro (Rolling)

### Additional Information

Thank you!

richdouglasevans:
Closes #5820

## Description
Adds a text transform to substitute/replace/expand ligatures.

https://github.com/user-attachments/assets/0fe7ab88-85d6-4d24-a33c-f79c6e0dfaf4

## Changes
I added another of the text transforms that we've had a spate of recently.

## Additional information

This is marked as a draft PR only because I think @wideaperture may want to nail down:

* the exact label for the UI menu item; currently it is _Substitute ligatures_
* firmly decide on the exact set of ligatures to be substituted/replaced/expanded
* decide where/how the transform is to be accessed
  * via the existing popup menu
  * a new _Transforms_ submenu of the existing _Edit_ menu
  * both?
  * something else entirely

Purely to facilitate manual testing the transform is actioned by an item in the context menu.

Tested on: Windows 11.

### DIFF
--- a/source/common/modules/markdown-editor/commands/transforms/substitute-ligatures.ts
+++ b/source/common/modules/markdown-editor/commands/transforms/substitute-ligatures.ts
@@ -1,0 +1,96 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        TextTransform
+ * CVM-Role:        CodeMirror command
+ * Maintainer:      Rich Douglas
+ * License:         GNU GPL v3
+ *
+ * Description:     This plugin contains a single text transformer.
+ *
+ * END HEADER
+ */
+import { transformSelectedText } from './transform-selected-text'
+
+type Substitution = {
+  target: string
+  replacement: string
+}
+
+const SUBSTITUTIONS_LIGATURE: ReadonlyArray<Substitution> = [
+  {
+    target: '\u00C6',
+    replacement: 'AE',
+  },
+  {
+    target: '\u00E6',
+    replacement: 'ae',
+  },
+  {
+    target: '\u0132',
+    replacement: 'IJ',
+  },
+  {
+    target: '\u0133',
+    replacement: 'ij',
+  },
+  {
+    target: '\u0152',
+    replacement: 'OE',
+  },
+  {
+    target: '\u0153',
+    replacement: 'oe',
+  },
+  {
+    target: '\uFB00',
+    replacement: 'ff',
+  },
+  {
+    target: '\uFB01',
+    replacement: 'fi',
+  },
+  {
+    target: '\uFB02',
+    replacement: 'fl',
+  },
+  {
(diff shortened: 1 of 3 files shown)


richdouglasevans:
Closing this 'cos I don't have the time right now to get this over the line.

PR #5868, opened 2025-08-28
Update danish

bstordrup:
## Description
Adding Danish translations where there were no translations, and updated a few other trnaslations

## Changes
Changed only in `static/lang/da-DA.po` 

## Additional information

### DIFF
--- a/static/lang/da-DA.po
+++ b/static/lang/da-DA.po
@@ -164,32 +164,36 @@ msgstr "Kan ikke hente SHA256 kontrolsummer: %s"
 
 #: source/app/service-providers/updates/index.ts:527
 msgid "Could not accept data for application update: Write stream is gone."
-msgstr ""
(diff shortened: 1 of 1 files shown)


nathanlesage:
Hi, thank you so much! Unfortunately, something appears to have gotten wrong with the PR. You should be able to see the error message. Once you are able to fix this, we can merge this!

bstordrup:
Could not parse file: Error parsing PO data: Invalid key name "Services" at line 934. This can be caused by an unescaped quote character in a msgid or msgstr value.

I will see if i can have a look soon.

bstordrup:
@nathanlesage, Found and fixed the issue. An ending quote was missing on line 931.

nathanlesage:
Looks good, linter doesn't complain — thank you!

PR #5870, opened 2025-08-29
Refactor PopoverWrapper `onClick` to ignore clicks on the toolbar button

benniekiss:
## Description

This PR enables clicking on popover toolbar buttons to close the popover.

## Changes

In the popover wrapper, the `onClick` method was refactored to ignore clicks on the popover toolbar button. This allows clicking the button to actually close the popover. 

Previously, clicking the toolbar button while the popover was open would cause the popover to close and immediately re-open.

## Additional information

Previous Behavior:

https://github.com/user-attachments/assets/7556715b-792a-44e1-9a23-ae837805b337

PR Behavior:

https://github.com/user-attachments/assets/c46a87be-48f6-4311-81d2-5222aae48a6f

Tested on:  macOS 15.6

### DIFF
--- a/source/win-main/PopoverWrapper.vue
+++ b/source/win-main/PopoverWrapper.vue
@@ -50,8 +50,8 @@ function onClick (event: MouseEvent): void {
     return
   }
 
-  // Ignore clicks on the popover button
-  // so that clicking it closes the popover
+  // Ignore clicks on the popover's target element
+  // so that clicking that closes the popover.
   if (props.target.contains(target)) {
     return
   }


source/win-main/PopoverWrapper.vue
```diff
@@ -36,18 +36,23 @@ onBeforeUnmount(() => {
...
   // Background for the following line:
   // https://github.com/Zettlr/Zettlr/issues/554
-  if (popupWrapper.value.contains(event.target as Node)) {
+  if (popupWrapper.value.contains(target)) {
+    return
+  }
+
+  // Ignore clicks on the popover button
+  // so that clicking it closes the popover
+  if (props.target.contains(target)) {
```
nathanlesage:
```suggestion
  // Ignore clicks on the popover's target element
  // so that clicking that closes the popover.
  if (props.target.contains(target)) {
```

Popovers can also occur on items, etc, so in those cases that would also work. Just to be semantically correct!

nathanlesage:
Wonderful!

PR #5871, opened 2025-08-29
Improve ToC drag and drop new line handling

benniekiss:
## Description

This PR improves ToC drag and drop new line handling. It sanitizes new lines in ToC sections to ensure appropriate spacing when moving them.

## Changes

the `moveSection` function was refactored to sanitize the section to ensure new lines are handled correctly.

Previously, moving sections would result in an additional new line above the moved section. This was because the section selection extended from the beginning of the section to the end of the last line, but this results in a blank line. By extending the selection to the beginning of the next section, we remove this additional line. 

Secondly, moving sections could result in them being appended to the end of the last line of the previous section, if they were not spaced correctly.

There is one edge case which I cannot seem to figure out, and it can be seen in the video below at about the 20 second mark. But if a section is moved above the previous section, but there is no new line between them, then the previous section will have no new line between it and the next section once it's moved.

## Additional information

Previous behavior:

https://github.com/user-attachments/assets/6640edff-18be-4500-9325-c32b28cf72c1

PR behavior:

https://github.com/user-attachments/assets/579f1c85-1735-4506-af59-c95ee2087dea

Tested on:  macOS 15.6

### DIFF
--- a/source/common/modules/markdown-editor/commands/move-section.ts
+++ b/source/common/modules/markdown-editor/commands/move-section.ts
@@ -1,6 +1,30 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        Table of Contents Commands
+ * CVM-Role:        Extension
+ * Maintainer:      Hendrik Erz
+ * License:         GNU GPL v3
+ *
+ * Description:     Commands to modify a document and its table of contents.
+ *
+ * END HEADER
+ */
+
 import type { StateCommand } from '@codemirror/state'
 import type { ToCEntry } from '../plugins/toc-field'
 
+/**
+ * This function returns a StateCommand which can be used to
+ * move a table of contents section.
+ *
+ * @param   {ToCEntry[]}    toc   The table of contents
+ * @param   {number}        from  The start line number of the section
+ * @param   {number}        to    The line number to move the section to
+ *
+ * @return  {StateCommand}        A StateCommand which moves the section.
+ */
 export function moveSection (toc: ToCEntry[], from: number, to: number): StateCommand {
   return ({ state, dispatch }) => {
     const entry = toc.find(e => e.line === from)
--- a/test/move-section.spec.ts
+++ b/test/move-section.spec.ts
@@ -2,12 +2,12 @@
  * @ignore
  * BEGIN HEADER
  *
- * Contains:        Tests for the addSpacesAroundEmdashes function
+ * Contains:        Tests for the moveSection function
  * CVM-Role:        TESTING
- * Maintainers:     Rich Douglas
+ * Maintainers:     Hendrik Erz
  * License:         GNU GPL v3
  *
- * Description:     This file tests the addSpacesAroundEmdashes function.
+ * Description:     This file tests the moveSection function.
  *
  * END HEADER
  */


nathanlesage:
> There is one edge case which I cannot seem to figure out, and it can be seen in the video below at about the 20 second mark.

I'm not entirely sure I properly saw that edge case.

That being said, I believe there is a good way to ensure the command works as expected, and that is adding a test case for it.

You could then also move the `moveSection` to an actual Codemirror command. This allows actual testing of the function (I seriously have no idea why it's a class method of the Markdown Editor, and what I was thinking).

So, do you think it would be possible to add the following two things:

1. Move the method into an actual codemirror command (suggest the `commands` folder. I wouldn't add this to the `markdown.ts` file because that's already very long.)
2. Add a test case to see what your edge case appears to be doing. If that test fails, I can have a look, too, and maybe I can figure out what goes wrong!

benniekiss:
I can definitely move it.

Regarding the edge case, I'm not so certain whether it should be, or can be, sanitized, but I'll see about creating a test for it. I don't think it would result in the `moveSection` command being the cause of any formatting errors, rather it doesn't sanitize a user-generated formatting error. It happens at the 23 second mark in the PR video.

nathanlesage:
Ahh, I see, the move section command didn't insert a new line when you exchanged the sections, I see it now.

Well, as long as the sections don't turn unusable (as it was before), I think it's fine for now. We can still see whether we find the cause and improve that (because I personally like that users are being shown how to properly space out headings; I never liked seeing headings without proper spacing, but that's a personal pet peeve of mine)!

benniekiss:
It looks like there already is a separate function with tests, do you remember if this was intended to replace the editor method? https://github.com/search?q=repo%3AZettlr%2FZettlr%20move-section&type=code

nathanlesage:
Oh boy, my memory is a sieve … (but thank you for discovering this!). So, I just checked the git blame, and that function hasn't been touched in five years. So it's safe to say that that was the original function, which is now dead code. (I [confirmed](https://github.com/Zettlr/Zettlr/commit/aafc5a499aa9d9363abf82cc970a02f1e4388c6c) that the now-used function is the one on the editor instance; while the unit tests use the now-defunct utility function that is not used in production.)

So what I would suggest is to just extract the `moveSection`-function from the class, and overwrite the `moveSection`-function with it, and adapt any failing tests.

…

… the more I think about the fact that I just completely forgot a relatively large and important utility function tells you enough about what a Swiss cheese my brain is.

benniekiss:
I know we talked about tests, but I don't think there are currently any tests for editor commands. The current implementation relies on the editor state/view.

I was able to move the function to a separate file in utils, but otherwise, I'm not sure where to go with restructuring this. It might require a larger refactor than anticipated.

nathanlesage:
Ah, looks better. Regarding tests: I believe someone recently already did some test harnessing for EditorViews to test commands; so you might be able to plug into that.

Otherwise: If it works for you, we might also postpone re-implementing the tests if that takes too long and deliver them sometime during the beta. But again, have a look in the test folder, I believe there's utilities for testing Editor Commands.

benniekiss:
Found them! Thank you for confirming that. I'll get something drafted before we merge this

benniekiss:
@nathanlesage I've done a little more refactoring and added some tests. If you could take a look at them when you get the chance, this should be good to merge once you're comfortable!

benniekiss:
Hey @nathanlesage, sorry to bug you again, but do you think this can get merged before the beta?

nathanlesage:
Gnarl, forgot about this PR again 🙄 Merging now! There will be plenty of additional betas in any case … Sorry!

benniekiss:
No problem! I just wanted to make sure this got in since its pretty user-facing

PR #5874, opened 2025-08-30
[Fix] Refactor text alignment keyboard shortcuts

### LINKED ISSUE #5860: `Ctrl-C` (Copy) keyboard shortcut not working on Nightly builds.
### Description

`Ctrl-c` frequently does not work to copy text to the clipboard on recent nightly builds. Selecting `Copy` from the right-click context menu or the `Edit` menu both work fine, and pasting with `Ctrl-V` works same as always. The `Ctrl-Alt-c` (Copy as HTML) and `Ctrl-x` (Cut) shortcuts continue to work as expected.

As you can see in this video, no console output is shown during the glitch:

https://github.com/user-attachments/assets/1ea99016-3aa3-450a-a80a-8dad43add409

At least one other user on Discord reports the same issue and additionally notes:

> select-by-cursor + y is not working in vi mode, when it should, _however_  visual mode + y works as expected.

### Reproducing

On a recent nightly build, select text in the editor and attempt to copy it to the clipboard using the `Ctrl-c` keyboard shortcut. Paste the clipboard contents. Rather than the expected text, the clipboard paste is either empty or contains text copied prior to the most recent attempt.

### Zettlr version

3.6.0-nightly (from August 18, 2025)

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Manjaro (Rolling)

### Additional Information

More discussion and screengrab videos are to be found in the \#help Discord channel ca. August 20th.

benniekiss:
## Description

This PR refactors the text alignment keyboard shortcuts so that they no longer conflict, notably with `Ctrl-C` copy and `Ctrl-L` insert ID.

Closes #5860

## Changes

A `Shift` key was added to each command so that they do not conflict with other main editor commands. 

Center alignment was changed to `Ctrl-Shift-E`, which matches LibreOffice and MS Word (`Ctrl-e`). `Ctrl-Shift-L` and `Ctrl-Shift-R` already match `Ctrl-L` and `Ctrl-R`, respectively.

macOS commands were added.

## Additional information

Tested on:  macOS 15.6


kyaso (commented):
Thank you for the fix, looks good to me! Good point with matching the bindings with MS Word / Libre. Again, from my side everything is fine, I'll let @nathanlesage have the last word, once he's back!

kyaso:
I just realized that `Ctrl-Shift-L` is also bound to "Copy ID", so there might a conflict still. One option would be to remove the "Copy ID" binding, because is there actually anyone who uses that? 😅 (the command would still be accessible through the Edit menu though). But then one could argue "if a feature is there, then there will be at least one user who relies on it"...

benniekiss:
Good catch! I would suggest `Alt`, but that conflicts with the dark mode toggle, which could also be changed. 

EDIT: From my searches, it doesn't seem `Ctr-Alt-D` is bound to anything, so I would offer changing the dark mode toggle to `Ctrl-Alt-D`, and modifying the keybinds to be `Ctrl-Alt-L`, `Ctrl-Alt-E`, and `Ctrl-Alt-R`, which I also don't believe conflict with anything else.

benniekiss:
I decided to change both Copy Link ID and Toggle dark mode. I wanted to keep the `Ctrl-Shift-*` mod for the alignment commands since it is more ergonomic to me, so I moved Copy Link ID to `Ctrl-Alt-L`, and I moved Toggle Dark Mode to `Ctrl-Alt-D`.

nathanlesage:
Thanks for the initiative! Since this would affect two shortcuts (I personally have Cmd+Alt+L for Dark mode switching already in muscle memory) I would like to ask you to do a small, informal survey on Discord, just to see if anybody objects vividly.

In addition, since keyboard shortcuts are relatively important to have correctly mapped, would you be so kind to also change the shortcuts on the docs repository, so that we can keep the documentation accurate here?

Thanks a bunch!

benniekiss:
Sure thing! Would you be opposed to just changing the Copy Link ID shortcut to something else? I Couldn't really think of a good, obvious alternative. I'll definitely ask on Discord when I get the chance.

nathanlesage:
I think we're safe to implement the survey response -> change both keyboard shortcuts as discussed!

benniekiss:
@nathanlesage Would actually test the darkmode and copy link ID commands? I don't think they are working for me because they're conflicting with system-wide commands. `Cmd-Alt-D` is opening the system dock, and I don't know what I'm looking for with the copy link ID--I can't seem to get it working on `develop` either

benniekiss:
`Cmd-Shift-D` works for dark mode, if we can't get around the system shortcut conflict.

EDIT: So the system shortcut can be changed and disabled, so I think it is OK for us to set the Zettlr toggle to `Cmd-Alt-D`

nathanlesage:
Ah damn, that's a good point, and I actually use Cmd-Alt-D quite often 😵‍💫

This is what I hate about shortcuts, there are just too many of them 🫠

My take at this moment would actually be: just remove the alignment shortcuts entirely. We can readd that via the custom shortcuts PR later on. Like, for real, if we will always have conflicts, just keep the existing ones and remove those alignment ones for now ...

benniekiss:
Sounds good to me! I'll update this to just drop the alignment commands. Would prefer me comment them out, but leave them in the code, or delete them entirely?

benniekiss:
I went ahead and commented out all of the relevant code sections to make the diff clear, but I can go ahead and just remove the code if it's too messy

nathanlesage:
Almost: I'd rather not have block comments but individual `//` line comments (with one additional line comment atop of those with a `TODO` keyword). That should make the diff more readable.

Also, the import can be completely removed; the linter will remind us that we have to re-import that command in the future :)

nathanlesage:
Wonderful, merci beaucoup!

PR #5875, opened 2025-08-30
[Fix] Ensure table delimiters include at least one hyphen

### LINKED ISSUE #5830: Setting Align Column Center breaks tables.
### Description

In the Nightly build, it is possible to cause table behaviour to completely break by setting Align Column Center. This causes both **rendering** and **behaviour** breakages, as seen in the below image.
Image

### Reproducing

1. Create a table
2. Turn on Align Column Center in one of the top rows.

Image

### Zettlr version

Zettlr 3.6.0-nightly (Build a7c1978)

### Installation Method

From the Website or GitHub

App Image

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Linux Mint Debian Edition

### Additional Information

_No response_

benniekiss:
## Description

This PR addresses a rendering/parsing issue when an empty cell is aligned. It ensures the delimiter contains at least one hyphen.

Closes #5830

## Changes

When an empty cell is center aligned, it results in a delimiter like so: `|::|`, however, it seems that the markdown parser expects there to be at least one hyphen, according to https://github.github.com/gfm/#tables-extension- 

> The [delimiter row](https://github.github.com/gfm/#delimiter-row) consists of cells whose only content are hyphens (-), and optionally, a leading or trailing colon (:), or both, to indicate left, right, or center alignment respectively.

Likewise, if a person handwrites a table with a single hyphen in the delimiter, further attempts to align an empty cell would result in a delimiter like `|:|`, which is ambiguous.

This PR ensures that there is at least one hyphen in the delimiter. I chose to sanitize to 3 characters so that center, right, and left alignment resulted in the same minimum number of characters.

The formatter also picked up trailing whitespace that was removed.

## Additional information

Current behavior:

https://github.com/user-attachments/assets/94ba1a30-c6f9-4aae-a2b3-35ed5ad69027

PR behavior:

https://github.com/user-attachments/assets/a09144c0-ccd3-411f-a8c5-0e643ba6fd1f

Tested on:  macOS 15.6

**EDIT: expanded the sanitation to account for handwritten tables with one-hyphen delimiters.**

### DIFF
--- a/source/common/modules/markdown-editor/table-editor/commands/tables.ts
+++ b/source/common/modules/markdown-editor/table-editor/commands/tables.ts
@@ -55,7 +55,7 @@ export function setAlignment (alignTo: 'left'|'right'|'center'): (target: Editor
       const delimOffsets = getDelimiterLineCellOffsets(delimLine, delimChar)
       const [ from, to ] = delimOffsets[idx]
       if (alignTo === 'left') {
-        return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(Math.max(to - from, 2)) }
+        return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(Math.max(to - from, 3)) }
       } else if (alignTo === 'right') {
         return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(Math.max(to - from - 1, 2)) + ':' }
       } else {


benniekiss:
Closing this in favor of #5877

PR #5873, opened 2025-08-29
[FEAT] Enable a scrolling toolbar on overflow

benniekiss:
## Description

This PR implements a scrollable area for the toolbar when its content overflows, such as with narrow window widths.

## Changes

The toolbar was refactored so that the buttons appear in a scrollable area. The toolbar only becomes scrollable when content overflows. When content overflows, buttons appear on either side of the toolbar indicating the overflow, and they can be clicked to scroll either left or right. When there is no remaining content to one side, the button will be disabled.

## Additional information

Closes #5022

Behavior:

https://github.com/user-attachments/assets/a1b76f30-f8b0-4c36-a452-c7bfdbd7855e

Tested on:  macOS 15.6 and Fedora 42 (aarch64)

PR #5876, opened 2025-08-31
fix: Consistent layout of toolbar on Windows

richdouglasevans:
The buttons to search for files and toggle the file manager are stacking vertically when the window is shrunk horizontally. This doesn't match the behaviour of those buttons on other platforms where no such stacking occurs.

## Changes
This change updates the CSS so that the layout behaviour is consistent -- no stacking -- across platforms.

https://github.com/user-attachments/assets/9fa85a29-50a7-4348-96e8-b1933a03c364

https://github.com/user-attachments/assets/bfa996f8-96fa-4b2a-a7ac-a68678b9c195

## Additional information
Noticed while reviewing #5873

The glory for this change belongs to @benniekiss 🙇‍♂️ 

Tested on: Windows 11


nathanlesage:
Looks good, thank you!

PR #5877, opened 2025-09-01
[Fix] Refactor table alignment handling

### LINKED ISSUE #5830: Setting Align Column Center breaks tables.

benniekiss:
## Description

This PR refactors table alignment handling to improve functionality and readability. It enables setting no alignment (`---`), tries to enforce a minimum of 3 characters per cell, fixes an error related to center alignment parsing and rendering, and fixes an error with table insertion related to alignment.  

Supersedes  #5875
Closes #5830 

## Changes

This PR addresses inconsistencies in how `left` alignment was handled. In several places, it was used when the actual alignment was unaligned, `---`. This PR corrects it so that all references to `left` alignment result in a delimiter with a colon, `:--`. To compensate, several functions and types were updated to accept an alignment of `null`, which corresponds to `---`, including the `Table` AST node

Cell padding was refactored to ensure that cells contain a minimum of 3 characters, including delimiter cells. See the previous PR for the rationale.

Column alignment on table insertion was refactored. Previously, the alignment was calculated based on the row index, not the column index. This resulted in errors when inserting a table that had more columns than rows, in which case the difference in columns would all be `right` aligned.

The code was generally refactored for readability. Long `if...else` statements were refactored into `switch` statements, and some nested loops were refactored to simpler list constructions.

## Additional information

Tested on:  macOS 15.6

### DIFF
--- a/source/common/modules/markdown-editor/table-editor/commands/tables.ts
+++ b/source/common/modules/markdown-editor/table-editor/commands/tables.ts
@@ -28,7 +28,7 @@ import type { SyntaxNode } from '@lezer/common'
  *
  * @return  {(EditorView): boolean}             A CodeMirror compatible command function
  */
-export function setAlignment (alignTo: 'left'|'right'|'center'): (target: EditorView) => boolean {
+export function setAlignment (alignTo?: 'left'|'right'|'center'): (target: EditorView) => boolean {
   return (target) => {
     const changes = mapSelectionsWithTables<ChangeSpec>(target, ctx => {
       const delimNodes = ctx.tableNode.getChildren('TableDelimiter')
@@ -36,33 +36,37 @@ export function setAlignment (alignTo: 'left'|'right'|'center'): (target: Editor
       if (delimNodes.length > 1) {
         node = delimNodes.find(node => target.state.sliceDoc(node.from, node.to).includes('='))
       }
-  
+
       if (node === undefined) {
         return undefined
       }
-  
+
       const delimLine = target.state.sliceDoc(node.from, node.to)
-  
+
       // TODO: Iterate over all ranges (but only once per row)
       const idx = findColumnIndexByRange(ctx.ranges[0], ctx.offsets.outer)
-  
+
       if (idx === undefined) {
         return undefined
       }
-  
+
       const delimChar = ctx.tableAST.tableType === 'grid' ? '+' : delimLine.includes('|') ? '|' : '+'
       const fillChar = ctx.tableAST.tableType === 'grid' ? '=' : '-'
       const delimOffsets = getDelimiterLineCellOffsets(delimLine, delimChar)
       const [ from, to ] = delimOffsets[idx]
-      if (alignTo === 'left') {
-        return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(to - from) }
-      } else if (alignTo === 'right') {
-        return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(to - from - 1) + ':' }
-      } else {
-        return { from: node.from + from, to: node.from + to, insert: ':' + fillChar.repeat(to - from - 2) + ':' }
+      // ensure that each delimiter contains at least 3 characters
+      switch (alignTo) {
+        case 'left':
+          return { from: node.from + from, to: node.from + to, insert: ':' + fillChar.repeat(Math.max(to - from - 1, 2)) }
+        case 'right':
+          return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(Math.max(to - from - 1, 2)) + ':' }
+        case 'center':
+          return { from: node.from + from, to: node.from + to, insert: ':' + fillChar.repeat(Math.max(to - from - 2, 1)) + ':' }
+        default:
+          return { from: node.from + from, to: node.from + to, insert: fillChar.repeat(Math.max(to - from, 3)) }
       }
     })
-  
+
     if (changes.length > 0) {
       target.dispatch({ changes })
       return true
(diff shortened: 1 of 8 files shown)


nathanlesage:
Thanks for this. However, I do have a question regarding your addition of `null` as an alignment value: Is this necessary? The default assumption of any Markdown table is left aligned text. I made the deliberate choice of then enforcing this, which might then add a left-aligned colon in the delimiter row to add verbosity, but I felt this makes the code a bit more compact, because there are only three, not four alignment options. So what was the rationale for this?

Then, you've used `forEach` to iterate over rows and columns in some places. Why so? I feel the for-let-i==0 construct is a bit more compact, especially since there's no callback function involved.

I do appreciate the switch statement, however, that is indeed a good improvement.

benniekiss:
I definitely think the `null` is necessary for the simple fact there are four delimiters. With the current implementation, there's no clear way to distinguish between `:--` and `---`.  While its implied that `---` could be left aligned, it is not technically. Likewise, I think it makes the code more confusing to bake in this implication. Explicitly setting `left` to mean `:--` is unambiguous and clear to the GFM table specification. The code was also inconsistent in whether `left` meant `---` or `:--`, which I can see causing problems down the road. 

Adding `null` to mean `---` just enforces correctness and makes the table alignments explicit. I think this is also important since Zettlr exports to other document formats that may not make this assumption.

Now, in terms of how these are rendered, there's nothing stopping the rendering from assigning a left-alignment to the `---` delimiter, and I can change the part in the code that sets this.

I think the only change to `forEach` was in the `calculateColSizes` function. I chose `forEach` because it was personally much easier to read and write. Currently, the loop moves through the AST in an inverted fashion, going through the number of columns then the number of rows, which was kinda difficult for me to understand at first. The `for (...)` is more compact, but I think that works against understanding what is going on.  

The `forEach` just makes the code cleaner and easier to read, for me at least, because we're no longer focusing on tracking where we are in the loop, rather just what we are doing within each loop. It also untangles the logic so we step through the list linearly.

nathanlesage:
> While its implied that `---` could be left aligned, it is not technically.

Agreed. I just checked all sources, and none mention that `---` is explicitly left aligned, so your point stands!

… although I'm now thinking … shouldn't we then maybe switch to `undefined` instead of `null`? This is definitely not a hill I want to die on because my personal usage of `undefined` vs. `null` is an utter mess. I've been trying to improve the semantic distinction, but … yeah … I don't know, semantically `undefined` sounds more correct ("Alignment is not specified/undefined"), but … I'll leave it up to you.

***

Now, for the `forEach`: I absolutely see your point. However, when no index is required, one can also use a `for (const row of rows)`. I do see your point that it's not a clear-cut, objective choice. Here are some arguments to ponder:

* Pro `for`-loop:
    * used across the entire codebase, which means consistency. This would be the very first occurrence of a `forEach`-loop in the entire codebase. I explicitly avoided `forEach` for everything.
    * `for`-loops are likely more optimized than calling a callback. You essentially have an anonymous scope vs. a function that has less guarantees for the JIT compiler (because you could pass whatever you want in there).
* Pro `forEach`-loop:
    * Newer, and might be faster
    * Subjectively more readable (although I personally don't think so, but I see that people can see it like so)

My personal cons for the `forEach`-loop:

* Easily confused with `map`-constructs. A `forEach`-loop cannot return anything, a `map` can.
* It is in my (subjective) opinion less clear if it interacts with the surroundings (filling up other arrays, etc.).

Now, I am 100% with you in that a `forEach`-loop does exactly what it says on the tin, and I agree that there is no objective reason to prefer any over the other.

However, having thought about all of this now, **I would like you to move back to `for`-loops, simply because `forEach` is used nowhere, and so in the big picture (the entire codebase), suddenly having `forEach`-loops when all the rest of the codebase uses `for`-loops will objectively be more confusing**, even if `forEach`-loops are slightly better readable.

Would you mind doing me this favor? Again, I think you have good and valid points, but I believe in the bigger picture it is easier to decide for for-loops. In any case, thank you for getting back and providing good arguments!

benniekiss:
I can definitely change it back to a `for` loop, and I agree with your point about manipulating lists from within the callback scope which exist outside of it. Just regarding your point about `forEach` in the code base, however, it is present, and in fact, there are several other callbacks that mimic the `forEach` style https://github.com/search?q=repo%3AZettlr%2FZettlr+forEach&type=code

benniekiss:
Oh, and to your first point, I think it should stay `null` rather than `undefined` since it's an intentional null value. I did go back and forth on this myself, but I think null is the right value.

nathanlesage:
> Oh, and to your first point, I think it should stay `null` rather than `undefined` since it's an intentional null value. I did go back and forth on this myself, but I think null is the right value.

Roger that!

And regarding the for loop: I feel it looks better; maybe it was the fact that it's two nested for-loops that put me off … in any case, you're right that I've been using forEach two or three other times as well. I'll have to investigate. Next time you use a `forEach` loop, I'll likely have a better opinion 🙈 

That being said — is there anything else we need to discuss regarding this? Or is this ready to merge?

benniekiss:
I think that covers everything for me! I think this should be good to go.

nathanlesage:
Roger that, then let's quickly merge that!

PR #5878, opened 2025-09-02
feat: Enable fuzzy autocomplete

kyaso:
## Description
This PR enables _fuzzy matching_ for autocomplete suggestions.

## Changes
In function `autocompleteSource()`:
- Always get _all_ possible entries
- Enable the `filter` flag

See changes in action below:

https://github.com/user-attachments/assets/f70aff09-c59c-4175-a5d3-72be58301b33

https://github.com/user-attachments/assets/47e49325-763c-47a6-9212-b4b3f1cdae06

Tested on: Ubuntu WSL

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,10 @@ the leading period of the extension, so do not write `$FILENAME.$EXTENSION`.)
   is already running.
 - **Feature**: Individual global search results can now be copied to the
   clipboard (#2070).
+- **Feature**: Autocomplete suggestions (links, tags, etc.) now use _fuzzy
+  matching_. You no longer need to type the exact text, but rather a few
+  letters from it will be enough to bring the best results to the top of the
+  list. (#5878)
 - **Change**: Snippets: The `$FILENAME` variable now does not contain the file
   extension anymore. Users who also want the extension should update their
   snippets to `$FILENAME$EXTENSION` (#4191).
--- a/source/common/modules/markdown-editor/autocomplete/index.ts
+++ b/source/common/modules/markdown-editor/autocomplete/index.ts
@@ -87,14 +87,13 @@ const autocompleteSource: CompletionSource = function (ctx): CompletionResult|nu
   }
 
   if (plugin !== undefined) {
-    const initialOptions = plugin.entries(ctx, ctx.state.doc.sliceString(startpos, ctx.pos).toLowerCase())
+    const initialOptions = plugin.entries(ctx, '')
     return {
       from: startpos,
       options: initialOptions,
-      filter: false,
+      filter: true,
       update: (current, from, to, ctx) => {
-        const query = ctx.state.doc.sliceString(from, to).toLowerCase()
-        current.options = plugin!.entries(ctx, query)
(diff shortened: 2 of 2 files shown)


nathanlesage:
Quick question, because this must've slipped my mind: won't this make sense to apply in general, and thus change the signature of the entries function to just omit the query thingy, or will this break other parts of the app?

kyaso:
> Quick question, because this must've slipped my mind: won't this make sense to apply in general, and thus change the signature of the entries function to just omit the query thingy, or will this break other parts of the app?

Good point. I've checked each `entries` implementation (`citations.ts`, `code-blocks.ts`, `files.ts`, `headings.ts`, `index.ts`, `snippets.ts`, and `tags.ts`), and all just return a filtered list of entries (based on the `query`). Since with this PR we are now handling the filtering inside the `autocompletionSource`, we could indeed change the `entries` methods to just return _all_ entries by default, thus making the `query` parameter obsolete.

kyaso:
Let me know what you think.

nathanlesage:
Yes, that sounds great! Makes the whole thing simpler!

nathanlesage:
Any updates?

kyaso:
I actually sat down again yesterday, and realized that we may have a problem with my suggested approach. From my understanding, the built-in filter only checks the `label` for any matches. However, the current manual filtering (based on query) for citations, snippets (+emoji), and files also incorporates other fields, such as `info`. Therefore, I'm afraid, that just using the built-in filter could potentially break things for those mentioned autocompletes.

One way to get around this would be to check which autocomplete plugin is active, and only enable the built-in filter for the "safe" plugins (such as tags). But I really do want to also have fuzzy filtering for _files_, and, in principle, it works (see the PR description), but that was only tested against the `label` field, I guess.

I just checked the [docs](https://codemirror.net/docs/ref/#autocomplete), and there is this `displayLabel` option. So, maybe, we could store all the relevant data (label, info, detail etc) into the `label` field, and provide a `displayLabel` to override the displayed label. Then, the built-in filter would match the "extended" `label` field, matching also with the strings coming from info/detail. Just an idea, haven't tried it yet.

nathanlesage:
Alright, then feel free to take another look, and comment here once you have made a decision. Please note that a lot of the autocomplete logic is still based on a much earlier version of CodeMirror, and in the past 2-3 years, many improvements in the API have been made, so that we can offload more and more parts to CodeMirror and remove the custom code for it!

nathanlesage:
I now thought about this. I think we definitely need the other pieces of information to match against (so that users can not only search for citekeys but also authors and titles; that users can not only search for filenames, but titles; etc.).

But the easiest way would be to stick as closely as possible to what CodeMirror already does, and I think it would be very easy to add at least the display label to the fuzzymatcher logic in the plugin.

See line: https://github.com/codemirror/autocomplete/blob/8f88a0e25ba84d44ecf0668d1c17347fce2fe45f/src/state.ts#L41

Where we can essentially also check other strings in the completion objects. Would you mind investigating this? It might then make sense to even propose a configurable PR (add a new option "checkDisplayLabel" or something like that, and if set, also check the corresponding Completion string in the fuzzymatcher).

PR #5883, opened 2025-09-03
[Fix] Style drop cursor with editor theme

benniekiss:
## Description

This PR adds CSS styling for the drop-cursor (shown when dragging selected text, etc.)

## Changes

The editor themes were updated to set the drop cursor color, and the main theme was updated to set the color in dark mode.

The CSS attributes were chosen based on the code here: https://github.com/codemirror/view/blob/b7c8f63f079c331fd48e59db84268feb626ae250/src/theme.ts#L125-L138

## Additional information

Currently, the drop cursor is unstyled, and it does not show up against the page in dark mode.

Current:

DropCursorLM

DropCursorDM

PR:

PR-DropCursorLM

PR-DropCursorDM

Tested on:  macOS 15.6


nathanlesage:
Looks good! Thank you!

PR #5890, opened 2025-09-09
[CSS] Fix macOS UX on scrollable toolbar

benniekiss:
## Description

~~This PR re-enables trackpad scrolling in the toolbar when it is overflowing, but maintains draggability when there is no overflow. It also increases the traffic light padding to provide a draggable space no matter the overflow state.~~
This PR fixes button clicking and window dragging in the toolbar on macOS.

## Changes

~~A class, `.is-overflowing` is dynamically added to the `toolbar` element depending on the overflow state.~~

~~The CSS was updated to configure dragging behavior based on the `.is-overflowing` class. The traffic-light padding was also increased to provide a larger draggable area.~~

Added `-webkit-app-region: no-drag;` to macOS toolbar buttons and `.toolbar-text` classes so that they can be clicked.

## Additional information

So it seems the electron drag handling is a bit more complicated. I tried to create a safe, draggable margin between the traffic lights and the scroll area, but the issue is when the scrollable area moves under this region, electron seems to detect it for no dragging, defeating the purpose. By ignoring dragging on buttons, we keep window dragging behavior, and the scroll area still responds to the trackpad when the scroll originates on a button.

Tested on:  macOS 15.6

### DIFF
--- a/source/common/vue/window/WindowToolbar.vue
+++ b/source/common/vue/window/WindowToolbar.vue
@@ -248,6 +248,7 @@ body.darwin {
   div#toolbar {
     // On macOS, there is no titlebar, and as such we need to make the toolbar draggable
     -webkit-app-region: drag;
+    button, .toolbar-text { -webkit-app-region: no-drag; }
 
     height: @toolbar-height;
     font-size: @font-size;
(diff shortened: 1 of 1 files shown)


nathanlesage:
Looks good!

But yes, you're right: electron dragging is weird. Essentially you'll need to create a rectangular area for dragging, and then "carve out" further rectangle boxes where clicking works again ...

Thanks for fixing the non clicking error I made!

PR #5892, opened 2025-09-09
[CSS] Increase size of toolbar icons

benniekiss:
## Description

This PR increases the size of formatting bar and toolbar icons to improve styling, and it syncs the `comment` icon between the toolbar and formatting bar.

## Changes

The size of toolbar icons was increased from `0.8em` to `1.0em`.

The `comment` and `code` icons of the formatting toolbar were changed to match the main editor toolbar. This could be confusing to users who are accustomed to the difference, but I think it's important to have them match.

## Additional information

I noticed that the toolbar icons were kind of thin and seemed small. Investigating https://clarity.design/documentation/spacing showed they were defaulting to about `0.8rem`. It's a personal preference, but I think increasing the icon size makes the toolbar and formatting bar look better.

 Current:

Screenshot 2025-09-09 at 16 46 54

PR:

Screenshot 2025-09-09 at 16 46 29

Tested on:  macOS 15.6 and Fedora 42

### DIFF
--- a/source/common/modules/markdown-editor/tooltips/formatting-toolbar.ts
+++ b/source/common/modules/markdown-editor/tooltips/formatting-toolbar.ts
@@ -62,12 +62,12 @@ function getToolbar (state: EditorState): Tooltip[] {
       const comment = document.createElement('button')
       comment.classList.add('formatting-toolbar-button')
       comment.setAttribute('title', trans('Comment'))
-      comment.innerHTML = '<cds-icon shape="code-alt"></cds-icon>'
+      comment.innerHTML = '<cds-icon shape="code"></cds-icon>'
 
       const code = document.createElement('button')
       code.classList.add('formatting-toolbar-button')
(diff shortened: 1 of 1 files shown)


nathanlesage:
… and you also exchanged two icons. I remember someone wanted some improvements there, but it should be its own PR.

Furthermore: I see your point; I just tried it out, and I can see why we might want to rethink this going forward.

*However*, I believe that this should be part of a broader discussion about the general layout of the app. I have in mind a somewhat larger rewrite of the entire UI at some point in the near to mid future which will unify the styling on all platforms. (i.e., move away from individual styles for all platforms and instead implement one design language that works cross-platform).

Improving icon sizing is definitely on the menu for then, so please keep this in mind, even though I don't think we should merge this right now at this point!

benniekiss:
Of course! Should I go ahead and open another PR to correct the code/comment icons?

nathanlesage:
Yes absolutely

PR #5894, opened 2025-09-10
[CSS] Style the global search pane

benniekiss:
## Description

This PR restyles the global search pane to increase spacing between elements.

## Changes

Added margin between the `form-control` label and input box.

Wrapped the `search` and `cancel` buttons in a flex container and created a gap between the buttons

Hid the `search-clear` element since it cannot be styled. 

## Additional information

Current:
Screenshot 2025-09-09 at 17 16 27

PR:

Screenshot 2025-09-09 at 17 16 06

Tested on:  macOS 15.6, Fedora 42

### DIFF
--- a/source/win-main/GlobalSearch.vue
+++ b/source/win-main/GlobalSearch.vue
@@ -555,6 +555,25 @@ body div#global-search-pane {
     border-bottom: 1px solid #ccc;
   }
 
+  p {
+    margin-top: 5px;
+    display: flex;
+    gap: 5px;
+  }
+
+  .form-control {
+    input {
+      margin-top: 5px;
+    }
+  }
+
+  // hide chromium 'search-clear' element since it cannot be styled
+  input[type="search"]::-webkit-search-cancel-button {
+    -webkit-appearance: none;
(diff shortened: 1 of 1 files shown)


nathanlesage:
… why would you remove the clear search element? I mean, sure, it can't be styled, but it's part of a11y elements so that screen readers etc. can select and clear the element. Essentially, this removes a part of the functionality from this thing.

The rest looks good!

benniekiss:
I hadn't considered the a11y functionality, but I thought it could be replaced at another time with a custom clear element. I can definitely add it back, though :)

nathanlesage:
> I hadn't considered the a11y functionality, but I thought it could be replaced at another time with a custom clear element.

We can absolutely see if we can replace that element with an accessible alternative that we can style. Tbh I also don't like that blue button thingy, but any aesthetic vs. accessible discussion should be won in favor of accessibility. Thank you!

benniekiss:
> but any aesthetic vs. accessible discussion should be won in favor of accessibility

I 100% agree

PR #5895, opened 2025-09-10
[CSS] Style popovers

benniekiss:
## Description

This PR restyles some of the popovers in the toolbar to increase element spacing.

## Changes

The `tags`, `export`, and `pomodoro` popovers were restyled to increase spacing between elements.

The export popover had its elements wrapped in a `div` so that the styling was applied only to the popover.

The pomodoro start and stop buttons were wrapped in a `div` so they could be centered and widened.

## Additional information

# Current
tags

export

pomodoro
pomodoro-timer

# PR

PR-tags

PR-export

PR-pomodoro
PR-pomodoro-timer

---

Tested on:  macOS 15.6 and Fedora 42

### DIFF
--- a/source/win-main/PopoverExport.vue
+++ b/source/win-main/PopoverExport.vue
@@ -1,6 +1,6 @@
 <template>
   <PopoverWrapper v-bind:target="target" v-on:close="$emit('close')">
-    <div class="exporter">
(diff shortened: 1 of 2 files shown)


nathanlesage:
Sure, looks good! Thanks!

PR #5898, opened 2025-09-13
[Feat] Implement `Intl.Segmenter` for word counting

### LINKED ISSUE #2363: [Enhancement] Word counting with UAX #29 word boundaries
### Description

Writing notes in Japanese with Zettlr works fine but word counting is far from a satisfactory behavior. In Japanese, you will not insert white spaces between words. Thus, Zettlr counts the whole paragraph as *a word.* Precise Japanese word counting is hard, but the [Unicode Standard Annex #29 defines a language-agnostic algorithm to determine word boundaries.](https://unicode.org/reports/tr29/#Word_Boundaries) This algorithm will not harm users that is already fine with the current behavior, and help users frustrated by the current behavior more or less.

### Proposed Changes

Implementing a word counting method based on the UAX 29's default word boundaries algorithm. Making that algorithm default or not depends on the runtime performance of it.

### Caveats

The UAX 29's default word boundaries algorithm is more complex than white space-based one. On the other hand, the algorithm is not optimal and [CLDR](http://cldr.unicode.org/)-based language-specific word boundaries algorithm will produce better results.

It is excellent to me if the CLDR-based method is implemented, but language-agnostic one is better than the status quo.

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Operating System Version

Windows 11 (Beta)

### Additional Information

_No response_

benniekiss:
## Description

This PR refactors the `getCleanedWords` function to separate text nodes using [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter). This enables correctly counting words in languages which do not split on whitespace.

This was proposed in #2363, and this PR implements the tests and functionality described in that issue.

Closes #2363

## Changes

The `getCleanedWords` function was refactored so that text nodes are split using `Intl.Segmenter` rather than on whitespace. 

Character counting was refactored to remove the `map` and `reduce` step, since we can get the character count when generating the word list. The `getCleanedWords` function was refactored so that it returns the word list and the character count, instead of just the word list.

An optional `locale` argument was added to the various word/char counting functions so that the locale can be explicitly set, otherwise, it falls back to get the `appLang` option from `window.config`

Tests were updated for correctness, and new tests were added to test Japanese word counting. The tests were also refactored to include more diagnostic information on failure, including the test number and the locale used.

## Additional information

The following test was actually incorrect, even though it was passing. This line is what I believe is all that should be counted, `# Heading\n\nLorem Ipsum dolor, sit amet`, and it is actually 29 characters, not 30. 

The new implementation counts the number of characters correctly (29).

```ts
{
  input: '---\ntitle: "Some title"\nkeywords:\n  - one\n  - two\n  - three\n...\n\n# Heading\n\nLorem Ipsum dolor, sit amet',
  locale: 'en',
  expectedWords: 6,
  expectedChars: 30
},
  ```

Tested on:

### DIFF
--- a/source/app/service-providers/fsal/util/file-parser.ts
+++ b/source/app/service-providers/fsal/util/file-parser.ts
@@ -29,7 +29,7 @@ import type {
   ZettelkastenTag
 } from '@common/modules/markdown-utils/markdown-ast'
 import { extractLinefeed } from './extract-linefeed'
-import { AppServiceContainer } from 'source/app/app-service-container'
+import { getAppServiceContainer, isAppServiceContainerReady } from '../../../app-service-container'
 
 // Here are all supported variables for Pandoc:
 // https://pandoc.org/MANUAL.html#variables
@@ -83,8 +83,7 @@ export default function getMarkdownFileParser (
       file.firstHeading = firstH1.content
     }
 
-    const _app = new AppServiceContainer()
-    const locale: string =  _app.config.get('appLang')
+    const locale: string | undefined = isAppServiceContainerReady() ? getAppServiceContainer().config.get('appLang') : undefined
 
     const counts = countAll(ast, locale)
 
--- a/source/common/util/counter.ts
+++ b/source/common/util/counter.ts
@@ -53,7 +53,9 @@ function prepareCounts (ast: ASTNode, locale?: string, from = 0, to?: number): {
   let chars = 0
   const words = textNodes
     .flatMap(node => {
-      // this count includes non-formatting characters and whitespace.
+      // this count includes non-formatting characters and whitespace,
+      // however, because the way text nodes are parsed, preceding whitespace
+      // may be excluded, such as for text following inline formatting.
       chars += node.value.length
 
       const segments = []

### CODE REVIEW

File: source/common/util/counter.ts [hunk 1]
```diff
@@ -49,12 +48,25 @@ function getCleanedWords (ast: ASTNode, from = 0, to?: number): string[] {
     })
   }
 
-  const plainText = textNodes.map(node => node.value)
-  return plainText
-    .join(' ')
-    .split(/[\s ]+/)
-    .filter(w => w.trim() !== '')
-    .filter(word => !interpunctionRE.test(word))
+  const appLang: string | undefined =  typeof window !== 'undefined' ? window.config?.get('appLang') : undefined
```
nathanlesage: Please remove this unnecessary check. The window object is modified with the config object before the DOM is even loaded via the preload; and TypeScript should pick this up from the `global.d.ts` ambient definitions.

source/common/util/counter.ts
```diff
@@ -49,12 +48,25 @@ function getCleanedWords (ast: ASTNode, from = 0, to?: number): string[] {
...
+  locale = locale !== undefined ? locale : appLang
```
nathanlesage: Would `locale ?? appLang` also work? Might make this a bit more nice.

Also, with the typeof check gone you could just plug the window config retrieval in here.

source/common/util/counter.ts
```diff
@@ -49,12 +48,25 @@ function getCleanedWords (ast: ASTNode, from = 0, to?: number): string[] {
...
+
+  const segmenter = new Intl.Segmenter(locale, { granularity: 'word' })
+
+  let chars = 0
+  const words = textNodes
+    .flatMap(node => {
+      const segments = []
+      for (const { segment, isWordLike } of segmenter.segment(node.value)) {
+        if (isWordLike === true) {
```
nathanlesage, line 63: We should add a comment that whitespace is not counted with this method. It wasn't before, but I just realized that will become important once anybody in the future looks at this code.

source/common/util/counter.ts
benniekiss: So I didn't have this check to begin with, but the `window` param was giving me `undefined` when the application starts up. I couldn't track down why, but I had to add these checks to make markdown files load at all.

source/common/util/counter.ts
benniekiss: Something like `// non-word characters and whitespace are excluded from the character count`?

source/common/util/counter.ts
nathanlesage: So essentially when the app started it would throw an error that window is undefined? That sounds like a serious bug, although I have no idea where it might come from.

source/common/util/counter.ts
nathanlesage: … oh wait. We might want to actually count non-word characters. If we ignore whitespace, that's fine, but we should definitely count everything else, at least in the character counter.

But … wait, the tests are all still passing, right? I'm a bit confused right now.

Oh wait, what happens if we, if `isWordLike` is false, simply add the length of whatever was in the segment to the char counter …?

source/common/util/counter.ts
benniekiss: Yeah, let me see if I can get a screenshot of it happening. 

It was weird in testing, because if I had the app running first and then made the change, no issue. But when starting the dev environment from scratch, markdown files would not load, but things like JSON did.

source/common/util/counter.ts
benniekiss: So non-word characters meaning punctuation, etc, that are not within the found words.

benniekiss: Every character in `segment` is counted, but this only includes characters that are within words, so punctuation, etc, is ignored.

source/common/util/counter.ts
benniekiss: Here's what happens with the line set like this:

`locale = locale ?? window.config?.get('appLang')`

This error pops up when I try to open a markdown file. When the app first starts, though, there is no error at all.

Screenshot 2025-09-14 at 11 52 27

benniekiss: this is the code path of that error message: https://github.com/Zettlr/Zettlr/blob/c097d147ef29290d3d0f4e618ab276ffd8747881/source/app/service-providers/commands/root-open.ts#L148-L159

nathanlesage: Okay, this smells like a file access error. The catch condition essentially is only triggered when there's an error in the getDescriptorForAnySupportedFile function.

Would you mind opening the logs and double-checking the exact error being thrown?

I remember that `<unknown>` is something that is often spat out by the FS subsystem in Node when there's an error with a file.

In any case, it has nothing to do with the window thing, and I'm 100% certain that this is a Heisenbug if it co-occurs with counting words.

benniekiss: Here's the error from logs when opening a file, but there's no error when starting the app, `yarn start --clean`

```
Could not open root /Users/benniemt/Dev/ws/Zettlr/resources/test/Rendering/code.md: window is not defined
Name: Error
Message: window is not defined
Error: window is not defined
```

nathanlesage: Ah wait, so there's some renderer function… OH WAIT; OF COURSE. The count words function is being used both in the main process AND in the renderer process. In one of the two, the config is used on the window object, on the other it's a service provider.

Solution: Require the application language, and pass it from the function callers; those will have access to EITHER the window OR the config provider, which will make it work. Sorry that it took me so long to figure this out!

So TL;DR: Don't access the window object in the word counter utility function :)

(Note: The Electron Node binary will be compiled with Intl support, so the segmenter should be available in the Node runtime, too.)

benniekiss: nice sleuthing! ill make those updates

source/common/util/counter.ts
nathanlesage: Alright, thanks for clarifying! So, there are three notions of counts that one will encounter in the wild: word counts, character counts, and character counts without spaces. Zettlr needs to correspond to these three concepts, because, while these are not always clearly defined, users will generally expect them to roughly align with each other.

A **word count** essentially maps onto what the Intl segmenter is doing here, so word counts are absolutely fine.

However, when it comes to character counts, this one is wholly independent of word counts.

Specifically, a general **character count** means to, quite literally, count all the characters in some text. There is one variation of character count that excludes spaces (or whitespace more generally).

I believe the latter one I nuked in one update, and this is fine. I have originally included both a character counter and a character counter without spaces in Zettlr because that's what Word offers ([source](https://support.microsoft.com/en-us/office/show-word-count-3c9e6a11-a04d-43b4-977c-563a0e0d5da3)).

What publishers will often impose is either a specified word limit, or a character limit, where the character limit just includes everything. I haven't seen any character count “without spaces” in quite a long time, so I believe it's perfectly fine to ignore that.

That being said, the character counter should not just count the characters inside real words, but all characters (excluding formatting characters; which is guaranteed by only working on text nodes).

In short, your new improved measure of a word count is fabulous and will heavily improve the accuracy of Zettlr's word count, but at the expense of a less accurate character count. Is there a cheap way of making use of the segmenter to extract all characters from the text nodes? If so, that would be awesome; otherwise we can also just do a MapReduce on the text nodes and calculate the length of all. (Iirc JavaScript's length property accurately accounts for multibyte characters and should be precise.)

source/app/service-providers/fsal/util/file-parser.ts
```diff
@@ -82,8 +82,9 @@ export default function getMarkdownFileParser (
       file.firstHeading = firstH1.content
     }
 
-    file.wordCount = countWords(ast)
-    file.charCount = countChars(ast)
+    const locale = undefined
```
nathanlesage: Pro-tip: The AppServiceContainer singleton can now be retrieved from anywhere in the app, meaning you can retrieve the correct locale here, too!

source/common/modules/markdown-editor/index.ts
```diff
@@ -720,6 +720,7 @@ export default class MarkdownEditor extends EventEmitter {
     const mainOffset = this._instance.state.selection.main.head
     const line = this._instance.state.doc.lineAt(mainOffset)
     const ast = markdownToAST(this._instance.state.sliceDoc(), syntaxTree(this._instance.state))
+    const locale = window.config.get('appLang') as string | undefined
```
nathanlesage:
```suggestion
    const locale: string = window.config.get('appLang')
```

source/common/modules/markdown-editor/plugins/statistics-fields.ts
```diff
@@ -22,7 +22,8 @@ import { countAll } from '@common/util/counter'
 export const countField = StateField.define<{ chars: number, words: number }>({
   create (state: EditorState) {
     const ast = markdownToAST(state.doc.toString())
-    return countAll(ast)
+    const locale = window.config.get('appLang') as string | undefined
```
nathanlesage:
```suggestion
    const locale: string = window.config.get('appLang')
```

source/common/modules/markdown-editor/plugins/statistics-fields.ts
```diff
@@ -32,9 +33,9 @@ export const countField = StateField.define<{ chars: number, words: number }>({
     if (!transaction.docChanged) {
       return value
     }
-
     const ast = markdownToAST(transaction.state.doc.toString(), syntaxTree(transaction.state))
-    return countAll(ast)
+    const locale = window.config.get('appLang') as string | undefined
```
nathanlesage:
```suggestion
    const locale: string = window.config.get('appLang')
```

source/common/util/counter.ts
```diff
@@ -68,8 +79,8 @@ function getCleanedWords (ast: ASTNode, from = 0, to?: number): string[] {
  *
  * @return  {number}         The number of words in the file.
  */
-export function countWords (ast: ASTNode, from = 0, to?: number): number {
-  return getCleanedWords(ast, from, to).length
+export function countWords (ast: ASTNode, locale: string | undefined, from = 0, to?: number): number {
```
nathanlesage:
```suggestion
export function countWords (ast: ASTNode, locale?: string, from = 0, to?: number): number {
```

We might want to rethink where we pass in `from`. I'm almost inclined to change the argument order to `ast, from, locale, to`, but I'm open to suggestions.

source/common/util/counter.ts
```diff
@@ -83,10 +94,8 @@ export function countWords (ast: ASTNode, from = 0, to?: number): number {
  *
  * @return  {number}         The number of characters in the file
  */
-export function countChars (ast: ASTNode, from = 0, to?: number): number {
-  return getCleanedWords(ast, from, to)
-    .map(w => w.length)
-    .reduce((prev, cur) => prev + cur, 0)
+export function countChars (ast: ASTNode, locale: string | undefined, from = 0, to?: number): number {
```
nathanlesage:
```suggestion
export function countChars (ast: ASTNode, locale?: string, from = 0, to?: number): number {
```

Also, same question as above

source/common/util/counter.ts
```diff
@@ -100,10 +109,10 @@ export function countChars (ast: ASTNode, from = 0, to?: number): number {
  *
  * @return  {{words, chars}}        Word and character counts
  */
-export function countAll (ast: ASTNode, from = 0, to?: number): { words: number, chars: number } {
-  const words = getCleanedWords(ast, from, to)
+export function countAll (ast: ASTNode, locale: string | undefined, from = 0, to?: number): { words: number, chars: number } {
```
nathanlesage:
```suggestion
export function countAll (ast: ASTNode, locale?: string, from = 0, to?: number): { words: number, chars: number } {
```

nathanlesage (changes requested):
We're getting there! Besides some minor things, I am currently thinking that we might want to shuffle some functions around. Right now, we have the `getCleanedWords`-function that utilizes the segmenter. However, the function doesn't anymore really just "get cleaned words," and instead provides essentially a character count and a list of words.

At the same time, the `countAll` function appears to be a bit superfluous right now, as it essentially just returns the return value of `getCleanedWords`, so my question would now be, whether we should move the body of the `getCleanedWords`-function into the `countAll` function, and furthermore remove the `countChars` and `countWords` function (to avoid people [I'm primarily thinking of myself] from calling both countChars and countWords, thus invoking countAll twice).

Thoughts? I know I'm coming around with more and more change requests, but I believe this to be also caused by simply my sluggishness when first implementing this … it looks like a lot of tech debt to me :/ If you don't want to do that cleanup, I can also do it!

nathanlesage:
P.S.: I just took a look, and the failing unit test is purely because I did some changes elsewhere. I now took the liberty to update your branch, which should now see passing unit tests.

benniekiss:
> Thoughts? I know I'm coming around with more and more change requests, but I believe this to be also caused by simply my sluggishness when first implementing this … it looks like a lot of tech debt to me :/ If you don't want to do that cleanup, I can also do it!

I think this would be fine! So condense everything into `getCleanedWords` and rename it to `countAll`? That makes total sense to me.

benniekiss:
I decided to keep `countAll` since it accesses ONLY the counts, and I renamed `getCleanedWords` to `prepareCounts`. Otherwise, I think I got the rest of your suggestions!

source/app/service-providers/fsal/util/file-parser.ts
```diff
@@ -82,8 +83,13 @@ export default function getMarkdownFileParser (
       file.firstHeading = firstH1.content
     }
 
-    file.wordCount = countWords(ast)
-    file.charCount = countChars(ast)
+    const _app = new AppServiceContainer()
```
nathanlesage: Do not instantiate a new AppServiceContainer. Instead, import the getter function from the app-service-container module and use that to retrieve the existing singleton.

benniekiss:
Two things before I think this is ready to merge. Currently, it looks like `AppServiceContainer` is still not available in [source/app/service-providers/fsal/util/file-parser.ts](https://github.com/Zettlr/Zettlr/pull/5898/files/516253e6288ef656605a288e97639629a05bde9b#diff-d6a55bde26dd33c56f0b70f808b11566b7658ca90171e7d3aeaedffbba6dbf99). Using `isAppServiceContainerReady`, it's always unavailable as we load the files, so the value is always `undefined`.

Secondly, character counting misses some preceding whitespace for instances where text follows inline formatting (bold, etc.). Is this an acceptable omission? It will result in slightly less accurate character counts, but including the preceding whitespace would also result in inaccurate character counts since it would include formatting whitespace.

nathanlesage:
> Is this an acceptable omission?

I saw this in your tests myself. I think yes. If it becomes a burden, improving that small part should be easier than trying to break our heads about this right now.

> Currently, it looks like AppServiceContainer is still not available in

Ah, yes. So the AppServiceContainer is made available only after the entire boot in the lifecycle. The file parser is called -- among other things -- during boot, meaning that at startup it will not have access to the container, but afterwards it will.

I don't want to make the container available before boot to avoid any race conditions where we have some code that accesses it while not all providers are booted, but I believe the config provider in particular (as well as the log provider) are special cases that should be available as early as possible, including during the boot cycle. (After all, both are booted much earlier for a reason).

hmhmhmhmhhm … so here's my take: Word and character counts must be precise when users actually look at them. This always implies having that file open. When the user opens a file, word and character counts are re-evaluated and the correct app lang is guaranteed to be available from config. The only thing that this influences is the cached word count, but I do not believe that there are many situations in which (a) the word count would suffer dramatically if we use a default lang of `en` if the app service container is not yet available and (b) where the user would just trust the cached info in the properties popover when it comes to submitting an important text.

So I think for now it should suffice to check for AppServiceContainer availability, and fall back to english if it's not available. That sounds like a good compromise without having to rip open large parts of the core of the app.

Going forward, we may consider making the config provider a singleton that can be accessed directly, if that is possible and our assumptions during the boot process are valid.

benniekiss:
> So I think for now it should suffice to check for AppServiceContainer availability, and fall back to english if it's not available. That sounds like a good compromise without having to rip open large parts of the core of the app.

I don't think we have to fallback to `en`, if we pass `undefined`, `Intl.Segmenter` does it's own fallback. If you agree, I think this should be good to merge!

nathanlesage:
Even better, especially if it checks for the users context, so sure, go ahead!

benniekiss:
my bad, I didn't mean to bug you with another review

nathanlesage:
No worries! But it's ready to merge now, right?

benniekiss:
Yep! Should be ready.

nathanlesage:
Perfect, thanks!

PR #5903, opened 2025-09-16
[Performance] Stricter pipe and grid table regex and parsing

benniekiss:
## Description

This PR implements a stricter pipe and grid table regex to improve parsing speed.

## Changes

The pipe and grid table regex were updated to better match the table syntax and reduce false-positives

~~The `pipeLeafParser` `nextLine` method was updated to test whether the next line matches the pipe table regex so that parsing finishes early on invalid tables.~~

## Additional information

In my performance testing, I noticed that the `pipeTableParser` was taking an enormous amount of time compared to the other parsers. In my testing document, the total parse time was about 9 seconds, of which the `pipeTableParser` was taking up about 3 seconds. With this implementation, the total parse takes about 4 seconds, and the `pipeTableParser` is no longer on the radar.

It seems that the previous regex, `/^(\|.+?\|)$|(.+?\|.+?)/`, was missing line ending markers around the second test, `^(.+?\|.+?)$/`. This change itself was enough to give a significant performance boost. I decided to reimplement the regex for a stricter test to reduce false-positives.

Tested on:  macOS 26


benniekiss:
My performance numbers. This is from a custom fork with patches, but the difference shows the impacts of this PR.
Current:

Screenshot 2025-09-16 at 16 13 53

PR:

Screenshot 2025-09-16 at 16 10 39

benniekiss:
A consequence of this change is that single-column tables like below are now valid:

```
Header
-------:
My cell
```

An alignment or pipe is required in the delimiter so that it's not registered as a Setext heading. 

I can change this to ignore these single-column tables, but I don't think they are out of spec. Likewise, regular single column tables are permitted.

```
| Header  |
|---------|
| My cell |
```

nathanlesage:
Just tried it out: The single-column-no-pipe table example that you provided is not detected by Pandoc, ergo I'd like those to not be detected as tables, regardless of alignment colon. Could you do this?

nathanlesage:
Otherwise, I'm taking any performance boost there is!

benniekiss:
Ok, so I updated it so there's at least two cells.

I also updated the other regexes, and the performance is now phenomenal! With all of the regexes refactored for strictness, here are my performance numbers. I'd say a decrease from 9 seconds to 0.2 seconds is excellent :)

Screenshot 2025-09-17 at 09 07 02

benniekiss:
~~Closing because https://github.com/Zettlr/Zettlr/pull/5903 has made most of this unnecessary~~ Wrong one :(

nathanlesage:
Oi, that's a lot – I just did some very coarse testing; I didn't find much issues, except that the pipe table RE will now just … match an entire table including empty lines? Would you mind testing that?

Other than that, I take any performance improvement I can get, and I couldn't immediately spot any missing details in the table rendering, but I may have overlooked edge cases.

See screenshots, first old, then new:

image

image

benniekiss:
Haha, that is not how I thought the regex worked. I have a fix

benniekiss:
These should handle new lines properly. We need the `/gm` flags to parse the multiline tables correctly.

One question: Is it ok to only exclude `\n`, or do we need to exclude `\r`, and any other newline types?

nathanlesage:
Do we really need the `gm`-flags? If I didn't overlook it, we only have individual lines, right? And I also just realized that my test was flawed yesterday, because the regex can ignore newlines as they will not occur in the source string (I believe).

Feel free to think about this; I'll go with whatever you say works out. We have to keep an eye on the table editor anyway until it works as flawless as possible … I've also caught a cold, so maybe my brain wasn't thinking straight yesterday.

Finally, to answer your question: Yes, all line endings are normalized to LF in the editor, there are no CR!

benniekiss:
I was rethinking the `/gm` as well - I think we can definitely do without `/g`, but I'm not certain about `/m`, I'll do a little more testing to verify!

Im sorry you're coming down with something. I hope you feel better soon!

benniekiss:
In my testing, we don't need the `/m` multiline flag, so both have been removed.

I also noticed that the grid table content parsing was not really accurate. What I added are essentially two negative lookaheads to avoid matching rows that include empty delimiters, `++`, and text within crosses, `+--invalid--+`. I believe this is correct, now. This improves the previous grid content parsing, which was not detecting mixed rows.

nathanlesage:
To get ahead of some of the PRs here, is this mergeable? I think some final things we can absolutely address during the beta.

benniekiss:
Yes, I believe so. I've tested it in the Table Editor test document, and there doesn't seem to be any issue.

nathanlesage:
Wonderful — there are definitely some bugs in the table editor itself, but again, I think it's perfectly fine :) Thank you!

PR #5910, opened 2025-09-19
Use LanguageTool API `data` parameter to reduce load; and filter words from local dictionary

benniekiss:
## Description

This PR changes the languagetool API call to pass the document as the `data` parameter, and it filters out languagetool responses which match words in the local user dictionary.

## Changes

By sanitizing the document into a list of `{ text: string}` and `{ markup: string }` objects, we can pass the JSON to the LanguageTool API, and it will only use the `text` objects to generate errors.

This reduces the load on the API service and increases the number of meaningful lints returned by the API.

The linter now filters out any matches for words that are in the local user dictionary.

## Additional information

~~Currently, it seems that `Task` nodes are not parsed correctly -- in my testing, `[ ]` is marked as a text node. I believe this is because we do not activate [all of the GFM extensions](https://github.com/lezer-parser/markdown?tab=readme-ov-file#user-content-gfm), we only activate the base GFM extension via `markdownLanguage`. There could be other causes, but I have not pinned down the problem.~~ So it actually seems to be an issue with the `parseChildren` method when generating the AST for list items. <-- This should be fixed now.

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/commands/language-tool.ts
+++ b/source/app/service-providers/commands/language-tool.ts
@@ -119,7 +119,7 @@ export default class LanguageTool extends ZettlrCommand {
    *
    * @return  {Promise<LanguageToolAPIResponse|string|undefined>}       The result
    */
-  async run (evt: string, arg: { text: string, language: string }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
+  async run (evt: string, arg: { text: string, data: string, language: string }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
     const {
       active,
       level,
@@ -130,13 +130,13 @@ export default class LanguageTool extends ZettlrCommand {
       customServer,
       provider
     } = this._app.config.getConfig().editor.lint.languageTool
-    const { text, language } = arg
+    const { text, data, language } = arg
 
     if (!active) {
       return undefined // LanguageTool is not active
     }
 
-    const searchParams = new URLSearchParams({ language, text, level })
+    const searchParams = new URLSearchParams({ language, data, level })
 
     // If the user has set the mother tongue, add it to the params
     if (motherTongue.trim() !== '') {
--- a/source/app/service-providers/commands/language-tool.ts
+++ b/source/app/service-providers/commands/language-tool.ts
@@ -21,7 +21,8 @@ import type { AppServiceContainer } from 'source/app/app-service-container'
 
 export interface Annotation {
   text?: string,
-  markup?: string
+  markup?: string,
+  interpretAs?: string
 }
 
 export interface AnnotationData {
--- a/source/app/service-providers/commands/language-tool.ts
+++ b/source/app/service-providers/commands/language-tool.ts
@@ -129,7 +129,7 @@ export default class LanguageTool extends ZettlrCommand {
    *
    * @return  {Promise<LanguageToolAPIResponse|string|undefined>}       The result
    */
-  async run (evt: string, arg: { data: AnnotationData, language: string }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
+  async run (evt: string, arg: { data: AnnotationData, language: string, disabledRules: string[] }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
     const {
       active,
       level,
@@ -140,13 +140,13 @@ export default class LanguageTool extends ZettlrCommand {
       customServer,
       provider
     } = this._app.config.getConfig().editor.lint.languageTool
-    const { data, language } = arg
+    const { data, language, disabledRules } = arg
 
     if (!active) {
       return undefined // LanguageTool is not active
     }
 
-    const searchParams = new URLSearchParams({ language, data: JSON.stringify(data), level })
+    const searchParams = new URLSearchParams({ language, data: JSON.stringify(data), level, disabledRules: disabledRules.join(',') })
 
     // If the user has set the mother tongue, add it to the params
     if (motherTongue.trim() !== '') {
--- a/source/app/service-providers/commands/language-tool.ts
+++ b/source/app/service-providers/commands/language-tool.ts
@@ -19,6 +19,15 @@ import { app } from 'electron'
 import { trans } from '@common/i18n-main'
 import type { AppServiceContainer } from 'source/app/app-service-container'
 
+export interface Annotation {
+  text?: string,
+  markup?: string
+}
+
+export interface AnnotationData {
+  annotation: Annotation[]
+}
+
 export interface LanguageToolAPIMatch {
   message: string // Long message (English)
   shortMessage: string
@@ -119,7 +128,7 @@ export default class LanguageTool extends ZettlrCommand {
    *
    * @return  {Promise<LanguageToolAPIResponse|string|undefined>}       The result
    */
-  async run (evt: string, arg: { text: string, data: string, language: string }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
+  async run (evt: string, arg: { data: AnnotationData, language: string }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
     const {
       active,
       level,
@@ -130,13 +139,13 @@ export default class LanguageTool extends ZettlrCommand {
       customServer,
       provider
     } = this._app.config.getConfig().editor.lint.languageTool
-    const { text, data, language } = arg
+    const { data, language } = arg
 
     if (!active) {
       return undefined // LanguageTool is not active
     }
 
-    const searchParams = new URLSearchParams({ language, data, level })
+    const searchParams = new URLSearchParams({ language, data: JSON.stringify(data), level })
 
     // If the user has set the mother tongue, add it to the params
     if (motherTongue.trim() !== '') {
@@ -171,6 +180,8 @@ export default class LanguageTool extends ZettlrCommand {
       server = server.substring(0, server.length - 1)
     }
 
+    const numCharacters = data.annotation.reduce((a, b) => a + (b.text?.length ?? b.markup?.length ?? 0), 0)
+
     const headers = {
       // NOTE: For debugging purposes, we send a custom User-Agent string. This
       // should help figure out potential problems in case Zettlr causes large
@@ -179,7 +190,7 @@ export default class LanguageTool extends ZettlrCommand {
     }
 
     try {
-      this._app.log.verbose(`[Application] Contacting LanguageTool at ${server} (using credentials: ${String(useCredentials)}); payload: ${text.length} characters`)
+      this._app.log.verbose(`[Application] Contacting LanguageTool at ${server} (using credentials: ${String(useCredentials)}); payload: ${numCharacters} characters`)
       const languages = await fetchSupportedLanguages(server)
       // NOTE: Documentation at https://languagetool.org/http-api/#!/default/post_check
       const result = await got(`${server}/v2/check`, { method: 'post', body: searchParams.toString(), headers })
--- a/source/common/modules/markdown-editor/linters/language-tool.ts
+++ b/source/common/modules/markdown-editor/linters/language-tool.ts
@@ -17,18 +17,11 @@ import { linter, type Diagnostic, type Action } from '@codemirror/lint'
 import { extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
 import { configField } from '../util/configuration'
 import type { LanguageToolAPIResponse, AnnotationData, Annotation } from '@providers/commands/language-tool'
-import { StateEffect, StateField } from '@codemirror/state'
+import { StateEffect, StateField, type Transaction } from '@codemirror/state'
 import extractYamlFrontmatter from 'source/common/util/extract-yaml-frontmatter'
+import type { ViewUpdate } from '@codemirror/view'
 
 const ipcRenderer = window.ipc
-
-export interface LanguageToolStateField {
-  running: boolean
-  lastDetectedLanguage: string
-  supportedLanguages: string[]
-  overrideLanguage: 'auto'|string
-  lastError: string|undefined
-}
 
 // store the local dictionary for later filtering
 const userDictionary: Set<string> = new Set()
@@ -55,6 +48,15 @@ ipcRenderer.on('dictionary-provider', (event, message) => {
   }
 })
 
+export interface LanguageToolStateField {
+  running: boolean
+  lastDetectedLanguage: string
+  supportedLanguages: string[]
+  overrideLanguage: 'auto'|string
+  lastError: string|undefined
+  disabledRules: string[]
+}
+
 export const updateLTState = StateEffect.define<Partial<LanguageToolStateField>>()
 
 export const languageToolState = StateField.define<LanguageToolStateField>({
@@ -78,7 +80,8 @@ export const languageToolState = StateField.define<LanguageToolStateField>({
       lastDetectedLanguage: 'auto',
       lastError: undefined,
       overrideLanguage,
-      supportedLanguages: []
+      supportedLanguages: [],
+      disabledRules: []
     }
   },
   update (value, transaction) {
@@ -89,6 +92,7 @@ export const languageToolState = StateField.define<LanguageToolStateField>({
         value.lastError = e.value.lastError
         value.supportedLanguages = e.value.supportedLanguages ?? value.supportedLanguages
         value.overrideLanguage = e.value.overrideLanguage ?? value.overrideLanguage
+        value.disabledRules = e.value.disabledRules ?? value.disabledRules
       }
     }
     return value
@@ -125,6 +129,34 @@ function adjustWhitespace (string1: string, string2: string): [ number, number ]
   const string2Padding = leadingWhitespace - Math.floor(leadingWhitespace / 2)
 
   return [ string1.length - string1Padding, string2Padding ]
+}
+
+// Hide the tooltip when the `Ignore Rule` action is selected.
+function hideOn (tr: Transaction): boolean | null {
+  for (const e of tr.effects) {
+    if (e.is(updateLTState)) {
+      if (e.value.disabledRules) {
+        return true
+      }
+    }
+  }
+
+  return null
+}
+
+// Re-run the linter when `ignoreRules` are updated.
+function needsRefresh (update: ViewUpdate): boolean {
+  for (const tr of update.transactions) {
+    for (const e of tr.effects) {
+      if (e.is(updateLTState)) {
+        if (e.value.disabledRules) {
+          return true
+        }
+      }
+    }
+  }
+
+  return false
 }
 
 /**
@@ -219,7 +251,8 @@ const ltLinter = linter(async view => {
     command: 'run-language-tool',
     payload: {
       data: annotations,
-      language: view.state.field(languageToolState).overrideLanguage
+      language: view.state.field(languageToolState).overrideLanguage,
+      disabledRules: view.state.field(languageToolState).disabledRules
     }
   })
 
@@ -267,9 +300,18 @@ const ltLinter = linter(async view => {
       source
     }
 
-    if (match.replacements.length > 0) {
-      const actions: Action[] = []
+    const actions: Action[] = []
+    actions.push({
+      name: 'Ignore Rule',
+      apply (view) {
+        const disabledRules = [...view.state.field(languageToolState).disabledRules]
+        disabledRules.push(match.rule.id)
 
+        view.dispatch({ effects: updateLTState.of({ disabledRules: disabledRules }) })
+      }
+    })
+
+    if (match.replacements.length > 0) {
       // Show at most 10 actions to not overload those messages
       let i = 0
       for (const { value } of match.replacements) {
@@ -292,7 +334,11 @@ const ltLinter = linter(async view => {
   }
 
   return diagnostics
-}, { delay: 2000 }) // Increase the delay to reduce server strain
+}, {
+  delay: 2000, // Increase the delay to reduce server strain
+  hideOn,
+  needsRefresh
+})
 
 export const languageTool = [
   ltLinter,
--- a/source/common/modules/markdown-editor/linters/language-tool.ts
+++ b/source/common/modules/markdown-editor/linters/language-tool.ts
@@ -14,11 +14,10 @@
  */
 
 import { linter, type Diagnostic, type Action } from '@codemirror/lint'
-import { extractASTNodes, markdownToAST } from '@common/modules/markdown-utils'
+import { extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
 import { configField } from '../util/configuration'
-import { type LanguageToolAPIResponse, type AnnotationData } from '@providers/commands/language-tool'
+import { type LanguageToolAPIResponse, type AnnotationData, type Annotation } from '@providers/commands/language-tool'
 import { StateEffect, StateField } from '@codemirror/state'
-import { type TextNode } from '@common/modules/markdown-utils/markdown-ast'
 import extractYamlFrontmatter from 'source/common/util/extract-yaml-frontmatter'
 
 const ipcRenderer = window.ipc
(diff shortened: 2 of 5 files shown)


source/app/service-providers/commands/language-tool.ts
```diff
@@ -118,7 +119,7 @@ export default class LanguageTool extends ZettlrCommand {
    *
    * @return  {Promise<LanguageToolAPIResponse|string|undefined>}       The result
    */
-  async run (evt: string, arg: { text: string, language: string }): Promise<[LanguageToolAPIResponse, string[]]|string|undefined> {
```
nathanlesage, line 122: If I saw it correctly, we won't need `text` here anymore, correct? It's only used for the character count, but that's easy to calculate from the `data` property.

Also: Please don't stringify that in the renderer, and rather do that here and properly type that property.

source/app/service-providers/commands/language-tool.ts
```diff
@@ -129,13 +130,13 @@ export default class LanguageTool extends ZettlrCommand {
...
     } = this._app.config.getConfig().editor.lint.languageTool
-    const { text, language } = arg
+    const { text, data, language } = arg
 
     if (!active) {
       return undefined // LanguageTool is not active
     }
 
-    const searchParams = new URLSearchParams({ language, text, level })
+    const searchParams = new URLSearchParams({ language, data, level })
```
nathanlesage: Stringify here instead

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -85,15 +85,38 @@ const ltLinter = linter(async view => {
...
 
-  // To avoid too high loads, we have to send a "pseudo-plain text" document.
-  // That will generate a few warnings that relate towards the Markdown syntax,
-  // but we are clever: Since we can extract the textNodes, we can basically
-  // ignore any warning outside of these ranges! YAY!
+  // To avoid too high loads, we sanitize the document
+  // according to the LanguageTool API `data` parameter.
+  // By doing so, we only generate errors for TextNode regions
+  // https://languagetool.org/http-api/swagger-ui/#!/default/post_check
+  type Annotation = { text?: string, markup?: string }
```
nathanlesage: Suggestion: Move this to the main process and import the type here. Also: Please collect types and interfaces at the top of the files for easy access.

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -85,15 +85,38 @@ const ltLinter = linter(async view => {
...
+  const annotations: { annotation: Annotation[] } = { annotation: [] }
+
+  let idx = 0
+  for (const node of textNodes) {
+    const from = node.from - node.whitespaceBefore.length
+
+    if (from - idx > 0) {
+      const markup: Annotation = { markup: view.state.sliceDoc(idx, from) }
+      annotations.annotation.push(markup)
```
nathanlesage:
```suggestion
      annotations.annotation.push({ markup: view.state.sliceDoc(idx, from) })
```

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -85,15 +85,38 @@ const ltLinter = linter(async view => {
...
+    }
+    const text: Annotation = { text: view.state.sliceDoc(from, node.to) }
+    annotations.annotation.push(text)
```
nathanlesage:
```suggestion
    annotations.annotation.push({ text: view.state.sliceDoc(from, node.to) })
```

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -85,15 +85,38 @@ const ltLinter = linter(async view => {
...
+    idx = node.to
+  }
+  // If the last TextNode does not extend to the end of the document,
+  // add the remainder as `markup`.
+  if (idx < view.state.doc.length) {
+    const markup: Annotation = { markup: view.state.sliceDoc(idx) }
+    annotations.annotation.push(markup)
+  }
+  const data = JSON.stringify(annotations)
```
nathanlesage: Don't stringify here

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -121,35 +144,24 @@ const ltLinter = linter(async view => {
...
-      ltSuggestions.matches.splice(i, 1)
-      i--
-    }
-  }
-
   // At this point, we have only valid suggestions that we can now insert into
   // the document.
+
+  // we want to filter out any words that are in our local dictionary
+  const localDictionary: string[] = await ipcRenderer.invoke('dictionary-provider', { command: 'get-user-dictionary' })
```
nathanlesage: This will have a heavy performance impact for large dictionaries. Recommendation: Keep it inside this file but outside the linter, and listen for update events from main. There are some other files that make use of the dictionary where I've implemented that.

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
```diff
@@ -41,6 +41,7 @@ const EMPTY_NODES = [
   'Document',
   'List',
   'ListItem',
+  'TaskMarker',
```
nathanlesage, line 46: What exactly does this do?

nathanlesage (changes requested):
Looks good in principle. Just so that I understand: What LanguageTool essentially asks you to do is to slice up a document into a consecutive series of markup and text objects and it will correctly provide offsets etc. like before, but just not check the markup parts of the document…?

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
benniekiss: So when I was testing, the task marker was being included as text, but I don't think that should be the case. Adding it to this list of `empty nodes` ensures that it's treated as markup.

benniekiss:
I think I should have covered everything you pointed out!

> Just so that I understand: What LanguageTool essentially asks you to do is to slice up a document into a consecutive series of markup and text objects and it will correctly provide offsets etc. like before, but just not check the markup parts of the document…?

This is correct! We don't need to do any post-filtering.

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
nathanlesage: I see, thanks for spotting this!

nathanlesage:
Excellent! Just to be sure: this is ready to merge, correct?

benniekiss:
Hmm I just did some more testing, and I realized that there are issues with inline code. Since the inline code is masked as markup, LT generates erroneous spacing errors

To make this work, we may have to include some special handling for inline formatting like code.

nathanlesage:
I don't quite understand what the issue is — how is inline code provided to LT, and what does it spit out then? Can you provide some minimal example input/output from the API?

benniekiss:
So this text:

```
Sample text with `inline` markup
```

Would be processed by LT like this:

```
Sample text with  markup
```

And it generates an error for extra spaces because of `with[__]markup`.

I might have something working to handle these edge cases. It's really an issue with how whitespace is associated with markup or text.

nathanlesage:
Ah I see. My naïve assumption now would be that if we only use the whitespace BEFORE that is accessible to the nodes, we should be able to circumvent this, no? Or does the whitespace also end up trailing after the text node?

benniekiss:
So I think this should handle the spacing issue.

There is really only one situation we need to sanitize -- inline markup where the spacing on either side is equal. Assuming that spacing should be balanced, we can ignore instances where the inline markup is surrounded by unequal spacing, since that should correctly be flagged as a spacing error. Likewise, if there's inline markup, but only one side has spacing, we don't need to sanitize anything. For example, we don't need to adjust spacing in this sentence: `my text with **b**old letters`. 

We need to sanitize inline markup with spacing on either side because this sentence: ```my text with `inline` markup``` would be interpreted as `my text with[..]markup`.

The implementation I came up with is rather simple -- take the whitespace from one side and distribute it between the strings. For single spaces, this results in one space at the end of the first string. And for more than one space, we distribute them across both strings so that any spacing errors generated will include the whole space context, not just one side of the markup.

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -14,11 +14,10 @@
  */
 
 import { linter, type Diagnostic, type Action } from '@codemirror/lint'
-import { extractASTNodes, markdownToAST } from '@common/modules/markdown-utils'
+import { extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
 import { configField } from '../util/configuration'
-import { type LanguageToolAPIResponse } from '@providers/commands/language-tool'
+import { type LanguageToolAPIResponse, type AnnotationData, type Annotation } from '@providers/commands/language-tool'
```
nathanlesage:
```suggestion
import type { LanguageToolAPIResponse, AnnotationData, Annotation } from '@providers/commands/language-tool'
```

(I like to have exports in the renderer from the main process be denoted as a single type import. If we accidentally import anything non-type, this will throw hard to trace errors.)

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -31,10 +30,42 @@ export interface LanguageToolStateField {
...
 }
 
+interface UserDictionaryStore {
+  value: Set<string>
+}
+
+// store the local dictionary for later filtering
+const userDictionary: UserDictionaryStore = {
+  value: new Set()
+}
```
nathanlesage: Why not use `const userDictionary: Set<string> = new Set()` and then use [`.clear()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/clear) to refresh the set? I see below you even use the `clear` method. So why the additional `.value`-complication?

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -31,10 +30,42 @@ export interface LanguageToolStateField {
...
+
+function updateUserDictionary (): void {
```
nathanlesage: Suggestion: `refreshUserDictionary` (since it's more of nuke-and-reload than updating with some new values)… or `resetUserDictionary`?

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -81,19 +152,88 @@ const ltLinter = linter(async view => {
...
+    const text: Annotation = {}
+
+    let from = node.from - node.whitespaceBefore.length
+    let to = node.to
+
+    text.text = view.state.sliceDoc(from, to)
+    const prevText = prevNode?.text ?? ''
+
+    // This function is meant to handle one following edge case with inline code.
+    // When we parse a string with inline code, we get the following structure:
```
nathanlesage, line 196: To be clear, this applies to any full non-text node, right? Including equations?

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -68,6 +99,46 @@ export const languageToolState = StateField.define<LanguageToolStateField>({
...
+
+  // Eventually, this may need to be expanded to handle other
+  // spacing constraints if they may arise. Currently, this
+  // will only handle cases where there is an equal amount of non-zero
+  // whitespace between the strings.
+  if (tailingWhitespace === 0 || leadingWhitespace === 0 || tailingWhitespace !== leadingWhitespace) {
+    return { string1, string2, adjust: [ 0, 0 ] }
+  }
+
+  const whiteSpace = Math.max(tailingWhitespace, leadingWhitespace)
```
nathanlesage: Based on the `if`-condition above, we don't need this `Math.max` call here.

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -68,6 +99,46 @@ export const languageToolState = StateField.define<LanguageToolStateField>({
...
+  // In case of an odd number of spaces, we bias towards the first string
+  //
+  // The end index of the first string:
+  const string1Padding = string1Trimmed.length + Math.ceil(whiteSpace / 2)
+  // The start index of the second string:
+  const string2Padding = leadingWhitespace - Math.floor(whiteSpace / 2)
+
+  return {
+    string1: string1.slice(0, string1Padding),
+    string2: string2.slice(string2Padding),
```
nathanlesage: Do we need to return pre-sliced strings? If I see it correctly, we can also directly trim the node string where necessary, no? Or, can we just return and immediately use the returned strings…?

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -81,19 +152,88 @@ const ltLinter = linter(async view => {
...
 
-  // To avoid too high loads, we have to send a "pseudo-plain text" document.
-  // That will generate a few warnings that relate towards the Markdown syntax,
-  // but we are clever: Since we can extract the textNodes, we can basically
-  // ignore any warning outside of these ranges! YAY!
+  // If the last TextNode does not extend to the end of the document,
+  // add the remainder as `markup`.
+  if (idx < view.state.doc.length) {
+    annotations.annotation.push({ markup: view.state.sliceDoc(idx) })
+  }
```
nathanlesage: Do we need this? Does this change LTs predictions? Otherwise we can just leave a comment here to indicate that trailing markup is ignored and save a few bytes to serialize and send off. (Which could be extremely large, think of a file ending with a huge code block).

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: I think this was me waffling on whether to iterate over the dict and call `.add` on each term, or just update the value as a whole -- I think I settled on just updating the whole value, in which case its easier to handle as an attribute on an object.

The `clear()` call is unnecessary, but I added it in case the `'get-user-dictionary' ` command failed, so rather than have a potentially incorrect dictionary, we have none at all.

I can definitely move it into a single `Set` object if you think that would be clearer

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: I had considered `resetUserDictionary` but it didn't seem right, but I had not thought of `refresh`, which I think is much clearer than either `update` or `reset`

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: Do you mean the comment? If so, it should apply to math, but I haven't tested that actually. It really only considers two text nodes and the white space between them, regardless of whatever markup is in between as well.

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: no, actually, we can just return the `adjust` values and slice the string in the main linter function.

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: That's a good point! I don't think we will -- the offsets should be correct.

benniekiss:
> This may give us a hint as to how they're dealing with Markdown, because I know that LT has just recently pushed an update that automagically accounts for Markdown here in GitHub comments

Hm, it may not be baked into the API. I've done some tests where it pushes the whole document and doesnt filter the errors, and it was still detecting errors related to things like list formatting. 

I'll do a couple more tests with the API and see what it generates. Maybe we dont need any filtering at all!

---

EDIT: I did some more tests, and while it does not really generate many erroneous errors, LT still parses in-line and fenced code blocks, and it generates typographical errors for lists.

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: Tested with math, and it does not generate errors

benniekiss:
> but would you mind giving it a test with various potential edge cases and adding them in a comment here?

Let me collect some test strings here, and I'll update with results:

- [ ] ```beginning of sentence with **markup**-combined with text``` <-- Expected: no errors
- [ ] ```text with `inline` code.``` <-- Expected: no errors
- [ ] ```sentence with mix_ed_ words``` <-- Expected: no errors
- [ ] ```my **bold**  sentence with unbalanced spacing after``` <-- Expected: errors
- [ ] ``` my  _italicized_ text with unbalanced spacing before``` <-- Expected: errors
- [ ] ```text with   `balanced`   spacing but incorrect number``` <-- Expected: errors

Erros generated by language tool:

Screenshot 2025-09-23 at 15 09 39

benniekiss:
Ok, I found some bugs with the whitespace -- it seems it's just an index adjustment issue, but here's a screenshot showing what's off. Up to this point, all of the lints are correctly positioned, so something happens in this block:

Screenshot 2025-09-23 at 15 36 19

EDIT: The issue is caused by formatting within blockquotes~~the header within the blockquote~~. But I have not tracked down *why*

benniekiss:
The above issue is solved with the latest commit. It was another node type missing from the `EMPTY_NODES` list.

source/common/modules/markdown-editor/linters/language-tool.ts
nathanlesage: Yeah I think that would be better. I'm not really a fan of single-property-objects. So unless there's a good reason for that, I'd prefer the set directly in the variable. Thanks!

nathanlesage:
Wow, it's almost as if when you're implementing an entire AST parser in a single PR you're bound to forget things 🙈  I'm impressed by how many issues you discover with this thing!

What about the "last block of markup in API call" issue from above? Did you find something?

(Also, jeez, I don't have to write any code, but I am feeling that I'm already falling behind your PR speed with just reviewing. You don't happen to actually be an entire office full of software developers?)

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
```diff
@@ -34,13 +34,15 @@ const EMPTY_NODES = [
   'SubscriptMark',
   'HighlightMark',
   'HeaderMark',
+  'Blockquote',
```
nathanlesage, line 37: Is this really just the formatting character of the Blockquote? It sounds weird that this doesn't follow the `Mark` naming convention that Marijn has with the rest of the nodes

benniekiss:
> What about the "last block of markup in API call" issue from above? Did you find something?

I did not see any issue!

> You don't happen to actually be an entire office full of software developers?

Haha, only a dev office of one :) I'm a contract worker, and I'm trying to move away from Office/Word, so I try to justify all this work as part of the 'day job'

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
benniekiss: I don't think so, I think that's `QuoteMark`. I think this `Blockquote` node is some higher-level structure. Excluding it, the content of the blockquote is still linted.

nathanlesage:
> Haha, only a dev office of one :) I'm a contract worker, and I'm trying to move away from Office/Word, so I try to justify all this work as part of the 'day job'

Ahh — still much more experience and time than I have! And I really appreciate it, I think it has been quite some time since Zettlr has received so many improvements! :)

benniekiss:
Well thank you for maintaining it! I'm just happy I'm able to contribute

benniekiss:
So this latest commit is open for consideration. It adds an `Ignore Rule` action which will add the LT rule to a list stored in the existing statefield. The list is reset every time the editor is mounted/unmounted, so the ignored rules can be reset by opening and closing the document, which I think is useful for this feature approach. This could be added to the PersistentView, so that switching document tabs does not reset the state.

By clicking the action, or otherwise updating the ignored rules, the linter will run again to clean up any other lints.

There is minimal styling on the Action, which has its background set as `OrangeRed`, but I'm not sure if the current approach of selecting it will pick up if the string is translated.

nathanlesage:
I think adding the ignore rule is already great. My suggestion to make my work easier would be to remove all the additional code and simply keep the no-op "action" as part of the commit with a "TODO" commit. I can then go and actually implement a persistent state in the main process.

Or, you know what, we can keep that in for now, I don't want you to throw away any work. I think it looks great! Does it work as advertised now? If so, we can merge.

benniekiss:
I think its working so far, except for the fact it resets when switching tabs, which is annoying to me. Also, as I was afraid, the button styling selector doesnt really work with translated strings.

To persist the list, all we have to do is get/set the value in the persistent state getter? If so, I can try out something simple later. 

Otherwise, I can easily move this to a separate PR to work out those issues.

nathanlesage:
> Also, as I was afraid, the button styling selector doesnt really work with translated strings.

What do you mean …?

> Otherwise, I can easily move this to a separate PR to work out those issues.

Or you can let me have this one, I know where the pieces are, and it should be fairly straight forward to quickly move the stuff into a persistent place!

benniekiss:
In terms of the button, I use the `aria-label` selector based on "Ignore Rule", but if we add i8n via `trans` to the label, the `aria-label` is the translated string, and the styling doesn't apply.

I'm happy to let you take over! I can go ahead and drop or revert the last commit to clear this up

nathanlesage:
Nah, keep it in, I can try to make it work! Regarding the theming; I would honestly keep it off for now. It's not really intended, although I'd be inclined that this would be a good feature for the linter plugin (to add, e.g., class names to the buttons). An option for the translation could be the [phrases plugin](https://codemirror.net/docs/ref/#state.EditorState%5Ephrases), but if we don't end up styling that one single button, I'd say we don't yet need that.

benniekiss:
If we dont style it, what are your thoughts on having it at the start vs end of the list?

nathanlesage:
Definitely end!

benniekiss:
Great! If you want to give this one more look, it's good to merge from my perspective.

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
nathanlesage: Please check if that breaks the HTML parser (e.g., by writing a document with a blockquote and "printing" it). I have the suspicion that doing this will break the HTML conversion.

nathanlesage:
I had one final question, but otherwise it looks good!

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
benniekiss: Would testing with the new print preview be enough? If so, I don't think blockquotes are picked up in develop, either.

I think the problem is because when blockquotes have internal formatting, the whitespace between the quotemark and the additional inline mark is essentially dropped? I'm still having a hard time tracking it down.

benniekiss: Ok, I think I figured it out! There's an issue in the html parser, but I think I solved it. Adding `Blockquote` to the empty nodes does not mess up parsing, if my fix is correct

benniekiss:
FWIW, I had a PR merged into codemirror/lint which should allow us to style individual actions (by assigning a class). So once that is released, we can add some styling to the ignore action (and others if we want)

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
benniekiss: Also, see the latest commit for the small fix

nathanlesage:
Beautiful, thanks for also improving CodeMirror for all :)

nathanlesage:
I've now had a look over your changes, and I think it looks fine. Is it ready to merge?

benniekiss:
I think so!

nathanlesage:
Wonderful, then thank you once more, I'll get to adding the ability to ignore rules for good!

nathanlesage:
@benniekiss I found an edge case!

The following section (taken from the Zettlr tutorial):

```markdown
Whenever you stumble upon `Cmd/Ctrl`, read this as
```

yields on sending to the API:

```markdown
Whenever you stumble upon , read this as
```

Which means that there's a comma surrounded by two spaces, which makes LanguageTool complain that there's a space in front of the comma.

benniekiss:
Oh such an obvious oversight on my part! 

I'm not sure the best way to handle this, since the inline code is basically being used as text.

We can tell language tool to "interpret" the markup as other things, so my first instinct would be to also extract inline code and math, and add additional "interpretAs" keys to those sections based on the interior text. I'm pretty sure this would fix those problems, and remove the need to adjust whitespace.

Inline code/math seems to be the biggest disrupter for this approach. I'm almost willing to go back to the old method of sending the whole document and filtering afterwards.

benniekiss:
Also, just a thought on the current persistence implementation, I had originally envisioned the "Ignore Rule" action to be more of a temporary, per-document ignore, with a separate list for globally disabled rules. This is similar to how Word lets you "ignore" a misspelled word without adding it to a dictionary.

Im not entirely against the action adding the rule to the global list, especially since it can be removed and makes it clearer which rules are active.

Just wanted your thoughts on the temporary vs permanent difference. If it stays how it currently is, I think the action should say "Disable" instead of "Ignore"

benniekiss:
> We can tell language tool to "interpret" the markup as other things, so my first instinct would be to also extract inline code and math, and add additional "interpretAs" keys to those sections based on the interior text. I'm pretty sure this would fix those problems, and remove the need to adjust whitespace.

This may be simpler than I thought, where for every "markup" object, we just set the "interpretAs" parameter to the same content, and that should solve all of the spacing issues. Ill do some testing on my end to check this.

---

EDIT: It's not as simple as passing the `interpretAs` parameter. When that's done, LT will then generate errors for the `interpretAs` text. So the most straight-forward way to remove these would be to include the inline code and math with the text, then filter those errors out later.

nathanlesage:
It's really not an obvious oversight — Markdown is difficult to parse! Don't be too hard on yourself.

I'll have a look at your fix soon!

Regarding your question:

This can be a bit iffy. Globally disabling rules is easy because we have a globally shared state. When it comes to ignoring certain rules, it's more difficult. The Word behavior is to quite literally only ignore the one specific instance of a rule. We could reflect this in the state by noting the from and to properties in a state field, and continuously mapping these positions through any changes that happen. This approach will work for as long as the user has the document open only once, and keeps it open. We could move this stuff into the EditorPane itself so that the remembering happens per editor pane, which would be probably satisfactory. However, then we would have no control over whether the document is being edited in other editor panes, again making the positions out of sync.

That's why I believe a sensible way is to -- for now -- only globally disable the rule. We can certainly rename the button accordingly. We should either way remove the rule ID, because many are quite long.

PR #5917, opened 2025-09-22
[Feat] Show line numbers and line endings

benniekiss:
## Description

This PR adds the ability to show line numbers and line endings (pilcrows)

## Changes

Two extensions were added to allow a user to enable showing line numbers and line endings.

The line numbers extension uses the built-in extension from codemirror.

The line ending extension adds a pilcrow widget which is enabled along with the `Highlight Whitespace` setting. It adds a DOM element at the end of each line.~~assigns a CSS class which shows the pilcrow at the end of lines. The CSS was adapted from the one in the documentation to not show a newline when a pilcrow is on a blank line.~~

## Additional information

Screenshot 2025-09-22 at 14 16 17

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/editor-extension-sets.ts
+++ b/source/common/modules/markdown-editor/editor-extension-sets.ts
@@ -314,7 +314,7 @@ export function getMarkdownExtensions (options: CoreExtensionOptions): Extension
     markdownSyntaxHighlighter(),
     syntaxExtensions, // Add our own specific syntax plugin
     renderers(options.initialConfig),
-    showLineNumbers(options.initialConfig.showLineNumbers),
+    showLineNumbers(options.initialConfig.showMarkdownLineNumbers),
     mdLinterExtensions,
     languageTool,
     // Some statistics we need for Markdown documents
--- a/source/common/modules/markdown-editor/plugins/line-numbers.ts
+++ b/source/common/modules/markdown-editor/plugins/line-numbers.ts
@@ -18,17 +18,11 @@ import {
 import {
   type Extension,
   Compartment,
-  StateEffect,
   EditorState
 } from '@codemirror/state'
 import { configUpdateEffect } from '../util/configuration'
 
 const extensionCompartment = new Compartment()
-
-/**
- * Toggle whether to show line numbers
- */
-export const showLineNumbersEffect = StateEffect.define<boolean>()
 
 /**
  * A TransactionExtender that reconfigures the line number compartment in response
@@ -41,11 +35,9 @@ const modeSwitcher = EditorState.transactionExtender.of(transaction => {
   for (const effect of transaction.effects) {
     // Allow updating both via the main config and a dedicated effect.
     if (effect.is(configUpdateEffect)) {
-      if (effect.value.showLineNumbers !== undefined) {
-        showLineNumbers = effect.value.showLineNumbers
+      if (effect.value.showMarkdownLineNumbers !== undefined) {
+        showLineNumbers = effect.value.showMarkdownLineNumbers
       }
-    } else if (effect.is(showLineNumbersEffect)) {
-      showLineNumbers = effect.value
     }
   }
 
--- a/source/common/modules/markdown-editor/util/configuration.ts
+++ b/source/common/modules/markdown-editor/util/configuration.ts
@@ -73,7 +73,7 @@ export interface EditorConfiguration {
   theme: MarkdownTheme
   margins: 'S'|'M'|'L'
   highlightWhitespace: boolean
-  showLineNumbers: boolean
+  showMarkdownLineNumbers: boolean
   countChars: boolean
 }
 
@@ -131,7 +131,7 @@ export function getDefaultConfig (): EditorConfiguration {
     theme: 'berlin',
     margins: 'M',
     highlightWhitespace: false,
-    showLineNumbers: false,
+    showMarkdownLineNumbers: false,
     countChars: false
   }
 }
--- a/source/win-main/MainEditor.vue
+++ b/source/win-main/MainEditor.vue
@@ -278,7 +278,7 @@ const editorConfiguration = computed<EditorConfigOptions>(() => {
     darkMode,
     theme: display.theme,
     highlightWhitespace: editor.showWhitespace,
-    showLineNumbers: editor.showLineNumbers,
+    showMarkdownLineNumbers: editor.showMarkdownLineNumbers,
     countChars: editor.countChars
   } satisfies EditorConfigOptions
 })
--- a/source/win-preferences/schema/editor.ts
+++ b/source/win-preferences/schema/editor.ts
@@ -284,8 +284,8 @@ export function getEditorFields (config: ConfigOptions): PreferencesFieldset[] {
         },
         {
           type: 'checkbox',
-          label: trans('Show Line Numbers'),
-          model: 'editor.showLineNumbers'
+          label: trans('Show line numbers for Markdown files'),
+          model: 'editor.showMarkdownLineNumbers'
         },
         {
           type: 'checkbox',
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -96,7 +96,7 @@ export interface ConfigOptions {
     showStatusbar: boolean
     showFormattingToolbar: boolean
     showWhitespace: boolean
-    showLineNumbers: boolean
+    showMarkdownLineNumbers: boolean
     defaultSaveImagePath: string
     enableTableHelper: boolean
     indentUnit: number
@@ -283,7 +283,7 @@ export function getConfigTemplate (): ConfigOptions {
       autoCloseBrackets: true,
       showLinkPreviews: true, // Whether to fetch link previews in the editor
       showWhitespace: false,
-      showLineNumbers: false,
+      showMarkdownLineNumbers: false,
       defaultSaveImagePath: '',
       citeStyle: 'regular', // Determines how autocomplete will complete citations
       enableTableHelper: true, // Enable the table helper plugin


nathanlesage:
I'm not a fan, to be honest.

First, until now we always showed line numbers if the file was a code file (JSON, YAML, LaTeX), but never when the file was a Markdown file. There were some calls that some users preferred to turn on line numbers for Markdown files in the settings, and I do believe we can definitely do that. But it is crucial that line numbers in Markdown files are separate from line numbers in code files. This is not the case how you implemented it.

> As an aside: now that we have a MarkdownEditor, an ImageViewer, and a PDFViewer, I was contemplating splitting up the Markdown vs. code editing into two separate components to have a clearer separation between Markdown files and code files. This makes it easier to determine which plugins are active in code files vs. markdown files; also we currently make use of some additional CSS that is outside of the editor for some effects, and I believe that is dirty.

Furthermore, I do not like that line ending visualizers are a separate setting. We do have a `showWhitespace` setting, and I strongly believe that pilcrows should be shown based on that setting. However, the original plugin does not allow that right now. I would go ahead and implement it directly in the corresponding plugin, because that saves a lot of space (and also brings benefits to other CodeMirror users). Finally, implementing this using decorations also has the benefit that we don't need external CSS rules, which I definitely like.

See: https://github.com/codemirror/view/blob/main/src/highlight-space.ts

benniekiss:
Ah I noticed I had overwritten the code line numbers at some point but thought I had reverted that! I can drop the pilcrows since I agree with your point about moving it into the white space extension.

Should I move forward with re-working the line numbers?

benniekiss:
I asked the codemirror maintainer if they would be open to adding a pilcrow extension to the main library, but they said they don't want to add more utilities, so this should be implemented in the 3rd party code. I can go ahead and split this into two PRs, and work the current pilcrow implementation into the whitespace extension we currently have, or I can drop the pilcrows entirely if you only want to deal with line numbers at the moment.

nathanlesage:
That's fair, and I suspected that he may want to keep the functionality lean. However, while researching this yesterday, I found that he simply uses individual mini-extensions for that. So what I believe we should do is that we should essentially just copy his code, and adapt it for pilcrows, and then just activate that new extension together with the highlight whitespace method. This is the least-intrusive method for this.

When it comes to line numbers, can you explicate what exactly you hope to achieve with this? Then I can give you targeted recommendations of how to implement this with the least amount of work necessary.

benniekiss:
Should I keep the pilcrow code within the highlight whitespace file? It would be the same process of assigning the right pilcrow class to lines, then writing the CSS.

I'm not sure what you mean about line numbers. I just want to make them toggleable for markdown files. The current implementation essentially copies the whitespace extensions setup. My workflow really benefits from having them available.

nathanlesage:
> Should I keep the pilcrow code within the highlight whitespace file? It would be the same process of assigning the right pilcrow class to lines, then writing the CSS.

Yes, please. And, do we absolutely need CSS, or can't we just add a decoration with the pilcrow symbol…? If we absolutely need CSS, I would like the CSS to be a theme that can be included together with the extension. This way, from the perspective of the editor, we only have to include a single extension. A great example is the typewriter extension: https://github.com/Zettlr/Zettlr/blob/develop/source/common/modules/markdown-editor/plugins/typewriter.ts

> I'm not sure what you mean about line numbers. I just want to make them toggleable for markdown files. The current implementation essentially copies the whitespace extensions setup. My workflow really benefits from having them available.

You had originally removed them from the core code extensions unless I completely overlooked that. That's why I was confused. The line numbers extension should always be included in the code extensions, and can then be added to the Markdown extension set conditional on the provided setting (and yeah, you can write analogous code to the whitespace extension).

_However_, in case you have some time on your hands for an experiment: I believe I understood the CodeMirror documentation in such a way that one can conditionally toggle extensions based on the field of some facet. Why do I mention this? Because I believe that our setup of Boolean preference -> enables/disables extension fits this workflow. This would then obliterate the need for (a) a compartment, and (b) this somewhat unhandy effect setup. Might in general make the code much more comprehensive.

source/common/modules/markdown-editor/editor-extension-sets.ts
```diff
@@ -27,7 +27,6 @@ import { Compartment, EditorState, type Extension } from '@codemirror/state'
 import {
   drawSelection,
   EditorView,
-  lineNumbers,
```
nathanlesage, line 30: Ahh, I think I see where our misunderstanding comes from. I want the `lineNumbers` extension to always be included in the `genericCodeExtensions` with no toggle. You can absolutely include your toggleable wrapper in the Markdown extension set.

You've accidentally modified the `coreExtensions` which are the **shared** extensions between both code and Markdown files. Maybe that was a misunderstanding because I named them core, not shared?

nathanlesage: NB: So essentially what you want:

1. Add the pilcrow extension to the whitespace wrapper I wrote.
2. Do not touch the core extensions — the pilcrow wrapper will automatically included because its part of the whitespace wrapper.
3. Do not remove the lineNumbers extension and keep it in the code extension set.
4. Import your custom toggleable line numbers extension in addition to the non-toggleable, and add that one to the Markdown extension set.

I hope now it makes more sense. I agree that this extension set file is a bit iffy to understand.

benniekiss:
> Yes, please. And, do we absolutely need CSS, or can't we just add a decoration with the pilcrow symbol…?

We don't actually -- I can make a `Decoration.widget` instance that applies all this via code. I have something that I think is much more compact.

> You had originally removed them from the core code extensions unless I completely overlooked that. 

Ah, yeah, I had meant to restore this and only add the toggle to the markdown editor -- with the current implementation, it should only be a couple of line changes. But I like the idea of looking into a more native-codemirror method for toggling.

let me push the updated changes if you want to give them a look over, and I'll start looking into the facet toggle.

source/common/modules/markdown-editor/plugins/line-numbers.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * A configurable line number renderer.
+ *
+ * @param   {boolean}      show  Initial setting for the highlighter
+ *                                    (default: false)
+ *
+ * @return  {Extension[]}             The extension.
+ */
+export function showLineNumbers (show?: boolean): Extension[] {
+  const initialSetting = show === true ? extensionCompartment.of([lineNumbers()]) : extensionCompartment.of([])
```
nathanlesage: I think the default behavior if no `show` value is given should be to actually show the line numbers. It seems weird to add an extension that defaults to not doing anything.

Also, can't we shorten the expression a bit by moving the expression into the extension compartment…?

source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
```diff
@@ -16,26 +16,82 @@ import {
...
- * to exchange the light or dark themes that the editor uses, or to switch
- * between the light and dark mode.
+ * A widget which adds a pilcrow, '¶', to the end of lines.
+ */
+class PilcrowWidget extends WidgetType {
+  toDOM () {
+    const span = document.createElement('span')
+    span.className = 'cm-pilcrow'
+    span.textContent = '¶'
+    span.style = 'color: #666'
```
nathanlesage: Mh, while I think the pilcrow symbol shouldn't be CSS as we discussed, I do believe that the coloring should be moved to a theme, just so that it's a bit more colocated (see the Typewriter plugin). It's gonna be a bit empty, but I think it makes the code more streamlined and easier to understand!

nathanlesage (commented):
Looks good! Just two mini things!

source/common/modules/markdown-editor/plugins/line-numbers.ts
benniekiss: How about something like 

```ts
const lineNumbersExtension = extensionCompartment.of(show ?? true ? [lineNumbers()] : [])
```

benniekiss:
I moved the styling to an editor theme, matching the color with the highlight whitespace color, and I added opacity so that it looks less like a character at the end of the line.

source/common/modules/markdown-editor/plugins/line-numbers.ts
```diff
@@ -0,0 +1,73 @@
...
+  EditorState
+} from '@codemirror/state'
+import { configUpdateEffect } from '../util/configuration'
+
+const extensionCompartment = new Compartment()
+
+/**
+ * Toggle whether to show line numbers
+ */
+export const showLineNumbersEffect = StateEffect.define<boolean>()
```
nathanlesage: We can remove this, since it's not used. This plugin exclusively depends on the config effect. While I'm slowly moving towards wanting to split that up, and make each plugin self-contained, that should be done in another, unified PR that changes this. For now, the config update effect suffices.

source/common/modules/markdown-editor/util/configuration.ts
```diff
@@ -73,6 +73,7 @@ export interface EditorConfiguration {
   theme: MarkdownTheme
   margins: 'S'|'M'|'L'
   highlightWhitespace: boolean
+  showLineNumbers: boolean
```
nathanlesage:
```suggestion
  showMarkdownLineNumbers: boolean
```

Just to make it more explicit that this is only intended to be used for Markdown files

source/common/modules/markdown-editor/util/configuration.ts
```diff
@@ -130,6 +131,7 @@ export function getDefaultConfig (): EditorConfiguration {
     theme: 'berlin',
     margins: 'M',
     highlightWhitespace: false,
+    showLineNumbers: false,
```
nathanlesage:
```suggestion
    showMarkdownLineNumbers: false,
```

Same as above

source/common/modules/markdown-editor/editor-extension-sets.ts
```diff
@@ -313,6 +314,7 @@ export function getMarkdownExtensions (options: CoreExtensionOptions): Extension
     markdownSyntaxHighlighter(),
     syntaxExtensions, // Add our own specific syntax plugin
     renderers(options.initialConfig),
+    showLineNumbers(options.initialConfig.showLineNumbers),
```
nathanlesage:
```suggestion
    showLineNumbers(options.initialConfig.showMarkdownLineNumbers),
```

Same as above

source/win-main/MainEditor.vue
```diff
@@ -278,6 +278,7 @@ const editorConfiguration = computed<EditorConfigOptions>(() => {
     darkMode,
     theme: display.theme,
     highlightWhitespace: editor.showWhitespace,
+    showLineNumbers: editor.showLineNumbers,
```
nathanlesage:
```suggestion
    showMarkdownLineNumbers: editor.showMarkdownLineNumbers,
```

Same as above

source/win-preferences/schema/editor.ts
```diff
@@ -282,6 +282,11 @@ export function getEditorFields (config: ConfigOptions): PreferencesFieldset[] {
           label: trans('Show formatting toolbar when text is selected'),
           model: 'editor.showFormattingToolbar'
         },
+        {
+          type: 'checkbox',
+          label: trans('Show Line Numbers'),
+          model: 'editor.showLineNumbers'
```
nathanlesage:
```suggestion
          label: trans('Show Line Numbers for Markdown files'),
          model: 'editor.showMarkdownLineNumbers'
```

Same as above + wording

benniekiss:
I think that should cover all your comments. This lgtm on my end!

nathanlesage:
Perfect -- merging.

PR #5921, opened 2025-09-24
[CSS] Cleanup print preview CSS

benniekiss:
## Description

This PR fixes some CSS related to print previews.

## Changes

The `line-height` parameter was changed from a percentage to a number. Setting it this way is the [recommended approach](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#number), likewise, it means other elements with different font sizes inherit the property correctly. It was set to `1.5` to meet [minimum accessibility recommendations](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#accessibility).

The `margin-top` parameter for headings is only applied to all but the first element. This removes additional margin at the top of the document.
 
## Additional information

Current:

Screenshot 2025-09-24 at 11 54 50

PR:

Screenshot 2025-09-24 at 11 54 37

Tested on:

### DIFF
--- a/source/win-print/App.vue
+++ b/source/win-print/App.vue
@@ -106,7 +106,7 @@ function handleClick (buttonID?: string): void {
 #print-container {
   width: 100%;
   border: none;
-  
+
   // Styles primarily copied from the Pandoc default HTML template to emulate
   // the old print preview:
   // https://github.com/jgm/pandoc/blob/main/data/templates/styles.html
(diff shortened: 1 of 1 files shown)


nathanlesage:
Great, BUT: I think a *little* bit less line height might be in order, and a different value for the headings; that H1 in your example looks atrocious :D 

Other than that, looks good!

benniekiss:
Here's `1.2` for both headings and text:

Screenshot 2025-09-24 at 17 07 21

And here's `1.2 `for text, and `1` for headings:

Screenshot 2025-09-24 at 17 07 10

I think going with `1.2` is best, and what I think was originally intended. `1` for headings is almost too small, as the bottom of the `y` nearly touches the top of characters on the line below

nathanlesage:
I agree, thank you!

PR #5923, opened 2025-09-27
[Fix] Add special handling for InlineCode (`` and $$) in LanguageTool pre-parsing

benniekiss:
## Description

This PR implements a simpler way to handle inline markup when sanitizing the payload for LanguageTool.

## Changes

When we sanitize the payload for LanguageTool, inline code and math pose unique issues due to how LT parses the spacing. Initially, I tried manually adjust the whitespace, however, there are too many edgecases.

This approach tries to simplify sanitization. By including the inline markup with the regular text and later filtering out diagnostics generated for those inline regions, LT processes the text correctly, and we can exclude any unwanted errors.

## Additional information

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/linters/language-tool.ts
+++ b/source/common/modules/markdown-editor/linters/language-tool.ts
@@ -14,7 +14,7 @@
  */
 
 import { linter, type Diagnostic, type Action } from '@codemirror/lint'
-import { extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
+import { extractASTNodes, extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
 import { configField } from '../util/configuration'
 import type { AnnotationData, Annotation, LanguageToolLinterRequest, LanguageToolLinterResponse } from '@providers/commands/language-tool'
 import { StateEffect, StateField, type Transaction } from '@codemirror/state'
@@ -22,6 +22,7 @@ import extractYamlFrontmatter from 'source/common/util/extract-yaml-frontmatter'
 import type { ViewUpdate } from '@codemirror/view'
 import { trans } from 'source/common/i18n-renderer'
 import type { LanguageToolIgnoredRuleEntry } from '@providers/config/get-config-template'
+import type { InlineCode } from '../../markdown-utils/markdown-ast'
 
 const ipcRenderer = window.ipc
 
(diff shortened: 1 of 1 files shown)


nathanlesage:
Do I read this PR correctly in that it approaches this using a dual-way? That is, you filter out most processing by providing `markup` chunks to LT, but treat inline code separately?

benniekiss:
Yes! Since we're adding the `InlineCode` nodes as text, there's a filter that excludes errors generated for those regions. Otherwise, everything else is still treated as markup.

benniekiss:
Responding to your other comment here just to centralize things:

> That's why I believe a sensible way is to -- for now -- only globally disable the rule. We can certainly rename the button accordingly. We should either way remove the rule ID, because many are quite long.

That makes complete sense to me! I think it's fine to keep it globally disabled, and I'm not too set on having a per-editor ignore list, that was just my initial direction so I wanted to clear it up.

I think removing the rule from the Action button is fine, but if that's the case, should the `source` parameter instead include the rule ID rather than the category? Currently, it shows something like `language-tool(mispelling)`, but it may be helpful to switch to the rule ID so users are more aware of what's being ignored. I'm open to keeping the category since the rule is shown in the preferences ignored list.

nathanlesage:
That's a great example of why it's sometimes important to think from the perspective of a user: for us it's pretty evident that the rule ID is kind of the only important info we really need, because technically it certainly is. But after having used the LT integration for about a year by now, I can say that my somewhat arbitrary decision to show the category and not the rule was amazing: it feels very natural, because you can quickly see whether you just violated a stylistic rule (which is something you can ignore more easily than a spelling mistake). Then, when it comes to displaying the rule ID on the disable button: again, from the context, the user sees that they don't want these types of errors marked anymore. They don't care about the exact ID, and I think it would even be fine to not show the rule ID in the settings, because what the user needs to decide whether to reenable a rule is to view the rules name!

That's just my two cents :D I hope I made sense!

benniekiss:
Totally! I agree that the tooltip is more useful with the category. I'll go ahead and remove the rule id from the button. I think it should still be in the settings, both so users can know what exactly is disabled, and in the case multiple rules with the same category are disabled, they can be differentiated.

nathanlesage:
I can't comment on the PR itself due to a bug, but one thing: wouldn't it make more sense to keep only one AST and extract nodes from there? I'm asking because your current approach may be a bit more performance heavy and may yield false results since you're re-parsing just a part of the document!

benniekiss:
> wouldn't it make more sense to keep only one AST and extract nodes from there

That definitely sounds like a better approach. 

Did I introduce the bug here?

nathanlesage:
I wasn't speaking of a bug! I just saw what your code did and was wondering if we couldn't just reuse the existing AST for extracting the InlineCode nodes, rather than reparsing the parts of the document again…?

benniekiss:
Oh, I was referencing you not being able to comment because of a bug, I was just curious

nathanlesage:
Ah that is likely the GitHub app, nothing on your end!

source/common/modules/markdown-editor/linters/language-tool.ts
```diff
@@ -283,7 +222,13 @@ const ltLinter = linter(async view => {
   // At this point, we have only valid suggestions that we can now insert into
   // the document.
   for (const match of ltSuggestions.matches) {
-    const word = view.state.sliceDoc(match.offset, match.offset + match.length)
+    const matchFrom: number = match.offset
+    const matchTo: number = match.offset + match.length
+
+    // Exclude diagnostics overlapping with our filtered regions
+    if (filterRegions.some(range => !(matchTo <= range.from || matchFrom >= range.to))) { continue }
```
nathanlesage: Small change: Why not use `codeNodes` directly here? After all, `filterRegions` is essentially the same. Would make it a bit more obvious where this variable is coming from!

nathanlesage (commented):
Looks good — one minor thing, but that's purely aesthetic. As long as it works we're golden!

source/common/modules/markdown-editor/linters/language-tool.ts
benniekiss: Agreed! Should be fixed.

nathanlesage:
Thank you once more!

PR #5924, opened 2025-09-27
[Enhancement] Improve spellchecker linting

benniekiss:
## Description

This PR implements hunspell functionality for the local dictionary and enables spelling suggestions for mispelled words.

## Changes

The dictionary provider was refactored to load and save the local dictionary as a hunspell dictionary. If an affix file does not exist, a barebones version is written at `user.aff` with the contents `SET UTF-8` whenever `persistUserDictionary()` is called. The file is only written if it does not exist.

`persistUserDictionary()` was refactored so that the first line of the written dictionary includes the number of terms in the list. I tried to implement this without that line, which I don't believe is technically necessary, but the local dictionary would not load without it.

Multiple methods were refactored to use the `async` hunspell functions, since Nodehun recommends against using the `Sync` versions.

The `add` function was refactored so that it now adds words to the runtime dictionary.

A `remove` function was added to remove words from the runtime user directory and remove them from the words list. Likewise, the `setUserDictionary()` function was refactored to use the remove method.

The `spellcheck` linter now returns spelling suggestions, currently limited at 3 per loaded dictionary. The linter now uses the `Intl.Segmenter` class to split text nodes into words rather than using regex.

## Additional information

This should allow users to write dictionaries with the full power of hunspell.

Regarding the refactored `setUserDictionary()` method, I'm not sure if there are cases where this kind of `add/remove` would get out of sync, so I'm happy to revert it to the old way of a complete reset. I went this approach to avoid having to reload the hunspell object.

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/dictionary/index.ts
+++ b/source/app/service-providers/dictionary/index.ts
@@ -96,6 +96,8 @@ export default class DictionaryProvider extends ProviderContract {
         return Promise.all(terms.map(t => this.suggest(t, limit)))
       } else if (command === 'add') {
         return Promise.all(terms.map(t => this.add(t)))
+      } else if (command === 'remove') {
+        return Promise.all(terms.map(t => this.remove(t)))
       }
     })
 
--- a/source/app/service-providers/dictionary/index.ts
+++ b/source/app/service-providers/dictionary/index.ts
@@ -400,14 +400,14 @@ export default class DictionaryProvider extends ProviderContract {
       return suggestions
     }
 
-    if (this._userHunspell !== undefined) {
-      const userSuggestions = await this._userHunspell?.suggest(term) ?? []
-      suggestions.push(...userSuggestions.slice(0, limit))
-    }
-
     for (const dictionary of this.hunspell) {
       const values = await dictionary.suggest(term) ?? []
       suggestions.push(...values.slice(0, limit))
+    }
+
+    if (this._userHunspell !== undefined) {
+      const userSuggestions = await this._userHunspell?.suggest(term) ?? []
+      suggestions.push(...userSuggestions.slice(0, limit))
     }
 
     return suggestions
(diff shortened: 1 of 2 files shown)


nathanlesage:
Dude, I just recently fixed a minor bug with the minimal amount of work possible after running into weird bugs, and you're just casually porting the entire provider to using hunspell directly — you're on fire! :D 

What would be the benefit of using a hunspell dictionary for user dictionaries? As far as I understand it, there's little benefit to it, isn't it?

benniekiss:
The biggest benefit is a user can provide flags to words if they supply their own dictionary and take full advantage of hunspell, and now the implementation returns spelling suggestions for words in the user dictionary.

nathanlesage:
Users can already import dictionaries via the GUI. How would they/could they set flags?

benniekiss:
I didnt actually realize there was a way to import dictionaries! Currently, there's no way for a user to add flags without manually editing the file, but I think the settings UI could be improved to allow manually adding words with an optional flag field.

Other than the dictionary, we could provide a more robust affix file, or a user could supply their own, to set case insensitive matching, etc. which would otherwise require reimplementation on our end, or the manual addition of more words.

I did this largely to make the "add to dictionary" function more useful, but I didn't know about importing dictionaries, so if I'm already maintaining one, its not too disruptive to just import it. I'm happy to restore the old list checking method if you think it would be best

nathanlesage:
I think for now it would definitely be good to keep the older functionality going — just because this one may introduce new bugs, and it's not (yet) needed or requested by many. I really appreciate your enthusiasm, and please don't take this personally. I am just thinking that in order to release the next version we should be tackling a few open bugs instead of cramping even more features in right now. Once the next version is out, there's lots of space for the next flood of feature PRs from your side :D 

But please, keep all of these ideas you have, because so far you've rarely done anything I don't appreciate in terms of added functionality!

benniekiss:
No problem! I'll close this, but I'll keep the branch going on my end for a later time

Issue #5926, opened 2025-09-29
Edge cases in new LanguageTool implementation


nathanlesage:
### Description

Hi @benniekiss, here are a few additional edge cases in the rewritten LT implementation. Most are due to inline code, such as citations (see image):

<img width="1470" height="431" alt="Image" src="https://github.com/user-attachments/assets/e10ec56a-a079-42d1-bc35-957d0581b84f" />

But there are now also new issues popping up surrounding headings, which LT didn't complain about before:

<img width="1516" height="301" alt="Image" src="https://github.com/user-attachments/assets/de330562-b7c2-49a8-886f-3c7d19dda864" />

### Reproducing

We may need to rethink what the linter is doing to the text. I know it's possible to vet Markdown with LanguageTool, because the browser plugin correctly checks, e.g., the GitHub content.

### Zettlr version

develop/HEAD

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

macOS 26

benniekiss:
I hadn't considered citations, and those can be addressed in a similar fashion to InlineCode.

The reason headings are now linted is because I changed the extraction to use `extractTextNodes` rather than `extractAstNodes('Text')`, which was previously only extracting paragraph text. I could restore the previous extraction, which would remove linting for headings, and a few other nodes like list items.

I've tested the API with a complete markdown document and no filtering. However, it still generates errors for inline code items, and it generates several typographical errors for things like lists. So filtering is still required on our end. I'm pretty sure the way the LT plugin handles Github markdown is through some kind of pre-filtering.

nathanlesage:
I think it's easier than we think. I just started a debugger of the LanguageTool extension to see how it would communicate with LanguageTool servers (since the developers should probably know best), and here are a few observations:

**Observation 1: Markdown is passed as-is No `markup` nodes.**

**Observation 2: The extension straight out disables the [`WHITESPACE_RULE`](https://community.languagetool.org/rule/show/WHITESPACE_RULE?lang=en)**

**Observation 3: If you start a new paragraph (== two LF) it will attempt to first only lint that one line, before occasionally re-checking the entire text.**

**Observation 4: The extension straight-up strips all markup, but appears to be remembering where the markup was, and accounts for this.**

[...]

PR #5928, opened 2025-09-29
[Fix] Revert LanguageTool pre-processing

benniekiss:
## Description

This PR reverts the document pre-processing before sending the payload to the API. Instead, the entire document is sent to the API, and responses are filtered afterwards based on extracted text nodes. This should hopefully clear up any edge cases in how documents are processed.

A later PR will implement an incremental linter for languagetool so that the entire document is not constantly sent to the API.

## Additional information

While this removes any preprocessing, I would consider implementing some to exclude code blocks, math blocks, and other block-level markdown that should not be linted in order to reduce API load, but that in itself may introduce other edge cases. Just a thought for consideration.

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/linters/language-tool.ts
+++ b/source/common/modules/markdown-editor/linters/language-tool.ts
@@ -14,9 +14,9 @@
  */
 
 import { linter, type Diagnostic, type Action } from '@codemirror/lint'
-import { extractASTNodes, extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
+import { extractTextnodes, markdownToAST } from '@common/modules/markdown-utils'
 import { configField } from '../util/configuration'
-import type { AnnotationData, Annotation, LanguageToolLinterRequest, LanguageToolLinterResponse } from '@providers/commands/language-tool'
+import type { LanguageToolLinterRequest, LanguageToolLinterResponse } from '@providers/commands/language-tool'
 import { StateEffect, StateField, type Transaction } from '@codemirror/state'
 import extractYamlFrontmatter from 'source/common/util/extract-yaml-frontmatter'
 import type { ViewUpdate } from '@codemirror/view'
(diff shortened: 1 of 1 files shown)


benniekiss:
In corralation with #5926

nathanlesage:
Thank you so much for doing this so quickly.

My two cents regarding the incremental parsing: I think nice would be to have a stateful LanguageTool service provider (thus, promoting the current command), which accepts lists of paragraphs (the Markdown parser thankfully provides paragraph nodes in the tree) and attempts to send as many of those to the server as possible. It could take into account potential rate limits on the API (which includes either keeping a tally, or being obnoxious and just sending stuff until the server responds with a 429, but probably both). For each chunk of paragraphs it receives back, it sends them to the requesting renderer, which in turn would need some state management to update the existing list of annotations instead of replacing them. Finally, this would also make incremental checking possible, where the linter checks for updates during ViewUpdates and extracts only changed paragraphs.

I have the feeling that choosing paragraphs as the main unit is a good balance between complexity and modularity. But take your time; I may start the beta soon, but we can always deliver improvements during the beta.

nathanlesage:
Ah, and lastly, this might be of interest for you: The LanguageTool server is indeed relatively smart. Here is one example:

image

Here it sees the citation and tells me that I should use singular instead of plural. The reason is that the underlying citation key (which the LanguageTool server sees) is `@Ergen2021` — so the server does see that there appears to be a name in it. Likewise, I have some multi-citekeys and then the server does not complain, so it appears to do some fancy text segmentation.

benniekiss:
The way I have it currently implemented doesn't touch the language tool command. I essentially create a statefield that accumulates changes since the last lint, and once linting is triggered, the ranges are expanded to the nearest text node on either end, diagnostics are mapped/filtered, text is checked, and once it's been processed, the accumulated changes are reset. I should have something up later today that's working. I've had it for a while, but it includes some other changes out of scope.

nathanlesage:
That sounds a bit more involved, but I'm happy to take a look!

PR #5929, opened 2025-09-29
[WIP] Implement an incremental linter for LanguageTool

benniekiss:
## Description

This PR implements an incremental linter for LanguageTool to reduce API server load and constant processing of the entire document.

It is currently a WIP to show a proof-of-concept and illustrate some of the necessary functionality for an incremental linter.

To-Do's include refactoring the language-tool provider so that it processes documents in chunks to address API ratelimits and length limits.

## Changes

Several utilities were added to help with processing changed regions, including diagnostic filtering and mapping, and context expansion.

The linter now operates over a set of `{ from, to }` regions which are computed based on an accumulated set of changed regions expanded to text node boundaries. The internal processing remains the same.

## Additional information

CC @nathanlesage 

Here's how I've been tackling an incremental linter. Since it's a WIP, feel free to go a different route. I hope this at least can give you a jumping off point!

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/linters/language-tool.ts
+++ b/source/common/modules/markdown-editor/linters/language-tool.ts
@@ -22,6 +22,7 @@ import extractYamlFrontmatter from 'source/common/util/extract-yaml-frontmatter'
 import type { ViewUpdate } from '@codemirror/view'
 import { trans } from 'source/common/i18n-renderer'
 import type { LanguageToolIgnoredRuleEntry } from '@providers/config/get-config-template'
+import { changesFieldEffectFactory, prepareDiagnostics, TEXTNODE_FILTER } from './utils'
 
 const ipcRenderer = window.ipc
 
@@ -101,6 +102,8 @@ export const languageToolState = StateField.define<LanguageToolStateField>({
   }
 })
 
+export const { effect: ltChangesEffect, field: ltChangesField } = changesFieldEffectFactory()
+
 // Hide the tooltip when the `Ignore Rule` action is selected.
 function hideOn (tr: Transaction): boolean | null {
   for (const e of tr.effects) {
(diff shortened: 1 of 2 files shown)


benniekiss:
Also, FWIW, this same approach works really well for the hunspell spellchecker, which doesn't rely so much on surrounding text context.

nathanlesage:
Looks like a smart solution, I like it. Feel free to keep at it.

Just two important considerations for going forward:

First, unit of analysis. If you only expand text nodes, you may retrieve partial sentences which LT may complain even more about than entire paragraphs. I really do believe that using paragraphs — while certainly being not perfectly efficient — are the easiest method, and I would like to stress this again. Text and natural language unfortunately do not work by rules of technical efficiency. I have been specializing in NLP over the past five years, and maximally efficient incremental linting as you are planning it here won't work — at least not if we want to make it useful to the users.

Second, context length: there are quite a lot of rules in the LT server that check broader context, such as word repetition/sentence start repetition. Those may require multiple paragraphs as input (because it is capable to check across paragraph edges), and I think this is important and very useful for academic writing.

I believe that we should try to maximize the amount of paragraphs we check per each request. Let's say that the server accepts at most 10 paragraphs, and we have a document of 22 paragraphs. Then we could first check the first ten, then use a sliding-window approach to check the middle 10, and then the last 10.

This will ensure that the LT server has broad context to check for repeated words in each paragraph. Again, this will double-check some paragraphs and yes, it is absolutely not efficient, but this is the only way we will be able to get accurate spell and grammar checking.

So I'm very happy with especially a lot of the bootstrapping you've already done around the linter, but here are some non-negotiable requirements for this PR:

* Work on entire paragraphs (or any logical unit of analysis that roughly approximates this, if it is technically easier).
* Check as many paragraphs at once as possible. This may involve auto-detecting the server capabilities (I wouldn't hard-code any numbers, and I would compartmentalize that to the main process and keep it out of the linter).
* Use a sliding-window approach to maximize the amount of context the server has access to. This is where you can try to be efficient: How can you check a very long document with as few passes as possible? Also, remember to merge all matches so that we don't lose repeated-word-matches for a paragraph that, in the second iteration has become the first paragraph the server sees.

benniekiss:
Totally agree with your thoughts on paragraph level segmentation.

I think the simplest would just be to detect block-level boundaries via newlines, but it'll probably have to be a two-step approach, block position than node position, so we don't end up in the middle of a code block, for example.

nathanlesage:
Have a look at the syntax tree the parser spits out; I'm pretty positive we won't have to deal with any new lines!

benniekiss:
Oh you're right. I thought the `Paragraph` node was just text lines, but it's the whole block. This already handles that case, then! I may have misspoke earlier. The expansion isn't to Text nodes, it's to the nearest `Paragraph` or `Blockquote` node (can be expanded to others). Currently, it expands to the first `Paragraph`/`Blockquote` node before and after the region of changed text, so if you were typing in a paragraph, it would expand to the first one before and after, regardless of how many other nodes are between them.

Otherwise, I think your other point about context length is valid, but that functionality is already configurable in the helper functions, and it can be made dynamically. Right now, it's fixed at 1 paragraph node in either direction, but this can be increased, like you said, to 10 paragraphs.

A sliding window would be great, especially for longer documents (which I work with a lot, and LT's limitations really impact my ability to use it). Would you implement that in the language-tool provider?

nathanlesage:
> A sliding window would be great, especially for longer documents (which I work with a lot, and LT's limitations really impact my ability to use it). Would you implement that in the language-tool provider?

Yes, I think we should divide the responsibilities like so:

The renderer should collect the text to be checked, and apply the matches returned from the API. It should only care about the paragraphs and applying matches. So it would essentially first send off the entire document in paragraph-blocks to main, and then subsequently smaller parts (i.e., only changed paragraphs plus some context, that we would have to arbitrarily decide upon. I'd argue 3 paras before and 3 after sounds like a good starting point that we can adjust where necessary).

The main process will receive various documents from potentially multiple editors, and will essentially just need to make sure to provide checks for the linters. That means it only needs to somehow determine how many characters the user can send, and the server will accept, and split up ONE request by a linter into potentially multiple requests to the server if the user uses LT free official, and once all requests are in, respond to the linter process.

There's probably a lot of small decisions to make apart from that …

nathanlesage:
Ah, and I just realized that this will probably involve a few architectural changes: First, for the renderer each request should only yield one single response. However, since that will be asynchronous to CodeMirror asking the linter to re-run, we probably will have to de-couple the linting process from the requests (i.e., when CodeMirror requests linting, schedule one, but immediately return the existing, cached diagnostics to CodeMirror for display, and update them behind the scenes as soon as the main process is done pulling the results from the server, etc…)

benniekiss:
Closing this, but I have the work continued in a separate branch

PR #5935, opened 2025-10-06
[Fix] Handle snippet tabstop removal

benniekiss:
## Description

This PR fixes two bugs related to tabstop removal from snippets 

## Changes

The `Escape` shortcut was fixed so that it now correctly removes all remaining tabstops. There was ony a small typo.

Tabstops are now removed if the context before them was deleted by mapping positions using `MapMode.TrackBefore`. This enables a user to manually select and remove tabstops.

## Additional information

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/autocomplete/snippets.ts
+++ b/source/common/modules/markdown-editor/autocomplete/snippets.ts
@@ -25,7 +25,8 @@ import {
   EditorSelection,
   Facet,
   type SelectionRange,
-  type EditorState
+  type EditorState,
+  MapMode,
 } from '@codemirror/state'
 import { Decoration, EditorView, WidgetType } from '@codemirror/view'
 import { type AutocompletePlugin } from '.'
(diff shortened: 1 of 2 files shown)


source/common/modules/markdown-editor/autocomplete/snippets.ts
```diff
@@ -161,9 +162,16 @@ export const snippetsUpdateField = StateField.define<SnippetStateField>({
     }
 
     // This monstrosity ensures that our ranges stay in sync while the user types
-    val.activeSelections = val.activeSelections.map(selection => {
-      return selection.map(transaction.changes)
-    })
+    val.activeSelections = val.activeSelections
+      .filter(selection => selection.ranges.some(r => {
+        if (transaction.changes.mapPos(r.from, -1, MapMode.TrackBefore) === null) {
```
nathanlesage: Shouldn't this be [`TrackDel`](https://codemirror.net/docs/ref/#state.MapMode.TrackDel)?

nathanlesage: Also, I think you can simplify the if/else construct to a direct return

nathanlesage (commented):
Looks good. The MapReduce loop can be made a bit more compact I believe, and there is a question, but other than that it's good to go!

source/common/modules/markdown-editor/autocomplete/snippets.ts
benniekiss: `TrackDel` requires the characters on both sides to be deleted, which can cause problems when a tabstop is at the end of a snippet context, and it's slightly less intuitive. Using `TrackBefore`, deleting the tabstop works like deleting a regular character. 

You're right about the direct return, I'll clean that right up!

nathanlesage:
Thank you! Merging!

PR #5936, opened 2025-10-06
Explicitly require `codemirror/lint` package

benniekiss:
## Description

This PR adds the `codemirror/lint` package to `package.json`.

## Changes

Zettlr depends on `codemirror/lint`, but it was absent from the `package.json`. I believe it was otherwise being installed through the `codemirror/lang-javascript` package. 

## Additional information

My main motivation for including it is that with `codemirror/lint v6.9.0`, individual actions can be styled throught the `markClass` attribute.

Tested on:

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@codemirror/lang-yaml": "^6.1.2",
     "@codemirror/language": "^6.11.3",
     "@codemirror/legacy-modes": "^6.5.1",
+    "@codemirror/lint": "^6.9.0",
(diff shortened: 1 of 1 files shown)


nathanlesage:
Whoops, I normally don't do that, and all packages that are imported somewhere in the code base should be explicitly required. Thanks for spotting this. Just one request: please delete the lock file and re-install everything, to ensure that no funky lock file conflicts happen (CodeMirror is unfortunately prone to this).

nathanlesage:
Thank you!

PR #5939, opened 2025-10-07
[Fix] Handle nested placeholders in snippets

benniekiss:
## Description

This PR fixes an issue with nested snippet placeholders, i.e. `Here is a ${1:nested ${2:placeholder}}`. Nested placeholders now generate correct selection ranges.

## Changes

The placeholder parsing logic was moved to a separate function. Now, placeholders with default text are parsed recursively to handle any nested placeholders.

Tests were added to test the `template2snippet()` function.

## Additional information

Currently, nested placeholders cause errors in the selection ranges due to the internal braces contributing to the external range, however, the internal braces are later removed by the parsing, so the external ranges become unsynced.

Here is a sample snippet to test:

`Here is a ${1:sample snippet with ${2:nested placeholders} in $3 various spots.}, it includes
more, ${4: and more default ${5:nested}} placeholders.`

Tested on:  macOS 26

### DIFF
--- a/test/setup.js
+++ b/test/setup.js
@@ -33,6 +33,25 @@ function mockBrowser () {
   // we copy it here in order for those tests not to fail.
   window.path = path
 
+  // Mock the window.config get() and set()
+  window.config = {
+    _data: new Map(),
+    get(key) {
+      if (key === undefined) return this._data
+      return this._data.get(key)
+    },
+    set(key, value) {
+      this._data.set(key, value)
+    }
+  }
+
+  // Mock the navigator.clipboard readText() and writeText()
+  navigator.clipboard = {
+    _data: '',
+    readText() { return this._data },
+    writeText(text) { this._data = text }
+  }
+
   global.window = window
   global.document = window.document
   global.requestAnimationFrame = function (callback) {


source/common/modules/markdown-editor/autocomplete/snippets.ts
```diff
@@ -290,7 +342,7 @@ async function replaceSnippetVariables (state: EditorState, text: string): Promi
   const hour = now.hour
   const minute = now.minute
   const second = now.second
-  const clipboard = await navigator.clipboard.readText()
+  const clipboard = await navigator.clipboard?.readText()
```
nathanlesage:
```suggestion
  const clipboard = await navigator.clipboard.readText()
```

source/common/modules/markdown-editor/autocomplete/snippets.ts
```diff
@@ -308,7 +360,7 @@ async function replaceSnippetVariables (state: EditorState, text: string): Promi
     CURRENT_SECONDS_UNIX: now.toSeconds(),
     UUID: uuid(),
     CLIPBOARD: (clipboard !== '') ? clipboard : undefined,
-    ZKN_ID: generateId(String(window.config.get('zkn.idGen'))),
+    ZKN_ID: generateId(String(window.config?.get('zkn.idGen'))),
```
nathanlesage:
```suggestion
    ZKN_ID: generateId(String(window.config.get('zkn.idGen'))),
```

test/snippets.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+          ranges: [{
+            anchor: 17,
+            head: 17
+          }]
+        },
+      ]
+    }
+  },
+  {
+    template: 'Test ${1:nested ${2:placeholders} for ${3:recursive ${4:parsing}}}',
```
nathanlesage, line 163: Would you mind also adding one test each for simple tabstops (`$4`) and default variable replacement (`${4:$VARIABLE}`)?

source/common/modules/markdown-editor/autocomplete/snippets.ts
benniekiss: So I had to add this check because the `config` and 'clipboard` are not available in the test environmemt.

test/snippets.spec.ts
benniekiss: Add those variables to this specific test, or separate tests? There's already one for simple tabstops at the top.

nathanlesage: Separate tests; essentially I want to make sure that nested placeholders with no-default and/or variable replacements also work in the new implementation!

source/common/modules/markdown-editor/autocomplete/snippets.ts
nathanlesage: Ah… this reminds me that we have to somehow get the JSDOM shim for the test environment running. That setup hasn't worked in years, because I never found the time to get to it. Essentially, it would be nice to be able to add both browser and node environments (separately/per test) to the mocha suite, but I haven't been able to get it to work…

benniekiss:
Added more tests for variables, nested placeholders with no-default tabstops, and escaped placeholders.

I also changed the mapping association which provides a better UX, and I added a `side` parameter to the `SnippetWidget` decoration so that stacked tabstops are ordered correctly.

nathanlesage:
Quick note: I have now presumably fixed the JSDOM document harness for the mocha tests. So the navigator clipboard should work (or we can add something in the `setup.js` file to make it work). When it comes to window.config, I think we might need an additional setup that mocks that module for the test suite. Then, no `?.` should be necessary anymore.

benniekiss:
~~So the tests still do not work, same issue. I did remove the check on the properties, so the tests will fail for the time being. Is the clipboard~~ I'm so sorry, I totally misread your comment, you already said this! 

Yes, `setup.js` needs something included to mock the clipboard, since it is still unavailable -- I noticed in the documentation it's only available in secure contexts, so it might just be disabled in the test environment.

benniekiss:
The latest commit mocks a minimal amount of the `window.config` and `navigator.clipboard` to make the tests function. They are now passing

nathanlesage:
Great! As long as the mock functions have the same function signature than the real thing that's all we need; after all they're tests :) Thank you for this work!

benniekiss:
You're welcome! This is good to merge on my end once you're ready.

nathanlesage:
Sooo, let's merge this quickly so that the amount of your open PRs gets a bit reigned in again haha

PR #5941, opened 2025-10-08
[CSS] Style `Disable Rule` LanguageTool linter action

benniekiss:
## Description

This PR adds styling for the LT `Disable Rule` diagnostic action so that it can be differentiated from other suggestion actions.

## Changes

Since `codemirror/lint v6.9.0`, actions can be assigned a CSS class. This PR assigns a CSS class to the suggestion actions and the disable rule action, `cm-ltSuggestAction` and `cm-ltDisableAction`, respectively. Additional styling was added to the disable rule action so that it stands apart from the other suggestions.

## Additional information

Screenshot 2025-10-08 at 10 46 05

Screenshot 2025-10-08 at 10 46 16

Tested on:  macOS 26


nathanlesage:
In light mode, the color looks okay (maybe a bit bright), but in the dark mode I can already see that seeing this red button might be a bit too much. Would it be possible to reduce the vibrancy of the color a bit? I think it's fine if it is faintly red, its only purpose is to prevent accidental clicks on it, and it's not a destructive action :)

benniekiss:
Yeah, definitely. I'll play around with some colors/saturation

benniekiss:
Here are a couple samples, I'm open to other suggestions:

**#dd1111ff** 
*matches tooltip red color*

Screenshot 2025-10-09 at 09 32 07
Screenshot 2025-10-09 at 09 32 18

**#FF6347**
*Tomato*

Screenshot 2025-10-09 at 09 29 34
Screenshot 2025-10-09 at 09 29 44

nathanlesage:
How about working from the gray and adding some red? I really think that the button shouldn't deviate from the UI of the other buttons too much. Again, it's not an irreversible action, it should just reduce the probability of clicking on the wrong button. I feel the top one is too aggressive, and the bottom one is a bit too yellow ... take a look at the token list in the attachment settings (advanced section)

benniekiss:
That one looks great to me!

Screenshot 2025-10-09 at 16 41 49
Screenshot 2025-10-09 at 16 42 02

nathanlesage:
Ahh yes! Red and identifiable, but not too strong and matches the other buttons :)

benniekiss:
Thanks for the feedback! Im terrible with color, so I appreciate it. Ill push the update in just a bit

nathanlesage:
Perfect, merging! :)

PR #5946, opened 2025-10-15
[Enhancement] Toggle export auto open from popover

### LINKED ISSUE #5930: Add a toggle for the `Automatically open successfully exported files` setting to the toolbar's export panel
### Description

First off, thank you so much for the new ability to check/uncheck `Automatically open successfully exported files` in `Preferences > Import and Export`. This has already been a major quality of life improvement while using Zettlr on Linux, as the editor would previously spawn a new PDF application window on every export, which could mean dozens of extraneous windows needing to be closed each time I was tweaking the look of an exported document.

The only additional enhancement I would ask for in this regard is to perhaps add a toggle switch to the setting to the toolbar's export panel. E.g.,

Image

This is because, while the new feature eliminates some great annoyances, there are still times when it's helpful to open the exported file automatically. For example, the _first_ time I export something, I'd like the exported document to open in the external viewer. It's on subsequent exports that the auto-export becomes troublesome, as it's spawning duplicate copies of the document I already have open in my PDF viewer. Being able to toggle the functionality right from the export panel would make it that much simpler to turn on open-externally when I first export a file and then flip it back to off on subsequent exports. Toggling the feature from the main settings menu is welcome functionality, but because the decision to launch or not launch an external viewer is often contextually specific to the immediate document I'm working on, and whether I already have it open in an external viewer, it can feel a bit weird to toggle a global setting.

### Proposed Changes

See above.

### Caveats

None.

No

### Zettlr Version

Nightly

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Manjaro (Rolling)

### Additional Information

Thank you!!!

### LINKED ISSUE #6002: Remember last used export profile on a per-file basis
### Description

I am currently writing tons of applications, but I still continue working on some papers. Zettlr currently only remembers the last-used export profile globally, but not on a per-file basis. However, some files require one, others different profiles. Having Zettlr remember the last-used export profile in a simple map of file path -> profile name would make exporting easier.

### Proposed Changes

Remember the last used export profile on a per-file basis so that Zettlr automagically selects the correct one for each individual file.

### Caveats

* When a user moves the file, this would invalidate the path. But worst case is that Zettlr exports with the wrong profile, and you'll have to do it a second time.
* Same when profiles are renamed or moved. Same worst case in this instance.

Yes

### Zettlr Version

Right off the develop branch

### Your Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Operating System Version

any

### Additional Information

I'm happy to implement this, but it will take a few days. If someone of you is interested, feel free to open a PR before that. I'd recommend doing a map in the configuration file (probably?) and doing a prune on every boot of no-longer existing file paths and profiles.

benniekiss:
## Changes

This PR adds a checkbox toggle to the Export popover which allows a user to control whether the exported document is automatically opened. It also saves the last selected export profile per file.

Closes #5930
Closes #6002

## Additional information

A checkbox was added to the Export popover bound to the `export.autoOpenExportedFiles` config value.

The label for the `Format` element was moved to a translation.

Tested on:  macOS 26

### DIFF
--- a/source/win-main/PopoverExport.vue
+++ b/source/win-main/PopoverExport.vue
@@ -66,16 +66,23 @@ const tempDirLabel = trans('Temporary directory')
 const cwdLabel = trans('Current directory')
 const askLabel = trans('Select directory')
 
+// This is used to limit the number of selected
+// profile to filename mappings in the config
+const PREVIOUSLY_SELECTED_PROFILE_LIMIT = 50
+
 ipcRenderer.invoke('assets-provider', { command: 'list-export-profiles' } as AssetsProviderIPCAPI)
   .then((defaults: PandocProfileMetadata[]) => {
     // Save all the exporter information into the array. The computed
     // properties will take the info from that array and re-compute based
     // on the value of "format".
     profileMetadata.value = defaults
-    // Get either the last used exporter OR the first element available
-    const lastProfile = selectedProfiles.value.get(props.filePath)
-    if (lastProfile !== undefined && lastProfile in availableFormats.value) {
-      format.value = lastProfile
+    // Get either the last selected exporter for the open file,
+    // the last used exporter, or the first element available
+    const lastProfile = selectedProfiles.value.find(item => item.filePath === props.filePath)
+    const profile = lastProfile ? lastProfile.profile : lastUsedProfile.value
+
+    if (profile in availableFormats.value) {
+      format.value = profile
     } else {
       format.value = profileMetadata.value[0].name
     }
@@ -96,9 +103,10 @@ const format = ref('')
 const exportDirectory = ref(configStore.config.export.dir)
 const autoOpenExport = ref(configStore.config.export.autoOpenExportedFiles)
 const profileMetadata = ref<PandocProfileMetadata[]>([])
+
 const customCommands = computed(() => configStore.config.export.customCommands)
-const openPaths = computed(() => configStore.config.openPaths)
-const selectedProfiles = computed(() => new Map(Object.entries(configStore.config.export.selectedProfiles)))
+const selectedProfiles = computed(() => configStore.config.export.selectedProfiles)
+const lastUsedProfile = computed(() => configStore.config.export.lastUsedProfile)
 
 const exportButtonLabel = computed(() => isExporting.value ? trans('Exporting…') : trans('Export'))
 const filename = computed(() => pathBasename(props.filePath))
@@ -137,22 +145,19 @@ watch(format, function (value) {
   const prof = profileMetadata.value.find(e => e.name === value)
   const cmd = customCommands.value.find(x => x.command === value)
 
+  const profile: string  = prof?.name ?? cmd?.command ?? lastUsedProfile.value
+  const filePath: string = props.filePath
+
   const newProfiles = selectedProfiles.value
-  newProfiles.set(props.filePath, prof?.name ?? cmd?.command ?? '')
+    // Remove any previous items with the same path
+    .filter(item  => item.filePath !== filePath)
+    // Clamp the list to the last N - 1 items since we will be pushing one
+    .slice(-PREVIOUSLY_SELECTED_PROFILE_LIMIT - 1)
 
-  configStore.setConfigValue('export.selectedProfiles', Object.fromEntries(newProfiles))
-})
+  newProfiles.push({ filePath, profile })
 
-watch(openPaths, function (value) {
-  // Prune any path no longer part of the open paths.
-  const newProfiles = selectedProfiles.value
-  for (const filename of newProfiles.keys()) {
-    if (!value.some(path => filename.startsWith(path))) {
-      newProfiles.delete(filename)
-    }
-  }
-
-  configStore.setConfigValue('export.selectedProfiles', Object.fromEntries(newProfiles))
+  configStore.setConfigValue('export.selectedProfiles', JSON.parse(JSON.stringify(newProfiles)))
+  configStore.setConfigValue('export.lastUsedProfile', profile)
 })
 
 function doExport (): void {


benniekiss:
Latest commit preserves the last selected profile per file by storing it in the config file. So far, this seems fine, especially since the `openPaths` are already stored in the file as well.

source/app/service-providers/config/get-config-template.ts
```diff
@@ -100,9 +100,9 @@ export interface ConfigOptions {
     cslLibrary: string
     cslStyle: string
     useBundledPandoc: boolean
-    singleFileLastExporter: string
```
nathanlesage, line 103: Please keep this setting (I'll explain in a comment below)

source/app/service-providers/config/get-config-template.ts
```diff
@@ -100,9 +100,9 @@ export interface ConfigOptions {
...
     exportQmdWithQuarto: boolean
     customCommands: Array<{ displayName: string, command: string }>
+    selectedProfiles: Map<string, string>
```
nathanlesage: I'd rather use a 2d-array of type `Array<[string, string]>` here. Prevents any oddities with the built-in map type. Also allows us to do something clever.

source/win-main/PopoverExport.vue
```diff
@@ -5,7 +5,7 @@
       <p><strong>{{ filename }}</strong></p>
       <SelectControl
         v-model="format"
-        v-bind:label="'Format'"
+        v-bind:label="formatLabel"
```
nathanlesage, line 8: Ooh, good find, thank you!

source/win-main/PopoverExport.vue
```diff
@@ -65,8 +73,8 @@ ipcRenderer.invoke('assets-provider', { command: 'list-export-profiles' } as Ass
     // on the value of "format".
     profileMetadata.value = defaults
     // Get either the last used exporter OR the first element available
-    const lastProfile = configStore.config.export.singleFileLastExporter
-    if (lastProfile in availableFormats.value) {
+    const lastProfile = selectedProfiles.value.get(props.filePath)
```
nathanlesage: Here, I want you to fetch the lastProfile using the selectedProfiles props.filePath, BUT if it is not found in the recently selected profiles, fall back to the global setting.

source/win-main/PopoverExport.vue
```diff
@@ -86,8 +94,11 @@ const emit = defineEmits<(e: 'close') => void>()
 const isExporting = ref(false)
 const format = ref('')
 const exportDirectory = ref(configStore.config.export.dir)
+const autoOpenExport = ref(configStore.config.export.autoOpenExportedFiles)
 const profileMetadata = ref<PandocProfileMetadata[]>([])
 const customCommands = computed(() => configStore.config.export.customCommands)
+const openPaths = computed(() => configStore.config.openPaths)
+const selectedProfiles = computed(() => new Map(Object.entries(configStore.config.export.selectedProfiles)))
```
nathanlesage: … with the entire thing being an array would make that a bit handier, too

source/win-main/PopoverExport.vue
```diff
@@ -119,7 +136,23 @@ watch(format, function (value) {
   // Remember the last choice
   const prof = profileMetadata.value.find(e => e.name === value)
   const cmd = customCommands.value.find(x => x.command === value)
-  configStore.setConfigValue('export.singleFileLastExporter', prof?.name ?? cmd?.command ?? '')
```
nathanlesage, line 122: Keep this line as the fallback, so that we update it accordingly

source/win-main/PopoverExport.vue
```diff
@@ -119,7 +136,23 @@ watch(format, function (value) {
...
+
+  const newProfiles = selectedProfiles.value
+  newProfiles.set(props.filePath, prof?.name ?? cmd?.command ?? '')
```
nathanlesage: Here, I'd like you to do the following: Search for the filePath in the config array. Is it in there? Then remove it. Then, add the filepath to the end or start of the array, and afterwards truncate it to … say … 50 entries. This has a few benefits.

1. It definitely keeps the list manageable, and will satisfy my anxiety with users who export a bunch of stuff.
2. It will at some point automatically remove no-longer-existing files, and thus keep the list clean
3. If a file drops out of that list, you're probably not going to re-export it all the time, and thus are happy to re-select a more appropriate profile
4. I intended this mostly as a convenience not always having to switch between profiles. The use case is that you have a small set of files you are constantly working on which may require different profiles, and thus a short list of "most recent files" is sufficient here.

source/win-main/PopoverExport.vue
```diff
@@ -119,7 +136,23 @@ watch(format, function (value) {
...
+
+  configStore.setConfigValue('export.selectedProfiles', Object.fromEntries(newProfiles))
```
nathanlesage: … also it prevents this here :P

source/app/service-providers/config/get-config-template.ts
benniekiss: My initial thought was to use an array, so I'm glad you pointed it out! Would you be fine with `Array<{ filepath: string, profile: string }>` instead of sub arrays? I think it makes it clearer what each position means and would serialize into the config file nicely.

source/app/service-providers/config/get-config-template.ts
```diff
@@ -100,9 +100,10 @@ export interface ConfigOptions {
     cslLibrary: string
     cslStyle: string
     useBundledPandoc: boolean
-    singleFileLastExporter: string
     exportQmdWithQuarto: boolean
     customCommands: Array<{ displayName: string, command: string }>
+    selectedProfiles: Array<{ filePath: string, profile: string }>
+    lastUsedProfile: string
```
nathanlesage, line 106: One thing I just realized … given you have renamed the setting anyhow, would you mind moving both new settings into the `export` key? I want to slowly but steadily get rid of top-level config keys if they can be grouped somehow sensibly!

benniekiss:
Nothing else in mind. This is good on my end!

nathanlesage:
There's this one unresolved thing, just name spacing the settings a bit more, then it's good from my side too!

benniekiss:
@nathanlesage Im not sure if you saw my other comment, but I think those keys are already namespaced under export, unless I'm looking at the wrong thing.

nathanlesage:
… this goes to show how detrimental being overworked can be 👀  Apologies!

PR #5947, opened 2025-10-15
[Fix] Handle quoted property values when parsing pandoc attributes

benniekiss:
## Description

This PR refactors the pandoc attribute parser so that it handles quoted property values, such as `style="font-family: 'Serif';"`

## Changes

The parser was refactored to use regex capture groups rather than splitting on white space and parsing the components. Currently, by splitting on whitespace, the parser incorrectly parses properties with quoted values.

The `parseAttributeNode` function in the markdown-ast `parse-children` module was refactored to re-use the pandoc attribute parser rather than re-implement it in the module.

The tests were updated to include quoted property values and to test a few more edge cases.

## Additional information

Tested on:  macOS 26


benniekiss:
Since I've been working on these refactors, the last commit now moves the attributes parser to use delimiters

nathanlesage:
Thank you, merging now!

PR #5952, opened 2025-10-16
[Performance] Debounce word counting

benniekiss:
## Description

This PR implements a debounced word counting plugin to improve typing performance.

## Changes

The word count computation was moved to a `ViewPlugin` so that it can be debounced using `setTimeout()`. It uses a `StateEffect` to update the existing `countField` value.

Event listeners were added for the new `updateWordCountEffect` so that the UI shows updated counts when they arrive rather than waiting for cursor activity.

## Additional information

Currently, word counts are recomputed every keypress. In larger documents (>50,000 words) this starts to add a small but noticeable, and increasing, delay. By debouncing the counting, typing performance is dramatically improved in large documents. 

Word counts are no longer live, however, they are updated the moment the transaction is dispatched.

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/index.ts
+++ b/source/common/modules/markdown-editor/index.ts
@@ -456,6 +456,8 @@ export default class MarkdownEditor extends EventEmitter {
     }
 
     this._instance.focus()
+
+    this.emit('loaded')
   }
 
   /**
--- a/source/win-main/MainEditor.vue
+++ b/source/win-main/MainEditor.vue
@@ -441,6 +441,13 @@ async function getEditorFor (doc: string): Promise<MarkdownEditor> {
   const editor = new MarkdownEditor(props.leafId, props.windowId, doc, documentAuthorityIPCAPI, undefined, persistentState)
 
   // Update the document info on corresponding events
+  editor.on('loaded', () => {
+    if (currentEditor === editor) {
+      windowStateStore.activeDocumentInfo = currentEditor.documentInfo
+      windowStateStore.tableOfContents = currentEditor.tableOfContents
+    }
+  })
+
   editor.on('change', () => {
     if (currentEditor === editor) {
       windowStateStore.tableOfContents = currentEditor.tableOfContents
@@ -504,9 +511,6 @@ async function loadDocument (): Promise<void> {
 
   mainEditorWrapper.value?.appendChild(newEditor.dom)
   currentEditor = newEditor
-
-  windowStateStore.tableOfContents = currentEditor.tableOfContents
-  windowStateStore.activeDocumentInfo = currentEditor.documentInfo
 
   currentEditor.setCompletionDatabase('tags', tags.value)
   currentEditor.setCompletionDatabase('snippets', snippets.value)


nathanlesage:
You guys are all writing too large documents.

(/s)

Will have a look later!

benniekiss:
Trust me, I wish I wasnt!

source/common/modules/markdown-editor/index.ts
```diff
@@ -299,20 +299,28 @@ export default class MarkdownEditor extends EventEmitter {
...
           this.emit('change')
         }
+
         if (update.focusChanged && this._instance.hasFocus) {
           this.emit('focus')
         }
+
         if (update.selectionSet) {
           this.emit('cursorActivity')
+          this.emit('docUpdate')
```
nathanlesage, line 309: Doesn't this also now need to be emitted from the update branch a few lines above?

source/common/modules/markdown-editor/index.ts
```diff
@@ -299,20 +299,28 @@ export default class MarkdownEditor extends EventEmitter {
...
-        // Listen for config updates, and parse them into the internal cache. We
-        // do it this way, because the editor itself is also capable of changing
-        // its configuration (e.g., via the statusbar). This way we ensure that
-        // both external updates (via setOptions) as well as internal updates
-        // both end up in our cache.
         for (const transaction of update.transactions) {
           for (const effect of transaction.effects) {
+            // Listen for word count updates
+            if (effect.is(updateWordCountEffect)) {
+              this.emit('docUpdate')
```
nathanlesage, line 316: Mh, I'm wondering if `docUpdate` semantically correctly describes when and how this event is being emitted ...?

source/common/modules/markdown-editor/plugins/statistics-fields.ts
```diff
@@ -14,31 +14,70 @@
...
+    return count(state)
   },
 
   update (value, transaction) {
-    // If someone provided the markClean effect, we'll exchange the saved doc
-    // so that, when comparing documents with cleanDoc.eq(state.doc), it will
-    // return true.
-    if (!transaction.docChanged) {
-      return value
+    for (let e of transaction.effects) {
```
nathanlesage:
```suggestion
    for (const e of transaction.effects) {
```

source/common/modules/markdown-editor/plugins/statistics-fields.ts
```diff
@@ -14,31 +14,70 @@
...
   },
 
   compare (a, b): boolean {
     return a.chars === b.chars && a.words === b.words
   }
 })
+
+export const countPlugin = ViewPlugin.fromClass(class {
+  private timeout: number | null = null
+  private delay = 750
```
nathanlesage: Please move this as a constant to the top of the file so that we can adjust this if necessary (e.g., something like `const WORD_COUNT_DELAY = 750`)

source/common/modules/markdown-editor/index.ts
benniekiss: I think `update` is the right implication since its emitting to a change in metadata about the document rather than its contents. I agree that `docUpdate` may not be the clearest naming, though. I went with `docUpdate` since this event is used only to sync the `windowStateStore.activeDocumentInfo` with the `currentEditor.documentInfo`. Maybe a better name would be `docInfoUpdate`?

source/common/modules/markdown-editor/index.ts
benniekiss: I originally had this also emitting for `update.docChanged` as well, but it caused to many performance issues, and I don't think it's necessary in this case, either way. The reason it was added here is because the event is used to update `windowStateStore.activeDocumentInfo`, which includes information about selection ranges.

benniekiss:
Latest commit should close #5996. I added an emit on the editor's `loadDocument` function so that the `MainEditor` process can update the word-count/TOC once the document is loaded. For some reason, the `MainEditor.loadDocument` function was no updating the info, even though it was attempting to set it.

PR #5953, opened 2025-10-16
[Enhancement] Refactor Footnote handling

benniekiss:
## Description

This PR mainly refactors the footnote cleanup handling so that it is debounced.

## Changes

The footnote cleanup handling was moved to a `ViewPlugin` so that it can be debounced after some delay in typing. This improves user typing performance. 

The logic was also refactored to address some bugs related to footnote removal. A notable change is in how footnotes are sorted -- previously, they were concated into one long string and inserted into the region contained by the first and last footnote. However, this causes issues when text occurs between footnotes, where the in-between text would be removed. ~~The new handling tries to avoid this kind of destructive action, but this could be further tested.~~ All footnote references are now moved to the end of the document.

~~The citation handler was updated so that it uses a `Set` when searching for valid locator labels. This is a much faster approach than using the `Array.find()`.~~

~~The footnote parser regex was updated to use character exclusion rather than non-greedy matching, which is much slower in general.~~

## Additional information

I don't use footnotes in my day-to-day, so my testing is quite limited to the dev environment files. I would appreciate others testing this to catch any other bugs!

In addition, I think it would be good to add a toggle for this behavior in the settings. I could see usecases where people do not want Zettlr automatically updated references.

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -133,6 +133,7 @@ export interface ConfigOptions {
     boldFormatting: '**'|'__'
     italicFormatting: '_'|'*'
     readabilityAlgorithm: 'dale-chall'|'gunning-fog'|'coleman-liau'|'automated-readability'
+    cleanupFootnotes: boolean
     lint: {
       markdown: boolean
       languageTool: {
@@ -334,6 +335,7 @@ export function getConfigTemplate (): ConfigOptions {
       readabilityAlgorithm: 'dale-chall', // The algorithm to use with readability mode.
       showStatusbar: true,
       showFormattingToolbar: true,
+      cleanupFootnotes: false,
       lint: {
         markdown: true, // Should Markdown be linted?
         languageTool: {
--- a/source/common/modules/markdown-editor/commands/footnotes.ts
+++ b/source/common/modules/markdown-editor/commands/footnotes.ts
@@ -17,31 +17,31 @@
 // 2. Cmd-Click them to edit them in place
 // 3. And, obviously, add and remove them
 
-import { syntaxTree } from '@codemirror/language'
-import { type ChangeSpec } from '@codemirror/state'
-import { type EditorView } from '@codemirror/view'
+import { ensureSyntaxTree, syntaxTree } from '@codemirror/language'
+import { ChangeSet, type ChangeSpec, type StateCommand } from '@codemirror/state'
 import { extractASTNodes, markdownToAST } from '@common/modules/markdown-utils'
 import { type Footnote, type FootnoteRef } from '@common/modules/markdown-utils/markdown-ast'
 
 /**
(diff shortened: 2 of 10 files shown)


benniekiss:
Another point that could be added -- but should footnotes be moved to the end of the document? That could simplify the sorting behavior a lot and ensure they are all kept in the same place.

nathanlesage:
Yes, I'd argue footnotes should always be moved to the end. That's where Zettlr has always placed any and all inserted footnotes, and it is recommended to do so. I'd argue we can implement this behavior, because nobody has yet complained about it, and it's the most natural.

nathanlesage:
(Also, I won't be looking at the source in the coming 4 days until after my defense, just as a heads up!)

benniekiss:
Latest commit moves all references to the end of the document, with numeric references before non-numeric.

Also, good luck on your defense! Wishing you well!

nathanlesage:
I'm currently going by train and want to give this PR the attention it deserves, so I will review it next week with a clear head. (I may implement some things in the code base in the meantime.)

Just a few thoughts as to my severely misguided ad-hoc fix: My main frustration was that footnotes form a single semantic unit, but are treated by Markdown as two separate syntactic elements. Adding no functionality in this regard would be bad, but overridden every transaction has its own drawbacks, as you have correctly observed.

Here are some unsorted thoughts that I've had over the past week:

* We should not alter transactions too much. Sometimes this is probably deserved, but not always; especially if information gets added or deleted. This way, we simply add another way for data loss (as you have experienced -- I'm sorry!)
* Users should have the freedom to choose which altercations they accept (which you implemented here already with the setting)
* We have to think about footnote workflows much more deeply, and integrate it better. For example, what my commit did and what your PR fixes is simply the cleaning up of footnotes. But what I completely missed and what might be even more important is a proper addition of the footnotes to the copy, cut, and paste actions (add footnote references during copy, place footnotes from clipboard content at the correct position, and so on).
* With this being said, I believe that your PR will already do some good improvements in this regard, but it is probably just one part of a more wholistic overhaul of the workflow.
* My suggestion would be to come up with a proper design for footnote handling in the editor, and only then start implementing/merging PRs. This would allow us to possibly tweak this PR according to the workflow.

Looping in @wideaperture because I know you're also a seasoned footnote-enjoyer ;)

benniekiss:
I think this PR should definitely cover most of your points.

Re: a more integrated workflow, I think my refactor to use statecommands is the right pattern, especially for cut/copy/paste, so those can maybe be added as additional commands, either with this PR or through a future one. However, anything that requires linking the note to the ref is going to require a full document analysis, so we should be careful about the performance implications.

Like 50% of this diff is just new tests, so don't be too put off!

benniekiss:
I messed up the merge conflicts! I'll re-open this once its cleaned up.

benniekiss:
I dropped the commits unrelated to footnote handling just so that this can stay focused for now.

PR #5955, opened 2025-10-18
Explicitly require @lezer/markdown

benniekiss:
## Description

This PR adds the `@lezer/markdown` dep to `package.json`. The package is used in several parsers, so it should be explicitly required. It was otherwise being included from `@codemirror/lang-markdown`

## Additional information

Currently working on a parser feature, which requires the latest release of `@lezer/markdown`, and I noticed it was not included in the `package.json`.

Tested on:


benniekiss:
~~superseded by #5964~~ wrong one

PR #5965, opened 2025-10-25
[Enhancement] Style blockquotes with a vertical bar

### LINKED ISSUE #5960: Disable italic rendering in quotation environnement of Berlin, Frankfurt & Karl-Marx-Stadt
### Description

**Berlin, Frankfurt and Karl-Marx-Stadt appearances** render quotation environment in italics. This is useful to discriminate it quickly, but when you are to *emphasis* certain parts of the quoted text, it is not visible in the Preview mode since  the whole will be rendered as italics. So for instance, the following markdown : 

`> This is a citation and *this part* is important`

… will be rendered as italics, with no discrimination of the emphasized part. Like : 

> *This is a citation and this part is important.*

… instead of (normal and correct behavior, like in Bielefeld & Bordeaux appearances) : 

> This is a citation and *this part* is important

I understand that, here, italics helps the user to discriminate quickly the environment, but the drawback is that you can’ discriminate emphasis inside of it.

### Proposed Changes

It seems that Berlin, Frankfurt and Karl-Marx-Stadt should be modified so that quotation environment are displayed in a normal shape, and not italics. It would be more transparent toward raw markdown in the background (but see caveats).

### Caveats

Italics actually function to help discrimination of environment, in an quite elegant fashion. If you look at Bielefeld and Bordeaux, you will note that quotation environment are in normal shape, not that indented, and so are more difficult to detect. In Preview mode at least, it is so close to normal text.

A choice has to be made for **all appearances** about how to display and help to discriminate a quotation environment. It seems to me that the traditional solution of a vertical line in the margin is the way to go. This line could be in the highlight color of the theme (e.g., the green of the Berlin theme). Is is simple, very clear, not ambiguous.

No

### Zettlr Version

None

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Debian 13

### Additional Information

Enhancement suggested by [this discussion](https://github.com/Zettlr/Zettlr/issues/5956#issuecomment-3427704120).

benniekiss:
## Description

This PR styles blockquotes with a vertical bar to indicate quote level across all themes.

Closes #5960 

## Changes

The `SpaceWidget` and `BufferWidget` were refactored to accept a `classList` parameter, which is used to add optional classes to the widget element.

The `hideFormattingCharacters` function was refactored so that `QuoteMark` decorations are assigned a `cm-quotemark` class.

Styling was added to the  `cm-quotemark` and `cm-quote` classes through an editor theme. The `cm-quotemark` styling adds a vertical bar to quotemarks to more clearly indicate quote level, and the `cm-quote` styling reduces the opacity to make blockquotes stand out from regular text.

Each theme was updated to remove any specifc blockquote styling, and the quotemarks were styled according to the theme's `primaryColor`.

Finally, the `rangeInSelection` is passed `true` for taking into account adjacent ranges. This makes it so that formatting characters appear if the cursor is next to one, even when it is on the outside of the node. While testing, I felt that this improved the UI for showing formatting characters.

## Additional information

Themes:

Screenshot 2025-10-25 at 11 27 21
Screenshot 2025-10-25 at 11 27 47

Screenshot 2025-10-25 at 11 27 26
Screenshot 2025-10-25 at 11 27 50

Screenshot 2025-10-25 at 11 27 33
Screenshot 2025-10-25 at 11 27 54

Screenshot 2025-10-25 at 11 27 37
Screenshot 2025-10-25 at 11 27 57

Screenshot 2025-10-25 at 11 27 41
Screenshot 2025-10-25 at 11 28 01

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-emphasis.ts
+++ b/source/common/modules/markdown-editor/renderers/render-emphasis.ts
@@ -15,7 +15,7 @@
 
 import { syntaxTree } from '@codemirror/language'
 import { type RangeSet, type Range } from '@codemirror/state'
-import { type ViewUpdate, type EditorView, ViewPlugin, Decoration, type DecorationSet, WidgetType } from '@codemirror/view'
+import { EditorView, type ViewUpdate, ViewPlugin, Decoration, type DecorationSet, WidgetType } from '@codemirror/view'
 import { rangeInSelection } from '../util/range-in-selection'
 import type { SyntaxNode } from '@lezer/common'
 
@@ -32,6 +32,7 @@ class BulletWidget extends WidgetType {
     const elem = document.createElement('span')
     elem.innerHTML = '&bull;'
     elem.classList.add('rendered-bullet')
+
     return elem
   }
 }
@@ -41,7 +42,7 @@ export class SpaceWidget extends WidgetType {
     super()
   }
 
-  eq (other: BulletWidget): boolean {
+  eq (other: SpaceWidget): boolean {
     if (this.node === undefined || other.node === undefined) {
       return false
     }
@@ -52,6 +53,8 @@ export class SpaceWidget extends WidgetType {
   toDOM (_view: EditorView): HTMLElement {
     const elem = document.createElement('span')
     elem.innerHTML = '&nbsp;'.repeat(this.numChars)
+    elem.classList.add('rendered-space')
+
     return elem
   }
 }
@@ -66,7 +69,7 @@ function hideFormattingCharacters (view: EditorView): RangeSet<Decoration> {
       to,
       enter (node) {
         // Do not hide any characters if a selection is inside here
-        if (rangeInSelection(view.state, node.from, node.to)) {
+        if (rangeInSelection(view.state, node.from, node.to, true)) {
           return
         }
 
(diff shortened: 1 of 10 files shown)


nathanlesage:
Uh, that looks nice. I'll take a look soon :) But so far couldn't spot anything wrong with it!

source/common/modules/markdown-editor/renderers/render-emphasis.ts
```diff
@@ -31,17 +31,22 @@ class BulletWidget extends WidgetType {
   toDOM (_view: EditorView): HTMLElement {
     const elem = document.createElement('span')
     elem.innerHTML = '&bull;'
+
+    if (this.classList !== undefined && this.classList !== '') {
+      elem.classList.add(this.classList)
```
nathanlesage: On Firefox it doesn't work if you provide a string with whitespace, so I'd suggest you provide a list of strings here, or simply set the classList here (since we only assign additional classes below?)

benniekiss: Makes sense, do you mean something like this?

```ts
elem.className = this.className ?? ''
elem.classList.add('rendered-bullet')
```

benniekiss:
Something I didn't test is line-wrapped block quotes, which look ugly given the current rendering. This may need a different approach to account for those.

Screenshot 2025-11-04 at 16 35 20

source/common/modules/markdown-editor/renderers/render-emphasis.ts
nathanlesage: Yes, or directly passing a list of strings:

```
constructor(..., readonly classNames: string[])
```

and assigning that with the destructor:

```
elem.classList.add(...this.classNames)
```

nathanlesage:
> Something I didn't test is line-wrapped block quotes, which look ugly given the current rendering. This may need a different approach to account for those.

My hunch would be to do something similar to the codeblock background lines, where you add a class to the ENTIRE line and style based on that?

benniekiss:
I think the code-background or visual-indent is a good template, but it definitely has to be some type of line decoration. The only tricky part is getting the right number of lines for each quote level

nathanlesage:
Can't you move through the syntax tree one by one and keep a tally of how many opening block quotes you have seen so far ...?

benniekiss:
> Can't you move through the syntax tree one by one and keep a tally of how many opening block quotes you have seen so far ...?

I meant more so creating and rendering the elements. I'm not so certain on the CSS necessary.

The latest commit refactors this some. I removed the custom class names, since those aren't really needed when using line decorations. This is more in the direction I want, but right now, it only puts one line regardless of quote level.

Screenshot 2025-11-05 at 13 00 26

benniekiss:
This still does not show multiple lines per quote level, but here's what the current work looks like. Mainly, the `visual-indent` plugin now accounts for the blockquote indent level, including the quotemarks.

Screenshot 2025-11-06 at 12 07 01

Something that I noticed is that the `visual-indent` plugin actually conflicts with the line-folding plugins. Here's what happens when I collapse the top-level blockquote:

Screenshot 2025-11-06 at 12 29 15 1

benniekiss:
I think I've hit a hard mental block in trying to style the vertical bar to match the number of quotemarks. I think a widget could be used, but I'm pretty sure we would lose spellcheck marks when its rendered.

Otherwise, I've cleaned up the `visual-indent` and `code-background `implementations. The only major refactor was only calculating the indent and backgrounds for the visbile ranges, currently, they are processed for the entire document. Also, folding blockquotes works great now!

Here's what the current implementation looks like:

Screenshot 2025-11-22 at 10 50 24
Screenshot 2025-11-22 at 10 50 31
Screenshot 2025-11-22 at 10 50 41

I think beyond the vertical bar, this PR has a few rather helpful fixes. I also think the issue this was targeting makes a good point about moving away from italics to style blockquotes, and I think using opacity would be a better option, which this implements. 

@nathanlesage I know you're busy, but I just wanted to put this on your radar. Would you be interested in the implementation so far? I could even drop the single vertical bar entirely so that blockquotes are indented and have decreased opacity only. I think the blockquote indent could also be increased to make nested quotes more distinct.

benniekiss:
I'm closing this because I believe the next release of `codemirror/view` will have a feature that's better suited for this implementation https://discuss.codemirror.net/t/codemirror-view-6-39-0-beta-1-looking-for-testers/9588/8

PR #5968, opened 2025-10-26
MD2HTML Improvements

nathanlesage:
## Description

I figured since @benniekiss is now churning out quite some code, it makes sense for me to also switch to PRs for somewhat bigger changes to avoid issues like the footnote mayhem recently.

This PR refactors the md2html function and will – once it's done – enable us to produce more complete HTML, which primarily includes generating a bibliography for the print preview.

## Changes

* Refactored the MD2HTML to return a Promise
* Unify option API for the function
* Adapt existing places across the codebase to use this new signature
* TODO: Enable bibliography generation

## Additional information

Tested on: macOS 26 Tahoe

### DIFF
--- a/test/markdown-to-ast.spec.ts
+++ b/test/markdown-to-ast.spec.ts
@@ -23,7 +23,7 @@ const TESTERS: Array<{ description: string, input: string, output: ASTNode }> = 
 |--|--|
 |No padding|  More padding  |`,
     output: {
-      type: 'Generic', name: 'Document', from: 0, to: 54, whitespaceBefore: '',
+      type: 'Document', name: 'Document', from: 0, to: 54, whitespaceBefore: '',
       children: [
         {
           type: 'Table', name: 'Table', from: 0, to: 54, whitespaceBefore: '',
@@ -70,7 +70,7 @@ const TESTERS: Array<{ description: string, input: string, output: ASTNode }> = 
 -|-
 C|D`,
     output: {
-      type: 'Generic', name: 'Document', from: 0, to: 11, whitespaceBefore: '',
+      type: 'Document', name: 'Document', from: 0, to: 11, whitespaceBefore: '',
       children: [
         {
           type: 'Table', name: 'Table', tableType: 'pipe', from: 0, to: 11, whitespaceBefore: '',
@@ -113,7 +113,7 @@ C|D`,
 
 This iss a *file* with two tpyos in here. asdaa *ss* adas word word.`,
     output: {
-      type: 'Generic',
+      type: 'Document',
       name: 'Document',
       from: 0,
       to: 82,


nathanlesage:
@benniekiss If you want to have a look, feel free to!

benniekiss:
I'll give your branch a test!

So I was poking around in the html conversion a bit ago, and I was curious as to why pandoc isn't used. Is it just difficult to call to the external binary in this context?

benniekiss:
Ok, so first thing I notice is this rendering of footnotes, where I assume they should be on the same line.

Screenshot 2025-10-26 at 13 38 59

I know why this is happening, and its related to the block parsing of footnote refs, and the `partialParse` function.

The logic of `partialParse` is kind of bugged in extracting the child nodes, but it's hard to fix because some parsers depend on the bugged parsing logic. In the footnote ref instance, it extracts the first child, which is a `Paragraph` node, which is not what we want for inline content.

We can get around this by a) refactoring the `partialParse` function or b) refactoring the footnote ref parsing to not need the partial parse. 

I think b) is the better option, since I think a composite block type parser would fit better, and it would natively handle internal element parsing.

I think a) would require some delicate logic to implement and would be more complicated than reimplementing the parser.

nathanlesage:
> Ok, so first thing I notice is this rendering of footnotes, where I assume they should be on the same line.

Yes, but fortunately I just had the idea of mounting the label into the footnote ref body, that should fix this issue.

> In the footnote ref instance, it extracts the first child, which is a Paragraph node, which is not what we want for inline content.

Indeed, we want that: each footnote reference should be a paragraph node. This is because technically footnotes can span multiple paragraphs, and while the logic of properly handling them is still a bit iffy, it should contain paragraphs.

> We can get around this by a) refactoring the partialParse function or b) refactoring the footnote ref parsing to not need the partial parse.

I am aware that the partialParse is a bit of a monkey patch to get sub-document parse trees, but how exactly would you refactor it exactly? What would you change?

benniekiss:
So I've been struggling with a composite parser, but I think the patch is way simpler:

Change these lines:

https://github.com/Zettlr/Zettlr/blob/be0ba67f2e2253185acedeea34783855ed966d9d/source/common/modules/markdown-editor/parser/footnote-parser.ts#L86-L87

To this:

```ts
    const treeElem = ctx.parser.parseInline(footnoteBody.join('\n'), from)
    const body = ctx.elt('FootnoteRefBody', from, to, treeElem)
```

It it solves the rendering/`Paragraph` node issue.

benniekiss:
> Indeed, we want that: each footnote reference should be a paragraph node. This is because technically footnotes can span multiple paragraphs, and while the logic of properly handling them is still a bit iffy, it should contain paragraphs.
> 
> >

I don't think it'll matter how the footnote ref is attached, but I'd be curious to find out if that fixed it! What I mean is that the child nodes get added to the `FootnoteRefBody` as `Paragraph` nodes, which is not what we want, that will wrap the inline content in a `<p>` tag. I experienced this same problem implementing a `PandocSpanParser`. We want the whole footnote ref to be one `Paragraph` element, but if there are internal `Paragraph` nodes, it will break up the single block.

benniekiss:
> I am aware that the partialParse is a bit of a monkey patch to get sub-document parse trees, but how exactly would you refactor it exactly? What would you change?

I would try to move away from the partial parser. I think the API it was trying to fill is pretty covered otherwise. 

Currently, this logic only takes the first child: 

```ts
  let treeElem = ctx.elt(innerTree, offset)
  const firstChild = innerTree.children[0]
  if (firstChild instanceof Tree) {
    treeElem = ctx.elt(firstChild, offset)
  }
```

So if we were parsing multiple paragraphs, only the first would be returned. It also doesn't work for inline parses because it returns regular text wrapped in `Paragraph` nodes, which are supposed to be block nodes. 

---

EDIT: FWIW, I think the footnote and math parser are the only uses of the `partialParse` function, and refactoring the math parser away from the `partialParse` is just one line.

nathanlesage:
Your suggestion with the `inlineParse` was almost right. I think I got it now, and the support for multi-paragraph footnote bodies is now much better. One can still not use block elements in there, but at least now the stuff actually works.

I'll see if I can quickly add the math parser into this PR as well.

I'm pretty sure the inline parse wasn't available when I first implemented that feature, OR if I just didn't know how to properly use it. In any case, thanks for the suggestion!

nathanlesage:
Okay I can't figure out how to switch the context parser from Markdown to stex mode. Any idea? If so, feel free to open another PR. And then we can dump the partialParse method entirely.

nathanlesage:
I'll be merging this part now, but again, please feel free to replace the partial parser in the math parser!

PR #5970, opened 2025-10-27
[Fix] Remove link parser

benniekiss:
## Description

This PR removes the link parser.

## Changes

The native codemirror link parser covers all of the functionality of the existing link parser except for allowing spaces in URLs, which is [against spec](https://spec.commonmark.org/0.31.2/#link-destination). I think at this point, moving into a 4.0 release, it should be OK to remove it entirely. While this may break some links for  users, I think its the right move to reduce the maintenance burden, improve functionality, and stay close to the markdown spec.

## Additional information

When refactoring the parser to handle nested images, a new bug popped up that caused errors with inline URLs within link text. It seems this is a harder bug to workaround given the tight integration of the link parsing in `lezer/markdown`, and it seems the only addition the project implementation provides is accepting spaces.

Tested on:  macOS 26


benniekiss:
I also was able to track the problem down to a conflict with the `Autolink` parsing, specifically how it determines the URL end here: https://github.com/lezer-parser/markdown/blob/98a892cadde1a157f07a653b5f8e5f7ffa74df04/src/extension.ts#L179-L195

So I think the way to fix this is to install the `Autolink` extension after this link parser, but I'm not sure how to do that givin the current parser configuration.

nathanlesage:
Hmmm, that complicates things. Disabling plain links with spaces will break a large majority of the app's users' files, since Pandoc allows this and it has tons of benefits in terms of ease of writing. I myself have a ton of links of this style ...

benniekiss:
ahh, I didn't consider checking pandoc. I agree than that this would be too disruptive.

FWIW, I think this issue is more to do with the `lezer` system's handling of autolinks and links in general. The reason it doesn't trigger for the native links is because the `Autolink` parser specifically checks the `hasOpenLink` attribute, which only triggers for `lezer`'s own links.

I think a proper fix would be to re-implement the `Autolink` parser as well, or disable it for now. I also opened an issue to see if it would be possible to expose a little more of the internal API in `lezer` to make extending the link parsing easier.

PR #5971, opened 2025-10-27
[Refactor] Refactor math parser

benniekiss:
## Description

This PR refactors both inline and block math parsers, and removes the `partialParse` function, which is no longer used in the codebase.

## Changes

The inline math parser was refactored to use delimiters.

The block math parser was refactored for readability. I don't believe the logic is changed.

A check was added to the math renderer `createWidget` function to first check if the cursor is within or adjacent to a math node. This improves UX when cursoring next to inline or block math content.

The `partial-parse` module was removed as its no longer used.

## Additional information

Tested on:  macOS 26


nathanlesage:
Now I feel stupid that I didn't realize one can simply throw a parse tree into the `elt`-function 🥴 But great that we can get rid of at least that monkeypatch, thank you!

benniekiss:
Thanks for catching those inline delimiter issues!

nathanlesage:
Of course, thank you for making the entire Codemirror Operation much simpler and easier to maintain!

PR #5976, opened 2025-10-29
[Refactor] Footnote parsing and rendering

benniekiss:
## Description

This PR refactors the footnote parsing and rendering.

## Changes

The footnote parser was refactored to use delimiters. The footnote reference parser was refactored to use a composite block parsing, which allows interior block nodes.

The footnote rendering was moved to a dedicated renderer rather than a CSS style. This enables nested nodes to be styled. I tried to use a class/theme based approach, but I don't think the class values are applied to child nodes, so this hard-coded style is the only way to get the styling to apply to the interior nodes.

The print page CSS was updated to add more styling to footnote reference body text.

## Additional information

Images:

Screenshot 2025-10-29 at 06 29 58
Screenshot 2025-10-29 at 06 42 24

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/parser/footnote-parser.ts
+++ b/source/common/modules/markdown-editor/parser/footnote-parser.ts
@@ -54,8 +54,10 @@ export const footnoteParser: InlineParser = {
       return -1
     }
 
+    const children = ctx.takeContent(opening)
+
     ctx.addDelimiter(FootnoteDelimiter, pos, pos + 1, false, true)
-    return ctx.addElement(ctx.elt('Footnote', delim.from, pos + 1, isInline ? ctx.takeContent(opening) : undefined))
+    return ctx.addElement(ctx.elt('Footnote', delim.from, pos + 1, isInline ? children : undefined))
   }
 }
 


nathanlesage:
> This enables nested nodes to be styled. I tried to use a class/theme based approach

For this, take a look at the MarkdownParser itself, specifically at how the Highlighter is implemented:

```ts
// NOTE: The convention {TagName}/... means that the corresponding styles
// from the syntax theme get assigned to all child nodes that are contained
// within this node as well. The default is to only style otherwise "empty"
// spans of plain text.
{ name: 'HighlightContent', style: { 'HighlightContent/...': customTags.HighlightContent } },
```

source/common/modules/markdown-editor/parser/footnote-parser.ts
```diff
@@ -12,101 +12,75 @@
...
 }
 
 export const footnoteRefParser: BlockParser = {
   name: 'footnote-refs',
   parse (ctx, line) {
-    const match = /^\[\^[^\s]+\]:\s/.exec(line.text)
-    if (match === null) {
-      return false
-    }
+    // This prevents footnotes from nesting into footnotes
```
nathanlesage, line 67: Does this also ensure that footnote refs must be at the top-level? Because that's how the Pandoc manual references that, and I'd argue that's a cleaner explanation of why we can only allow footnote references at the top-level.

source/common/modules/markdown-editor/parser/footnote-parser.ts
```diff
@@ -12,101 +12,75 @@
...
+
+    line.moveBaseColumn(match[0].length)
 
-    const wrapper = ctx.elt('FootnoteRef', refFrom, to, bodyElems)
-    ctx.addElement(wrapper)
+    return null
+  },
 
+  endLeaf (_ctx, line, _leaf) {
+    return /^\[\^[^\s\^\[\]]+\]:\s/.test(line.text)
```
nathanlesage: I just checked the manual for the BlockParser, and it appears to me as if we can simply check for ANY non-indented line here. Every line that is part of a reference must be indented by at least four characters, so essentially, any non-empty line that is not indented at least by four spaces will automatically end the footnote. Might make this check easier. Or does this already account for the base indent here …?

source/common/modules/markdown-editor/parser/footnote-parser.ts
```diff
@@ -12,101 +12,75 @@
...
+  }
+}
+
+export function footnoteComposite (ctx: BlockContext, line: Line, _value: number): boolean {
+  // If the line is indented, or the line is empty and the next line is indented.
+  if (line.indent >= 4 || (/^\s*$/.test(line.text) && /^([ ]{4,}|\t)/.test(ctx.peekLine()))) {
```
nathanlesage, line 85: Does `peekLine` return a line object? If so, can't we remove the regex by checking `ctx.peekLine().indent`?

source/common/modules/markdown-editor/parser/markdown-parser.ts
```diff
@@ -232,7 +232,12 @@ export default function markdownParser (config?: MarkdownParserConfig): Language
         // spans of plain text.
         { name: 'HighlightContent', style: { 'HighlightContent/...': customTags.HighlightContent } },
         { name: 'Footnote', style: customTags.Footnote },
-        { name: 'FootnoteRef', style: customTags.FootnoteRef },
+        {
+          name: 'FootnoteRef',
+          style: customTags.FootnoteRef,
```
nathanlesage: See the `HighlightContent` implementation for a better way that hopefully does not require a new renderer.

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -648,11 +648,7 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
       return astNode
     }
     case 'FootnoteRef': {
-      const p = node.getChild('Paragraph') // First para
-      if (p === null) {
-        return genericTextNode(node.from, node.to, markdown.substring(node.from, node.to), getWhitespaceBeforeNode(node, markdown))
-      }
-      const label = p.getChild('FootnoteRefLabel')
+      const label = node.getChild('FootnoteRefLabel')
```
nathanlesage, line 651: Did you check the tooltip for footnotes? Since the format has changed a bit now, we may have to change that one as well again.

nathanlesage (changes requested):
In principle, looks very clean, and I appreciate that we can move closer to how Codemirror handles the syntax tree. Just a few notes!

source/common/modules/markdown-editor/parser/footnote-parser.ts
benniekiss: This isn't to end the footnote ref block, this is to interrupt paragraph and other leaf blocks. So if a line matches, then it ends the paragraph and starts the footnote ref. I did this so that footnotes immediately following regular text without a blank line are detected, but I did not check if this is what pandoc allows.

source/common/modules/markdown-editor/parser/footnote-parser.ts
benniekiss: unfortunately it returns a string, not the line object.

source/common/modules/markdown-utils/markdown-ast/index.ts
benniekiss: Latest commit should fix this!

source/common/modules/markdown-editor/parser/footnote-parser.ts
benniekiss: I think so? `ctx.depth` refers to all parents, so it insures footnotes cannot be within another block element, I believe.

nathanlesage: Wonderful, thank you!

source/common/modules/markdown-editor/parser/footnote-parser.ts
nathanlesage: Aaaahhhh, I see, that makes sense. I just tested it, and nope, Pandoc requires blank lines above and below the footnote reference.

source/common/modules/markdown-editor/parser/footnote-parser.ts
nathanlesage: Sad, but thanks for confirming!

source/common/modules/markdown-editor/parser/footnote-parser.ts
```diff
@@ -44,11 +46,19 @@ export const footnoteParser: InlineParser = {
...
 
+    // Inline footnotes can contain markup, however, identifier footnotes cannot.
+    const isInline = ctx.char(delim.from) === 94 // 94 === '^'
+
+    // Finally, check if the identifier is valid
+    if (!isInline && !validFootnoteRe.test(ctx.slice(delim.to, pos))) {
+      return -1
+    }
+
     const children = ctx.takeContent(opening)
```
nathanlesage: How expensive is this function? If so, might it make sense to refactor to return conditionally? Like "if inline, then takeContent, otherwise return an empty element"?

source/common/modules/markdown-editor/parser/footnote-parser.ts
benniekiss: Great, that means this can just be removed entirely

source/common/modules/markdown-editor/parser/footnote-parser.ts
benniekiss: I'll see if we don't have to run this command, but I believe for delimiter parsing, if we don't do a `takeContent`, then anything parsed before this point will be added to the tree as its own node, so if there's inline markup, it'll be included separately

benniekiss: I don't think it is any expensive call, though. It doesn't retrigger parsing except for resolving interior delimiters.

benniekiss: Ok, so it doesn't seem `takeContent` is needed to avoid duplicate nodes, however, since pandoc footnote identifiers can contain any character except `\s ^ [ ]`, we need to call `takeContent`, otherwise any characters that could be inline formatting interrupt the node.

For example, though probably not the best idea, a valid identifier could be `[^my*star*]`, but without `takeContent`, this will not parse as a footnote.

nathanlesage: Roger, I rest my case! Thanks!

nathanlesage:
With that being said, all my comments have been addressed and from my side we can merge. Is it good to go from your side, too?

benniekiss:
I'm comfortable with where things are. LGTM on my end!

PR #5977, opened 2025-10-30
[Fix] Footnote refactor fixes

benniekiss:
## Description

This PR addresses two issues with the new footnote refactor: stacked footnote reference parsing, i.e., references with no blank lines between them, and an edge case regarding invalid identifier footnotes and links.

## Changes

The inline parser was refactored to `return -1` when it is adding an opening delimiter for identifier (`[^1]`) type footnotes. This enables the built in link parser to add its own delimiters, and enables us to fallback to link parsing when the identifier is otherwise invalid. For instance, `[^in valid](example.com)` will now correctly parse as a URL. This is the behavior of pandoc, as well.

I believe this is an appropriate approach, but it does feel a little experimental.

The block parser had the `endLeaf` method restored. This was removed during review, however, I found that it's necessary to include it given the structure of the blocks in the composite node. Since the footnote ref is a composite, it includes other blocks, and it technically starts a paragraph. Because paragraphs end on a blank line, if the footnote references are stacked, then the body text will be parsed to the next blank line because the internal paragraph is never interrupted. By adding the `endLeaf`, we can end paragraph blocks only if they are direct children of the footnote reference.

## Additional information

I hadn't realized its possible to enable this kind of fallback parsing by returning `-1` instead of the delimiter position, so with this new approach, I should have a fix for the `sloppy-parser` as well.

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/parser/footnote-parser.ts
+++ b/source/common/modules/markdown-editor/parser/footnote-parser.ts
@@ -18,6 +18,8 @@ const FootnoteDelimiter: DelimiterType = {}
 
 const validFootnoteRe = /^[^\s\^\[\]]+$/
 
+const footnoteRefRe = /^\[\^[^\s\^\[\]]+\]:\s/
+
 export const footnoteParser: InlineParser = {
   name: 'footnotes',
   before: 'Link', // [^1] will otherwise be detected as a link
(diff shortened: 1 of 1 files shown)


nathanlesage:
Looks good! One additional thing that occurred to me yesterday: does the footnote insertion command now still work, given the new structure of the DOM?

benniekiss:
Quickly tested with `Mod-Alt-R`, and footnotes are still inserted!

nathanlesage:
Is the numbering also correctly updated? 👀 I remember a few days ago it wasn't and I forgot to fix it then

benniekiss:
Yeah, the ordering seems to be correct -- if I insert a footnote between others, it correctly adjusts the following numbers

nathanlesage:
In that case, nothing to add -- thank you!

Issue #5938, opened 2025-10-07
Directory > Properties > Sort by time: does not auto refresh


digbyphonic:
### Description

In the ideal situation the working repository sidebar folder view with `Properties` set to `Sort by time` would automatically update.

e.g. saving a file would move it to the top if `Sort by time` and `ascending` is selected in `Properties`.

Instead, the order does not refresh until right-click repository in working area to bring up `Properties` dialogue. 

### Reproducing

Save new version of file.

File does not automatically change its timed sort order in `Repositories` sidebar folder view.

### Zettlr version

3.6.0

### Installation Method

From the Website or GitHub

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Windows 11 Pro 24H2

BabakFarshchian:
I can confirm this. I think this is the same for both panels, file browser but also ToC panel to the right. For ToC I always have to click on the panel, then click on the note area to refresh the ToC. 

This applies also to 4.0 beta 1.

Issue #5920, opened 2025-09-24
Link incorrectly reports file does not exist


larrytalley:
### Description

I link an existing note via [[e1906-1910-78085]] which points to an existing file. When I hover over the link it reports "File does not exist". If I then open file, input the file name, and select the file, it will of course open the file.

### Reproducing

This began to happen to me about 3 days ago. Prior to that it hadn't been apparent to me. Now it happens often. I don't know what it different now.

### Zettlr version

Zettlr 3.6.0 UUID: 7f1e155e-93ca-4c36-b780-197d80a0c859

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Mac OS 13.7.8 (22H730)

### Additional Information

There doesn't appear to be anything in the log that directly corresponds in time to my observation of the bug, but, here are a few messages from the log from earlier in the day:

[12:06:10]
[R] [Main Window (656e4a1a-97c3-4f31-a27f-5fb9a22087b9)] Measure loop restarted more than 5 times (index.js:2)
[12:06:54]
[WorkspaceManager] Root /Users/ltalley/Mailboats Project/d1938-1942-78074.md has been removed
[12:29:43]
[R] [Main Window (656e4a1a-97c3-4f31-a27f-5fb9a22087b9)] MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 documents-update listeners added. Use emitter.setMaxListeners() to increase limit (sandbox_bundle:2)

larrytalley:
https://github.com/user-attachments/assets/e2987718-d55a-4beb-91cf-83649f63e36a

larrytalley:
The inability of Zettlr to find notes (files) by name is also affecting partial match. Prior to this bug, I could open a link via "[[" and type a partial match for another note, and Zettlr would respond with a list of matches and I could then select from that list. Now that many of the notes (files) in my collection are not able to be found by Zettlr, I don't get much value from the partial match, and I realize how often I had used that feature!

nathanlesage:
I'll be looking into this! (Or someone else, whoever is faster)!

larrytalley:
FYI I have another Macbook, a smaller model that I use for travel. It has the same files (I use Syncthing to keep my Zettlr library synchronized.) 

[...]

Issue #5784, opened 2025-05-31
file manager does not update


thekswenson:
### Description

The filemanager is not updated when a file is moved.

### Reproducing

1. create a new folder
2. move an existing file to that folder via drag and drop

notice that the filemanager is not updated.  the same happens when I move a file using the system filemanager.

### Zettlr version

3.5.0

### Installation Method

Other (please specify below)

### Installation Method (Other)

.deb

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Ubuntu 25.04

thekswenson:
This may be related to:

https://github.com/Zettlr/Zettlr/issues/5773
https://github.com/Zettlr/Zettlr/issues/5594

siwei-li:
I have the same issue on macOS 15.5

I'd need to unload the workspace and open it again whenever there's any change, including but not limited to: renaming .md files, creating/deleting/moving folders and files

Issue #5773, opened 2025-05-21
Zettlr File Manager is not showing changes made in a Github folder


polgarp:
### Description

I have a Workspace which is also a Github repository (when I open the properties, this is also visible there). When I make any change here (creating a new file, renaming a file, creating a new folder, deleting a file etc). The File Manager doesn't refreshes and shows the previous state. I have this issue on Windows and Mac too. The changes are visible if I open a local folder as a Workspace.

### Reproducing

1. Clone a Github repository locally.
2. Open the folder in Zettlr as a new workspace.
3. Create a new file in the workspace.
4. The new file is not visible in the File Manager.

### Zettlr version

3.4.4

### Installation Method

From the Website or GitHub

### Your Platform

- [x] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [x] ARM 64bit (includes Apple Silicon)

### Operating System Version

Windows 10 & macOS 14.4.1

peterwjacko:
I had the same issue yesterday. I'll add my report here rather than opening a new issue. Let me know if a new issue is better.

## Description

New files are not shown in the file manager pane. The workspace is a git repo and one of the subdirs is a Zettlr "project" but unsure if that makes a difference. Opening the file individually works, but then it's not possible to close the file. 

It could have been a coincidence, I also couldn't re-open a workspace (or create new workspaces) like in #5331 and #5332.

I enabled debug mode to see if I could find a log or messages but nothing looked out of place to my untrained eye.

**Issue resolved with "Developer reload"** 

Unfortunately I can't seem to reproduce it now and everything works as expected.

## Reproducing

1. Open git repo root dir as "Workspace"
2. "Enable project" on one subdir (e.g. `../repo_root/project1`)
3. Create new file with `⌘ + n`
4. Use Finder pane (launched by Zettlr) to navigate to `project1` dir
5. Save file with custom name (e.g. `my-document.md`)
6. The new file launches in a new Zettlr tab but does not appear in the file manager pane (confirmed file is present in system file manager)
8. Make some edits & save file - still nothing in file manager

Attempted:

1. Quitting the app
2. Rebooting OS
3. Reinstalling app (from `brew`)

Resolved the issue with:

1. Enable `Debug mode`
2. `Developer > Reload (F5)`
3. File manager now shows changes

## Zettlr version

**Zettlr Version:** 3.4.4
**Build number:** 28b034a
[...]

PR #5978, opened 2025-10-30
refactor: FSAL v2

nathanlesage:
## Description

This PR aims to fully rewrite the entire FSAL a second time. This time, the goal is to get rid of any brittle tree structures that we essentially do not need, and instead treat the entire file system as a long list, and defer any tree-like recursion to the Vue renderer. Because that's the only time we actually need a tree: when we display the file tree.

The reason for this is simple: When I wrote the first file system layer back in 2017, I had zero plan of how all of this works, and so I essentially mistrusted anything, and kept every descriptor I needed always loaded.

When I rewrote the layer the first time to create the FSAL, I went a bit bolder, but still decided to keep an entire file system tree in memory. And not just that, there's copies of that file system tree across all main windows. You can imagine how much memory that eats up.

This time, the goal is to treat the filesystem as the single source of truth, and only load from disk what we absolutely need. Any changes during runtime will be deferred to chokidar, which now works flawlessly since a few years.

By doing so, I hope to achieve the following goals:

* Get rid of any nested tree structures that are difficult to navigate and brittle to patch
* Only load what is absolutely necessary for the functioning
* By getting rid of any trees, I can treat the loaded files as a single large list wherever necessary, which makes dealing with it much simpler. I'm thinking of the tag provider, the link provider, and anything else that extracts information from this tree.
* Get rid of the ugly recursion in the FSAL that requires instantiating and passing around an unwieldy Markdown file parser that, en passent, also doesn't work properly.
* Improve performance considerably, both in terms of CPU time, simply because there's less work to do, and in terms of memory consumption, because there's no weird trees hanging around.
* Make the code much simpler by removing all the bloated tree-patching functionality.
* Fix issues pertaining to the app not being able to pick up on all changes that happen on the file system.
* Bypass all providers and grab the necessary information directly from the renderer via the FSAL.

## Changes

* change!: Completely get rid of the brittle tree structures. Instead, everywhere directories are treated as simple lists, and only the renderers will, where necessary, recursively treat them as trees for visual purposes.
* change!: Rewrite the `workspaces`-pinia store in the renderer. It now only keeps a list of all loaded paths, and all descriptors, patching these lists as FSAL events come in with no tree merging.
* change!: Completely remove the `WorkspaceProvider`.
* refactor: The file system watchers now report all events in raw form, and the FSAL will immediately broadcast them. This interestingly has led to the updating of file items in the file manager to be much faster and feel almost instant.
* chore: Removed a large amount of old, dead code in the directory module (`FSALDir`) that hasn't been in use.
* Many commands are now much leaner, since they essentially just trigger raw filesystem actions instead of going through the bloated tree update procedures.
* change!: No recursive file tree reading on parsing directories. This means that **`DirDescriptor`s no longer have a `children` property**! Instead, anyone who needs the children of a folder needs to make a second round trip to the main process. There is only one instance where an entire file tree will be read in. This is only meant so that we have a list of all loaded files (for the file autocomplete, for example), and then it will only completely read in the paths of the file tree, not its descriptors.
* change: Directory sorting is now completely deferred to each renderer element. Every `TreeItem` and the `FileList` will now sort its own children (recursively in the case of the `FileList`) according to general and particular sorting rules as defined in the settings and on the directory. This fixes a bunch of sorting issues we have had.
* fix: Small but until now unnoticed bug where the config provider would never emit an update event when the `openPaths` (=roots/workspaces) have changed.
* feat: Debounce splash screen update to at most 60fps from the main process to not overload the IPC pipe with large amounts of updates. (The splash screen now sees all individual files as they are being re-indexed by the FSAL.)
* chore: Improved typing of some file system commands (the `arg` property).
* refactor: Rename `obj` --> `item` in both `TreeItem` and `FileItem`.

### Fixes:

#5938
#5920
#5784
#5773
#5594
#5184
#5345

(if you find more, feel free to ping me!)

### Still TODO:

- [x] Proper sorting of children in the file list (essentially here we need a recursive array sorter)
- [x] Some more testing
- [x] A few performance tests and improvements

### Caveats

There are some performance bottlenecks, but fortunately these only seem to come from the fact that I perform a few too many IPC calls. So by now (at least on my relatively fast computer) the biggest performance bottleneck seems to be the IPC pipe which works much better with larger chunks of data.

Tested on: macOS 26

### DIFF
--- a/source/app/app-service-container.ts
+++ b/source/app/app-service-container.ts
@@ -154,7 +154,6 @@ export class AppServiceContainer {
 
     await this._informativeBoot(this._fsal, 'FSAL')
 
-
     await this._informativeBoot(this._recentDocsProvider, 'RecentDocsProvider')
     await this._informativeBoot(this._assetsProvider, 'AssetsProvider')
     await this._informativeBoot(this._cssProvider, 'CSSProvider')


nathanlesage:
So, I have now finalized work on this. Apologies that it has become so large, but it definitely felt refreshing that now the entire logic is much simpler.

I have tested most of the parts, and they work as expected. But a few additional tests might be good. However, we have the nightlies and then the beta phase for comprehensive testing with a larger user base.

@kyaso @benniekiss Anyone wants to give this a ride? Again, any review doesn't have to be fully comprehensive, we have some time to test after merging, too!

source/app/service-providers/commands/dir-project-export.ts
```diff
@@ -13,15 +13,13 @@
  */
 
 import ZettlrCommand from './zettlr-command'
-import objectToArray from '@common/util/object-to-array'
```
nathanlesage, line 16: Completely removed since we no longer have any tree structures that need to be flattened

source/app/service-providers/commands/dir-remove-project.ts
```diff
@@ -22,11 +22,12 @@ export default class DirRemoveProject extends ZettlrCommand {
...
   /**
     * Remove the project of a directory
-    * @param {String} evt The event name
-    * @param  {Object} arg The hash of a directory.
+    *
+    * @param  {string} evt  The event name
+    * @param  {any}    arg  The payload
     */
-  async run (evt: string, arg: any): Promise<void> {
-    let dir = this._app.workspaces.findDir(arg.path)
```
nathanlesage, line 29: The commands have been entirely moved from calling `this._app.workspaces` to `this._app.fsal`, which implements the same functionality, but on the bare metal file system syscalls instead of the tree structure.

source/app/service-providers/commands/dir-rename.ts
```diff
@@ -24,11 +24,12 @@ export default class DirRename extends ZettlrCommand {
...
    * Rename a directory
-   * @param {String} evt The event name
-   * @param  {Object} arg An object containing hash of containing and name of new dir.
+   *
+   * @param  {string}  evt The event name
+   * @param  {any}     arg An object with the path for the source dir and the new directory name.
    */
-  async run (evt: string, arg: any): Promise<boolean> {
-    const sourceDir = this._app.workspaces.findDir(arg.path)
+  async run (evt: string, arg: { path: string, name: string }): Promise<boolean> {
```
nathanlesage, line 31: Also a bunch of type improvements

source/app/service-providers/commands/dir-rescan.ts
```diff
@@ -25,7 +25,8 @@ export default class DirRescan extends ZettlrCommand {
     * @param {String} evt The event name
     * @param  {Object} arg The path of the descriptor
     */
-  async run (evt: string, arg: any): Promise<void> {
-    await this._app.workspaces.rescanForDirectory(arg.path)
+  async run (event: string, _arg: any): Promise<void> {
+    // DEBUG: DEPRECATED
+    throw new Error('rescanForDirectory not re-implemented')
```
nathanlesage, line 30: This hasn't been functional since ages and nobody has complained. That's only neessary for a very small set of people who have files on a thumb drive, so that they can plug and unplug it. OOS for this PR.

source/app/service-providers/commands/dir-sort.ts
```diff
@@ -14,21 +14,23 @@
...
+    * @param  {string}  evt  The event name
+    * @param  {any}     arg  An object containing both a path and a sorting type
     */
-  async run (evt: string, arg: any): Promise<boolean> {
-    const dir = this._app.workspaces.findDir(arg.path)
+  async run (evt: string, arg: { path: string, sorting: SortMethod }): Promise<boolean> {
+    const dir = await this._app.fsal.getAnyDirectoryDescriptor(arg.path)
     if (dir !== undefined) {
-      await this._app.fsal.sortDirectory(dir, arg.sorting)
+      await this._app.fsal.changeSorting(dir, arg.sorting)
```
nathanlesage, line 33: Sorting now exclusively happens in the renderer, because elsewhere it doesn't matter. I have pushed everything I could to the edge of the program (meaning defer computation of anything until it's absolutely necessary)

source/app/service-providers/commands/file-find-and-return-meta-data.ts
```diff
@@ -57,7 +57,7 @@ export default class FilePathFindMetaData extends ZettlrCommand {
     // The filename can contain a `#`, indicating a specified heading in the target file
     const filename = arg.includes('#') ? arg.slice(0, arg.indexOf('#')) : arg
     // Quick'n'dirty command to return the Meta descriptor for the given query
-    const descriptor = this._app.workspaces.findExact(filename)
+    const descriptor = await this._app.fsal.findExact(filename)
```
nathanlesage, line 60: Not completely happy with this, but it's better than nothing. Might also fit into a command, but OOS.

source/app/service-providers/commands/file-rename.ts
```diff
@@ -79,12 +81,11 @@ export default class FileRename extends ZettlrCommand {
     // Thus, we need to check two conditions: Whether the user has requested a
     // case change only, and whethere there is a DIFFERENT file at that new
     // place.
-    const newPathFile = this._app.workspaces.findFile(newPath)
     const caseChangeOnly = newName.toLowerCase() === file.name.toLowerCase()
 
     if (await this._app.fsal.pathExists(newPath)) {
       // The file system reports the newPath already exists.
-      if (caseChangeOnly && (newPathFile === undefined || newPathFile === file)) {
+      if (caseChangeOnly && await pathExists(newPath)) {
```
nathanlesage, line 88: Sometimes, rewriting has significantly reduce the complexity of the logic :D

source/app/service-providers/commands/index.ts
```diff
@@ -121,9 +121,7 @@ export default class CommandProvider extends ProviderContract {
   async run (command: string, payload: any): Promise<any> {
     // FIRST: Try to run a minimal command for which its own custom function
     // wouldn't make sense.
-    if (command === 'get-statistics-data') {
```
nathanlesage, line 124: Statistics is now 100% renderer-based

source/app/service-providers/commands/rename-tag.ts
```diff
@@ -34,7 +34,7 @@ export default class RenameTag extends ZettlrCommand {
     const newName: string = arg.newName
 
     // First, retrieve all files from the FSAL
-    const allFiles = this._app.workspaces.getAllFiles()
+    const allFiles = (await this._app.fsal.getAllLoadedDescriptors())
```
nathanlesage, line 37: I think there are a few places where we could use this small helper function, but which I have forgotten. If you spot anywhere in this PR, feel free to ping.

source/app/service-providers/config/index.ts
```diff
@@ -369,6 +369,7 @@ export default class ConfigProvider extends ProviderContract {
       this.sortPaths()
       this._container.set(this.config)
       this._emitter.emit('update', 'openPaths')
+      broadcastIpcMessage('config-provider', { command: 'update', payload: 'openPaths' })
```
nathanlesage, line 372: This bug is actually unrelated to this PR, but I only realized it now: the config provider never informed the renderers if the user has added a workspace. I have no idea why it worked previously.

source/app/service-providers/fsal/fsal-attachment.ts
```diff
@@ -16,8 +16,9 @@ import path from 'path'
 import { promises as fs } from 'fs'
 import type { OtherFileDescriptor } from '@dts/common/fsal'
 import { shell } from 'electron'
+import type FSALCache from './fsal-cache'
 
-export async function parse (absPath: string): Promise<OtherFileDescriptor> {
+export async function parse (absPath: string, cache: FSALCache): Promise<OtherFileDescriptor> {
```
nathanlesage, line 21: Attachments can now be cached. It's not much, but it makes retrieval of the cached descriptors a thousand times faster. (Went from about 15s to load the entire workspace tree to ca. 8s b/c the files don't have to be fetched from disk anymore.)

source/app/service-providers/fsal/fsal-directory.ts
```diff
@@ -54,19 +45,6 @@ const PROJECT_TEMPLATE: ProjectSettings = {
   }
 }
 
-/**
- * Sorts the children-property of "dir". NOTE that this is an internal helper
- * function to ensure possible API changes are properly respected in the future.
- *
- * @param   {DirDescriptor}  dir  A directory descriptor
- */
-function sortChildren (
```
nathanlesage, line 63: This file included a bunch of dead code. I think I at some point wanted to push all functionality into these helper modules, but it's just unnecessary for many parts, especially since I never implemented any other storage adapters into the FSAL (yet).

source/app/service-providers/links/index.ts
```diff
@@ -31,23 +32,23 @@ export default class LinkProvider extends ProviderContract {
...
+    ipcMain.handle('link-provider', async (event, message) => {
       const { command } = message
 
       if (command === 'get-inbound-links') {
         // Return whatever links to the given file
         const { filePath } = message.payload
         return {
           inbound: this.retrieveInbound(filePath),
-          outbound: this.retrieveOutbound(filePath)
+          outbound: await this.retrieveOutbound(filePath)
```
nathanlesage, line 51: These things were somewhat sneaky: Since FS access is now 100% async, there were a few Promise-returns to the renderer. There may be some more lurking that I didn't spot.

nathanlesage: The problem is that Promises can't be serialized through the IPC pipe, that's when there will be odd errors thrown. Took me a while to figure out.

source/app/service-providers/workspaces/index.ts
nathanlesage, line 1: Workspaces are now completely gone.

source/app/util/splash-screen.ts
```diff
@@ -29,6 +29,8 @@ let splashScreen: BrowserWindow|undefined
 let initSplashScreenMessage = ''
 let initSplashScreenPercent = 0
 
+let debounceTimeout: NodeJS.Timeout|undefined
```
nathanlesage, line 32: Pro: I can now show exactly which file is currently being indexed.

Con: Pressing EVERY update over the IPC pipe has made the IPC a bottleneck, so I'm limiting updates to max. 60fps.

source/app/app-service-container.ts
```diff
@@ -34,8 +34,10 @@ import TargetProvider from '@providers/targets'
 import TrayProvider from '@providers/tray'
 import UpdateProvider from '@providers/updates'
 import WindowProvider from '@providers/windows'
-import WorkspaceProvider from '@providers/workspaces'
 import { dialog } from 'electron'
+import { closeSplashScreen, showSplashScreen, updateSplashScreen } from './util/splash-screen'
```
nathanlesage, line 38: The splash screen never really fit in elsewhere. Now the splash screen also indicates booting of other providers.

One NOTE is an interference with the firstStartGuide that the config provider is showing.

source/common/vue/form/elements/ProgressControl.vue
nathanlesage, line 1: Strictly speaking OOS for this PR; but I always wanted to give the progress bar an indeterminate status, which I did with this.

One thing I couldn't figure out is why the animation doesn't work. Maybe because it's an non-standard pseudo-prop?

source/pinia/workspace-store.ts
```diff
@@ -0,0 +1,253 @@
...
+ * 
+ * @throws if the path is not a directory
+ */
+async function readDirectory (absPath: string): Promise<AnyDescriptor[]> {
+  return await ipcRenderer.invoke('fsal', { command: 'read-directory', payload: absPath })
+}
+
+// In order to avoid frequent updates of the workspaceMap on initial load, we
+// retrieve the bulk immediately on load, and then only patch where necessary.
+async function retrieveInitialUpdate (rootPaths: string[], workspaceMap: Ref<Map<string, string[]>>, descriptorMap: Ref<Map<string, AnyDescriptor>>) {
```
nathanlesage, line 71: Retrieving workspaces and descriptors as one humongous bulk and then updating once during startup has decreased load time with my full directory tree (the one I use in production, not the few test files; ca. 7,000 files) from ca. 12 seconds to 500ms, which shows how huge the overhead from IPC calls is.

benniekiss:
Just a quick question, but with the removal of the workspace provider, is the concept of workspaces still around? I've been working on a workspace-level configuration feat, and I was wondering if this would completely change that.

source/pinia/workspace-store.ts
nathanlesage, line 71: Retrieving all workspaces and descriptors in one big swoop dramatically reduces load times from several seconds to a few milliseconds. I was honestly surprised to see how much overhead the IPC calls and the Vue reactivity was adding with ~7,000 files.

source/win-main/file-manager/util/retrieve-children-and-sort.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Utility function that recursively sorts the various contained directories
+ * within the base descriptor according to the correct settings and then returns
+ * them as a flat list sorted accordingly.
+ *
+ * @param   {AnyDescriptor}      descriptor      The base descriptor
+ * @param   {AnyDescriptor[][]}  allDescriptors  The registry of descriptors
+ *
+ * @return  {AnyDescriptor[]}                    The sorted list of descriptors.
+ */
+export function retrieveChildrenAndSort (descriptor: AnyDescriptor, allDescriptors: AnyDescriptor[], sorter: GenericSorter): AnyDescriptor[] {
```
nathanlesage, line 14: Filtering and sorting is now centralized in the renderer

source/win-main/file-manager/FileItem.vue
```diff
@@ -10,13 +10,13 @@
       ref="displayText"
       v-bind:class="{
         'list-item': true,
-        project: obj.type === 'directory' && obj.settings.project !== null,
-        selected: selectedFile !== undefined && obj.path === selectedFile.path,
-        active: activeFile !== undefined && obj.path === activeFile.path,
+        project: item.type === 'directory' && item.settings.project !== null,
```
nathanlesage, line 13: I have consistently updated `obj`->`item` in both FileItem and TreeItem.

source/app/service-providers/fsal/fsal-attachment.ts
nathanlesage, line 21: Attachments are now cached, too, since I've switched the FSAL to aggressively return cached descriptors upon request, instead of accessing each file on disk again. This increases the cache size, but that was never too big anyways.

source/app/service-providers/fsal/index.ts
```diff
@@ -583,4 +800,62 @@ export default class FSAL extends ProviderContract {
...
+   * 
+   * NOTE: This function will already exclude dotfiles and ignored directories,
+   * so this function is safe to consume in terms of what Zettlr should display.
+   *
+   * @param   {string}             directoryPath  The absolute path to parse
+   *
+   * @return  {Promise<string[]>}           Returns a list of the entire directory
+   */
+  public async readDirectoryRecursively (directoryPath: string): Promise<string[]> {
+    const contents = (await fs.readdir(directoryPath, { withFileTypes: true }))
```
nathanlesage, line 821: Here I'm deferring much more to syscalls which can return `dirent` objects much faster than my `fs.access`-calls would be. This is also the reason why I have to restrict this function to only accept directory paths and need to guard against file paths above in this file, since otherwise we would need isFile/isDir checks on every recursive call, making this function much slower than necessary.

source/app/util/splash-screen.ts
nathanlesage, line 1: Strictly speaking OOS but I've moved the Splash screen to the AppServiceProvider now, where I believe it fits better, and this way we can even show if something else needs a long time for booting up.

source/app/service-providers/commands/dir-new-project.ts
```diff
@@ -22,11 +22,12 @@ export default class DirNewProject extends ZettlrCommand {
...
-    * @param {String} evt The event name
-    * @param  {Object} arg The hash of a directory.
+    *
+    * @param  {string}  evt  The event name
+    * @param  {any}     arg  The payload
     */
-  async run (evt: string, arg: any): Promise<void> {
-    let dir = this._app.workspaces.findDir(arg.path)
+  async run (evt: string, arg: { path: string }): Promise<void> {
+    const dir = await this._app.fsal.getAnyDirectoryDescriptor(arg.path)
```
nathanlesage, line 30: Most changes in the commands refer to moving from workspaces to the FSAL. This also makes the logic simpler in some places. Also, a lot of proper typing in the arg.

source/app/service-providers/links/index.ts
```diff
@@ -31,23 +32,23 @@ export default class LinkProvider extends ProviderContract {
...
+  constructor (private readonly _logger: LogProvider, private readonly _config: ConfigProvider, private readonly _fsal: FSAL) {
     super()
 
     this._fileLinkDatabase = new Map()
     this._idDatabase = new Map()
     // TODO: Add a set of duplicate IDs so we can inform the user so they can
     // fix this
 
-    ipcMain.handle('link-provider', (event, message) => {
+    ipcMain.handle('link-provider', async (event, message) => {
```
nathanlesage, line 43: Since the entire FSAL access is now async, I have stumbled a few times upon accidentally sending Promises to the renderer, which fails as those can't be serialized over the IPC pipe.

source/common/vue/form/elements/ProgressControl.vue
```diff
@@ -2,9 +2,15 @@
   <div class="progress-bar-container">
     <progress
       v-bind:max="props.max ?? 100"
-      v-bind:value="props.value ?? 0"
+      v-bind:value="props.indeterminate === true ? undefined : props.value ?? 0"
```
nathanlesage, line 5: Strictly speaking OOS but I always wanted to add an indeterminate state to the progress bar.

If anyone knows why the animation isn't playing, I'd be absolutely stoked!

nathanlesage:
> Just a quick question, but with the removal of the workspace provider, is the concept of workspaces still around? I've been working on a workspace-level configuration feat, and I was wondering if this would completely change that.

Yes, absolutely. Workspace === directory. The "WorkspaceProvider" was essentially my attempt at retaining the tree structure of the file system, because I didn't trust the file system. Now that I do, we don't need that logic that essentially has been entirely reimplemented via the FSAL.

Feel free to open an issue regarding your planned feature/PR and we can discuss it there. I'm happy to imbue Workspaces with more functionality if that makes sense!

benniekiss:
Oh that's great to here. I gave this PR a test run and nothing seemed immediately broken! I'll see if I can give this a lookover in a little as well.

I'll open a WIP PR for the workspace config after you get this merged so I can clean up any conflicts. It's ~~mostly working

nathanlesage:
Wonderful! Once this here is merged I'm also gonna merge a bunch of your PRs. Sorry again that I apparently am only capable of creating monster PRs :D But at least it's not vibe coded!

benniekiss:
You have a negative diff here, so really its only a tiny massive.

benniekiss:
Btw, I don't think #5966 is closed by this. I can still repro the issue. But I do have a fix in #5983

nathanlesage:
As long as we have a fix that's perfect ;)

nathanlesage:
Also, feel free to comment on the performance -- is it similar, or even a bit faster?

benniekiss:
It definitely feels snappier, but I'm never working with many documents at once. It also seems to improve the editor performance some? But that may just be a perspective thing.

benniekiss:
So one thing I noticed, which is really quite small, is that when the onboarding window pops up, the loading screen appears shortly after on top of the onboarding window, and it doesn't start loading until the user clicks back into the onboarding window and clicks get started. 

Is it possible to make the loading screen wait to appear until after the onboarding window is closed? I think that would be a better UX

https://github.com/user-attachments/assets/ea6eb0f1-2692-4645-b900-ce6a83282500

nathanlesage:
> Is it possible to make the loading screen wait to appear until after the onboarding window is closed?

Yes, I also noticed that. Essentially, I only need to defer the loading screen until after the config provider has booted. Give me a sec

nathanlesage:
> It definitely feels snappier, but I'm never working with many documents at once. It also seems to improve the editor performance some? But that may just be a perspective thing.

Doesn't surprise me — fetching the list of all descriptors is now much faster because no roundtrip to the main process is needed. That's why I rewrote the entire thing now: We need 99% lists, not trees. Updating trees is a pain, and by keeping everything as a list, we save tons of compute, and make the code less complex. The only "complex" piece of code now is in the FileTree and FileList, that do need some form of tree visualization.

nathanlesage:
@benniekiss Does this otherwise look mergable to you?

benniekiss:
I want to give it one more performance test, because there was something I noticed in loading a really large and nested directory that could be addressed here, but I'm not sure if I tested it correctly.

nathanlesage:
Roger, just let me know :)

benniekiss:
I noticed that when I open a directory that takes a long time to load, in my case because of nested directories and large files, the UI gets locked up, and there isn't any indication of what work is being done. 

However, I noticed that on `develop`, when I load the directory, a progress bar shows up right at the end as its finished indexing, however, one does not show up on this branch, probably because of the performance improvements you realized.

Is it possible to move the progress bar window so that it appears when I add the workspace? There seems to be some other process that triggers it to appear right at the end, and on this branch, it doesn't appear at all.

Another helpful feature would be to add a `cancel` button to the progress window to gracefully interrupt loading the workspace

nathanlesage:
You mean the entire UI doesn't do anything? That means the main process is locked up, too, which indicates that even though the syscalls are non blocking, they are so fast that the main event loop can't do much between requesting and receiving the file from the operating system. Hm.

I don't wanna reinstate the old behavior since this was strictly speaking a (benevolent) bug, as a splash screen should only appear in the beginning of booting the app.

We possibly have two options: either move all the indexing into a worker thread to not block the main process, or display a full-window loading screen. I'd argue moving it to the background into a worker thread would be the best idea. But this should also probably be a second PR. You wanna open a quick issue so we don't forget this before the launch of 4.0?

benniekiss:
Its definitely not a behavior introduced here, so a separate issue/pr totally makes sense, and I'll open one in a bit.

To clarify -- the whole UI does lock up, I get the spinning color wheel. In the logs, it starts slowing down when creating the "shards" for the FSAL cache

nathanlesage:
Wait ... my, I think the shards are blocking calls ... would you mind trying out what happens if you use fs promises for creating the shards? What must be synchronous is the access, but the creation can be async I believe.

But alright, let's merge this one (after all, how often does one load a new workspace), and figure this out along the way. Thank you so much!

PR #5964, opened 2025-10-24
[Fix] Refactor link and image parsing

benniekiss:
## Description

This PR refactors the `sloppy-parser` link parser and updates the image and link renderers to match.

Closes #5959
Closes #5962

Supersedes #5963

## Changes

The `sloppy-parser` was renamed to `link-parser` so it's clearer what the parser actually does. Some of the difficulty in tracking down the closed issues was not recognizing the `sloppy-parser` as a link parser.

The parser was refactored to use `lezer` delimiter parsing, which simplifies and speeds up parsing.  It relies on new APIs in version 1.5.1. The structure of the link and image nodes was also updated to match the codemirror nodes. Notably, this means that the `LinkLabel` node is no longer assigned to what is the link or image text -- this was actually an incorrect application of the node type, which refers to the text enclosed in brackets like so: `[link text][link label]`.

The image and link renderers were updated to handle this new node structure. The image renderer was also updated to not fail when alt text is missing, i.e., ``.

## Additional information

This PR requires that #5955 is merged first, since it depends on the APIs in that version of `lezer/markdown`.

Tested on:  macOS 26

PR #5982, opened 2025-10-31
Refactor sloppy parsing 2.0

### LINKED ISSUE #5959: Markdown badge links are not rendered correctly
### Description

First of all, thank you for your work on Zettlr — it’s a great tool for writing in Markdown.

I noticed that badge images display correctly, but **adding a link around the badge breaks the rendering**.
Here’s a minimal example to reproduce the issue:

```markdown
<!-- ✅ Works -->
![Updated](https://img.shields.io/badge/Updated-13/10/2025-darkgreen?)

<!-- ❌ Doesn't work -->
[![Updated](https://img.shields.io/badge/Updated-13/10/2025-darkgreen?)](https://pandoc.org/)
```

Expected behavior:
The second example should render as a clickable badge that links to `https://pandoc.org/`.

Actual behavior:
The badge is not rendered correctly when wrapped in a link.

### Reproducing

**Steps to reproduce:**

1. Open a Markdown file in Zettlr.
2. Paste the code above.

```markdown
[![Updated](https://img.shields.io/badge/Updated-13/10/2025-darkgreen?)](https://pandoc.org/)
```
3. View the preview.

### Zettlr version

3.6.0

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Ubuntu 22.04

### Additional Information

_No response_

benniekiss:
## Description

This PR refactors the sloppy link parser to use delimiters.

Closes #5959 (again)

## Changes

See #5964 for additional information.

The `sloppy-parser` was renamed to `pandoc-link-parser` for clarity. This is because the main difference is to implement spaces in link URLs, which is supported by pandoc. 

The original PR was reverted due to conflicts with `Autolink` and `Link` parsing. How this PR differs is that instead of returning the open delimiter end positions after parsing, we return `-1` so that the default link parser can also add delimiters. 

This is necessary for two reasons. The first is so that we do not have to completely re-implement link parsing, including bare brackets, `[my link]`, or labeled links, `[my link][my label]`. The other reason is to work around a check in the `Autolink` parser. Otherwise, link text which ends in a valid url containing path elements will not parse the closing bracket correctly. For example, `[Example: www.example.com/my-path]` will not parse as a link because `www.example.com/my-path]` is parsed as an `Autolink`. 

This last instance is because the `Autolink` parser has a [check to ignore brackets](https://github.com/lezer-parser/markdown/blob/63872036b2a41f0bc916f7d60e33e32c6e62b835/src/extension.ts#L219-L222) only if `cx.hasOpenLink` is `true`, which is only the case if the default link parser adds delimiters.

## Additional information

Tested on:  macOS 26


source/common/modules/markdown-editor/parser/pandoc-link-parser.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (!delim) { return -1 }
+
+    const isLink = delim.to - delim.from === 1
+    let linkContents = ctx.takeContent(opening)
+
+    ctx.addDelimiter(PandocLinkDelimiter, pos, pos + 1, false, true)
+
+    // Remove nested links, which are invalid
+    if (isLink) {
+      const nodeType = ctx.parser.nodeSet.types.find(node => node.is('Link'))?.id
```
benniekiss, line 69: I wonder here if there is an easier way to get the node id number. I previously hard coded its value, but I do not think it is static, so any adjustment of the node definitions would change the number.

PR #5983, opened 2025-11-01
[Fix] Sanitize file titles from H1 headings

### LINKED ISSUE #5966: Links made in first header not populating graph or backlinks.
### Description

In a peculiar turn, none of the internal inks between files are functioning as they should (beyond being clickable links). This means that, as most of my notes consist solely of a title, there are only one or two connections on my graph. This also means that notes which are connected don't show up in "related".

### Reproducing

1. Create a note.
2. Create a header like; # One must beware the [[20251024174808|fallacy of compositon]] when switching from [[20251024162651|microeconomics]] to [[20251024162701|macroeconomics]].
3. Open graph view.
4. None of the notes are connected.
5. Open Related
6. None of the notes are related
7. Click or over on link
8. Link is connected

### Zettlr version

3.6.0

### Installation Method

Flathub

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Zorin OS LTS

### Additional Information

_No response_

### LINKED ISSUE #5980: Unlink internal links in file manager, search, auto-label, etc.
### Description

Currently, in the file manager, searches, auto complete, and the new internal label feature, links and all text are shown entirely un-rendered.

This is what it looks like in my file explorer

Image

and an example note to explain why there might be a leading link in a note, plus the autocomplete bar.

Image
Image

It really messes up the auto label especially

Image

### Proposed Changes

This could be augmented in one of two ways. The first is that the "Titles" should simply be parsed and stripped of internal links, in the same way that "Export/InternalLinks/Unlink" functions. This can either happen right off the bat in the behind the scene title grabbing, or only when these file titles are being displayed. A setting for this would be fine.

As a little side note, stripping them of italics and whatnot as if they were passed through ctrl-shft-v would be nice but a little different.

TLDR: Titles should be stripped of links but leave the link text or label.

### Caveats

This may interfere with someone else's workflow, particularly users who use non UID file names, but as this only removes the brackets and labels it should not affect them at all? 

If there are any complaints, I am glad to have this simply be a setting.

No

### Zettlr Version

Beta (if applicable)

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Zorin OS

### Additional Information

_No response_

benniekiss:
## Description

This PR adds sanitization to H1 headings when they are extracted for file display.

Closes #5980
Closes #5966 

## Changes

Instead of assigning `firstHeading` to the heading node content, the plain text is extracted using `extractTextNodes`. This removes any markup that is in the heading.

The `extractTextNodes` function was refactored to include the `ZettelkastenLink` title, which matches `Link` and `Image` nodes.

## Additional information

Title: 

Screenshot 2025-11-01 at 13 11 28

Current:

Screenshot 2025-11-01 at 13 11 35

PR:

Screenshot 2025-11-01 at 13 05 35

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -65,31 +65,45 @@ export function extractASTNodes (ast: ASTNode, nodeType: ASTNodeType, filter?: (
     return []
   }
 
-  if (ast.type === nodeType) {
-    return [ast]
-  } else if (ast.type === 'FootnoteRef' || ast.type === 'Highlight' || ast.type === 'ListItem' || ast.type === 'Generic' || ast.type === 'Emphasis') {
-    let returnNodes: ASTNode[] = []
-    for (const child of ast.children) {
-      returnNodes = returnNodes.concat(extractASTNodes(child, nodeType, filter))
+  let returnNodes: ASTNode[] = []
+
+  switch (ast.type) {
+    case nodeType: {
+      returnNodes.push(ast)
+      break
     }
-    return returnNodes
-  } else if (ast.type === 'OrderedList' || ast.type === 'BulletList') {
-    let returnNodes: ASTNode[] = []
-    for (const item of ast.items) {
-      returnNodes = returnNodes.concat(extractASTNodes(item, nodeType, filter))
-    }
-    return returnNodes
-  } else if (ast.type === 'Table') {
-    let returnNodes: ASTNode[] = []
-    for (const row of ast.rows) {
-      for (const cell of row.cells) {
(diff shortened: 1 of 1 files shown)


benniekiss:
The latest commit refactors the ast extractors into a switch statement, matching the node parser structure and making it much more readable in my opinion. 

I also believe I was able to fix #5966. All that was missing was the `Header` node in the children extractor guard.

richdouglasevans:
Conan, what is best in life? 🤔

> "To crush your enemies, see them driven before you, and to see another zarjaz PR from BennieKiss!"

I am McLovin' your work Bennie, rock on! ✌️ LGTM.

PR #5986, opened 2025-11-04
[Performance] Refactor citation parser to use sets

benniekiss:
## Description

This PR improves parsing performance by refactoring the citation parser to use sets rather than array lookups.

## Changes

The locator label logic was refactored to use set lookups. Two sets are maintained, `sanitizedLocatorLabels`, which converts each key's label values into a set, and `allValidLocatorLabels`, which adds all labels into a flat set.

The locator validation logic was refactored to make use of the sets. It now:
1. slices the string to the first space character (ensuring that the locator ends in a space)
2. Lowercases only the sliced section
3. Checks the sets using the lowercased text

## Additional information

In my small benchmarking tests, this doubles citation parsing performance! I tested using the test citation rendering file, since I don't regularly use citations.

Tested on:  macOS 26


nathanlesage:
Looks good! And if you already noticed a performance improvement with two citation items, it will probably stack up for documents with a few dozens. Thank you!

PR #5994, opened 2025-11-05
[Refactor] Move FSAL cache to async

benniekiss:
## Description

This PR moves the FSAL cache code to `async` functions.

## Changes

The FSAL cache code was migrated to use `async` functions.

A couple of external functions which use the FSAL cache were updated to match the new async pattern.

## Additional information

The main motivation was to take advantage of async `readFile` and `lstat`. In my tests loading a large workspace, the time went from ~~89 seconds to 3 seconds to load.~~

EDIT: The claim in speed improvements may be erroneous, as after a few restarts on the develop branch, I no longer experienced the same long load times that led me down this path. Both implementations provide approximately the same performance, in my small amount of testing.

Async is not my strong suit, so I may have missed something in the port!

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/fsal/index.ts
+++ b/source/app/service-providers/fsal/index.ts
@@ -157,7 +157,7 @@ export default class FSAL extends ProviderContract {
    * @param   {EventName}  event    The event name
    * @param   {string}     absPath  The absolute path for this event
    */
-  private async emitChokidarEvent (event: EventName, absPath: string): Promise<void> {
+  private emitChokidarEvent (event: EventName, absPath: string): void {
     if (event === 'all' || event === 'raw') {
       return this._logger.error('[FSAL] Cannot emit events "all" or "raw" -- wrong chokidar setup!')
     }
@@ -171,7 +171,8 @@ export default class FSAL extends ProviderContract {
     }
 
     // Regardless of the event, it will invalidate that particular cache entry.
-    await this._cache.del(absPath)
+    this._cache.del(absPath)
+      .catch(err => this._logger.error(`[FSAL Cache] Failed to delete key: ${absPath}`, err))
 
     // In unlink-events, there won't be a descriptor.
     if (event === 'unlink' || event === 'unlinkDir') {
@@ -214,8 +215,8 @@ export default class FSAL extends ProviderContract {
         } else {
           // Start watching the root path.
           const watcher = new FSALWatchdog(this._logger, this._config)
-          watcher.on('change', async (event, absPath) => {
-            await this.emitChokidarEvent(event, absPath)
+          watcher.on('change', (event, absPath) => {
+            this.emitChokidarEvent(event, absPath)
           })
           watcher.watchPath(rootPath)
           this.watchers.set(rootPath, watcher)


source/app/service-providers/fsal/fsal-cache.ts
nathanlesage, line 25: Do we need the `fs` object? Or can we access everything from the promises API? If so, we could move to the `import { promises as fs } from 'fs'` pattern that we're using in other places of the app as well. If not, please disregard

source/app/service-providers/fsal/fsal-code-file.ts
```diff
@@ -39,9 +39,8 @@ function applyCache (cachedFile: CodeFileDescriptor, origFile: CodeFileDescripto
  * @param {CodeFileDescriptor} origFile The file to cache
  */
 function cacheFile (origFile: CodeFileDescriptor, cacheAdapter: FSALCache): void {
-  if (!cacheAdapter.set(origFile.path, structuredClone(origFile))) {
-    throw new Error(`Could not cache file ${origFile.name}!`)
-  }
+  cacheAdapter.set(origFile.path, structuredClone(origFile))
+    .catch(() => { throw new Error(`Could not cache file ${origFile.name}!`) })
```
nathanlesage: This is an unhandled rejection in the making. Could you trace the `cacheFile` calls to see if we can move that to async, too? I'm pretty sure it's only being called from within async functions, too, so that should be a quick fix. (Same applies to the other instances)

source/app/service-providers/fsal/index.ts
```diff
@@ -171,7 +171,7 @@ export default class FSAL extends ProviderContract {
     }
 
     // Regardless of the event, it will invalidate that particular cache entry.
-    this._cache.del(absPath)
+    this._cache.del(absPath).catch(err => console.error(err))
```
nathanlesage: This should be a `log` call, not a console log. We'd need to be able to track them in the logging.

benniekiss: Can we move the `emitChokidarEvent` function to promises? It would remove the need for the `catch()`.

 It is only called in these lines, and I believe the `watch` callback can be made `async`:

https://github.com/Zettlr/Zettlr/blob/13a9c2eb8377855386495cea09ac313347c269ae/source/app/service-providers/fsal/index.ts#L215-L221

source/app/service-providers/fsal/fsal-cache.ts
benniekiss: It's definitely possible, but its a little more of a refactor. I'll push a separate commit with those changes: https://github.com/Zettlr/Zettlr/pull/5994/commits/68f94aed7220788e34f09768231b988f4566b56d

benniekiss: I moved the `_datadir` existence check to the `persist` method because the async calls won't work in the constructor, and it makes more sense to me to check the dir each time we try to save the cache

source/app/service-providers/fsal/fsal-code-file.ts
benniekiss: Easy enough! I handled the `fsal-file` and `fsal-code` instances, which I believe is all there is.

source/app/service-providers/fsal/index.ts
nathanlesage: What does this function have to do with the async of the caching…? I don't see the connection right now 👀

source/app/service-providers/fsal/index.ts
```diff
@@ -214,8 +214,8 @@ export default class FSAL extends ProviderContract {
         } else {
           // Start watching the root path.
           const watcher = new FSALWatchdog(this._logger, this._config)
-          watcher.on('change', (event, absPath) => {
-            this.emitChokidarEvent(event, absPath)
+          watcher.on('change', async (event, absPath) => {
```
nathanlesage: This is another unhandled rejection in the making, since the `on` method of the FSALWatchdog unfortunately is not an EventEmitter `on`. Furthermore, I just doublechecked, and event emitters do not allow async callbacks, so while you can give them a function that returns a promise, that promise will get thrown into the ether if there is any rejection happening.

My suggestion would really be to run the `emitChokidarEvent` synchronously, and if there's an exception, log it in a `catch` branch. Like so:

```ts
this._cache.del(absPath)
  .catch(err => this._logger.error('Describe the error', err))
```

I also think that this makes more sense, simply because this handler should never break anything. It needs to run, and if there is any error, it's better to just log it rather than hoping that it gets caught somewhere towards the top of the call chain.

nathanlesage (commented):
I've now seen what you mean with making the `emitChokidarEvent` function async, but unfortunately that won't work as expected :/

I think, simply keeping the `catch`-handlers and logging any errors is the most suitable approach here.

source/app/service-providers/fsal/index.ts
benniekiss: Thanks for explaining that! I'll revert the change.

benniekiss:
This is good to go on my end!

nathanlesage:
Looks good! Let's go with it!

Issue #4638, opened 2023-09-11
Allow subtags seperated by /


trutzig89182:
### Checklist

- [X] Zettlr does not possess this feature (if it does, but the feature doesn't work out for you, propose an enhancement instead.)
- [X] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)
- [X] I have searched the documentation to make sure I did not overlook it.
- [X] I have searched both open and closed issues to see if someone has already proposed this.

### Feature Proposal

It would be helpful if Zettlr would allow nested tags like f.i.: #organization/space or #rationality/economic

This allows to define domains to better structure knowledge. Also it can help to deal with concepts that are markedly different in different contexts. For instance, it might make not a lot of sense to relate the Club of Rome report on the limits of growth to the same category "environment" as thoughts about organizations in their societal environment from a system theoretical perspective. This is an issue that can be better addressed if nested tags are available. 

### Example Workflow

If you write a note and find that a tag you want to assign to a note is taken by the word in a context which does not fit with what you are writing about, you might consider giving some extra context that better delineates the tag, by nesting the tag.
Also, if you dig deeper into one area of topics, it might be helpful to group your tags together. Otherwise the tag list can become very long and it can become easy to loose track of all the related tags.
Organisational_Functionalism
Networkorganisations
Organisation_as_prison
organisational_Hiearchy
organisational_Architecture 
…

### Caveats

Of course it can be confusing to have the tags "space" in a domain "organisation" and for instance "physics". This requires some extra effort to keep the tags organised. Also, overusing it, would make the tags more or less useless, as they are supposed to bring different ideas together. 
However, these are mainly questions of how to productively use nested tags.

### Do you wish to attempt implementing this yourself?

No

### Platforms

- [X] Windows
- [X] macOS
- [X] Linux

ghost:
Bumping

This functionality is essentially implemented, we just need graphical and autocomplete enhancements.

benniekiss:
I'm curious about implementing this since I'm going through each of the internal syntax parsers, but I wanted a little more clarity about the design goal. 

So is the goal to have `#organization/space` end up as two tags, `organization` and `space`? Essentially a shorthand for `#organization #space`? Or should there be more structure involved?

EDIT: Another structure could be that for a tag of `#organization/space/group`, tags would be created for each item in the tree, so `#organization/space/group`, `#organization/space`, and `#organization`

ghost:
[...]

PR #5997, opened 2025-11-08
Refactor ZknLink and ZknTag parsing

benniekiss:
## Description

This PR refactors the ZknLink and ZknTag parsers and node structure, and it implements namespaced subtags, i.e. `#org/group` creates `#org` and `#org/group` tags

## Changes

The ZknLink parser was refactored to use delimiters. A `ZknLinkMark` node was added and applied to the opening and closing brackets. 

The ZknTag parser was refactored. The `ZknTagContent` node was replaced with a `ZknTagMark` node to better align with the semantics of other nodes. `allowedCharsBefore` was refactored to a regex test, and `tagRe` was refactored to a simpler pattern, which I believe still captures the goal of the original. The parser was also refactored to support nested sub tags, as described below.

~~The Zettelkasten tags now support nested namespaces, as described in #4638. I went forward with this implementation because the diff to implement it was not that large. However, this would be a large change in the syntax and functionality of tags. For that reason, all of the implementation is contained within 26d4be0507a9e208ef98da831f3dd54a9e1a7255, so this commit can be dropped if so desired. It also means that the diff for that commit shows all that's required in addition to the base commit.~~ separated out 

Finally, a few places throughout the code were updated to match the new link and tag structure. ~~I also organized and deduplicated the list of `EMPTY_NODES` in `parse-children.ts`~~

## Additional information

Tested on:  macOS 26


nathanlesage:
Great Initiative, as always! But I'd like to let this one simmer a bit, primarily because I'm worried what might break with this refactor. Also, the 4.0 changelog is ridiculously long already so it wouldn't hurt to push this one a bit further down the line :)

benniekiss:
Totally fine! This is mostly for completeness since these are the last two parsers to be refreshed.

nathanlesage:
Since you keep this updated so diligently, would you mind separating the refactoring from the sub tag-support? I don't have the headspace for thinking about sub tags right now, and that way we can already merge your improvements beforehand!

benniekiss:
Definitely. I'll keep this up without the subtags

nathanlesage:
Thank you so much!!

Issue #5826, opened 2025-08-06
Citations are replaced by their rendered versions on overwriting selections


nathanlesage:
### Description

I just found a very curious bug in the current nightly release while writing my background section. If I select a part of a paragraph which I wish to delete, and then start typing, Zettlr will replace all citations after the selection end within that paragraph with their rendered counterparts. This does not happen if I delete the section first, and only then start typing (i.e., not "overwriting" the selection).

### Reproducing

Before (includes the selection):

<img width="1513" height="230" alt="Image" src="https://github.com/user-attachments/assets/f29ec460-799f-44f8-a1ea-5a3fae2888c7" />

After (typing a single space, which replaces the selection):

<img width="1506" height="161" alt="Image" src="https://github.com/user-attachments/assets/84e3abef-4c9e-40a4-a605-83d8230fec51" />

Note how the two citations at the end of the paragraph suddenly turn into their rendered counterparts. Indeed, it is literal text, and not a simple visual bug.

### Zettlr version

3.6.0-nightly (August 2026)

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [x] ARM 64bit (includes Apple Silicon)

### Operating System Version

macOS Sonoma 15.6

### Additional Information

I have absolutely no idea what is going on here.

LunaciaDev:
This bug is affecting all markup that remove its markup in the rendered version, not just Citation, with same reproduction step.

https://github.com/user-attachments/assets/a7bfa502-67b1-45cb-a4ea-3efa8b08af9c

nathanlesage:
This is super weird; I have no idea what might have caused that ...

LunaciaDev:
The commit [`e1a9e8e`](https://github.com/Zettlr/Zettlr/commit/e1a9e8ee6266d37cad60f9b3c6bc524efd2a44b8) introduced this issue. Reverting that commit and running `yarn install` seems to resolve it.

benniekiss:
I bet this is solved by https://github.com/codemirror/dev/issues/1630

nathanlesage:
Ahh wonderful — would you mind keeping an eye for the next view release? I've noticed the patch is not yet released, just so that we can hopefully ship that one also with 4.0 :)

PR #6001, opened 2025-11-09
[Fix] Pilcrow cleanup

benniekiss:
## Description

This PR cleans up the pilcrow rendering, ~~fixing an issue where a literal pilcrow would be inserted into the text when selecting and deleting multiple lines.~~

## Changes

~~Instead of applying the pilcrow character via `textContent`, it is applied via the CSS `content` parameter. This prevents it from being inserted into the actual text body when multiple lines are selected and then deleted. I am not sure why this is happening, but with this fix, I am unable to reproduce the problem.~~

`span.setAttribute('aria-hidden', 'true')` was added to the widget to not interrupt screen readers.

The `side` paremeter of the pilcrow widget was increased to `10000`, the max value, so that it appears after any other widget that may appear at the end of the line. I noticed this when using snippets.

Pilcrows are now only added after lines in regular text, ignoring items such as code blocks.

## Additional information

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
+++ b/source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
@@ -41,6 +41,7 @@ class PilcrowWidget extends WidgetType {
   toDOM () {
     const span = document.createElement('span')
     span.className = 'cm-pilcrow'
+    span.textContent = '¶'
     span.setAttribute('aria-hidden', 'true')
 
     return span
@@ -99,8 +100,7 @@ const pilcrowPlugin = ViewPlugin.fromClass(class {
 })
 
 const pilcrowTheme = EditorView.baseTheme({
-  '.cm-pilcrow::after': {
-    content: '"¶"',
+  '.cm-pilcrow': {
     color: '#aaa',
     opacity: '0.5'
   }


nathanlesage:
Do you think this is related to https://github.com/Zettlr/Zettlr/issues/5826

?

benniekiss:
Definitely the same issue. Maybe a regression in `codemirror/view` -- seems it was fixed in an unreleased version. I'll revert the CSS `content` changes, but I'll keep the others.

nathanlesage:
> seems it was fixed in an unreleased version.

Do you have a pointer for me? I have been wondering if this was my mistake, or if it was an issue in CodeMirror? If the latter, it would be nice because purely logical it makes more sense to have text content as actual text (or html) content rather than move stuff into CSS properties, also for a11y reasons.

source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
```diff
@@ -49,16 +52,30 @@ class PilcrowWidget extends WidgetType {
...
-      builder.add(line.to, line.to, pilcrowDeco)
-      pos = line.to + 1
-    }
+    syntaxTree(view.state).iterate({
+      from,
+      to,
+      enter (node) {
+
+        // Don't show pilcrows in code text
+        if (node.name === 'CodeText') { return false }
```
nathanlesage, line 67: Interesting decision. Is this how other apps also work..? Or would we be the trend setters here? I mean I can see the logic, just for clarification.

benniekiss:
> > seems it was fixed in an unreleased version.
> 
> Do you have a pointer for me? I have been wondering if this was my mistake, or if it was an issue in CodeMirror? If the latter, it would be nice because purely logical it makes more sense to have text content as actual text (or html) content rather than move stuff into CSS properties, also for a11y reasons. 

See my comment on the other issue! I think the upstream issue and the associated patch addresses it.

source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
benniekiss: I went this route just from thinking how Zettlr handles code text in general. It felt like a more internally consistent approach. Also, when looking at code blocks, it just kinda felt ... off to see pilcrows within them. I did not look at other editors, but I will!

nathanlesage: Thanks :) Essentially, I'm just wondering what would break immersion more — having pilcrows everywhere or except code blocks. (Also: We may want to include comments in whitespace rendering, since I frequently use comments to comment out entire paragraphs…).

You know what, let me ask on Discord what others think.

benniekiss: Good call! I did try checking obsidian and typora, but neither actually have the ability to show pilcrows, or at least I did not find the option.

nathanlesage: Hah, peasants! (/s)

nathanlesage:
Sorry for the delay, it's quite busy here atm ... merging!

PR #6009, opened 2025-11-12
[Fix] Sanitize existing tags to lowercase in `replaceTags`

### LINKED ISSUE #6008: Tag manager rename tag function broken (DANGER)
### Description

The tag manager has an option to rename tag. While this seems to work, it is only within the tag manager that the change is applied. 
Actually the tags are no changed in the files. 
So once you do this, it is impossible to locate the new or old tags. the new  because it doesnt actually exist, the old because you didnt have the foresight to write it down and trusted a button and text entry for Rename tag would actually rename tag. 
After deleting every conceivable json file and cache i could, nothing resolved the issue. i have had to delete my entire config folder to reset the tags to their originals.
Either fix this bug so that it works or remove the option, so people like me dont spend 2 hours trying to find imaginery tags and then have to reset their system

### Reproducing

go to the tag manager - use the rename function built in (make it something unique so that it isnt shared with an existing tag) - do a search for the new renamed tag. you can search using zettlr, obsidian, vs code or terminal. you wont find it becuase it doesnt exist. the tags have not actually been renamed but tag manager will never show the real tag again, just the false renamed tag Version 3.6.0 (3.6.0) mac

### Zettlr version

Version 3.6.0 (3.6.0) mac

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit

### Operating System Version

macos 15.7.1 (24G231)

### Additional Information

_No response_

benniekiss:
## Description

This PR fixes support for non-lowercased tags

Closes #6008

## Changes

All tags are lowercased when comparing in the `replaceTags` function

Additional tests were added for case-insensitive replacement

## Additional information

Tested on:

### DIFF
--- a/source/common/util/replace-tags.ts
+++ b/source/common/util/replace-tags.ts
@@ -28,6 +28,8 @@ import type { ZettelkastenTag } from '@common/modules/markdown-utils/markdown-as
  * @return  {string}                  The new document
  */
 export default function replaceTags (markdown: string, oldTag: string, newTag: string): string {
+  const sanitizedOldTag = oldTag.toLowerCase()
+
   // Since tags can occur in a frontmatter as well as in the document text, we
   // have the following combinations of possibilities:
   // 1. If the oldTag contains spaces --> Can only be in the frontmatter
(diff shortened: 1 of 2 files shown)


nathanlesage:
Would a fix for that issue not rather involve case insensitive search and replace ...?

There's a unit test for the tag replacer function, but I believe I only used lowercase tags...

nathanlesage:
Oh no, I was wrong during tag rename the tag is exactly like that. The question is: do we want to make tags case sensitive? I'd argue that introduces more errors than it fixes.

benniekiss:
I get what you mean. I think the replace function could probably be refactored instead to case-insensitive replacement if you want to keep it case insensitive.

PR #6010, opened 2025-11-12
[Fix] Update rendering mode immediately on statusbar change

### LINKED ISSUE #6007: Minor bug with switch Raw/Preview button
### Description

Image

When you click the Switch between Raw and Preview button, the effect is not immediate.
You must click on the text to apply the change.

### Reproducing

The bug seems perfectly reproducible.
As you can see in the image, I am in Preview mode, but the # symbol is clearly visible.
If I click on the text, the rendering is applied.

### Zettlr version

Zettlr 4.0.0-beta.2

### Installation Method

From the Website or GitHub

deb package

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Ubuntu 22.04

### Additional Information

_No response_

benniekiss:
## Description

This PR reconfigures the renderer handling so that changing the rendering mode from the statusbar immediately updates the editor.

Closes #6007

## Changes

Functionally, the `renderingModeToggle` was updated to use `dispatch` events to update the config rather than `window.config.set`, and an empty dispatch was added so that the changed mode is applied without further interaction from the user.

Additionally, the code was refactored for maintainability, notably moving the configuration of renderers to its own function.

## Additional information

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/renderers/index.ts
+++ b/source/common/modules/markdown-editor/renderers/index.ts
@@ -24,7 +24,7 @@ import { renderMermaid } from './render-mermaid'
 import { renderTables } from '../table-editor'
 import { renderIframes } from './render-iframes'
 import { renderEmphasis } from './render-emphasis'
-import { configField, type EditorConfiguration } from '../util/configuration'
+import { configField, configUpdateEffect, type EditorConfiguration } from '../util/configuration'
 import type { EditorView } from '@codemirror/view'
 import { hasMarkdownExt } from 'source/common/util/file-extention-checks'
 import { trans } from 'source/common/i18n-renderer'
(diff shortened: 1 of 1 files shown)


source/common/modules/markdown-editor/renderers/index.ts
```diff
@@ -138,7 +103,13 @@ export function renderingModeToggle (state: EditorState, _view: EditorView): Sta
...
     title: trans('Enable or disable the preview mode for Markdown files by clicking'),
     onClick () {
-      window.config.set('display.renderingMode', config.renderingMode === 'preview' ? 'raw' : 'preview')
+      const mode = config.renderingMode === 'preview' ? 'raw' : 'preview'
+      view.dispatch({ effects: configUpdateEffect.of({ renderingMode: mode }) })
+      // We dispatch an empty transaction here so that the renderer
+      // update takes effect immediately when clicking. Otherwise, it would
+      // take an additional transaction, i.e. user interaction,
+      // for the renderer changes to take effect.
+      view.dispatch({})
```
nathanlesage: I believe we can get rid of this, because if I'm not mistaken, in the current state of this PR, the global config will NOT be affected by this, correct? Because if you do set the global state (via window.config.set) this will indirectly trigger an update because the config will be updated after a round trip.

That is the reason why a single click in the settings window will automatically update the rendering state.

nathanlesage (commented):
I had a question, but other than that it looks good to me!

source/common/modules/markdown-editor/renderers/index.ts
benniekiss: Hmm I tested without and it did not update the editor view--it still required the user to click somewhere in the editor. I also dont believe changes in the settings pane are immediately applied, at least from what I recall from my tests.

Maybe there's a different root cause then?

nathanlesage: I just tried it a few times; something fishy is going on.

So, setting the global config using window.config is definitely non-negotiable because otherwise Zettlr won't remember the setting anyways.

But when it comes to updating: It turns out that CodeMirror only sometimes immediately updates. The thing is the following:

- Set the rendering mode in the global config
- Global config receives the request, updates, and emits an update event
- The Pinia store receives the request, and updates its config
- Vue's reactivity propagates the change into the MainEditor component
- The MainEditor sets the config changes in the MarkdownEditor class
- The config change in the MarkdownEditor class is applied with a transaction effect, which is then picked up, among other things, by this transaction extender.

Which means that it could be that the logic I use in the MarkdownEditor to apply the config may be broken somehow. I've had this issue once already. Would you mind having a look at the config updating logic?

benniekiss: definitely! Ill revert the `window.config` change here as well

benniekiss: Okay, so I think I figured it out --- the transaction extender was not tracking `configUpdateEffect`, so in essence, it was always one transaction behind. In essence, it would go:

1. Config update transaction occurs
2. Extender updates the `renderCompartment` using the existing config
3. transaction goes through, config is updated

With the new refactor, it tracks `configUpdateEffect`, and it falls back to the old config values. So it goes:

1. Config update transaction occurs
2. Extender updates the `renderCompartment` using the values from `configUpdateEffect` and falls back to the previous config
3.  transaction goes through, config and renderers are updated

Please take a look at the new handling! I've tested it, and renderer settings take effect immediately, both in the statusbar and in the settings window.

nathanlesage:
Looks good now! I've given it a good look, and it seems a very good solution! Thank you! :)

PR #6012, opened 2025-11-13
[Fix] Fix remaining type warnings

benniekiss:
## Description

This PR fixes all of the remaining type warnings.

## Changes

A lot of type casting, mostly `as string`.

The `NodeError` type was exported.

## Additional information

There are no more type warnings or lint errors :)

Tested on:  macOS 26


nathanlesage:
Uh nice! I've been wanting to do that ever since the last big migration that caused all of those … five years ago :D 

Thank you!

PR #6025, opened 2025-11-24
[Fix] Use inline renderer for images

benniekiss:
## Description

This PR changes the image renderer from using a block renderer to an inline renderer

## Changes

In cleaning up the pilcrow handling, I noticed that lines with rendered images were showing two pilcrows. The reason is because the images were rendered as block widgets, but I believe they should be rendered as inline widgets given their position in the syntax structure. 

I've tested this and did not notice any changes in general behavior, other than the pilcrows now showing the correct number.

I've also corrected a small TODO to fix an `any` type

## Additional information

Tested on:  Fedora 43


nathanlesage:
This makes sense, given that images are inline nodes, and are only treated as "implicit" figures by Pandoc. Never noticed that; this should also make image handling actually inline. Thank you!

PR #6026, opened 2025-11-24
allow passing `Tree` as `null` to `markdownToAST`

benniekiss:
## Description

This PR refactors the `markdownToAST` function to accept `null` as the `Tree` parameter. This matches the output of several tree-getting methods and functions

## Changes

Several calls to `markdownToAST` were updated to reuse the tree, if available.

The `EMPTY_NODES` list was organized and changed to a `Set`

## Additional information

Tested on:  Fedora 43

### DIFF
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -35,7 +35,7 @@ export { md2html } from './markdown-to-html'
  *
  * @return  {ASTNode}            The root node of the AST
  */
-export function markdownToAST (markdown: string, tree: Tree|null, parserConfig?: MarkdownParserConfig): Document|ASTNode {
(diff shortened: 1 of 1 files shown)


nathanlesage:
Can't we just refactor the method to `tree: Tree|null` and pass `null` explicitly in the cases where we don't want this? Given how JavaScript is, I'm always uneasy if we do a falsy-check with possible null/undefined variables (so `tree === null` is preferable to `!tree`). I wonder why the linter didn't complain about this, since I always believed explicit checks are appropriate …

benniekiss:
That makes sense to me! I'm not sure why the linter didn't complain either, maybe because `Tree` is always truthy?

nathanlesage:
> That makes sense to me! I'm not sure why the linter didn't complain either, maybe because `Tree` is always truthy? 

Probably because the rule originally was inside one of the standard sets that we use, and they dropped it at some point. We can always manually re-add it!

PR #6030, opened 2025-11-29
[Fix] Improve math parsing validation

benniekiss:
## Description

This PR fixes two issues with inline single-delimiter math parsing I experienced while drafting a snippet in the main editor:

1. When a line contains a dollar sign, and the next line starts with a dollar sign, the content between is interpreted as math text.

2. When the closing delimiter is followed by a digit, the content is still interpreted as math despite the delimiter being invalid, according to pandoc docs.

It also better aligns the block math parser with pandoc so that blank lines are no longer valid within math blocks.

## Changes

The whitespace tests were moved so that the tests are only performed when needed. This reduces unnecessary calculations.

The following changes are from what I understand of the [pandoc docs on the delimiters](https://pandoc.org/MANUAL.html#extension-tex_math_dollars): 

The whitespace checks were refactored to a regex test for any whitespace character using `\s`. I changed this because I interpreted the pandoc docs, `must have a non-space character`, to mean any whitespace.

An additional test was added for closing delimiters to ensure they are not followed by a digit.

The whitespace and following-digit tests are ignored if the delimiter is for display math, `$$`, which do not have limitations on preceding or following characters.

The math block parser had an additional check added to make sure that no lines within the block are blank.

## Additional information

 

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/parser/math-parser.ts
+++ b/source/common/modules/markdown-editor/parser/math-parser.ts
@@ -66,7 +66,7 @@ export const inlineMathParser: InlineParser = {
     if (!isInlineDisplayMath && (/\s/.test(ctx.slice(pos - 1, pos)) || /\d/.test(ctx.slice(pos + 1, pos + 2)))) {
       // However, if this is an invalid closing delimiter, we need to check if
       // it would be a valid  opening delimiter and add it to the tree if it is.
-      const invalidOpening = !isInlineDisplayMath && /\s/.test(ctx.slice(pos + 1, pos + 2))
+      const invalidOpening = /\s/.test(ctx.slice(pos + 1, pos + 2))
 
       // Either return -1 due to an invalid delimiter, or return the end position of the delimiter
       return  invalidOpening ? -1 : ctx.addDelimiter(MathDelimiter, pos, pos + delimLength, true, false)


benniekiss:
After doing some tests with pandoc, there's some inconsistent behavior. Newlines after the opening delimiter invalidate the inline math, but newlines before the closing delimiter do not. For example:

Invalid:

```
$
x^2+y*(3+2)=z^2$
```

Valid: 

```
$x^2+y*(3+2)=z^2
$
```

Maybe this is a pandoc bug? I feel like the newline before the closing delimiter should also be invalid.

nathanlesage:
@tarleb Do you have more insight on this behavior? I think it makes sense to align the Zettlr parser as much as possible to the Pandoc behavior.

tarleb:
Huh, interesting. That's probably a bug, as newlines before the closing `$` should be forbidden, too.

benniekiss:
FWIW I opened an issue here https://github.com/jgm/pandoc/issues/11311

benniekiss:
The pandoc issue has been cleared up! This implementation now matches the pandoc output (once a new version is released)

source/common/modules/markdown-editor/parser/math-parser.ts
```diff
@@ -36,32 +39,32 @@ export const inlineMathParser: InlineParser = {
...
     const opening = ctx.findOpeningDelimiter(MathDelimiter)
 
     // Since there was no opening delimiter, this is a potential opening
-    if (opening === null && followedBySpace) {
-      return -1
-    } else if (opening === null) {
-      return ctx.addDelimiter(MathDelimiter, pos, pos + delimLength, true, false)
+    if (opening === null) {
+      // Inline opening delimiters cannot be followed by a space, but display math delimiters can
+      const invalidOpening = !isInlineDisplayMath && /\s/.test(ctx.slice(pos + 1, pos + 2))
```
nathanlesage, line 49: This would not work properly if at the end of a line, would it? Also, what about `some text $$asd$$`? Wouldn't that return false, too?

source/common/modules/markdown-editor/parser/math-parser.ts
```diff
@@ -93,7 +95,11 @@ export const blockMathParser: BlockParser = {
     const blockStart = ctx.lineStart
     const from = ctx.lineStart + line.text.length + 1
 
-    while (ctx.nextLine() && !blockMathRE.test(line.text)) {
+    while (ctx.nextLine()) {
```
nathanlesage, line 103: Why move the conditions into the loop? Seems legit, however.

nathanlesage (commented):
Just had a question, because the logic you've used seems a bit more involved, and just want to make sure I understand it!

nathanlesage:
Found a discrepancy between Pandoc and Zettlr:

`This is some test $x=1$3. And another test $y=2$.` detects as math from the first to the last dollar sign, showing an error. Pandoc, on the other hand, fully disregards both the first dollar signs, and properly renders the second pair.

source/common/modules/markdown-editor/parser/math-parser.ts
benniekiss: So if the check is put in the `while` condition, it becomes `while (ctx.nextLine() && !blankLineRe.test(line.text) && !blockMathRE.test(line.text))`, and this is a little more difficult to wrap my head around given the negations and multiple required conditions. Moving it into the loop was clearer to me.

source/common/modules/markdown-editor/parser/math-parser.ts
benniekiss: Hm, let me test this again. I think I tested inline display math, and it rendered. 

When I wrote this condition, the logic was that if marker is display math, the first condition will be false, so `invalidOpening` is now false, i.e. a valid opening

benniekiss: Can confirm `some text $$asd$$` renders `asd` as display math

benniekiss:
> Found a discrepancy between Pandoc and Zettlr:
> 
> `This is some test $x=1$3. And another test $y=2$.` detects as math from the first to the last dollar sign, showing an error. Pandoc, on the other hand, fully disregards both the first dollar signs, and properly renders the second pair.

The same happens with spaces around the delimiters. I think handling this case will actually make the parsing simpler, but it's going to be a significant refactor. I have an idea though and should have something up in a bit

source/common/modules/markdown-editor/parser/math-parser.ts
nathanlesage: Roger!

benniekiss:
Ok, I think the solution was actually much simpler, no refactoring needed. This handles the edge case you presented. I think this is good to merge on my end,

PR #6031, opened 2025-11-30
[Feat] Add pandoc div and span parsing and rendering -- Part 1

benniekiss:
## Description

This PR implements a syntax parser and renderer for pandoc fenced divs and bracketed spans.

See #5848 for more information.

## Additional information

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
+++ b/source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
@@ -7,9 +7,9 @@
  * Maintainer:      Bennie Milburn
  * License:         GNU GPL v3
  *
- * Description:     This renderer displays Pandoc divs and spans,
- *                  rendering the attributes defined for the node
- *                  as they would be displayed by pandoc.
+ * Description:     This renderer displays Pandoc divs and spans using
+ *                  Decorations, rendering the attributes defined for
+ *                  the node as they would be displayed by pandoc
  *
  * END HEADER
  */


source/common/modules/markdown-editor/parser/pandoc-div-span-parser.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        Pandoc Span and Div Parser
+ * CVM-Role:        InlineParser, BlockParser
+ * Maintainer:      Hendrik Erz
```
nathanlesage: Add your name here, this is mostly intended to provide some recognition (especially for your hard work!)

source/common/modules/markdown-editor/parser/pandoc-div-span-parser.ts
```diff
@@ -0,0 +1,122 @@
...
+      return -1
+    }
+
+    if (next !== 93) { // 93 === ']'
+      return -1
+    }
+
+    // There are no valid attributes, so return
+    const match = pandocSpanClosingRe.exec(ctx.text.slice(pos - ctx.offset))
+    if (!match?.indices?.groups) { return -1 }
```
nathanlesage: Can you move these ifs across the PR into three lines? Makes them a bit more verbose, but it's always easier to read. I'm also curious why the linter rule didn't catch this that I added due to this …

benniekiss:
Ill get to these fixups in just a bit. But jsyk, I'm still waiting for a fix in `lezer/markdown` to get released so that the print preview can render correctly. Do you want to push these before that gets released, or wait? The issue is that the closing div mark, `:::`, shows up in the print preview

nathanlesage:
Ah, well we can wait. I'm planning on releasing 4.0 on Christmas, if my documentation overhaul works in time, and if the fix doesn't get released until then, we can push it to 4.1. No stress on my part. Better wait and do it right than push half-baked things!

nathanlesage (changes requested):
Sooo, I now finally had some time to look over this PR. I understood it, and it seems to make perfect sense.

Just two small requests:

1. Please add a few more comments to the new parsers, especially towards the start of the block and composite parser I feel it is a bit less intuitive to understand what is going on there.
2. Add a few general comments on what the Div renderer is supposed to render (i.e. how you want those to see if the user activates rendering)

This will make it easier to understand how this works at a future point, and I am pretty confident future us will appreciate it!

Code wise, all looks great!

benniekiss:
I think this should be good to go! I added comments throughout the parser and renderer, and I fixed some spec-compliance stuff along the way--mostly whitespace before div marks, which is not allowed

nathanlesage:
I just discovered that you even apply any arbitrary classes, meaning that in theory users could implement class styling both in HTML export templates as well as the custom CSS. I believe people may enjoy this new capability. I'm just unsure if we can push an entirely new feature before 4.0… mhhhh.

benniekiss:
> as well as the custom CSS

Heh, that was my main motivation. I have several docx custom styles that are mapped to CSS styles, so it helps to be able to see the "live" rendering of what the docx is approx going to look like when using `{custom-style="..."}`

I'm fine waiting until after 4.0, but it would be nice to get a wider test base with the new release. 

We could also add it as an `experimental` option by removing the settings window config option to enable rendering and only allow enabling rendering via manual edits of the config file.

nathanlesage:
hmhm … I'm almost tempted to push this in, because — I keep forgetting it, but there's also the necessity of a horizontal rule parser, and it feels as if it makes sense. Also, I didn't promise this to be the "last" beta, so we technically can continue. I'd want to release on no later than Dec. 26 for historic reasons, but there's still some time to test.

You know what, eff it. Merging as soon as the checks complete.

nathanlesage:
Haha — and I found an issue immediately 😂 I'm sorry!

The following should render (empty div):

image

benniekiss:
I'll look into that!

benniekiss:
Found the issue, pushing a fix shortly

nathanlesage:
lol, I was so close to fixing it myself! Almost detective work, I'm looking forward to seeing how you solved it

PR #6032, opened 2025-11-30
[Feat] Add utilities for pandoc divs and spans -- Part 2

benniekiss:
## Description

This PR adds utilities for pandoc fenced divs and bracketed spans, including formatting bar improvements and a popover utilty.

See #5848 for more information

## Additional information

Tested on:

### DIFF
--- a/test/parse-pandoc-attributes.spec.ts
+++ b/test/parse-pandoc-attributes.spec.ts
@@ -16,9 +16,9 @@ import { deepStrictEqual } from 'assert'
 import { ParsedPandocAttributes, parsePandocAttributes } from 'source/common/pandoc-util/parse-pandoc-attributes'
 
 const tests: Array<{ input: string, output: ParsedPandocAttributes|'logs-error' }> = [
-  { input: 'width=50%', output: 'logs-error' }, // Missing braces. See NOTE below.
+  { input: 'width=50%', output: { properties: { width: '50%' } } }, // Attributes without braces
   { input: '{width=50%}', output: { properties: { width: '50%' } } }, // Simple parsing
-  { input: '{ width = 50% }', output: {} }, // Spaces between = not supported
+  { input: ' width = 50% ', output: {} }, // Spaces between = not supported
   { input: '{ width= 50% }', output: {} }, // Spaces between = not supported
   { input: '{ width =50% }', output: {} }, // Spaces between = not supported
   { input: '{width=10}', output: { properties: { width: '10px' } } }, // Parsing width without unit
@@ -27,7 +27,7 @@ const tests: Array<{ input: string, output: ParsedPandocAttributes|'logs-error' 
   { input: '{key=some long value}', output: { properties: { key: 'some' } } }, // Parsing malformed unquoted values
   {
     // Longer test with all available classes
-    input: '{#some-id .class1 .class2 width=50% height=25 disabled=false style="font-size: 12px;"}',
+    input: '#some-id .class1 .class2 width=50% height=25 disabled=false style="font-size: 12px;"',
     output: {
       id: 'some-id',
       classes: ['class1', 'class2'],
@@ -51,27 +51,8 @@ const tests: Array<{ input: string, output: ParsedPandocAttributes|'logs-error' 
 
 describe('Utility#parsePandocAttributes()', function () {
   for (const test of tests) {
-    // NOTE: Changed on January 10, 2026 by Hendrik Erz. Reason: Parsing link
-    // attributes should not throw errors, since this function is also called in
-    // parsing contexts, and throwing errors would completely abort the parsing,
-    // rendering entirely white documents that cannot be edited. Instead, the
-    // function has been changed to simply return an empty record and log the
-    // error to the console.
-    if (test.output === 'logs-error') {
-      it(`should log an error for string "${test.input}"`, function () {
-        const oldError = console.error
-        let hasLoggedError = false
-        console.error = (...args) => {
-          if (args[0] instanceof Error) hasLoggedError = true
-        }
-        parsePandocAttributes(test.input)
-        deepStrictEqual(hasLoggedError, true)
-        console.error = oldError
-      })
-    } else {
-      it(`should parse the attribute string "${test.input}"`, function () {
-        deepStrictEqual(parsePandocAttributes(test.input), test.output)
-      })
-    }
+    it(`should parse the attribute string "${test.input}"`, function () {
+      deepStrictEqual(parsePandocAttributes(test.input), test.output)
+    })
   }
 })


nathanlesage:
Quick heads up: Going to postpone this to after 4.0!

source/app/service-providers/config/get-config-template.ts
```diff
@@ -226,6 +227,7 @@ export interface ConfigOptions {
     showNewFileButton: boolean
     showPreviousFileButton: boolean
     showNextFileButton: boolean
+    showMarkdownFenceButton: boolean
```
nathanlesage: Should we really call it "fence", given that that is usually referred to for the fenced code blocks?

source/app/service-providers/config/get-config-template.ts
```diff
@@ -341,6 +343,7 @@ export function getConfigTemplate (): ConfigOptions {
       inputMode: 'default', // Can be default, vim, emacs
       boldFormatting: '**', // Can be ** or __
       italicFormatting: '_', // Can be * or _
+      highlightFormatting: 'span', // Can be 'span' or ==
```
nathanlesage: The default should be `==`, given that this is how it has worked for the past years.

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -72,6 +72,45 @@ function insertLinkOrImage (target: EditorView, type: 'link'|'image'): void {
     .catch(err => console.error(err))
 }
 
+/**
+ * Helper to insert pandoc fenced divs
+ *
+ * @param   {EditorView}      target        The target view
+ * @param   {'link'|'image'}  attributes    Attributes to assign to the fenced div
+ */
+function insertPandocFence (target: EditorView, attributes: string): void {
```
nathanlesage: Same as above: Shouldn't we rename this to something more apt, like `pandocDiv`?

In addition: In hindsight, shouldn't we provide a dictionary for the attributes instead of the raw string? The raw string implies that the caller knows that the curly brackets will be assigned by the function. With a dictionary, it would be easier to control what gets fed into this function. In addition, I'd argue we probably will use a function to turn a dict into an attribute string quite often around the code base, wouldn't you agree?

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -72,6 +72,45 @@ function insertLinkOrImage (target: EditorView, type: 'link'|'image'): void {
...
+    const startLine = target.state.doc.lineAt(range.from)
+    const endLine = target.state.doc.lineAt(range.to)
+
+    const changes: ChangeSpec[] = []
+
+    // If the startLine is not empty, insert a newline after the opening mark
+    if (!/^\s*$/.test(startLine.text)) { opening = opening + '\n' }
+
+    // If the previous line is not empty, insert a newline before the opening mark
+    const prevLine = target.state.doc.line(startLine.number - 1)
```
nathanlesage, line 99: Suggestion to move `prevLine` and `nextLine` directly after the `startLine` and `endLine`

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -310,18 +349,42 @@ export function applyComment (target: EditorView): boolean {
...
  * @return  {boolean}             Whether the command was applicable
  */
-export function toggleHighlight (target: EditorView): boolean {
+export function applyHighlight (target: EditorView): boolean {
   if (!viewContainsMarkdown(target)) {
     return false
   }
 
-  applyInlineMarkup(target, '==', '==')
+  const markup: string = target.state.field(configField, false)?.highlightFormatting ?? '=='
```
nathanlesage, line 364: A hint that the highlight formatting should indeed be defaulting to `==` :P

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -502,3 +565,59 @@ export function applyTaskList (target: EditorView): boolean {
...
+/**
+ * Format the inputs into a string to be inserted into a
+ * Pandoc attributes marker, `{#id .classes key=value}`.
+ *
+ * @param     {string}    identifier    The identifier to assign. Whitespace will
+*                                       be replaced by '-', and '#' will be prepended if
+*                                       it is not already
+ * @param     {string}    classes       The classes to include. A '.' will be prepended to
+ *                                      each word, split on whitespace, if it is not already
+ * @param     {string}    attributes    Key=Value attributes to include.
```
nathanlesage: A hint that these attributes really should come in the form of a dictionary.

Also, nota bene: I think that we can add a key `classes` that should accept a list of strings, and a special key `id` that turns its value into `#{value}`. Maybe a custom type, like so:

```ts
interface PandocAttributeDict {
  id?: string
  classes?: string[]
  class?: string // For ease of access
  [key: string]: string // Arbitrary attributes
}
```

This could then, in the future, be more strictly typed for special attributes such as width or height or so.

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -502,3 +565,59 @@ export function applyTaskList (target: EditorView): boolean {
...
+ *
+ * @returns   {string}                  The formatted string
+ */
+export function formatPandocAttributes (identifier: string, classes: string, attributes: string): string {
```
nathanlesage: Above's comment would also make this function less hairy, I'd argue, because while this does look clean, I believe a better way would be to have a strict dictionary data type, and then just a set of functions to convert between that data type and an attribute string.

source/common/modules/markdown-editor/util/configuration.ts
```diff
@@ -122,6 +123,7 @@ export function getDefaultConfig (): EditorConfiguration {
     },
     boldFormatting: '**',
     italicFormatting: '_',
+    highlightFormatting: 'span',
```
nathanlesage: Same as above: Please default to `==`

nathanlesage (changes requested):
Looks good and adds a ton of value, there are just a few minor things to change (plus one potentially larger refactor).

benniekiss:
Its been a minute since I really looked through this, but I agree with pretty much all your comments! 

I'll dedicate some time to this over the next couple days to respond in detail

source/common/pandoc-util/parse-pandoc-attributes.ts
```diff
@@ -25,6 +25,62 @@ export interface ParsedPandocLinkAttributes {
   properties?: Record<string, string>
```
nathanlesage, line 17: Wouldn't it make more sense to add the properties as strings on the object directly, rather than nesting them below `properties`, especially given that width and height are also properties?

source/common/pandoc-util/parse-pandoc-attributes.ts
```diff
@@ -25,6 +25,62 @@ export interface ParsedPandocLinkAttributes {
...
 }
 
+/**
+ * Formats a `PandocAttributes` object into a useable HTML string.
+ *
+ * @param attributes
+ */
+export function formatPandocAttributes (attributes: ParsedPandocAttributes): string {
```
nathanlesage, line 25: Suggestion for this function: Provided you agree with my comment above to move any arbitrary properties to the top-level instead of nesting them under `properties`, couldn't we do it something like this:

```ts
attributes
.entries()
.map((prop, value) => {
  if (prop == 'id') { // Special case handling
    return `#${value}`
  }

  // Default
  return `${prop}="${value}"`
})
.join(' ')
```

source/common/pandoc-util/parse-pandoc-attributes.ts
```diff
@@ -41,9 +97,9 @@ const pandocAttributeRe = /#(?<id>[\w\-_]+)|\.(?<class>[\w\-_]+)|(?<attr>(?<key>
  *
  * @param   {string}  attrString  The attribute string (e.g., `{width=50%}`)
  *
- * @return  {ParsedPandocLinkAttributes}  The parsed string
+ * @return  {ParsedPandocAttributes}  The parsed string
  */
-export function parseLinkAttributes (attrString: string): ParsedPandocLinkAttributes {
+export function parsePandocAttributes (attrString: string): ParsedPandocAttributes {
   attrString = attrString.trim()
   if (!attrString.startsWith('{') || !attrString.endsWith('}')) {
```
nathanlesage: I'm wondering if this check is even necessary at all…? Doesn't the function already kind of presuppose that the string you want to parse is already an attribute string? If we'd not require the string to be surrounded by curly brackets, this might make the logic simpler. Or is there anything that will break down if we remove this check?

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -502,3 +566,34 @@ export function applyTaskList (target: EditorView): boolean {
...
+ * @return  {boolean}                 Whether the command was applicable
+*/
+export function applyPandocDivOrSpan (target: EditorView, type: 'div'|'span', attributes: ParsedPandocAttributes): boolean {
+  if (!viewContainsMarkdown(target)) {
+    return false
+  }
+
+  const attributeString = formatPandocAttributes(attributes)
+
+  switch (type) {
```
nathanlesage: Why not a simple if-else construct here?

nathanlesage (changes requested):
We're getting there. I really like how this is developing! :)

source/common/pandoc-util/parse-pandoc-attributes.ts
benniekiss: Honestly, I think it makes more sense to keep them under the `property` attribute and remove the dedicated `width` and `height` elements. Properties can be arbitrarily named, so keeping them under a parent object seems safer to me.

However, this structure is pretty embedded in the code, so any changes are going to require refactors in several places

source/common/pandoc-util/parse-pandoc-attributes.ts
benniekiss: Its only necessary insofar as other parts of the code use the function expecting the braces to be required, but I'm not against removing the check. It would require changes to some syntax parsers, however

source/common/pandoc-util/parse-pandoc-attributes.ts
nathanlesage: In other words, you're suggesting to only keep attributes that literally require changes (`id=id` --> `#id` and `class=class` --> `.class`) at the top level of the property, and everything else under `properties`?

benniekiss: Yeah, essentially just moving `width` and `height` under `properties`

nathanlesage: Roger, then let's do this for the time being :)

source/common/pandoc-util/parse-pandoc-attributes.ts
benniekiss: Maybe instead of making this a test, we just perform the character stripping, otherwise passing the string?

nathanlesage (approved):
So, now, finally. Looks good!

Issue #6016, opened 2025-11-15
Pop-up list of notes when adding links upon typing [[ does not work


BabakFarshchian:
### Description

Zettlr provides the possibility to write [[ and see a pop-up list of notes to link to. This pop-up windows fails to show up sometime, more often since the beta 4 version.

### Reproducing

Inside a note:
- Type [[ to activate the popup menu with a list of existing notes.
- Sometimes the list does not show up.
- Wait for the note to save, or click ctrl-save.
- Type [[ again, the list now shows up.

### Zettlr version

Version 4 beta 2 latest nightly built

### Installation Method

From the Website or GitHub

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Windows 11

### Additional Information

I am not able to recreate this every time. But I have a feeling that it happens when I do quick edits, especially when working inside lists.

nathanlesage:
Likely a small error in how the database is populated when a document gets opened; I have touched that logic during the recent FSAL rewrite. Should be a hopefully simple fix!

nathanlesage:
This has been fixed in #6039

PR #6039, opened 2025-12-02
[Fix] Update file database in `loadDocument`

benniekiss:
## Description

This PR fixes an issue where autocomplete for zettelkasten links would not show up.

## Changes

A call to `updateFileDatabase` was added in the `loadDocument` method so that the file database is populated when the document is loaded. Otherwise, the user would have to update the `useH1` or `useTitle` config options, or cause an fsal update, such as by saving the document, for the autocomplete to populate. 

## Additional information

Tested on:  macOS 26


benniekiss:
Probably resolves #6016

nathanlesage:
Since all other databases are updated in that call, that's a bullseye bug you found! Thank you!

PR #6041, opened 2025-12-05
[Enhancement] Configure showing hidden files

### LINKED ISSUE #5949: Zettlr Crashes on Start With Symlinks in Directory Tree
### Description

Just had a bug (on the most recent develop branch) where I added a softlink (`ln -s source target`) into the directory tree for Zettlr, and subsequently I couldn't start Zettlr. The start was aborted with an error that the app couldn't load the main window. This is spurious because it was certainly caused by the symlink (it worked again after removing the symlink), but somehow the actual error got swallowed in the boot process.

### Reproducing

Have a symlink in your loaded workspaces.

### Zettlr version

develop

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit

### Operating System Version

Tahoe

### Additional Information

_No response_

### LINKED ISSUE #6040: File List shows `.git`-folder
### Description

While rewriting and updating the documentation, I noticed that the file list shows the entire contents of a `.git`-folder, if present:

Image

It should be noted that the tree view already filters this out. *However*, a `.git`-folder should not even be parsed by Zettlr. Instead, it should just set a flag indicating that the folder contains a git-repository.

### Reproducing

Have a git repository and look inside the file tree.

### Zettlr version

4.0.0-beta.3

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit

### Operating System Version

Sequoia

### Additional Information

This is not really a bug inside the file list. This should've been caught in the main process/FSAL.

benniekiss:
## Description

This PR allows a user to configure whether hidden files are shown in the sidebar and file manager.

Closes #6040
Closes #5949

## Changes

A configuration option was added to the advanced settings pane allowing a user to configure whether hidden files and directories (essentially dot files) are shown in the file manager and sidebar.

Additional extensions configured in the advanced settings pane now show up in the file manager.

Hidden file detection was refactored to be consolidated within the file sorting methods of the `TreeItem`, `FileList`, and `workspace-store`. Now, the FSAL indexing includes all files.

A check was added to the `readDirectoryRecursively` and `readDirectory` FSAL methods to check if the directory is a `.git` dir, in which case parsing is skipped.

`ignoreDir` handling was consolidated, and the FSAL watchdog was updated to use the `ignoreDir` function. The regex to ignore dot files was commented out so that changes to dot files are picked up when a user has them visible.

The `getFilesystemMetadata` function was refactored to be more streamlined and remove references to recursive parsing, which is now no longer handled by this function. The call to `fs.stat` was changed to `fs.lstat` to detect symlinks and prevent crashes when they are in the tree. This led to the target being `stat`'d, but then the subsequent call to `readFile` would fail because it tries to read the link path, not the target path.

## Additional information

Tested on:

### DIFF
--- a/source/app/service-providers/fsal/index.ts
+++ b/source/app/service-providers/fsal/index.ts
@@ -822,7 +822,7 @@ export default class FSAL extends ProviderContract {
     }
 
     const { files } = this._config.get()
-    const ignoreDotFiles = !(files.dotFiles.showInFilemanager || files.dotFiles.showInSidebar)
+    const ignoreDotFiles = !files.dotFiles.showInFilemanager && !files.dotFiles.showInSidebar
 
     try {
       const contents = (await fs.readdir(directoryPath, { withFileTypes: true, recursive: true }))
@@ -856,7 +856,7 @@ export default class FSAL extends ProviderContract {
     }
 
     const { files } = this._config.get()
-    const ignoreDotFiles = !(files.dotFiles.showInFilemanager || files.dotFiles.showInSidebar)
+    const ignoreDotFiles = !files.dotFiles.showInFilemanager && !files.dotFiles.showInSidebar
 
     try {
       const children = await fs.readdir(absPath, { withFileTypes: true })


source/app/service-providers/fsal/fsal-directory.ts
```diff
@@ -85,24 +84,33 @@ async function persistSettings (dir: DirDescriptor): Promise<void> {
...
       // We have a project, so we need to sanitize the values (in case
       // that there have been changes to the config). We'll just use
       // the code from the config provider.
       dir.settings.project = safeAssign(settings.project, PROJECT_TEMPLATE)
     }
+
     if (JSON.stringify(dir.settings) === JSON.stringify(SETTINGS_TEMPLATE)) {
       // The settings are the default, so no need to write them to file
-      await fs.unlink(configPath)
+      await fs.unlink(configPath).catch(() => {})
```
nathanlesage: Why would you catch these errors here? I'd expect them to be at least logged somewhere…?

source/common/util/ignore-dir.ts
```diff
@@ -16,20 +16,29 @@ import path from 'path'
 // Ignored directory patterns
 import { ignoreDirs } from '../data.json'
```
nathanlesage: I believe pulling in those regexps here into this file is a good idea, since they aren't referenced anywhere else. (Also this data file is a remnant from way back, where nothing was consolidated)

source/common/util/file-extention-checks.ts
```diff
@@ -124,6 +134,17 @@ export function hasDataExt (filePath: string): boolean {
...
 }
 
+/**
+ * Check if the given filename is a hidden file.
+ *
+ * @param   {string}   filename  The filename to check
+ *
+ * @return  {boolean}            Whether the file is a hidden file
+ */
+export function isHiddenFile (filename: string): boolean {
```
nathanlesage: My personal suggestion: rename to `isDotFile`. This is both closer to the truth and makes reasoning about the code more easy.

source/common/util/file-extention-checks.ts
```diff
@@ -132,8 +153,7 @@ export function hasDataExt (filePath: string): boolean {
  *
  * @return  {boolean}            Whether the filePath has any recognized ext.
  */
-export function hasAnyRecognizedFileExtension (filePath: string): boolean {
-  return hasMdOrCodeExt(filePath) || hasImageExt(filePath) ||
-    hasPDFExt(filePath) || hasOpenOfficeExt(filePath) ||
-    hasMSOfficeExt(filePath) || hasDataExt(filePath)
+export function hasAnyRecognizedFileExtension (filePath: string, extensions: string[] = []): boolean {
+  const ext = [ ...ALL_EXT, ...extensions ]
+  return hasExt(filePath, ext)
```
nathanlesage:
```suggestion
  return hasExt(filePath, [ ...ALL_EXT, ...extensions ])
```

benniekiss:
> Okay, so, I think I get this PR. In general, I'd argue we should move from hidden files to dot files, just because it's more accurate and won't confuse Windows users, since there "hiding" files involves setting an attribute.

I had actually originally had this as `dotFiles` and moved to `hiddenFiles` just because there may be other non-dotfile files we would want to hide? But that was just me overthinking the future. I can get all `hiddenFiles` throughout the PR moved back to `dotFiles`.

> Then, it doesn't seem as if this PR actually ignores `.ztr-directory` files, or does it?

No, `.ztr-directory` is not ignored and it is visible when `hiddenFiles` is enabled. Should it be ignored? I can make that happen.

source/app/service-providers/fsal/fsal-directory.ts
benniekiss: Oh you're right, I don't think this is necessary here. 

I added it below to catch cases where `unlink` fails due to the file not existing, but the `if` block I put before should handle that case, and I don't think it's necessary below either.

nathanlesage:
I would like to separate the logic of detecting dot-files, and whether they are affected by the show/hide section. We should probably add a simple array to hide specific dot-files and folders; see also issue #6061. I believe there could be three "layers" of dot files:

1. Default: Everything is just hidden.
2. "Dot-files" are shown except those that users of Zettlr commonly deal with. I strongly believe this should be a hard-coded array of dot-files that should never appear. (".git", ".quarto", ".config", and so on)
3. User-defined "dot-files" are hidden.

Settings-wise it might make sense to capture it like so:

* Keep the "show dot files" in the file types matrix as you have already implemented.
* Add a new setting "Hide common dot-files/folders" (feel free to get creative with the exact wording) that is on by default. If "dot files" are hidden in general, it won't have an effect, but if dot files are shown, this filters a few common dot files and folders; i.e.: "most" dot files are shown, but the hard-coded one are not.

Optionally (but I'd argue we should not implement before we actually need it), we could also add a token list-input (like for the attachment extensions) where users could provide additional dot files that should be hidden in addition.

benniekiss:
I think user defined files are out of scope for this, and I dont really have interest in implementing that here.

Regarding always hidden files (.git, etc), should those always be hidden--i.e, no setting to make them visible, or should there be a separate setting to make them visible? 

I can get behind always hidden, but a separate setting in my opinion would only confuse users who have to track down two settings for essentially the same functionality.

benniekiss:
Also, in what layer should the always hidden files be filtered?

Currently, git folders are filtered in the FSAL layer because we dont want to recurse into them, which causes issues when parsing certain "exotic" files within the folder.

However, every other file/folder is filtered at the ui level

nathanlesage:
I'm fine with no option for that. Again, Zettlr only provides a view over the file system, and Is argue users can also always use the computer file browser to view those hidden folders.

I'd also say all hidden files should be filtered at the FSAL level, that is, not even be parsed. This reduces the memory footprint and avoids potential footguns where hidden folders contain a bazillion plain text files that are of no concern to Zettlr. The less we parse, the better. That was also why I was initially hesitant of such a prominent setting...

benniekiss:
Thats definitely doable!

Do you have a list in mind of files/folders to always ignore? It can start off simple, too

danieltomasz:
As the author of https://github.com/Zettlr/Zettlr/issues/6061, I agree that hiding hidden or dot files could be expanded to completely ignore them.  Regarding hardcoding the list of files, could it be  just a setting for files or regular expressions that Zettlr ships with?  These could then be overriden by the user either through the settings interface or by editing a text configuration file (if the settings  inteface complexity shouldnt be increased and there is no need for new panel) .

 I don’t expect this to be a common occurrence but sometimes someone might want at least some of these files accessible in very rare cases.

benniekiss:
So again, I'm not going to implement user-configured hidden files in this PR. I think there's more discussion to be had on the UI and UX of that feature.

Regarding what kind of files are included in the hardcoded list, I would really only want these to be files and folders users have no business editing regardless -- i.e, people should not be messing around in the `.git` folder, and theres really no need to give a user access to this in the UI, same with the `.ztr-directory` file. I think the list should also include other VCS folders (`.hg` for mercurial), and maybe some other common artifact patterns. I don't think this list will end up being long or that intrusive. Otherwise, it would defeat the purpose imo.

nathanlesage:
I concur with @benniekiss here. Also, there will be many more releases!

As to a potential starting list of what should always be ignored, we might want to start with common gitignore lists and see what we find that might commonly end up in user directories! I don't think that list will be very long, though. One additional thing would be `.DS_Store` that we should ignore

danieltomasz:
if I can add to the list a quite often folders to exclude can be `.obsidian` and `.quarto` (.git for sure)

benniekiss:
Last commit is a WIP to implementing the hard-coded list and filtering at the FSAL level. I believe its 90% there, I just haven't tested extensively.

I renamed the `ignore-dir` file and function to `ignore-path`, etc. because the function is used to check more than just dirs, but rather every file path. I also moved the `isDotFile` function into this file. It felt more at home there, and the previous location in the `file-extension` file was sort of out of place since it was not checking an extension.

Even though we filter at the FSAL level, we still need filtering at the UI level. Since the show dot files config option has two parts -- the file manager and the sidebar -- there may be instances where dotfiles are in the tree but shouldn't be shown in a particular context.

One question I had was how to address watchdog handling. Currently, I re-use the `ignorePath` function in the `chokidar` `ignore` option, and because of that, I believe `ztr-directory` handling is likely broken ATM. What I was wondering -- do we need to have this `ignored` option set, or does the watchdog already ignore the files since we filter them in the FSAL layer?  It is not complicated to have separate watchdog handling, but it means the paths to ignore needs to be separated out. Another question I had was whether we still need to emit events for changes to a `.git` directory -- I couldn't find any clear functionality that depended on these events.

nathanlesage:
Everything that is filtered out at the FSAL level also needs to be ignored in the watchdog, especially since changes to git folders would emit a huge amount of events that can have a performance impact on Zettlr because the main loop is busy sending out events into the void.

I just had another question: do we generally need to distinguish between dot files and dot folders?

benniekiss:
> Everything that is filtered out at the FSAL level also needs to be ignored in the watchdog, especially since changes to git folders would emit a huge amount of events that can have a performance impact on Zettlr because the main loop is busy sending out events into the void.

Currently, `.git` dirs are still watched with special handling here, but that should be removed? https://github.com/Zettlr/Zettlr/blob/0904e1cdab00ceecf7da622ff87cfc250fadf905/source/app/service-providers/fsal/fsal-watchdog.ts#L112-L124

> I just had another question: do we generally need to distinguish between dot files and dot folders?

I don't think so, and there's no distinction in the code currently.

nathanlesage:
> Currently, .git dirs are still watched with special handling here, but that should be removed?

Given that Zettlr may need to react to those changes, we should leave it. But anything that Zettlr doesn't need should be ignored. If we have to extend the list of exceptions in the future so be it!

source/app/service-providers/fsal/index.ts
```diff
@@ -124,7 +124,7 @@ export default class FSAL extends ProviderContract {
     await this.syncRoots()
 
     this._config.on('update', (which: string) => {
-      if (which === 'openPaths') {
+      if (which === 'openPaths' || which === 'files.dotFiles.showInFilemanager' || which === 'files.dotFiles.showInSidebar') {
```
nathanlesage, line 129: Do I read this correctly that you want to add the ability to open dotfiles as workspaces? If so, can you tell me what the use-case is?

source/app/service-providers/fsal/index.ts
```diff
@@ -234,6 +234,13 @@ export default class FSAL extends ProviderContract {
     }
   }
 
+  private ignorePath (absPath: string): boolean {
+    const { files } = this._config.get()
+    const showDotFiles = files.dotFiles.showInFilemanager || files.dotFiles.showInSidebar
+
+    return ignorePath(absPath) || (showDotFiles ? false : isDotFile(absPath))
```
nathanlesage: What exactly is this pattern? I don't quite understand how it is supposed to work?

nathanlesage: Oh, I see – two functions of the same name. Please resolve this, otherwise, future-us won't be happy.

source/app/service-providers/fsal/index.ts
```diff
@@ -824,13 +840,20 @@ export default class FSAL extends ProviderContract {
...
+
     const contents = (await fs.readdir(directoryPath, { withFileTypes: true }))
       .filter(dirent => {
-        return (dirent.isFile() && !dirent.name.startsWith('.')) ||
-          (dirent.isDirectory() && !ignoreDir(dirent.name))
+        return ((dirent.isFile() || dirent.isDirectory()) && !this.ignorePath(dirent.name))
       })
       .map(dirent => {
         const childPath = path.join(directoryPath, dirent.name)
+        if (this.ignorePath(dirent.name)) {
```
nathanlesage: This should be a no-op, since you already exclude this in the `filter`, no?

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Description:     Utility functions to check for ignored file paths.
+ *
+ *
+ * END HEADER
+ */
+
+// A list of path patterns that should be ignored
+// by the FSAL layer and Chokidar watchdog process
+const IGNORE_PATH_RE: RegExp[] = [
+  /(?:^|[\/\\])\.DS_Store$/i, // macOS directory files
```
nathanlesage: Can we go back to constructing a list of regexp's like before? This would make adapting the non-changing part of the regexps simpler, and possibly make the list of file types we ignore easier to ignore, too?

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Check if the given filename is a dot file or folder.
+ *
+ * @param   {string}   filePath  The path to check
+ *
+ * @return  {boolean}            Whether the file is a dot file or folder
+ */
+export function isDotFile (filePath: string): boolean {
+  return /(?:^|[\/\\])\./.test(filePath)
```
nathanlesage: Wouldn't it be safer to use `path.basename().startsWith('.')`?

nathanlesage: Ah I just saw that you also include that thing in the renderer. Don't we at least then need an expression that checks for end-of-string (`$`)?

source/win-main/file-manager/FileList.vue
```diff
@@ -155,41 +164,51 @@ const getDirectoryContents = computed<RecycleScrollerData[]>(() => {
...
-        if (!files.msoffice.showInFilemanager) {
-          continue
-        }
-      } else if (hasOpenOfficeExt(sortedDescendants[i].path)) {
-        if (!files.openOffice.showInFilemanager) {
-          continue
-        }
-      } else if (hasDataExt(sortedDescendants[i].path)) {
-        if (!files.dataFiles.showInFilemanager) {
+    if (sortedDescendants[i].type === 'directory') {
```
nathanlesage: Is there a way to simplify this nested if-expression?

source/win-main/file-manager/TreeItem.vue
```diff
@@ -250,28 +259,35 @@ const secondaryIcon = computed(() => filteredChildren.value.length > 0 ? 'angle'
...
+      return 'file'
     } else if (hasOpenOfficeExt(props.item.path)) {
-      return 'file' // fileExtIcon.replace('EXT', props.obj.ext.slice(1, 4))
+      return 'file'
     } else if (hasDataExt(props.item.path)) {
-      return 'file' // fileExtIcon.replace('EXT', props.obj.ext.slice(1, 4))
+      return 'code'
+    } else if (hasExt(props.item.path, attachmentExtensions)) {
+      return 'file-group'
+    } else if (props.item.name === '.ztr-directory') {
```
nathanlesage: Please remove, this should display the warning below, because that file should NEVER appear in the file tree.

source/app/service-providers/fsal/index.ts
benniekiss: No, this is just a trigger to re-index the files, no? Since we exclude dotfiles at the FSAL level, when we enable/disable the option to show them, we need to reindex to capture them.

source/app/service-providers/fsal/index.ts
benniekiss: This is really ugly I agree. Is `ignoreFilePaths` still too similar? or maybe reverse the boolean and call it `handlePath`?

source/app/service-providers/fsal/index.ts
benniekiss: Oh that makes sense to me! I'll remove it and test.

source/common/util/ignore-path.ts
benniekiss: I initially used `startsWith`, but I wanted this to be able to handle absolute paths. I don't think we need a line terminator since it's only checking whether the string starts with `.` or any `/.` token, meaning a path part starts with a dot.

source/win-main/file-manager/FileList.vue
benniekiss: I moved this to a filter/map, which is a little more streamlined

source/common/util/ignore-path.ts
benniekiss: That makes sense to me! This is also going to have to be refactored to address watchdog handling.

benniekiss:
> Given that Zettlr may need to react to those changes, we should leave it. But anything that Zettlr doesn't need should be ignored. If we have to extend the list of exceptions in the future so be it!

So I did some testing to see what was emitted for git dirs, and I don't think anything relies on those types of event, and I don't think the code actually picks them up anyways. In a separate terminal, I tried doing some git commands on a folder opened as a workspace in Zettlr, and no events were emitted, doing things like `touch test.md; git add test.md; git commit -m 'test'`, etc. did not emit the `.git` watchdog events. I think it should be safe to remove this special handling, as I don't think its needed, and I'm not sure its functional.

Regarding `.ztr-directory`, though, that does have events appropriately picked up, and I think it is used by the Zettlr for monitoring, so that should stay in.

nathanlesage:
Great – yes, changes to the `.ztr-directory`-file need to be caught by the FSAL so that Zettlr can update its sorting, display and other config settings for the corresponding directory. If the `git`-changes aren't caught, then we can remove any related code for now. At some point we might want to capture them, but for now it's fine if we don't.

benniekiss:
Last commit builds the regex programmatically, which I think addresses all your points. I believe this is good to go on my end

nathanlesage:
Great, thank you so much! I will be merging this after 4.0 stable comes out, simply because I don't want to risk any potential delays. I'll release Zettlr 4.0 on Dec 26, so this thing will be merged this year.

FYI: I have taken the liberty of cherrypicking the fix for #5949 from this PR just to get that one bugfix over the bridge, too 👀

benniekiss:
@nathanlesage If you could PTAL at the last two commits, otherwise this is good on my end.

The last ~~two~~ three changes essentially remove the `isFile` and `isDir` check from the FSAL indexing since the `openFiles` and `openWorkspaces` already do that filtering. it also moves dotfile detection within the `ignorePath` function

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,89 @@
...
+// be watched for changes. Will be converted
+// to a RegExp object
+const IGNORE_PATHS: string[] = [
+  '\\.ztr-directory', // Zettlr project settings
+  ...WATCHDOG_IGNORE_PATHS
+]
+
+// Matches either the start of the line or
+// the first letter after a path separator.
+const re_prefix = '(?:^|[\\\/\\\\])'
```
nathanlesage:
```suggestion
const prefixRE = '(?:^|[\\\/\\\\])'
```

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,89 @@
...
+
+// Matches either the end of the line
+// or an internal segment of the path.
+const re_suffix = '(?:$|[\\\/\\\\])'
```
nathanlesage:
```suggestion
const suffixRE = '(?:$|[\\\/\\\\])'
```

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const WATCHDOG_IGNORE_RE: RegExp[] = WATCHDOG_IGNORE_PATHS
+  .map(re => new RegExp(`${re_prefix}${re}${re_suffix}`, 'i'))
```
nathanlesage:
```suggestion
  .map(re => new RegExp(`${prefixRE}${re}${suffixRE}`, 'i'))
```

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const IGNORE_PATH_RE: RegExp[] = IGNORE_PATHS
+  .map(re => new RegExp(`${re_prefix}${re}${re_suffix}`, 'i'))
```
nathanlesage:
```suggestion
  .map(re => new RegExp(`${prefixRE}${re}${suffixRE}`, 'i'))
```

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Check if the given filename is a dot file or folder.
+ *
+ * @param   {string}   filePath  The path to check
```
nathanlesage:
```suggestion
 * @param   {string}   filePath  The path to check (can be absolute or relative)
```

source/common/util/ignore-path.ts
```diff
@@ -0,0 +1,89 @@
...
+export function ignorePath (filePath: string, ignoreDotFiles: boolean = true): boolean {
+  if (ignoreDotFiles && isDotFile(filePath)) {
+    return true
+  }
+
+  if (IGNORE_PATH_RE.some(re => re.test(filePath))) {
+    return true
+  }
+
+  return false
```
nathanlesage:
```suggestion
  return IGNORE_PATH_RE.some(re => re.test(filePath))
```

source/app/service-providers/fsal/fsal-directory.ts
```diff
@@ -128,34 +136,22 @@ export async function parse (currentPath: string): Promise<DirDescriptor> {
     settings: JSON.parse(JSON.stringify(SETTINGS_TEMPLATE))
   }
 
+  try {
+    dir.isGitRepository = (await fs.lstat(path.join(dir.path, '.git'))).isDirectory()
+  } catch (err: any) {}
+
```
nathanlesage, line 142: Can't we do a simple `isDir`-check? That would avoid this try/catch contraption.

source/app/service-providers/fsal/fsal-watchdog.ts
```diff
@@ -61,10 +43,10 @@ export default class FSALWatchdog {
     this._config = config
 
     const options: ChokidarOptions = {
-      ignored: ignoreDirs,
+      ignored: WATCHDOG_IGNORE_RE,
       persistent: true,
       ignoreInitial: true, // Do not track the initial watch as changes
-      followSymlinks: true, // Follow symlinks
+      followSymlinks: false, // Zettlr does not support symlinks
```
nathanlesage, line 49: This is a beautiful diff to read.

source/app/service-providers/fsal/index.ts
```diff
@@ -823,23 +817,29 @@ export default class FSAL extends ProviderContract {
...
-          return Promise.resolve([])
-        }
-      })
+    if (!await this.isDir(directoryPath)) {
+      throw new Error(`[FSAL] Cannot read path ${directoryPath}: Not a directory!`)
+    }
 
-    return [ directoryPath, ...(await Promise.all(contents)).flat() ]
+    const { files } = this._config.get()
+    const ignoreDotFiles = !(files.dotFiles.showInFilemanager || files.dotFiles.showInSidebar)
```
nathanlesage: Isn't this suggestion equal? Would avoid easy to miss brackets.

```suggestion
    const ignoreDotFiles = !files.dotFiles.showInFilemanager && !files.dotFiles.showInSidebar
```

source/app/service-providers/fsal/index.ts
```diff
@@ -855,12 +855,27 @@ export default class FSAL extends ProviderContract {
       throw new Error(`[FSAL] Cannot read path ${absPath}: Not a directory!`)
     }
 
-    const children = await fs.readdir(absPath)
+    const { files } = this._config.get()
+    const ignoreDotFiles = !(files.dotFiles.showInFilemanager || files.dotFiles.showInSidebar)
```
nathanlesage: Same as above comment?

source/common/util/file-extention-checks.ts
```diff
@@ -132,8 +142,6 @@ export function hasDataExt (filePath: string): boolean {
  *
  * @return  {boolean}            Whether the filePath has any recognized ext.
  */
-export function hasAnyRecognizedFileExtension (filePath: string): boolean {
-  return hasMdOrCodeExt(filePath) || hasImageExt(filePath) ||
-    hasPDFExt(filePath) || hasOpenOfficeExt(filePath) ||
-    hasMSOfficeExt(filePath) || hasDataExt(filePath)
+export function hasAnyRecognizedFileExtension (filePath: string, extensions: string[] = []): boolean {
+  return hasExt(filePath, [ ...ALL_EXT, ...extensions ])
```
nathanlesage:
```suggestion
export function hasAnyRecognizedFileExtension (filePath: string, customExtensions: string[] = []): boolean {
  return hasExt(filePath, [ ...ALL_EXT, ...customExtensions ])
```

nathanlesage:
Also, I appreciate that you took the time to fix a bunch of newly introduced `any`-warnings while you were at it!

source/common/util/ignore-path.ts
benniekiss: python brain 🤪

source/app/service-providers/fsal/fsal-directory.ts
benniekiss: `isDir` is sync, which is why I avoided it here, but otherwise I have no problem using it.

source/app/service-providers/fsal/index.ts
benniekiss: The current version short circuits, but otherwise they are the same end result

benniekiss:
Just want your final opinions on the `isDir` and `!files.dotFiles.showInFilemanager && !files.dotFiles.showInSidebar` changes, otherwise this is good to go from my end!

source/app/service-providers/fsal/index.ts
benniekiss: Totally glossed over that your suggestion also short circuits, so Im gonna merge the change!

nathanlesage:
Thanks a bunch! Looks good now – merging

PR #6045, opened 2025-12-05
[Fix] Refactor visual-indent plugin

benniekiss:
## Description

This PR refactors the visual indent plugin to fix issues related to collapsed text and to account for blockquote indentation levels.

## Changes

The tabstop replacement decorator was refactored to use a separate `MatchDecorator`.

Indentation is skipped for collapsed text regions. This was causing visual issues for blockquotes when they were collapsed.

Indentation for blockquotes is calculated based on the number of `>` characters so that text is aligned within the innermost `>`.

The logic of the renderer was reduced to one `for` loop and simplified.

The `.cm-foldPlaceholder` styling was moved to `main-overrides`, and the background was made transparent to improve styling.

### Current:

Screenshot 2025-12-05 at 18 33 59

Screenshot 2025-12-05 at 18 34 11

---

### PR:

Screenshot 2025-12-05 at 18 33 32

Screenshot 2025-12-05 at 18 33 39

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/plugins/visual-indent.ts
+++ b/source/common/modules/markdown-editor/plugins/visual-indent.ts
@@ -31,6 +31,16 @@ import { SpaceWidget } from '../renderers/render-emphasis'
 // not to induce any problems when adjusting the indentation
 const BASE_PADDING = 6
 
+// These regexes are used to determine indentation level
+// Leading whitespace characters
+const WHITESPACE_RE = /^([ \t]*)/
+
+// List and task marker formatting characters
+const LISTMARK_RE = /^(\s*(?:[+*-](?:\s\[[x\s]\])?|\d+\.)\s)/
+
+// Blockquote formatting characters
+const QUOTEMARK_RE = /^((?:[ ]{0,3}\>)+[ ]?)/
+
 // Since tab characters have no fixed width in the editor,
 // we need to render every tab with the equivalent number of space
 // characters to prevent jumping in the editor.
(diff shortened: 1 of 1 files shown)


source/common/modules/markdown-editor/plugins/visual-indent.ts
```diff
@@ -15,167 +15,121 @@
...
+        continue
       }
 
-      ranges.push(
-        Decoration
-          .replace({ widget: new SpaceWidget(tabSize) })
-          .range(line.from + j, line.from + j + 1)
-      )
-    }
+      // First determine count the indent based on leading whitespace
```
nathanlesage:
```suggestion
      // First determine the indent based on leading whitespace
```

benniekiss:
> Looks good so far. I'm a bit anxious to have this break again. Did you do some excessive testing to see that it holds up even in absurd conditions…?

Yeah, I tested with a lot of copy-pasted block quotes, randomly indented paragraphs, and indented code. I tried scrolling fast, jumping around in the doc, and other things to break the indent, but I did not notice any thing going wrong. I could've overlooked something, but I felt I was fairly thorough.

nathanlesage:
Okay, let's figure out the Pandoc div parsing first, then we can push this and see what breaks.

source/common/modules/markdown-editor/plugins/visual-indent.ts
```diff
@@ -15,167 +15,121 @@
...
-      if (char === '\t') {
-        tabs++
-      } else if (char === ' ') {
-        spaces++
-      } else {
-        break
-      }
-    }
+      // Second, determine the offset based on list or blockquote elements.
+      const listMatch = /^\s*((?:[+*>-](?:\s\[[x\s]\])?|\d+\.)\s)/.exec(line.text)
```
nathanlesage: This also matches blockquotes, no? There's a `>` in the regexp

source/common/modules/markdown-editor/plugins/visual-indent.ts
```diff
@@ -15,167 +15,121 @@
...
+      const quoteMatch = /^([ ]{0,3}\>)+[ ]?/.exec(line.text)
 
-    // Second, determine the offset based on list elements.
-    const match = /^\s*((?:[+*>-](?:\s\[[x\s]\])?|\d+\.)\s)/.exec(line.text)
-    const listMarker = match !== null ? match[1].length : 0
+      const formatCharOffset = quoteMatch ? quoteMatch[0].length : listMatch ? listMatch[1].length : 0
```
nathanlesage: Maybe a comment to describe how that offset differs from the leading spaces…?

source/common/modules/markdown-editor/plugins/visual-indent.ts
```diff
@@ -15,167 +15,121 @@
...
+      // line as a key, we cause (a) duplicated measurements (`* One` and `* Two`
+      // require the same indentation) and (b) cause cache misses which results in
+      // jumpy behavior when the user adds novel characters to the line.
+      const measurementKey = line.text.slice(0, columnLineTextStart).replace('\t', ' '.repeat(view.state.tabSize))
+
+      view.requestMeasure({
+        read (view) {
+          const pos = line.from + columnLineTextStart
+
+          const folded = foldedRanges(view.state).iter(pos)
```
nathanlesage, line 95: Please add a comment describing that you're here searching for folded regions to not indent those accidentally

source/common/modules/markdown-editor/theme/main-override.ts
```diff
@@ -124,6 +124,10 @@ export const mainOverride = EditorView.baseTheme({
   },
   '&dark .cm-highlight': {
     backgroundColor: '#ffff0060',
+  },
+  '.cm-foldPlaceholder': {
```
nathanlesage: Please move this change to a separate PR, since it changes the background color of the fold markers.

source/common/modules/markdown-editor/plugins/visual-indent.ts
benniekiss: It matches a single `>`, I assumed this was just another listmarker, but I don't think that's right. I'll just remove that character from the regex

benniekiss:
I refactored the regex handling into something much more straightforward, and I moved the regex to module level constants, otherwise, I reverted the fold styling and added some comments. This is good to go on my end

nathanlesage:
Perfect, looks good! Let's merge so that we can move to the final beta!

PR #6046, opened 2025-12-06
[Performance] Refactor heading and code syntax highlighting

benniekiss:
## Description

This PR refactors the heading and code syntax highlighting, removing the `syntax-extensions.ts` file, moving the code decoration to a renderer plugin, and moving the heading styles to native codemirror classes.

## Changes

The `syntax-extensions.ts` file was removed. It seemed out-of-place in the parser folder, and the code was somewhat dated. Likewise, it was not performant, as it was unnecessarily applying decorations over the whole document.

The code syntax logic was moved to its own plugin under the `renderers` folder since it follows the same structure as other renderers. The `code-block-line` style was dropped, as it was not really used. The styling for code was moved into this plugin from the `MainEditor`, consolidating it in one place. 

Heading styles were refactored from using the custom `size-header-[1-6]` classes to using the native codemirror `cm-header-[1-6]` classes. This removes the need for a separate plugin to apply the classes, and it makes it less complicated for users to style headings.

Several `ViewPlugin` renderers had their `update` methods refactored to only conditionally update the decorations if the document or viewport changed to reduce unnecessary recomputations.

## Additional information

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-emphasis.ts
+++ b/source/common/modules/markdown-editor/renderers/render-emphasis.ts
@@ -169,7 +169,7 @@ export const renderEmphasis = ViewPlugin.fromClass(class {
   }
 
   update (update: ViewUpdate): void {
-    if (update.docChanged || update.viewportChanged) {
+    if (update.docChanged || update.viewportChanged || update.selectionSet) {
       this.decorations = hideFormattingCharacters(update.view)
     }
   }
--- a/source/common/modules/markdown-editor/renderers/render-headings.ts
+++ b/source/common/modules/markdown-editor/renderers/render-headings.ts
@@ -64,7 +64,7 @@ export const renderHeadings = ViewPlugin.fromClass(class {
   }
 
   update (update: ViewUpdate): void {
-    if (update.docChanged || update.viewportChanged) {
+    if (update.docChanged || update.viewportChanged || update.selectionSet) {
       this.decorations = hideHeadingMarks(update.view)
     }
   }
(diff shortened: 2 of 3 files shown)


source/common/modules/markdown-editor/renderers/render-code.ts
```diff
@@ -0,0 +1,84 @@
...
+  update (update: ViewUpdate): void {
+    if (update.docChanged || update.viewportChanged) {
+      this.decorations = getCodeHighlighter(update.view)
+    }
+  }
+}, {
+  decorations: v => v.decorations
+})
+
+const codeTheme = EditorView.baseTheme({
```
nathanlesage, line 81: Please re-implement the variables for the colors. The reason is that this will make it much easier in the future to re-adapt the theme where necessary, as it's easier to read than color codes. I think the simplest would just be to use JavaScript variables for this.

benniekiss:
> Also, did you check the code editor whether some of these changes also would apply there…? If nothing needs to change, then all good!

The code editor is used for the css editor, snippets, etc, right? I did not notice any changes.

nathanlesage:
Then let's get this in! Your PR backlog is growing a lot ... again ...!

benniekiss:
I'm hoping to slow down soon! This recently deluge was mostly from the work I realized during the blockquote attempt, so it's nice that work didn't end up entirely a dead-end

nathanlesage:
Oh absolutely! I'm always in favor of any code I don't have to write myself. It's an incredible relief having you around. Last week I could use the free time to fully rewrite the documentation, for example.

PR #6050, opened 2025-12-07
Cleanup asset window styling

benniekiss:
## Description

This PR synchronizes the styling of the assets window. 

## Changes

The save button handling was synchronized across windows, adding a timeout to the CSS button, and adding the error handling call to the other save buttons.

The Lua filters tab was moved after the Import/Export tabs since they are related in usage.

Some titles were moved to admonitions to better align the styling between tabs, and the CSS tab was refactored to match the structure of the other tabs. 

Styling was unified between the tabs.

## Additional information

Screenshots: 

Screenshot 2025-12-07 at 15 31 34

Screenshot 2025-12-07 at 15 31 41

Screenshot 2025-12-07 at 15 31 52

Screenshot 2025-12-07 at 15 31 59

Tested on:  macOS 26

### DIFF
--- a/source/win-assets/CustomCSS.vue
+++ b/source/win-assets/CustomCSS.vue
@@ -93,7 +93,7 @@ function saveCSS (): void {
       setTimeout(() => { savingStatus.value = '' }, 1000)
     })
     .catch(err => {
-      savingStatus.value = trans('Saving failed')
+      savingStatus.value = trans('Could not save changes')
       console.error(err)
     })
 }
--- a/source/win-assets/DefaultsTab.vue
+++ b/source/win-assets/DefaultsTab.vue
@@ -271,7 +271,7 @@ function saveDefaultsFile (): void {
       setTimeout(() => { savingStatus.value = '' }, 1000)
     })
     .catch(err => {
-      savingStatus.value = trans('Saving failed')
+      savingStatus.value = trans('Could not save changes')
       console.error(err)
     })
 }
--- a/source/win-assets/FilterTab.vue
+++ b/source/win-assets/FilterTab.vue
@@ -229,7 +229,7 @@ function saveFilter (): void {
       setTimeout(() => { savingStatus.value = '' }, 1000)
     })
     .catch(err => {
-      savingStatus.value = trans('Saving failed')
+      savingStatus.value = trans('Could not save changes')
       console.error(err)
     })
 }
--- a/source/win-assets/SnippetsTab.vue
+++ b/source/win-assets/SnippetsTab.vue
@@ -195,7 +195,7 @@ function saveSnippet (): void {
       setTimeout(() => { savingStatus.value = '' }, 1000)
     })
     .catch(err => {
-      savingStatus.value = trans('Saving failed')
+      savingStatus.value = trans('Could not save changes')
       console.error(err)
     })
 }


source/win-assets/CustomCSS.vue
```diff
@@ -1,18 +1,27 @@
 <template>
-  <div id="custom-css">
-    <p id="custom-css-info" v-html="customCSSInfo"></p>
+  <div id="css-container">
+    <ZtrAdmonition v-bind:type="'info'">
```
nathanlesage: These properties can be moved to a simple non-v-bind

source/win-assets/CustomCSS.vue
```diff
@@ -78,30 +89,55 @@ function saveCSS (): void {
   })
     .then(() => {
       lastLoadedCSS.value = editorContents.value
-      savingStatus.value = ''
+      setTimeout(() => { savingStatus.value = trans('Saved!') }, 1000)
```
nathanlesage: This is something I found a bit weird in the beginning. I believe we should say "Saved" as soon as it is, indeed saved – immediately upon having the `then` promise resolve.

source/win-assets/CustomCSS.vue
```diff
@@ -78,30 +89,55 @@ function saveCSS (): void {
...
-  overflow: auto; // Enable scrolling, if necessary
-  padding: 10px;
-  width: 100vw;
+#css-container {
+  padding: 0px 10px;
   height: 100%;
   display: flex;
   flex-direction: column;
 
+  .admonition {
```
nathanlesage: I have thought long about it; I think it's nicer to wrap it in a div here instead of accessing a class name that may change at any time because it comes from a component.

(Short reasoning: I want to slowly move components to having no explicit margin, so that it can instead be provided by the containing element. The current status is a bit too opinionated and I didn't see that issue coming when I did my big UI rewrite a few years ago.)

source/win-assets/SnippetsTab.vue
```diff
@@ -302,11 +305,15 @@ function openSnippetsDirectory (): void {
 }
 
 #snippets-container {
-  padding: 10px;
+  padding: 0px 10px;
```
nathanlesage: …can't we define one class to use? Even if we have repeated code then, we would still see that the divs do the same thing

source/win-assets/SnippetsTab.vue
```diff
@@ -302,11 +305,15 @@ function openSnippetsDirectory (): void {
...
   height: 100%;
   display: flex;
   flex-direction: column;
 
+  .admonition {
```
nathanlesage: Same as above

nathanlesage (changes requested):
A few marginalia, otherwise I can get behind this.

source/win-assets/CustomCSS.vue
benniekiss: I added it initially because you will see a very quick flicker of `saving ...` otherwise, and it gives a slightly jarring experience. But I can support showing `saved` right when it happens.

benniekiss:
Ok, I think I covered all your points! I moved all of the styling to `App.vue`, and I unified all the class names. I assigned classes to the admonitions as well so that it's not using the component class name.

benniekiss:
If you're amenable to this change, I removed all of the margin and padding from the `TextControl` element. This removes the need for some styling in the asset window, and it also improves the styling of the text control elements in the preferences window.

source/win-assets/DefaultsTab.vue
```diff
@@ -259,11 +266,14 @@ function saveDefaultsFile (): void {
   } as AssetsProviderIPCAPI)
     .then(async () => {
       lastLoadedEditorContents.value = editorContents.value
-      setTimeout(() => { savingStatus.value = trans('Saved!') }, 1000)
+      setTimeout(() => { savingStatus.value = trans('Saved!') }, 0)
```
nathanlesage: Do we need that timeout?

source/win-assets/FilterTab.vue
```diff
@@ -222,10 +225,13 @@ function saveFilter (): void {
   } as AssetsProviderIPCAPI)
     .then(() => {
       lastLoadedEditorContents.value = editorContents.value
-      setTimeout(() => { savingStatus.value = trans('Saved!') }, 1000)
-      setTimeout(() => { savingStatus.value = '' }, 2000)
+      setTimeout(() => { savingStatus.value = trans('Saved!') }, 0)
+      setTimeout(() => { savingStatus.value = '' }, 1000)
+    })
+    .catch(err => {
+      savingStatus.value = trans('Saving failed')
```
nathanlesage:
```suggestion
      savingStatus.value = trans('Could not save changes')
```

benniekiss:
> Would you mind checking across the app that nothing that uses the TextControl now looks off? I wouldn't be surprised if there were some obscure places that rely on those paddings …

I checked the search pane, preferences window, toolbar popovers, stats window/graph view, and file and folder properties window, and nothing seemed off to me

nathanlesage:
Alright, then let's do it; I'll be testing this thing out until the full release; if I notice anything I'll complain :D

PR #6053, opened 2025-12-08
[Fix] use `line.moveBase`, not `ctx.nextLine` in pandoc div parser

benniekiss:
## Description

This PR fixes a parsing issue in the pandoc div parser related to how the parsing was moved forward after the block was detected.

## Changes

Instead of using `ctx.nextLine`, we call `line.moveBase` with the tracked offset to move parsing forward. This is the correct way, as the docs explicitly state not to use `ctx.nextLine` in composite block parsers (oops). 

This fixes a parsing issue where an empty div would not parse the closing mark correctly:

```
::: {#ref}
:::
```

## Additional information

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/parser/pandoc-div-span-parser.ts
+++ b/source/common/modules/markdown-editor/parser/pandoc-div-span-parser.ts
@@ -158,7 +158,15 @@ export const pandocDivParser: BlockParser = {
     return null // composite blocks require returning `null` on success
   },
 
-  endLeaf: (_ctx, line, _leaf) => {
+  endLeaf: (ctx, line, _leaf) => {
+    // Opening marks can come one after the other without requiring
+    // a blank line in between. So we only interrupt if the line matches
+    // the opening mark if the parent is a `PandocDiv`. Otherrwise,
+    // only the closing mark can interrupt other nodes.
+    if (ctx.parentType().name === 'PandocDiv') {
+      return pandocDivClosingRe.test(line.text) || pandocDivOpeningRe.test(line.text)
+    }
+
     return pandocDivClosingRe.test(line.text)
   }
 }


nathanlesage:
Does this work with nested comments? That you're checking beginning of the line is … wait, this is not a problem, is it?

Does the following parse correctly (from Pandoc manual):

```
:::::::::::::: {.columns}
::: {.column width="40%"}
contents...
:::
::: {.column width="60%"}
contents...
:::
::::::::::::::
```

nathanlesage:
Ah, and also, I found that the attribute parser requires a space between the Div, but Pandoc does not, so the following would be a correct pandoc div:

```
:::{#refs} // <-- Note there's no space between div and attr
:::
```

benniekiss:
> Does the following parse correctly (from Pandoc manual)

It should now! It wasn't before because I had required a newline before the opening mark, but that seems to only be required for any other text except for the opening div mark.

> Ah, and also, I found that the attribute parser requires a space between the Div, but Pandoc does not, so the following would be a correct pandoc div

Fixed as well! That was a misunderstanding of the spec on my part

benniekiss:
> It should now! It wasn't before because I had required a newline before the opening mark, but that seems to only be required for any other text except for the opening div mark.

Let me clarify -- the `columns` class is not rendered as columns, but the nodes parse correctly. The css for `columns` would have to be added to the renderer plugin and print preview window.

nathanlesage:
Yes yes, the columns should be actually visually parsed because that would break the editor, but I now checked your changes locally, and it all parses well as far as I can see!

I think we can merge, unless you've found something else?

nathanlesage:
(P.S.: I'm impressed by the inline parsing, that's quite nice)

benniekiss:
Nah, I haven't noticed anything else... yet!

PR #6056, opened 2025-12-09
[Fix] Improve table editor jumping and other small improvements

benniekiss:
## Description

This PR refactors the table editor to improve scroll jumping when selecting cells, transaction filtering, and performance

## Changes

To address scroll jumping, a height cache was implemented. I noticed that the image renderer used a similar approach, and I thought, hey, this may work here! And it does marginally improve jumping behavior! However, there are still issues, such as when a really long table is loaded into a document, and the user scrolls all the way to the bottom of the table and clicks into it for the first time (i.e., no cached height), the table will still jump up; however, subsequent clicks into the table do not jump nearly as far, if at all. This can reproduced by opening the table stress test file, scrolling to the bottom, and clicking in the last cell.

`estimatedHeight` still falls back to a crude mean-row-height estimate, but this is necessary to prevent severe jumping if there is no cache. I reduced the value from `500` to `100` as a more reasonable fallback since egregiously large numbers also cause issues.

The transaction filter was refactored. I moved the `docChanged` test was moved to the top, and the `Disallowing transaction` test was refactored. I changed it because I noticed I could still delete the cell boundaries in the table editor, so I refactored the test to make sure that the mapped character is not deleted.

I added an `ensureSyntaxTree` call to try and get the whole tree, otherwise falling back to the incomplete tree. This is a similar strategy as the TOC renderer.

I added some small, but rather performant, changes to increase performance, mostly reducing loops and moving some variable accesses to constants.

Removed the old, unused `render-tables` file.

## Additional information

I think this overall really improves table editor behavior! I tested this using the development test files, and I tried very hard to break things. I may have overlooked something here, but so far, these changes have really improved the table editor imo.

Other issues I've noticed -- both in this PR and in develop -- is that when the window is narrow, causing lines to wrap into very tall rows, typing in cells at the bottom of the document causes significant flickering. This is alleviated by increasing `meanRowHeight`, but that's not really a viable long-term solution, and causes other issues. Increasing the window size resolves the flickering. Interestingly, this flickering doesn't seem to happen in the middle of a table.

I would appreciate anyone pulling and testing this PR to see how the UX is for real-world use cases.

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/table-editor/widget.ts
+++ b/source/common/modules/markdown-editor/table-editor/widget.ts
@@ -30,9 +30,39 @@ import { configField } from '../util/configuration'
 
 /**
  * This holds the last measured height of each rendered table to provide
- * values for the `estimatedHeight` getter.
+ * values for the `estimatedHeight` getter. It implements a simple LRU
+ * cache so that the cache size does not get out of hand, otherwise we
+ * would end up with an entry for essentially every position in every
+ * open document.
  */
-const TABLE_HEIGHT_CACHE = new Map<string, number>()
+const TABLE_HEIGHT_CACHE = new (class {
+  private readonly cache = new Map<string, number>()
+  private readonly maxSize = 100 // Limit the number of entries. Currently an arbitrary number.
+
+  get (key: string): number|undefined {
+    const value = this.cache.get(key)
+    if (value !== undefined) {
+      // Refresh the key's position
+      this.cache.delete(key)
+      this.cache.set(key, value)
+    }
+    return value
+  }
+
+  set (key: string, value: number): void {
+    // Refresh the key's position
+    if (this.cache.has(key)) {
+      this.cache.delete(key)
+    // Prune the cache entries
+    } else if (this.cache.size === this.maxSize) {
+      const first = this.cache.keys().next().value
+      if (first !== undefined) {
+        this.cache.delete(first)
+      }
+    }
+    this.cache.set(key, value)
+  }
+})()
 
 // This widget holds a visual DOM representation of a table.
 export class TableWidget extends WidgetType {
@@ -80,6 +110,8 @@ export class TableWidget extends WidgetType {
     return tableAST.rows.length * this.meanRowHeight
   }
 
+  // By setting the cache key to the node's `from` position,
+  // we stabilize the cache while edits happen within the table.
   private get cacheKey (): string {
     return `${this.node.from}`
   }


source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
```diff
@@ -33,7 +33,7 @@ export function dispatchFromSubview (mainView: EditorView): (tr: Transaction, su
       if (userEvent !== undefined) {
         annotations.push(Transaction.userEvent.of(userEvent))
       }
-      mainView.dispatch({ ...tr, annotations })
+      mainView.dispatch(tr, { annotations })
```
nathanlesage, line 39: Why this? The original code overwrote just the annotations property; this one now dispatches two separate transactions, correct?

source/common/modules/markdown-editor/table-editor/subview.ts
```diff
@@ -38,6 +38,10 @@ const ensureBoundariesFilter = EditorState.transactionFilter.of((tr) => {
     return tr // Do not mess with synchronizing transactions
   }
 
+  if (!tr.docChanged) {
```
nathanlesage: Ahhh, probably difficult. This transaction filter should ALSO disallow MOVEMENTS that go outside the subview boundaries. Unless you have a good reason for this change, I'd advise against it.

source/common/modules/markdown-editor/table-editor/subview.ts
```diff
@@ -79,41 +69,34 @@ const ensureBoundariesFilter = EditorState.transactionFilter.of((tr) => {
...
-    if (from < cellFrom || to > cellEndAfter) {
-      console.log('Disallowing transaction', { from, cellFrom, to, cellTo, cellEndAfter })
+
+    const mappedFrom = tr.changes.mapPos(cellFrom, -1, MapMode.TrackBefore)
+    const mappedTo =  tr.changes.mapPos(cellTo, 1, MapMode.TrackAfter)
+
+    const wasDeleted = mappedFrom === null || mappedTo === null
+
+    if (wasDeleted || from < mappedFrom || to > mappedTo) {
+      console.log(`Disallowing transaction: Selection: ${from} - ${to} | Cell: ${cellFrom} - ${cellTo} | Mapped: ${mappedFrom} - ${mappedTo}`)
```
nathanlesage, line 77: Can you explain the changes you made here? It looks mostly like refactoring, so I just want to be sure I'm with you on the nuances.

source/common/modules/markdown-editor/table-editor/widget.ts
```diff
@@ -69,14 +80,23 @@ export class TableWidget extends WidgetType {
...
     try {
       const { wrapper, table } = generateEmptyTableWidgetElement()
       const tableAST = parseTableNode(this.node, view.state.sliceDoc())
       if (tableAST.type !== 'Table') {
         throw new Error('Cannot render table: Likely malformed')
       }
+
       updateTable(table, tableAST, view)
+
+      const height = table.getBoundingClientRect().height
```
nathanlesage: Does this work if the element is not yet mounted to the DOM?

source/common/modules/markdown-editor/table-editor/widget.ts
```diff
@@ -184,11 +207,17 @@ export class TableWidget extends WidgetType {
    * @return  {DecorationSet}         The DecorationSet
    */
   public static createForState (state: EditorState): DecorationSet {
-    const newDecos: Array<Range<Decoration>> = syntaxTree(state)
+    const markdown = state.sliceDoc()
```
nathanlesage: Can't we just keep this in the `parseTableNode`?

source/common/modules/markdown-editor/table-editor/widget.ts
```diff
@@ -398,10 +428,19 @@ function updateRow (
       // So, here we enforce that the main selection is definitely somewhere
       // inside the table cell *content*.
       const sel = view.state.selection.main
-      let newFrom = Math.max(sel.from, cell.from)
-      newFrom = Math.min(newFrom, cell.to)
-      let newTo = Math.max(sel.to, cell.from)
-      newTo = Math.min(newTo, cell.to)
+      let { from: newFrom, to: newTo } = sel
```
nathanlesage: Did you change the logic here, or is this simply a refactor? I wonder if that wouldn't be a good candidate for a `clamp`-function…

nathanlesage (changes requested):
A few open questions

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
benniekiss: Ok looking back at it, the typing is confusing. I initially assumed both were `TransactionSpec`s, but `tr` is a transaction, so I think it gets coerced into a `TransactionSpec`? Either way, since I assumed they were transaction specs, if we pass multiple to `dispatch`, [they are merged internally.](https://codemirror.net/docs/ref/#view.EditorView.dispatch)

However, if this is treating them as a list of transactions, and each is a separate dispatch, we still need to append the annotations to the existing ones, rather than overwrite them.

source/common/modules/markdown-editor/table-editor/subview.ts
benniekiss: Makes sense!

source/common/modules/markdown-editor/table-editor/subview.ts
benniekiss: Before, it was not checking whether the position had actually been deleted, so it was possible to put the cursor in a cell and continue pressing backspace until a table delimiter was deleted. Now, if the position is deleted by the transaction, the transaction is invalid.

This is also the only place we need the mapped positions, so I moved the calculation into this check.

I removed the `&& cellFrom !== cellTo && cellTo > 0` condition because this isn't really necessary from what I found, and it was incorrect from what I understood. It was comparing the cell size prior to the transaction, but we should be checking whether the current cell size is non-empty, if at all.

source/common/modules/markdown-editor/table-editor/widget.ts
benniekiss: I found that to be a big bottleneck and was one of the reasons I started working on this

source/common/modules/markdown-editor/table-editor/widget.ts
benniekiss: I did not want to change the logic. I actually meant to revert this change but forgot to. just me re-writing for understanding.

benniekiss: Are you opposed to simply combining the statements, 

```ts
      const newFrom = Math.min(Math.max(sel.from, cell.from), cell.to)
      const newTo = Math.min(Math.max(sel.to, cell.from), cell.to)
```

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
nathanlesage: I believe you Are correct. My worry is only whether there may be annotations we should throw away...? If that's not the case then we can go with that, yes!

source/common/modules/markdown-editor/table-editor/widget.ts
nathanlesage: Interesting -- I wouldn't have thought that this simple change made such a difference!

source/common/modules/markdown-editor/table-editor/widget.ts
benniekiss: No, it does not! I hadn't thought to check. It's just setting it to `0` here. However, moving the cache update into the `view.requestMeasure` call works, but I'm not clear on whether it goes into the `write` or `read` method

source/common/modules/markdown-editor/table-editor/widget.ts
benniekiss: big is a bit exaggerated in the grand scheme of things, but I was surprised it would have such an impact too!

source/common/modules/markdown-editor/table-editor/widget.ts
benniekiss: Oh, adding the update to the `requestMeasure` call and setting a `key` value improves things even more, I think.

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
benniekiss: Another point that I'll look into is whether the merge is an overwrite or a union

source/common/modules/markdown-editor/table-editor/widget.ts
nathanlesage: Not at all! Also, some context from where that "clamp" idea came from: I read this RFC from PHP a few days ago: https://wiki.php.net/rfc/clamp_v2

And essentially this is the same pattern 😂  But I am absolutely not opposed to combining that, no!

source/common/modules/markdown-editor/table-editor/widget.ts
nathanlesage: Well, maybe add a small comment to indicate that this is a load-bearing variable, just so that we don't undo that change lightly in the future, because I know myself (especially my bored self that wants to procrastinate…)!

source/common/modules/markdown-editor/table-editor/widget.ts
nathanlesage: I think the idea is that during a `write` phase you can mess with the DOM sizes (which we don't want to), and during a `read` phase the DOM is guaranteed to be unchanged, so that's where we should measure the height for the cache!

source/common/modules/markdown-editor/table-editor/subview.ts
nathanlesage: Wow, with a clear mind after sleeping well the code suddenly started to make sense! Thanks for the clarifications! :)

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
nathanlesage: I think CodeMirror merges by calculating a delta between the transactions for a slight speed-up, so I'd imagine annotations to be a union, but I may be wrong…

benniekiss:
Ok, last commit is a little bit of a breakthrough. I do not get flickering at the bottom of the table any longer, and editing in the table seems super stable.

There are two remaining points I would like to finish up before merging this. First, better cache invalidation and flushing so that stale entries are eventually removed (I want an LRU cache). ~~Second, `node.from` is probably an ok cache key, but I would like to try to find something a little more unique to each table that won't be affected by changes in the text, if that even exists. I.e., is there something about a particular table node that remains the same even as the table changes?~~ I think using just `node.from` is fine, actually, since the cache is only needed when typing within the table

source/common/modules/markdown-editor/table-editor/util/data-exchange.ts
benniekiss: Looked through the codemirror code, and annotations are [concated](https://github.com/codemirror/state/blob/c24cde9d0152d3882ef811a6579d0de90767d7a0/src/transaction.ts#L325) together. 

Currently, I don't think there are any annotations that need to be discarded, so I would opt for the merge approach, and switch to the override approach if the need arises. I could see how removing all other annotations may cause hard-to-pin bugs to pop up, as well, but currently, the `syncAnnotation` is the only annotation I can find referenced in the code

nathanlesage: I concur!

nathanlesage:
> Ok, last commit is a little bit of a breakthrough. I do not get flickering at the bottom of the table any longer, and editing in the table seems super stable.

What you did there was literally what I assumed you meant with a new cache key :D 

So, what do you think we need to merge this still? The sooner we get a final beta out, the better!

source/common/modules/markdown-editor/table-editor/subview.ts
```diff
@@ -94,26 +86,17 @@ const ensureBoundariesFilter = EditorState.transactionFilter.of((tr) => {
...
+    if (fromA >= cellFrom && toA <= cellTo) {
+      // Ensure that no newlines will be inserted into the table cell
+      const ins = inserted.toString()
+      const safeInsertion = ins.replace(/\n+/g, ' ')
+      safeChanges.push({ from: fromA, to: toA, insert: safeInsertion })
     }
   })
 
-  return shouldOverrideTransaction ? { ...tr, changes: safeChanges } : tr
+  return { ...tr, changes: safeChanges }
```
nathanlesage, line 99: Can we really do that? If `safeChanges` is empty, this would remove all existing changes in the Transaction, no?

source/common/modules/markdown-editor/table-editor/widget.ts
```diff
@@ -184,28 +218,30 @@ export class TableWidget extends WidgetType {
...
    */
   public static createForState (state: EditorState): DecorationSet {
-    const newDecos: Array<Range<Decoration>> = syntaxTree(state)
+    // We try to retrieve the full syntax tree, and if that fails, fall back to
+    // the (possibly incomplete) syntax tree.
+    const tree = ensureSyntaxTree(state, state.doc.length, 500) ?? syntaxTree(state)
+    // Constantly calling `sliceDoc()` within the tree traversal
+    // below has some negative performance impacts, so we extract
+    // the markdown text outside of the loop
+    const markdown = state.sliceDoc()
```
nathanlesage, line 259: OH MY GOD I am such a dumbass. OF COURSE we're going to see performance impacts, I didn't see that the sliceDoc call was inside a loop 🤦🏼 

I'm sorry I gave you a hard time!

source/common/modules/markdown-editor/table-editor/subview.ts
benniekiss: If safe changes is empty, that means all changes occured outside of the cell range, and therefore `shouldOverrideTransaction` would be set to `true` anyways, returning the empty list. Likewise, every change within the cell range is pushed, so we don't need to check whether the newline replacement changed anything

benniekiss:
> So, what do you think we need to merge this still? The sooner we get a final beta out, the better!

I want to make sure that the cache doesn't explode in size, since it'll create a new entry for every position in the document, essentially. I have an LRU implementation, just need to push it

source/common/modules/markdown-editor/table-editor/subview.ts
nathanlesage: Gotcha, thank you!

source/common/modules/markdown-editor/table-editor/widget.ts
benniekiss: Haha it's no problem :) Even in the loop I wouldn't have expected it to have such an impact

nathanlesage:
Roger that, then I'll use the time to check the PR out locally and give it a spin.

benniekiss:
Alright, this is good on my end. Let me know if you noticed anything in testing

nathanlesage:
Nothing found, works pretty well! Let's merge this one.

ariadesu:
Is this in 4.0.0? Mine still jumps.

benniekiss:
@ariadesu please open a new issue, preferably with a sample document where this happens

PR #6068, opened 2025-12-19
[Performance] Improve citation sidebar performance

benniekiss:
## Description

This PR seeks to improve the performance of the citation sidebar by moving the ast parsing of citations into the FSAL layer.

## Changes

Similar to how Zkn links and tags are parsed at the FSAL level, an attribute was added to the `MDFileDescriptor` for citations parsed from the document content. Currently, this is solely used by the references sidebar.

The reason to move parsing into this level is because of the cascading performance impacts of calling the ast parser in the sidebar logic. Since that logic is called essentially every time the editor focus changes, the file contents are retrieved and the ast is parsed, even if the file has not changed. The parsing is also an synchronous process. Likewise, the contents retrieved are the last saved contents, so moving the logic to the FSAL parsing layer does not diminish functionality from my understanding.

I reduced the citation `.map().flatMap()` logic to one `flatMap()` to reduce iterations. ~~And OOS but in the same spirit, I reduced the footnote retrieval logic in the `md2html` function to one iteration as well.~~

## Additional information

I do not have any citation setup to actually test the performance. I made these changes based on my experience with the `markdownToAst` performance in other areas of the code, and I recognized that the references sidebar is probably impacted by it.

I would greatly appreciate it if anyone is willing to benchmark this change to see if it actually has an impact. 

 I think this change makes sense in line with how links and tags are handled, but I'm open to other thoughts.

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/fsal/util/file-parser.ts
+++ b/source/app/service-providers/fsal/util/file-parser.ts
@@ -77,7 +77,7 @@ export default function getMarkdownFileParser (
     file.links = links.map(link => link.target)
 
     const citations = extractASTNodes(ast, 'Citation') as CitationNode[]
-    file.citations = citations.flatMap(node => node.parsedCitation.items.map(item => item.id))
+    file.citekeys = citations.flatMap(node => node.parsedCitation.items.map(item => item.id))
 
     file.firstHeading = null
     const headings = extractASTNodes(ast, 'Heading') as Heading[]
--- a/source/app/service-providers/fsal/fsal-file.ts
+++ b/source/app/service-providers/fsal/fsal-file.ts
@@ -83,7 +83,7 @@ export async function parse (
     id: '', // The ID, if there is one inside the file.
     tags: [], // All tags that are to be found inside the file's contents.
     links: [], // Any outlinks
-    citations: [],
+    citekeys: [],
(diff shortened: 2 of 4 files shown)


source/app/service-providers/fsal/util/file-parser.ts
```diff
@@ -77,6 +76,9 @@ export default function getMarkdownFileParser (
     const links = extractASTNodes(ast, 'ZettelkastenLink') as ZettelkastenLink[]
     file.links = links.map(link => link.target)
 
+    const citations = extractASTNodes(ast, 'Citation') as CitationNode[]
+    file.citations = citations.flatMap(node => node.parsedCitation.items.map(item => item.id))
```
nathanlesage: I'd argue we should call this variable `citekeys` because that's essentially what we extract, not the full citations!

nathanlesage (commented):
Looks good! Only one small naming request! :)

source/app/service-providers/fsal/util/file-parser.ts
benniekiss: Totally agree! Ill push the change in a bit

nathanlesage (approved):
Wonderful, thank you!

Issue #6018, opened 2025-11-16
Switching between ToC, references, and related files panels takes considerable time


BabakFarshchian:
### Description

The panels to the right have become quite slow in the new Beta 4. Switching between them takes some time, and even sometimes hangs the whole app and I get "app not responding" message in Windows.

### Reproducing

- Open a note.
- Click on the different panes in the right-side panel, switching between ToC, references, and related files.
- Sometimes, the switching can take more than 10-12 seconds while the application becomes non-responsive.

### Zettlr version

Version 4 Beta 2 latest nightly

### Installation Method

From the Website or GitHub

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Windows 11

### Additional Information

It might be that this is a caching issue since the operation becomes faster after the first time that this delay happens.

jgsw:
I can confirm something similar with the latest Nightly, on MacOS 15.6.1. For clarity this does not occur for me in 4.0.0-Beta2, only the subsequent Nightly. 
Slowdown is not as extreme as reported above -- but there is a lockup of around 2 - 3 seconds on opening a large file (~10K words with lots of references.) I confirm that this slowdown is *much* worse the first time you switch to a tab, so is probably caching related.
I also experienced slowdowns while typing on Nightly: sometimes Zettlr becomes unresponsive for about half a second, and the words typed then appear. This was in a large file (10K words, with lots of references), but one for which 4.0.0-Beta2 is perfectly responsive.

BabakFarshchian:
I still experience this in the latest nightly. Especially the first time you click on references, ToC or related files it takes a long time, and if you click again you get a "not responding" message in the window title bar.

benniekiss:
Hi @BabakFarshchian do you still see this performance issue on `4.1.0`? If you do, has it improved at all?

BabakFarshchian:
Now it refreshes within approximately 1 second also in large files of around 7000 words. It is also quite consistent. Before it could refresh very quickly or would take several seconds. Now it seems to be 1 seconds consistently. 1 second is acceptable, although not fluent.

benniekiss:
Thank you! Other than word count, does your document have a lot of internal links (`[[links]]`), citations, or headings?

BabakFarshchian:
[...]

PR #6072, opened 2025-12-22
[Performance] Reduce filesystem operations in FSAL provider

### LINKED ISSUE #6057: [v4.0.0-beta.4] Significant CPU usage when opening document with large number of ZKN links
### Description

When opening a markdown file with a lot of ZKN links `[[ ]]` (>500), there is a constant noticeable CPU usage by Zettlr. Of course, I can't give comparable numbers on the usage, as every computer is different, but on my machine, even when idling (i.e. Zettlr is opened in the foreground, but we are not interacting), Zettlr stays at the top in task manager with 10-15% CPU utilization.

A few observations:
- When switching tabs, e.g. to a smaller (i.e. few to no links) file, the CPU usage doesn't drop.
- When Zettlr is started with a _small_ file as the first tab being opened, everything is fine.
  - However, when the first tab upon start is a _large_ file (i.e. a lot of links), the CPU usage goes up, and stays up.
- When I start Zettlr with a small file, and keep switching tabs only between two small files, the CPU spike is not observed.
  - note that there still was a tab with the _large_ file; just never being switched to
- When closing the tab of the large file, the CPU usage doesn't drop

### Reproducing

1. Make sure to start Zettlr with a small file (e.g. empty)
2. Take note of Zettlr's idle CPU usage
3. Create/Open a new file with 500+ links (see https://github.com/Zettlr/Zettlr/issues/6057#issuecomment-3634149017)
4. After saving/opening, you should already see an increase in CPU usage by Zettlr
5. Try out my observations from above

### Zettlr version

v4.0.0-beta.4

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Windows 11 Pro (10.0.26100 Build 26100

benniekiss:
## Description

This PR reduces several file system operations in the FSAL layer to improve performance

Improves and possibly closes #6057 and #6018

## Changes

~~The descriptors load in the `reindexFiles` method are now cached in a `_loadedDescriptors` attribute. This dramatically improves performance of the `findExact` method, which re-computes descriptors for all loaded files on each call. Currently, the `findExact` method is the only place where `_loadedDescriptors` is used. This was causing performance issues when a file had several zkn links, and hundreds of files were opened in Zettlr.~~ 

So I found that the `reindexFiles` method does not get called often enough, so when files are added or removed, `_loadedDescriptors` will quickly become out-of-sync with the actual list of available files. Instead, I re-implemented the `findExact` logic within the link provider. By still calling `getAllLoadedDescriptors`, the list of available descriptors is up to date, and it can be reused across function calls, greatly reducing the performance impact.

Several calls to `pathExists`, `isFile`, and `isDir` were removed because lower level functions already handled the check. Instead, they were replaced with `try...catch` blocks. This also improves performance. In my ~500 file workspace, FSAL indexing currently takes about 450ms, but with this patch, it is about 250ms to reindex.

## Additional information

~~I believe `_loadedDescriptors` could replace `getAllLoadedDescriptors` entirely, but I wasn't certain.~~

~~I tested this in a folder of about 500 files, and I did not notice an increase in RAM usage.~~

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/links/index.ts
+++ b/source/app/service-providers/links/index.ts
@@ -22,6 +22,7 @@ import type LogProvider from '@providers/log'
 import path from 'path'
 import type FSAL from '../fsal'
 import type ConfigProvider from '../config'
+import type { MDFileDescriptor } from 'source/types/common/fsal'
 
 /**
  * This class manages the coloured tags of the app. It reads the tags on each
@@ -93,6 +94,34 @@ export default class LinkProvider extends ProviderContract {
   }
 
   /**
+   * Finds the descriptor for a file query based on a ZKN id, file name, or file
+   * base path. This logic is copied from the FSAL `findExact` method to avoid
+   * the performance impacts of calling `getAllLoadedDescriptors` on every
+   * execution.
+   */
+  private async findExact (query: string, descriptors: MDFileDescriptor[]): Promise<MDFileDescriptor|undefined> {
+    const { zkn } = this._config.get()
+    const idRe = getIDRE(zkn.idRE, true)
+
+    const isQueryID = idRe.test(query)
+    const hasMdExt = hasMarkdownExt(query)
+
+    for (const descriptor of descriptors) {
+      if (isQueryID && descriptor.id === query) {
+        return descriptor
+      }
+
+      if (hasMdExt && descriptor.name === query) {
+        return descriptor
+      }
+
+      if (descriptor.name === query + descriptor.ext) {
+        return descriptor
+      }
+    }
+  }
+
+  /**
    * Retrieves a set of links to the file given as argument
    *
    * @param   {string}    sourceFilePath  The source file's path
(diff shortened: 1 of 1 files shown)


nathanlesage:
Ahh, totally forgot to comment! I had a look and from the outset it looks like a bunch of decent improvements! I'll merge this as soon as 4.0 is out; with your other PRs and a few things I have on my mind we'll be able to get out a 4.0.1 quickly after, and catch a few bugs we've overlooked for 4.0 too.

benniekiss:
sounds great.

merry xmas eve to you, as well!

nathanlesage:
Oh, yes, Merry Christmas! 🙈

source/app/service-providers/links/index.ts
```diff
@@ -134,8 +136,34 @@ export default class LinkProvider extends ProviderContract {
 
     const outboundLinks: string[] = []
 
+    const { zkn } = this._config.get()
+    const idRe = getIDRE(zkn.idRE, true)
+
+    const loadedDescriptors = (await this._fsal.getAllLoadedDescriptors())
+      .filter(descriptor => descriptor.type === 'file')
+
+    const findExact = async (query: string) => {
```
nathanlesage: Could you move this into a proper `private` function within the Link Provider? Inline functions are fine if they are small, but as they become a bit bigger, I like having them separately.

nathanlesage (changes requested):
Looks good, just one small refactor, but I appreciate the much lower disk accesses here.

source/app/service-providers/links/index.ts
benniekiss: Makes sense to me. Taken care of in the last commit

nathanlesage:
Looks good! Merci beaucoup!

PR #6080, opened 2025-12-27
feat: Move openPaths to openFiles and openWorkspaces

### LINKED ISSUE #6071: [Develop] Separate `openPaths` into `rootFiles` and `workspaces`
### Checklist

- [x] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)

### Feature Proposal

I propose to separate out files and folders that are currently still held within a single array property in the config into two separate properties, `rootFiles` and `workspaces`.

This has two primary benefits: First, it turns the implicit distinction between files and folders explicit (like it should have been for years at this point). And second, it allows us to do better housekeeping. Since some users prefer to keep workspaces on a USB key, we have to retain workspaces even if they cannot be found (so that the user can plug in the USB key which then turns those entries "active"). However, the same does not apply for files.

This actually means that many users likely have hundreds of dead entries in the `openPaths` property from files that they have since deleted. I usually don't see that because I use Zettlr "as intended" (ha-ha, I'm the intention here, I know), but recently I saw this with a few colleagues who showed me their config files.

### Example Workflow

Not applicable.

### Caveats

The tricky part will be to properly implement this without causing a ton of ghost bugs. I want to take this opportunity to think about who's responsible for maintaining the configuration key, and how the changes can be propagated. In addition, what I'd love to do would be an automatic check that, whenever a new workspace is opened, that files that are opened standalone but which are actually part of that new workspace, will be automatically "closed" from the files section.

Yes

### Platforms

- [x] Windows
- [x] macOS
- [x] Linux

### Additional Information

**Please do not attempt to do this, I need to be doing this personally.**

nathanlesage:
## Description

This PR splits up the `openPaths` property into two new properties, `openFiles` and `openWorkspaces`. This makes it easier to detect whether paths are *supposed* to be files or folders, makes consolidating simpler, and also allows for a better path management in general.

## Changes

Since this property is deeply ingrained into the code base, changing it required many changes throughout the code base. I tried to keep existing code working as-is where possible.

## Open Questions

There are still some decisions to make.

1. [x] I generally want to namespace whatever I can (simply because it logically groups preferences together), so I moved it under a new section `fileManager`. The reason is that there are a bunch of non-namespaced filemanager properties that I really want to add to a `fileManager` group. Thinking about it, I'm not sure if the workspaces/root files really belong into that property, though. <mark>Question: Would possibly a group `app` make more sense? Or is this then too close to `system`? Or should we add it to `system` and perspectively rename `system` to `app`?</mark>
2. [x] Currently, the configuration still emits the old `openPaths` change event that some parts of the app are listening to. I thought about emitting two separate events, but most of the code will just re-fetch everything and this seems to be fine. However, the question is what to call this event (I do want to move away from `openPaths` in general).

**Any suggestions?**

Tested on: macOS Tahoe 26.2

Closes #6071

### DIFF
--- a/scripts/test-gui/index.mjs
+++ b/scripts/test-gui/index.mjs
@@ -95,7 +95,7 @@ async function prepareEnvironment (argv) {
 
   info('Creating new configuration file from test-config.yml ...')
   const cfg = await makeConfig()
-  cfg.fileManager = {
+  cfg.app = {
     openFiles: files,
     openWorkspaces: workspaces
   }
--- a/source/app/service-providers/commands/dir-rename.ts
+++ b/source/app/service-providers/commands/dir-rename.ts
@@ -43,7 +43,7 @@ export default class DirRename extends ZettlrCommand {
     try {
       // Before renaming the dir, let's see if it is a workspace. Because if it
       // is, we have to close it first.
-      const { openWorkspaces } = this._app.config.getConfig().fileManager
+      const { openWorkspaces } = this._app.config.getConfig().app
       const isRoot = openWorkspaces.includes(sourceDir.path)
 
       if (isRoot) {
--- a/source/app/service-providers/commands/file-rename.ts
+++ b/source/app/service-providers/commands/file-rename.ts
@@ -108,7 +108,7 @@ export default class FileRename extends ZettlrCommand {
 
       // Before renaming the file, let's see if it is a root file. Because if it
       // is, we have to close it first.
-      const { openFiles } = this._app.config.getConfig().fileManager
+      const { openFiles } = this._app.config.getConfig().app
       const isRoot = openFiles.includes(file.path)
 
       if (isRoot) {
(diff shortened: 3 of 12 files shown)


benniekiss:
> Thinking about it, I'm not sure if the workspaces/root files really belong into that property, though

I could see them fitting under the `files` key. 

Otherwise, I think keeping a `fileManager` namespace is fine, esp given all the other `fileManager` keys

nathanlesage:
Thanks for chiming in! I'll be thinking about this a bit and then we can merge, unless I made a mistake somewhere! :)

nathanlesage:
I've now given it a thought and weighed the alternatives. I don't feel like the properties would fit under the `files` key as this is semantically very tightly bound with settings pertaining to whatever happens within the workspaces (which files are displayed, etc.). Also, I think the file manager settings should be separate from the open files and workspaces, because the open roots are really standing even above the file manager.

I have now decided to introduce an `app` namespace. This feels best for all the as-of-yet-unsorted config settings on the root level of the config. I have this hunch that `system` is still something else.

I mean at this point it's probably just voodoo and gut feelings, so as long as it works, I'm gonna leave it like that …

PR #6081, opened 2025-12-28
Update de-DE.po to fix translation

ruediger101:
- Fix incorrect translation of "column" (Spalte instead of Zeile)
- Improve consistency by changing translation of "Delete row"


nathanlesage:
Thank you!

PR #6082, opened 2025-12-29
feat: Add Visual Indicators for Long-Running Tasks

### LINKED ISSUE #5414: No visual feedback after successful project exports
### Description

When I try to print files, in plain text the end of line characters don't print, meaning all lines run together. I use tab/quote text, and the return char works fine in them.

When I try to Export a project, after projectifying, (and setting project type - ARE ALL THESE STEPS necessary, why can't I do this on -the-fly if not set) and control-clicking export, nothing happens. At all. I can export an idividual file (wich doesn't have the return chars either, but at least works).

I want to be able to print entire projects or export them (so that I can print them) but it just doesn't work.

### Reproducing

Write some verses... print them off... and you don't get the line breaks

### Zettlr version

Version 3.2.2 (3.2.2)

### Installation Method

Other (please specify below)

Can't remember... website probably

### Your Platform

- [ ] Windows
- [X] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit
- [X] ARM 64bit (includes Apple Silicon)

### Operating System Version

14.5 (23F79)

### Additional Information

_No response_

nathanlesage:
## Description

Sometimes, tasks simply can take long. This has correctly been identified and discussed at length, e.g., in issue #5414. This PR now sidesteps a bit of that thought process intentionally. The reason is that, with some recent changes, we are moving a lot of heavy lifting to different parts of the app which means that we need some more visual indications for long-running tasks. As projects become heavier and the information we are extracting from files during the caching process grows, these processes are bound to take more time. I myself have noticed that, if I open a new workspace, or export a somewhat larger project, the experienced “lag” between performing the action, and the result becoming visible is disconcerting. At the same time, I've started getting more and more reports by people noticing that the whole adding a workspace with existing files is really unsettling, because nothing seems to happen visually.

To address these more and more pressing concerns I now made a decision for the project (although there are still UX considerations). I have now implemented a minimal viable solution for the discussion in #5414 for giving users some visual feedback.

## Changes

1. Implement a provider that maintains "long-running tasks". Other commands\* such as project exports or workspace indexing can now "register" a long-running task with the provider.
2. Registering such a task will provide the callers with a "Task" object that they can interact with, and update its information and progress. Once the task is finished, the caller "ends" the task.
3. On the renderer side, a new Pinia store with accompanying toolbar button and popover now provides these tasks to the user. Users can see for how long these tasks have been running, what the current status is, and afterwards whether they completed successfully.
4. The Pinia store automatically synchronizes with the task list in the main process
5. A new "tasks" toolbar button appears when there are LRTs. When there are currently active tasks, it shows a badge.
6. When the user clicks on the toolbar button, a popover opens that shows them all tasks, their status and run time.
7. There is a button allowing the user to clear any completed tasks (which will make the toolbar button disappear if there are none left).

Quick screenshot:

image

## Additional information

LRTs are meant for informative purposes only. They literally just form a communication between processes where there simply is not much of a user feedback possible and the user. Essentially, they are logs that look a bit nicer and are less technical.

\* *Regarding the asterisk under "Changes"*: It makes sense to have these LRT respond to user actions. However, one long-running task in response to user actions is the reindexing triggered by the FSAL when the user adds a new workspace. This is weird, and it shows because the LRT provider should really not be provided to any PROVIDERS, but should rather only be used in the COMMANDS section of the main process. This shows how the line between a provider action and user action can sometimes be blurry. It would be great if we could move the reindexing LRT-logic into the corresponding `root-open` command. But that would be a somewhat bulky refactor we likely need to do in a separate PR. My dream would be that any reindexing at the start of the app is shown in the splash screen (as is the case right now), and any additional reindexing, which really just happens on adding new workspaces is exclusively triggered by the accompanying command. However, since we have a very event-driven conceptualization of the entire filesystem interaction, it's difficult, because the FSAL currently only passively responds to changes to the config, which ALSO directly influences the file manager's display of workspaces. It's a bit messy process-wise.

Tested on: macOS Tahoe 26.2

Closes #5414.


nathanlesage:
There it is — more than 1,000 lines of code for 576 pixels of color:

image

nathanlesage:
@benniekiss Would you be willing to do a review of this PR? I think it's pretty complete now. (You can ignore the iris indicator code, that already went through… a lot :D)

benniekiss:
I am quite busy the next two weeks, but I'll see if I can get any comments in in the meantime

nathanlesage:
No stress!

PR #6084, opened 2026-01-01
[FIX] fix snippet test YEAR bug

benniekiss:
## Description

Happy new year, this PR fixes a bug in the snippet test code where the year number was hardcoded to `25`.

## Changes

Use variable substitution to determine the year dynamically in the variable-replacement snippet test.

## Additional information

I had already added the variable `YEAR`, but did not actually implement it in the test string :P

Tested on:


nathanlesage:
Ahaha, I love it :D 

Thanks, and happy new year!

PR #6086, opened 2026-01-02
Update ja-JP.po 2026-01-02

coolvitto:
Compatible with version 4.0.0

## Additional information

Tested on:

### DIFF
--- a/static/lang/ja-JP.po
+++ b/static/lang/ja-JP.po
@@ -3,39 +3,40 @@
 # This file is distributed under the same license as the Zettlr package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 #
+#, fuzzy
 msgid ""
 msgstr ""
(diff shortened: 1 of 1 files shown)


coolvitto:
By the way, I would like to check the translation content on the app, but if it will take some time until the next release, is it possible to download just the updated ja_JP.pak?

nathanlesage:
I just checked the code, and unfortunately the old translation import code won't work anymore…

What you could do is set up everything locally and launch the app from there with your updates in place. However, I'll be releasing a new update soon!

nathanlesage:
Ah, also what you can do is use the nightly! A new nightly will build in about 10 minutes, so it will be available in circa one hour at https://nightly.zettlr.com! I'll merge this — thank you for improving the translation!

coolvitto:
I have confirmed that the Japanese version has been updated successfully. 
Thank you very much !!!

Issue #6087, opened 2026-01-02
Consider building Windows on ARM / WOA / Arm64) support


stucka:
### Description

More computers are getting built with Snapdragon chips, which have Windows running on Arm64. These include some Microsoft Surface products I think have been out for years, but these chips are increasingly showing up in more mainstream locations like Best Buy and Costco.

There are amd64 versions being built for Windows. There are arm64 versions for Linux and Mac. There are some 32-bit versions for several operating systems. But there are no versions for Windows on Arm (WOA).

The build pipeline might be relatively easy to adjust, but there may be some code peculiarities that need to get patched, e.g.:
https://www.electronjs.org/docs/latest/tutorial/windows-arm

### Proposed Changes

Build for Windows on ARM64.

### Caveats

I am largely ignorant here.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Operating System Version

Windows 11 on Arm64

PR #6089, opened 2026-01-03
Document available win-arm processes

stucka:
README listed five older OS-CPU combos but not Windows on ARM.

## Description

Documents in README.md the existence of the win-arm builds.

## Changes

Documentation only.

## Additional information

Tested on: ... Windows 11 on ARM 64.  =)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   collecting citation keys from the document only in the main process.
 - Improved math parsing to fix certain cases (#6030).
 - Update `de-DE` translation (#6081).
+- Documented in README.md the existence of `win-arm` build routines.
 
 ## Under the Hood
 


stucka:
For the record, this would have precluded me from opening and then closing #6087 =)

nathanlesage:
Does the build-process work on an ARM computer, including the dictionaries? I'm asking because that's why we have stopped shipping Windows on ARM — because we couldn't compile the binaries.

Also, please don't modify the Changelog for changes in the README file, other than that, if it works, we can re-add the info!

stucka:
@nathanlesage , my apologies -- I missed your reply here.

Yes, it appears to build completely. Just not properly. I'm going to open an issue on the other problems and move this one back to draft on the premise you don't want to tell people to use the flags if they're not ready yet.

I think I changed the changelog based on something I saw saying every change should be noted in the changelog. If readmes aren't substantive enough, that language might could use a review.

stucka:
Documented in https://github.com/Zettlr/Zettlr/issues/6119

nathanlesage:
If you can get together a documentation outlining how to build on ARM I'm happy to re-document and fix issues in this regard!

stucka:
I will look! I'm well outside my usual range on this, but I will look. =)  Thanks for all you do!

Issue #3241, opened 2022-02-20
[Enhancement] Remember collapsed title


gautierlabarre:
### Description

Hi, 

When you open a file and collapse some titles, close the file and reopen the file, it would be nice to have all the previously collapsed sections still collapsed.

On big files it can really improve the readability, and avoid scrolling too much if you open/close this file frequently.

### Proposed Changes

Whenever a file is open, Zettlr search for collapsed title and collapse them automatically.
There could be an option in the settings to avoid this behavior, if people don't like it :) 
Also a shortcut could be nice to collapse/open all titles at once

### Caveats

Only caveat i can see is, if it's "poorly" implemented the content might flicker like : display/hide really fast. Zettlr should display content once it has already collapsed sections.

### Do you Wish to Attempt Implementing this Yourself?

No

### Zettlr Version

Stable (most recent version)

### Your Platform

- [ ] Windows
- [ ] macOS
- [X] Linux

cjpoor:
This enhancement would make Zettlr more useful for long-form writing. Even better would be the implementation of shortcuts which fold the text to preset levels as in [org-mode](https://orgmode.org/worg/org-tutorials/org4beginners.html#orgb94f796).

cjpoor:
This issue of remembering the fold (etc) status of a file might be addressed more simply by the use of ["session" files as used by Vim](https://vimhelp.org/starting.txt.html#views-sessions)

cjpoor:
VS Code, using electron, also has the feature of [remembering fold status](https://github.com/microsoft/vscode/issues/26157).

wosym:
I've only recently started using Zettlr, but yes, this is indeed one of the things that I'm missing desperately. Wouldn't say it's making it a showstopper for me, causing me to abandon Zettlr for some other tool, but... it's pretty close :) 

If you have huge files, with lots of logically organized "areas" of information, then it can become very inefficient if you can't have all "areas" folded. Of course, you can fold them yourself, but if you'd have to do that each time, then this again is a big waste of time. Especially if you have hundreds of "areas". 

Alternatives to remembering the collapsed title would be:
1) implementing an option to "fold-all" (most modern editors have this)
2) Having setting that configures the default fold-mode when opening a file: folded/unfolded.

whereswaldodude:
This is really a killer issue not to have. Doesn't remember state, ok, I get that would take a bit more. No collapse all to top level right click menu? That would take, what, a day at the very most to implement?

[...]

PR #6115, opened 2026-01-08
[enhancement] improve code folding

benniekiss:
## Description

This PR adds code folding to YAML and PandocDiv nodes, and it preserves folded regions across editor focus changes

## Changes

A `foldNodeProp` was added to the markdown parser, which calculates fold regions for pandoc divs and yaml frontmatter.

A `foldedRanges` attribute was added to `EditorViewPersistentState`, which holds a decoration set of the currently folded ranges in the editor. When the document is loaded, the range set is iterated over, and a fold effect is dispatched for each range. This preserves the folded ranges across editor focus changes.

The styling for the `foldPlaceholder` was moved to the `main-overrides` file, and the background was made transparent. Previously, the background color was set to match the old editor background color, but that was updated in v4, and now the fold placeholder is out-of-sync. Making it transparent removes the need to keep it in sync. A border element was added to the dark mode placeholder, and this matches the light mode styling.

## Additional information

### DIFF
--- a/source/common/modules/markdown-editor/theme/main-override.ts
+++ b/source/common/modules/markdown-editor/theme/main-override.ts
@@ -129,7 +129,6 @@ export const mainOverride = EditorView.baseTheme({
   '.cm-foldPlaceholder': {
     backgroundColor: 'transparent',
   },
-  // Shown when a region is folded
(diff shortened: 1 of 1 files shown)


nathanlesage:
This appears to fix #3241.

PR #6120, opened 2026-01-10
[Fix] Pandoc div fixes

benniekiss:
## Description

This PR fixes some issues with pandoc div parsing

## Changes

The pandoc attribute regex was loosened to reduce complexity and reduce false-negative attributes

The nesting level for pandoc divs was refactored into a separated function, `getNestingLevel`. This is needed to accurately reflect the nesting level when other block elements are in the parent tree. It resolves issues where the closing mark, `:::`, was not parsed correctly following elements such as list items and blockquotes.

The TOC plugin was changed so that `PandocDiv` nodes are recursed in addition to top-level `Document` nodes. This fixes an issue where headings within pandoc divs were not included in the TOC.

## Additional information

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/plugins/toc-field.ts
+++ b/source/common/modules/markdown-editor/plugins/toc-field.ts
@@ -127,7 +127,7 @@ function generateToc (state: EditorState): ToCEntry[] {
 
   tree.iterate({
     enter (node) {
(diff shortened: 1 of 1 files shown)


nathanlesage:
Looks good, merci!

PR #6121, opened 2026-01-10
[Enhancement] re-enable pandoc div rendering

benniekiss:
## Description

This PR re-enables pandoc div rendering by using the new [`BlockWrapper`](https://codemirror.net/docs/ref/#view.BlockWrapper) api 

## Changes

The pandoc div and span renderers were separated because `BlockWrappers` have a similar, but different, structure and typing than `Decorations`.

The div renderer was updated to use `BlockWrapper`s, which essentially wraps the given range in a separate HTML element. This provides better styling than `Decoration.line` because the entire content is wrapped in one parent element.

An additional blockwrapper was added around the opening and closing lines of the pandoc div to essentially revert the styling of parent divs. I did this so that those regions do not receive *most* styling from the parent pandoc div, making them easier to pick out when div rendering is enabled. This is a very simple override, and there is still some styling that applies. Another approach would be to use a widget across these regions, but that would have to be applied in a separate renderer function since those are applied as `Decoration` objects.

## Additional information

Screenshot 2026-01-10 at 13 04 10

*the `lightGreen` section is a blockquote, which is why the font color is different*

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
+++ b/source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
@@ -113,12 +113,13 @@ function createDivDecorations (view: EditorView): RangeSet<BlockWrapper> {
           return
         }
 
+        const classes = info ? [view.state.sliceDoc(info.from, info.to)] : []
+
         const attributes = attrs ? parsePandocAttributes(view.state.sliceDoc(attrs.from, attrs.to)) : {}
-        const classes = attributes.classes ?? []
         const id = attributes.id ?? ''
 
-        if (info) {
-          classes.unshift(view.state.sliceDoc(info.from, info.to))
+        if (attributes.classes) {
+          classes.push(...attributes.classes)
         }
 
         const wrapper = BlockWrapper.create({


benniekiss:
Column rendering almost works. However, It messes with the cursor positioning below the rendered columns, probably due to using `flex` containers

Screenshot 2026-01-10 at 18 53 00

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -101,35 +62,127 @@ function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
...
-  // The classes `.mark`, `.underline`, and `.smallcaps` are used by pandoc
   EditorView.baseTheme({
+    // This must be set to `display: block` so that the
+    // attributes are applied correctly.
+    'pandoc-div-wrapper': {
+      display: 'block !important',
+      // Prevent user-provided styling from altering the dom-layout
+      flex: 'initial !important',
+      height: 'initial !important',
+      width: 'initial !important',
```
nathanlesage, line 187: Can't we get rid of the `!important`? We are already warning users that they need to be careful with Custom CSS, so I see no need to have that here. In addition, since the user CSS is always inserted last, it will always be able to override any styles we have anyways.

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -7,21 +7,21 @@
...
 import { syntaxTree } from '@codemirror/language'
 import type { Range, RangeSet } from '@codemirror/state'
-import { Decoration, EditorView, ViewPlugin, type DecorationSet, type ViewUpdate } from '@codemirror/view'
+import { BlockWrapper, Decoration, EditorView, ViewPlugin, type DecorationSet, type ViewUpdate } from '@codemirror/view'
 import { parsePandocAttributes } from 'source/common/pandoc-util/parse-pandoc-attributes'
 import { rangeInSelection } from '../util/range-in-selection'
-import type { SyntaxNode } from '@lezer/common'
 
-function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
+function showSpanDecorations (view: EditorView): RangeSet<Decoration> {
```
nathanlesage: Suggestion: This doesn't "show" inasmuch as it "creates" or "computes" these decorations, so I'd argue we should rename this function accordingly.

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -101,35 +62,127 @@ function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
...
         ranges.push(deco.range(from, to))
-      },
+      }
     })
   }
 
   return Decoration.set(ranges, true)
 }
 
+const overrideWrapper = BlockWrapper.create({ tagName: 'pandoc-div-info-wrapper' })
```
nathanlesage: Why would you call it `overrideWrapper`…?

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -101,35 +62,127 @@ function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
...
+
+function showDivDecorations (view: EditorView): RangeSet<BlockWrapper> {
```
nathanlesage: Same as with the "show" spans function above

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -101,35 +62,127 @@ function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
...
+          return
+        }
+
+        const attributes = attrs ? parsePandocAttributes(view.state.sliceDoc(attrs.from, attrs.to)) : {}
+        const classes = attributes.classes ?? []
+        const id = attributes.id ?? ''
+
+        if (info) {
+          classes.unshift(view.state.sliceDoc(info.from, info.to))
+        }
```
nathanlesage: This looks a bit … dirty(?) to me? Are info strings contained within the attributes?

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -101,35 +62,127 @@ function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
...
+
+        const wrapper = BlockWrapper.create({
+          tagName: 'pandoc-div-wrapper',
```
nathanlesage, line 126: Any benefits to using a custom tag instead of a `div`, except increased semantic meaning? Are there any downsides we should keep in mind?

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
```diff
@@ -101,35 +62,127 @@ function showDivSpanDecorations (view: EditorView): RangeSet<Decoration> {
...
+          },
+        })
+
+        // Only style the content within the marks
+        const fromLine = view.state.doc.lineAt(node.from)
+        const toLine = view.state.doc.lineAt(node.to)
+
+        ranges.push(wrapper.range(fromLine.to + 1, toLine.from - 1))
+        ranges.push(overrideWrapper.range(fromLine.from, fromLine.to))
+        ranges.push(overrideWrapper.range(toLine.from, toLine.to))
```
nathanlesage: Do I understand it correctly that you're creating two nested divs, one outside, and one inside? Do I understand it correctly that the `overrideWrapper` effectively just overrides styles based on the classes and attributes, whereas the `wrapper` itself styles the info string…? If so, we may want to think about the naming of these two to ensure we won't get confused in 6 months!

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
benniekiss: This is to prevent css declared in the div attribute from altering the dom. Sometimes the attributes are auto-generated or are required for proper rendering in pandoc, but we dont want them taking effect in the editor renderer

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
benniekiss: It overrides styling of the parent div, but I could try and come up with a better name

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
benniekiss: no, info strings are parsed separately, same as code-info strings:

```
::: info {#id}
...
:::
```

The class needs to be inserted first to take priority

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
benniekiss: I dont actually know if setting `tagName: 'div'` would create a regular div... I'll test, but my instinct is to keep it a separate element name to make it clear what is providing the wrapper

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
benniekiss: I'll clarify this in the code, but this essentially creates three sibling divs:

```
::: {#id}     <-- first (overrideWrapper)

...content... <-- second

:::           <-- third (overrideWrapper)

```

Two reasons for this:

The first reason is to remove any styling from the mark nodes (`:::`) when they are nested. This makes it easier to read and find the boundaries of nested divs.

The second is that applying the styling over just the content makes it easier to see how the styling is rendered. This is most evident when using something like 

```
::: {style="background-color: yellow; border-radius: 10px;"}

content

:::
```

The rounded corners will appear within the div marks. Otherwise, they would be rounded at the divmarks, and since background styling is removed from them, the rounded corners would not be visible.

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
nathanlesage: Aaaaaah, oh god, I didn't realize this is merely the string classes. I'd still argue that a cleaner pattern is to assign the info-class first, and then assign the attribute classes using a spread operator afterward. This ensures the insertion logic is the "right way"

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
nathanlesage: Got it!

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
nathanlesage: Fair, I understand that, thank you!

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
benniekiss: renamed to the more descriptive `pandocDivMarkWrapper`

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
nathanlesage: Gotcha, this makes more sense now!

source/common/modules/markdown-editor/renderers/render-pandoc-div-span.ts
nathanlesage: Thank you, but with your explanation overrideWrapper made absolute sense, too. But I guess since it wasn't immediately obvious, this rename might be warranted :)

nathanlesage:
Looks good now, ready for merging, thank you very much once more!

PR #6122, opened 2026-01-10
[Enhancement] Render blockquotes with a vertical bar

### LINKED ISSUE #5960: Disable italic rendering in quotation environnement of Berlin, Frankfurt & Karl-Marx-Stadt

benniekiss:
## Description

This PR adds a blockquote renderer which adds a vertical bar, indicating quote level. 

Closes #5960

## Changes

A renderer was added for blockquotes, which wraps the node in a `BlockWrapper` element. This renderer is enabled with `renderEmphasis` to tie in with the hidden quotemarks.

Theme styling was updated so that the `primaryColor` is assigned to the blockquote border color.

`QuoteMarks` are now replaced with a `hiddenDeco` rather than the `SpaceWidget`, and the hidden range was increased to account for trailing whitespace. 

The `SpaceWidget` had the type of its `eq` method fixed, and a class is now added, `rendered-space`, matching the `rendered-bullet` class of the `BulletWidget`.

## Additional information

Screenshot 2026-01-10 at 13 19 22

Screenshot 2026-01-10 at 13 19 28

### DIFF
--- a/source/common/modules/markdown-editor/renderers/render-blockquotes.ts
+++ b/source/common/modules/markdown-editor/renderers/render-blockquotes.ts
@@ -2,14 +2,13 @@
  * @ignore
  * BEGIN HEADER
  *
- * Contains:        renderPandoc
+ * Contains:        renderBlockquotes
  * CVM-Role:        View
  * Maintainer:      Bennie Milburn
  * License:         GNU GPL v3
  *
- * Description:     This renderer displays Pandoc divs and spans using
- *                  Decorations, rendering the attributes defined for
- *                  the node as they would be displayed by pandoc
+ * Description:     This renderer adds a vertical bar
+ *                  to the left edge of blockquotes
  *
  * END HEADER
  */


nathanlesage:
This looks pretty nifty, I must say! I want to let that simmer for a bit and release a 4.1 first (and fix our code signing), just to be on the safe side since this does quite a bit to the rendering. But I'm actually looking forward to that!

nathanlesage (approved):
Looks good, we can merge this! :)

PR #6125, opened 2026-01-11
[Enhancement] Custom default profile improvements

benniekiss:
## Description

This PR enables setting custom lua readers in default profiles, and it allows specifying custom output and input file locations using `output-file` and `input-files`.

## Changes

The profile reader filter was extended to allow readers ending in `.lua`. This enables support for custom readers 

When generating the defaults file for export, the plugin values no longer override those already set in the profile. From what I could find, the only plugin values provided are `input-files` and `output-file`. I ran into this problem because I have an export profile that uses a custom `output-file` location, but since it was getting overriden, the profile was not working correctly.

## Additional information

Tested on:  macOS 26

### DIFF
--- a/source/app/service-providers/commands/exporter/default-exporter.ts
+++ b/source/app/service-providers/commands/exporter/default-exporter.ts
@@ -39,7 +39,6 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions, 
   }
   const defaults = await ctx.writeDefaults(options.profile.name, defaultKeys)
 
-  console.log('OG TARGET: ', target)
   // Check that the defaults profile did not change
   // `output-file`, and if it did, update the target.
   if (defaults['output-file'] !== target) {
@@ -52,7 +51,6 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions, 
     }
   }
 
-  console.log('TARGET: ', target)
   // Run Pandoc
   const pandocOutput = await ctx.runPandoc(defaults)
 
--- a/source/app/service-providers/commands/exporter/default-exporter.ts
+++ b/source/app/service-providers/commands/exporter/default-exporter.ts
@@ -30,17 +30,31 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions, 
   // First file determines the name of the exported file.
   const firstName = path.basename(options.sourceFiles[0].name, options.sourceFiles[0].ext)
   const title = (options.defaultsOverride?.title !== undefined) ? sanitize(options.defaultsOverride.title, { replacement: '-' }) : firstName
-  const target = path.join(options.targetDirectory, `${title}.${extension}`)
+  let target = path.join(options.targetDirectory, `${title}.${extension}`)
 
   // Get the corresponding defaults file
   const defaultKeys = {
     'input-files': sourceFiles,
     'output-file': target
   }
-  const defaultsFile = await ctx.writeDefaults(options.profile.name, defaultKeys)
+  const defaults = await ctx.writeDefaults(options.profile.name, defaultKeys)
 
+  console.log('OG TARGET: ', target)
+  // Check that the defaults profile did not change
+  // `output-file`, and if it did, update the target.
+  if (defaults['output-file'] !== target) {
+    const parsed = path.parse(defaults['output-file'] as string)
+    target = path.join(parsed.dir, parsed.name + `.${extension}`)
+
+    if (!path.isAbsolute(target)) {
+      target = path.join(options.targetDirectory, target)
+      defaults['output-file'] = target
+    }
+  }
+
+  console.log('TARGET: ', target)
   // Run Pandoc
-  const pandocOutput = await ctx.runPandoc(defaultsFile)
+  const pandocOutput = await ctx.runPandoc(defaults)
 
   // Make sure to propagate the results
   return {
--- a/source/app/service-providers/commands/exporter/default-exporter.ts
+++ b/source/app/service-providers/commands/exporter/default-exporter.ts
@@ -39,16 +39,25 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions, 
   }
   const defaults = await ctx.writeDefaults(options.profile.name, defaultKeys)
 
-  // Check that the defaults profile did not change
-  // `output-file`, and if it did, update the target.
+  // Update `target` if the defaults profile changed`output-file`
   if (defaults['output-file'] !== target) {
-    const parsed = path.parse(defaults['output-file'] as string)
-    target = path.join(parsed.dir, parsed.name + `.${extension}`)
+    // Remove any internal `..` and `.` paths
+    target = path.normalize(defaults['output-file'] as string)
 
+    // If the target is a relative path, resolve it to the target directory
+    // and sanitize any potential path traversals.
     if (!path.isAbsolute(target)) {
-      target = path.join(options.targetDirectory, target)
-      defaults['output-file'] = target
+      target = path.resolve(options.targetDirectory, target)
+
+      // Make sure that the resolved path still falls under `targetDirectory`
+      // to prevent potentially insecure path traversals.
+      if (!target.startsWith(options.targetDirectory)) {
+        const parsed = path.parse(target)
+        target = path.join(options.targetDirectory, parsed.base)
+      }
     }
+
+    defaults['output-file'] = target
   }
 
   // Run Pandoc
--- a/source/app/service-providers/commands/exporter/default-exporter.ts
+++ b/source/app/service-providers/commands/exporter/default-exporter.ts
@@ -37,7 +37,7 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions, 
     'input-files': sourceFiles,
     'output-file': target
   }
-  const defaults = await ctx.writeDefaults(options.profile.name, defaultKeys)
+  const defaults = await ctx.loadDefaults(options.profile.name, defaultKeys)
 
   // Update `target` if the defaults profile changed`output-file`
   if (defaults['output-file'] !== target) {
--- a/source/app/service-providers/commands/exporter/default-exporter.ts
+++ b/source/app/service-providers/commands/exporter/default-exporter.ts
@@ -17,6 +17,7 @@ import sanitize from 'sanitize-filename'
 import type { ExporterOptions, ExporterPlugin, ExporterOutput, ExporterAPI } from './types'
 import { WRITER2EXT } from '@common/pandoc-util/pandoc-maps'
 import { parseReaderWriter } from '@common/pandoc-util/parse-reader-writer'
+import normalizePath from 'source/common/util/normalize-path'
 
 export const plugin: ExporterPlugin = async function (options: ExporterOptions, sourceFiles: string[], ctx: ExporterAPI): Promise<ExporterOutput> {
   if (typeof options.profile === 'string') {
@@ -41,23 +42,7 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions, 
 
   // Update `target` if the defaults profile changed`output-file`
   if (defaults['output-file'] !== target) {
-    // Remove any internal `..` and `.` paths
-    target = path.normalize(defaults['output-file'] as string)
-
-    // If the target is a relative path, resolve it to the target directory
-    // and sanitize any potential path traversals.
-    if (!path.isAbsolute(target)) {
-      target = path.resolve(options.targetDirectory, target)
-
-      // Make sure that the resolved path still falls under `targetDirectory`
-      // to prevent potentially insecure path traversals.
-      if (!target.startsWith(options.targetDirectory)) {
-        const parsed = path.parse(target)
-        target = path.join(options.targetDirectory, parsed.base)
-      }
-    }
-
-    defaults['output-file'] = target
+    defaults['output-file'] = normalizePath(defaults['output-file'] as string, options.targetDirectory)
   }
 
   // Run Pandoc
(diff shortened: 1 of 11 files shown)


benniekiss:
I made some more changes to improve the defaults writing.

Instead of passing the filename between `writeDefaults` and `runPandoc`, the raw object is shared between function calls, and `runPandoc` now handles the temporary file write. This enables more dynamic default setting, especially for `output-files`. 

I also corrected a lot of `any` types by writing a couple bespoke ones and reusing them in several places.

This also improves the pandoc stdout/stderr output logging by joining the output with newlines, making it far more readable. I havent yet tested the `stderr` output, though.

benniekiss:
Also, I think it may be worth considering filling out and documenting the `metadata.zettlr` key. I think this could be used to override the top-level default settings, so users do not have to write custom export profiles, and they can instead declare overrides in the yaml frontmatter -- this was inspired by some discord comments

nathanlesage:
> Also, I think it may be worth considering filling out and documenting the `metadata.zettlr` key. I think this could be used to override the top-level default settings, so users do not have to write custom export profiles, and they can instead declare overrides in the yaml frontmatter -- this was inspired by some discord comments

To respond to this specifically: I deliberately opted not to do so for now, because I want to avoid growth of options here. Also I am not happy with the existing choices. I would want to have a plan for what to support, maybe taking inspiration from RMarkdown and Quarto for this. Then we can also document this, but it would be more deliberate. The current options are mere fixes for weird issues we had in the past.

benniekiss:
~~I don't think quarto or rmarkdown support pandoc's default file variables in the frontmatter, but they do support some additional custom variables.~~ Actually, reading the discussion below seems to indicate quarto does support this? I just tried looking through the documentation to find any explicit mention, but did not find anything. EDIT: Ok, they have them documented under each format here: https://quarto.org/docs/reference/formats/html.html

This discussion is worth reading through in terms of allowing yaml-frontmatter to control the defaults https://github.com/jgm/pandoc/issues/5870

However, the pandoc support aspect is unimportant for zettlr, since we can parse the yaml and apply it to a defaults file before even running pandoc.

nathanlesage:
Yes, that mess is precisely why I have tried to avoid exposing too many Zettlr-specific frontmatter variables -- I saw how messy customization in the frontmatter has gotten with both RMarkdown and Quarto.

I think the main problem really is that Pandoc uses defaults file metadata first and then frontmatter variables second. I think this is due to a fundamental problem to decide what purpose both defaults files and YAML frontmatter serve.

RMarkdown and Quarto have -- in my opinion -- chosen the worst possible solution and just have added a ton of YAML frontmatter configurability, completely sidestepping the defaults file mechanism.

In an ideal world, a defaults file should specify options for the exported format AND provide defaults (for metadata) that can be overridden by a frontmatter.

I think the simplest way would be to just check the frontmatter and, if there is a metadata variable that is also defined in the defaults file, delete that value from the defaults file so that the YAML takes precedence. I think this reversal of precedence for the metadata values would already make it much easier to deal with exports for the users.

Then, the same for any other values that Zettlr provides, such as in files and out file name.

benniekiss:
> I think the simplest way would be to just check the frontmatter and, if there is a metadata variable that is also defined in the defaults file, delete that value from the defaults file so that the YAML takes precedence. I think this reversal of precedence for the metadata values would already make it much easier to deal with exports for the users.

I think this is the right approach here, as well. I'm fine implementing that in this PR, as it would rely on some of the changes I've already introduced.

An additional helpful feature would be an autocomplete plugin to provide autocomplete for the available metadata/default file options, and this could also be added to the export/import profile editor

source/app/service-providers/commands/exporter/index.ts
```diff
@@ -90,11 +90,11 @@ export async function makeExport (
 
   // This is basically the "plugin API"
   const ctx: ExporterAPI = {
-    runPandoc: async (defaults: string) => {
+    runPandoc: async (defaults: PandocDefaults) => {
       return await runPandoc(logger, defaults, options.cwd)
     },
-    writeDefaults: async (filename: string, overrides: Record<string, unknown> = {}) => {
-      return await writeDefaults(filename, overrides, config, assets, options.defaultsOverride)
+    writeDefaults: async (filename: string, overrides: PandocDefaults = {}) => {
```
nathanlesage: Since this function no longer writes the defaults, maybe rename to `loadDefaults`?

source/app/service-providers/commands/exporter/default-exporter.ts
```diff
@@ -30,17 +30,29 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions,
...
+  const defaults = await ctx.writeDefaults(options.profile.name, defaultKeys)
+
+  // Check that the defaults profile did not change
+  // `output-file`, and if it did, update the target.
+  if (defaults['output-file'] !== target) {
+    const parsed = path.parse(defaults['output-file'] as string)
+    target = path.join(parsed.dir, parsed.name + `.${extension}`)
+
+    if (!path.isAbsolute(target)) {
+      target = path.join(options.targetDirectory, target)
```
nathanlesage: Doesn't this open the route to path escape attacks? I'd suggest explicitly extracting `path.basename(target)` here.

source/app/service-providers/commands/exporter/pdf-exporter.ts
```diff
@@ -45,10 +45,24 @@ export const plugin: ExporterPlugin = async function (options: ExporterOptions,
...
+  // Check that the defaults profile did not change
+  // `output-file`, and if it did, update the target.
+  if (defaults['output-file'] !== pdfFilePath) {
+    const parsed = path.parse(defaults['output-file'] as string)
+    pdfFilePath = path.join(parsed.dir, parsed.name + '.pdf')
+    htmlFilePath = path.join(parsed.dir, parsed.name + '.html')
+
+    if (!path.isAbsolute(pdfFilePath)) {
+      pdfFilePath = path.join(options.targetDirectory, pdfFilePath)
+      htmlFilePath = path.join(options.targetDirectory, htmlFilePath)
```
nathanlesage: Same as above

source/win-assets/DefaultsTab.vue
```diff
@@ -152,7 +152,7 @@ const visibleItems = computed(() => {
       // Retrieve which one we need to check
       const readerWriter = (props.which === 'import') ? e.writer : e.reader
       const parsedReaderWriter = parseReaderWriter(readerWriter)
-      return SUPPORTED_READERS.includes(parsedReaderWriter.name)
+      return parsedReaderWriter.name.endsWith('.lua') || SUPPORTED_READERS.includes(parsedReaderWriter.name)
```
nathanlesage: Maybe we could add a new property on the parsed reader/writer to indicate if this is a custom reader/writer based on a Lua filter? I guess we should be able to use this info a few more times.

nathanlesage (commented):
Looks good so far! But one question: Could you explain to me what the `output-file` change is doing here exactly? And what you're intending with this? I am somehow having a hard time understanding the diff (but that's likely just me). Thank you!

benniekiss:
So currently, defaults files don't allow setting an explicit output-file key because it's always overridden by the plugin values.

The plugin output-file value is also used as the `target`, which is a path passed at the end of the pandoc processing to open the file after exporting. Since `target` is generated before the defaults file is parsed, we need to update the value if an explicit output-file was provided in the defaults, or else the post processing will get a non existent filename. 

The changes here make sure that the `target` value is aligned with `output-file`

source/win-assets/DefaultsTab.vue
benniekiss: Makes sense to me!

nathanlesage:
Ahhh true, now the additional handler in the `writeDefaults` section makes sense. Although I would argue we should always forcefully overwrite the input files, because if the user even sets this to an empty value, the current implementation would just error out. And since the input files are always dynamically computed, I'd exempt it from the override-only-when-non-existant.

benniekiss:
that makes sense to me! I'll have that patched in a bit, along with your other comments

benniekiss:
I need to do some testing with the pdf exporter, but this is almost there! I added the `isCustom` attribute to the reader/writer parser, which is `true` if the reader/writer name ends in `.lua`. 

Also, I believe I've sanitized `../../` instances in paths to limit path-traversal. I don't believe this is such a big security issue, since defaults files are input by the user, and writing to arbitrary locations should be allowed IMO. However, I see the most benefit when using a temporary directory for export, as this could *maybe* cause unexpected issues if it tried to write to a parent path

nathanlesage (commented):
Looks good and clean, I think it can be merged.

Just one concern regarding UX: Do you think it would be good to announce to people if the profile has an `output-file` set? I'm asking because that would override the directory choice in the exporter popover. Maybe a small admonition saying something along the lines of "Warning: This profile specifies an output-file. The directory setting will have no effect."?

benniekiss:
I think thats a great idea.

Maybe an admonition both on the profile editor and in the export popover?

nathanlesage:
My, I'm not sure if the profile editor is necessary, since there people would be able to find it, but in the export popover it would directly affect a visible setting.

Also, maybe in the export selector in the project preferences (something like "At least one profile you selected specifies an output-file").

I think those warnings make most sense where people select them, not necessarily where they edit them.

benniekiss:
Here's what the admonitions look like, I think the text could be made a little bigger, and I'm not sure about using `info` vs `warning` for the admonition.

Screenshot 2026-06-27 at 14 43 23
Screenshot 2026-06-27 at 14 43 38

I'll note that the path that is displayed is not sanitized as it is when the export actually happens. I'm undecided if it should show the raw path, or the one that is sanitized -- if it should show the sanitized path, it'll require a small refactor to extract the sanitation from the export command to its own function.

benniekiss:
The target directory isn't available outside of the export function context, and on the surface it seems quite intensive to refactor it to make it available elsewhere, so I think showing the raw path, un-normalized and as it appears in the profile, is fine.

I still refactored the path normalization to its own function, just to dedupe some logic and make it available in case it needs to be used for user-facing items in the future

benniekiss:
I think it would make sense to disable the output folder selection if the output path is absolute (in the export popover)

nathanlesage:
> I think it would make sense to disable the output folder selection if the output path is absolute (in the export popover)

I just realized that as well when looking at your screenshot.

I don't think we need to actually show the user a path here, it's sufficient to tell them *why* the output controls are disabled. If they don't know what this is, they'll probably directly go to the export profiles settings and look it up there.

Regarding the info vs warning, I first was in the warning-camp. However, an info might be more warranted, since in most cases this should be expected, and this should merely form a "heads up!" so the user knows what's going to happen.

benniekiss:
> I first was in the warning-camp. However, an info might be more warranted, since in most cases this should be expected, and this should merely form a "heads up!" so the user knows what's going to happen.

This was my line of thinking too. A warning makes it seems like something is wrong, and this is more to remind the user about a currently-unseen setting

benniekiss:
I think keeping the output path in the admonition is better than leaving it out. I played around with a few compositions, and without the path, it leaves more questions about what I'm being told, IMO.

I added another warning admonition if the target directory is disabled. It felt a little more appropriate to me to make it a warning, as functionality is disabled because of the setting.

I am not tied to the copy in the admonitions. I had a hard time coming up with something clear and concise

Screenshot 2026-06-28 at 08 44 41
Screenshot 2026-06-28 at 08 45 31

benniekiss:
Just realized one last thing, this only supports custom lua writers, but it should probably support custom lua readers, as well.

nathanlesage:
We definitely should only have one admonition with info. And the reason I'm hesitant with the path is if it's really long. I don't want the info to become longer than the rest of the popover.

Also one thing: I think the admonition should go below the rule to make it appear adjacent to the setting instead of the profile. Makes more sense.

I'll give this and the other PR a review hopefully tonight, then I think we can merge!

benniekiss:
> And the reason I'm hesitant with the path is if it's really long. 

Long filenames can wrap with `word-break: break-all`, and an example is below.

> Also one thing: I think the admonition should go below the rule to make it appear adjacent to the setting instead of the profile. Makes more sense.

Makes sense!

I still think the output path should be included. If there's just `Directory selection is disabled because output file path is absolute` begs the question, what is the path? And now users have to open the assets window, open the profile, and search through the profile to find `output-file` to get the setting. I think adding the filename surfaces really helpful information, especially if you are working with a lot of profiles and forget where each one outputs.

Here are some examples of the styling with two admonitions still, but I'll see about condensing them into one:

Screenshot 2026-06-28 at 10 23 36
Screenshot 2026-06-28 at 10 23 56

nathanlesage:
> And now users have to open the assets window, open the profile, and search through the profile to find output-file to get the setting.

I mean, yes, I see that point. The question is, how often do the users want to see that paths, versus just being reminded that they provided it? Difficult to judge just from theory. But if you have a strong opinion about that, I think we can include the path. Worst case people complain and we have to pull it later on, which shouldn't be a big issue.

Finally, I strongly believe we should only provide a single admonition in the popover. How about one on top of the dir selector that reads: 'Directory selection is disabled because this profile specifies "%s"'?

benniekiss:
That works for me!

Screenshot 2026-06-29 at 08 13 49
Screenshot 2026-06-29 at 08 14 55

How does this look?

benniekiss:
Before this is merged, I'd like to try and address the linked issue above related to over writing the export file.

I'd propose outright preventing over writing the file, instead exporting to a new filename with a randomized (or stable) prefix.

Or maybe provide a prompt to overwrite, or write to a new filename

nathanlesage:
Please don't. This will lead to many, many situations in which users suddenly end up having 20x the same file.

Almost no user has faced this overwriting issue before -- in the past seven years this happened to ca. 3-4 people. So while I agree a warning is in order, for 99% of the people the behavior should not change. So instead I'd rather give users a much more simple choice in the form of an ok/cancel modal, as I described in the issue.

Both alternatives (letting users set a new filename and simply creating a new unused filename) are going to be detrimental. Letting them choose a new filename will induce a lot of friction, and most need to rapidly re-export, so the overwriting is a feature. And as I said, creating suffixes after the filename will lead to file graveyards.

The solution I would like gives everyone a single warning, which most users will immediately disable again, so that it's only used by the few who actually need it, because they use Zettlr differently.

PR #6126, opened 2026-01-11
[Enhancement] Allow custom dictionaries with non-bcp47 names

benniekiss:
## Description

This PR allows dictionaries to be named with non-BCP 47 language code names.

## Changes

The language code check was removed.

## Additional information

I have a lot of domain-specific dictionaries, so limiting the available dictionaries to only those with appropriate language codes was too restricting. 

Tested on:  macOS 26

### DIFF
--- a/source/common/util/enum-dict-files.ts
+++ b/source/common/util/enum-dict-files.ts
@@ -17,7 +17,6 @@
 
 import { type Candidate } from './find-lang-candidates'
 import path from 'path'
-import * as bcp47 from 'bcp-47/index.js'
 import fs from 'fs'
 import { app } from 'electron'
 import isDir from './is-dir'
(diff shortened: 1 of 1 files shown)


nathanlesage:
How easy is it for users to accidentally break the dictionary provider if they provide malformed affix files? I'm asking because while simple dictionaries are easy, the Hunspell syntax can be somewhat error-prone. As long as loading an accidentally malformed dictionary doesn't break the app, this change is fine by me!

benniekiss:
I think malformed files crash the app, but that's entirely separate from acceptable names since nothing prevents a user from using a custom affix with a bcp-47 name

I can go ahead and add a guard for those cases though, I'm pretty sure it's just a `try...catch...` in the dictionary provider

nathanlesage:
> I think malformed files crash the app, but that's entirely separate from acceptable names since nothing prevents a user from using a custom affix with a bcp-47 name

Sometimes, I forget to think like users :D 

Alright, let's merge that thing. Try-catch is always nice, if you are bored in the next days or so!

benniekiss:
I'll definitely get something in

benniekiss:
@nathanlesage ok so I was mistaken -- I don't think malformed affixes crash the app. The crashes I was seeing were in a dev branch of mine, and testing with intentionally broken affixes didn't seem to cause any issue on `develop`.

nathanlesage:
Perfect, thanks! I guess we need to thank the hunspell devs for making the lib robust :D

PR #6142, opened 2026-01-20
[Fix] Correct `pathDirname` output for root-level directories

### LINKED ISSUE #6105: Search accross all files no longer gives any result
### Description

Search accross all files no longer gives any result: I type a word like I did before, no result, while the word exists in multiple files

### Reproducing

Just search accross all files

### Zettlr version

4.0.0.

### Installation Method

From the Website or GitHub

Update from previouis version

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Windows 11

### Additional Information

_No response_

benniekiss:
## Description

This PR fixes an issue with `pathDirname` which was stripping the final path separator if the provided path was a root-level path, such as `C:\Notes` or `/notes`, which would be returned as `C:` or ``

Closes #6105

## Changes

A check was added to `pathDirname` to ensure the final path separator is not removed by first checking if the path is absolute and whether the first and last found path separator are at the same position.

Two test cases were added which failed before the patches.

## Additional information

I would appreciate verification on a windows platform if this fixes issues related to search. I do not have a windows machine.

Tested on:  macOS 26

### DIFF
--- a/test/renderer-path-polyfill.spec.ts
+++ b/test/renderer-path-polyfill.spec.ts
@@ -36,6 +36,7 @@ const pathDirnameTesters = [
   '.\\test',
   'C:\\',
   'C:\\Notes',
+  '/notes',
   '/',
   '\\\\'
 ]


nathanlesage:
Ah gracias, and that should correctly give us back search for windows users who have a workspace at a volume root, correct?

benniekiss:
I'm pretty confident it will, but I cannot confirm myself since I dont have a windows machine

nathanlesage:
That's what unit tests are for 😎 I'd argue we yolo it and then let users test a manual nightly! I'll merge in a few moments

benniekiss:
couldnt agree more!

PR #6151, opened 2026-01-26
[chore] Remove redundant json folding code

benniekiss:
## Description

This PR removes the json code folding code since the `codemirror/lang-json` package already implement folding.

## Additional information

Tested on:  macOS 26


nathanlesage:
That was likely because back when I implemented that, the package didn't support code folding!

Issue #6067, opened 2025-12-19
Pilcrow/line-ending marker appears on a new blank/continuation line inside an indented code block


SuYanqi:
#### Expected Behavior
- No pilcrow (¶) line-ending markers are rendered on any lines that are part of the indented code block, including the newly created continuation/blank line after pressing Enter.

#### Actual Behavior
- A pilcrow/line-ending marker is rendered on the new continuation/blank indented code line (visible near the start/indent area; bbox # 7).

#### Notes
This is introduced by Pull Request https://github.com/Zettlr/Zettlr/pull/6024 , namely Commit 8dcaf8386dea7ccfeb1e0ada7fef61c897b5cd42

#### Screenshot Before https://github.com/Zettlr/Zettlr/pull/6024
<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/4e2b0741-9b4f-47b7-9c17-c5056036ed90" />

#### Screenshot After https://github.com/Zettlr/Zettlr/pull/6024
<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/87d583cd-84d8-4c8a-85d8-cf0cce7e55a6" />

#### Steps to Reproduce
1. Open the Markdown editor.
2. Enable “Show whitespace and line endings (¶)”.
3. In a new document, type a normal line and create a blank line.
4. Type an indented code line starting with four spaces, e.g.: '    echo "indented code 1"'
5. Press Enter to create the next indented code line (continuation/blank line).

### Zettlr version

Version 4.0.0-beta.5 (4.0.0-beta.5)

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [x] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit
- [x] ARM 64bit (includes Apple Silicon)

### Operating System Version

ubuntu:22.04, macOS 15.3.2 (24D81)

benniekiss:
This is not a bug, this is just how the markdown formatting and syntax node parsing works. The line after the indented code with only four whitespaces is not technically part of the indented code block until you add a non-whitespace character to the line. You can see this when you export to HTML using pandoc.

nathanlesage:
[...]

PR #6152, opened 2026-01-26
[Enhancement] Improve pilcrow styling

benniekiss:
## Description

This PR includes several pilcrow improvements, including no longer drawing them on code-block marker lines (\`\`\`) and no longer triggering line wrapping, 

## Changes

Pilcrows are no longer drawn on lines for code block or pandoc div formatting marks, i.e. \`\`\` and `:::`. This was discussed in #6067

Additionally, pilcrows are no longer drawn on folded lines. This prevents two pilcrows from showing up on a line when a range is folded.

Pilcrows no longer trigger wrapping to the next line. This prevents the pilcrow from wrapping by itself, ending up on the line alone. Because pilcrows may now overflow the text boundary, the `cm-content { overflow-x }` CSS property was changed from `hidden` to `visible`.

The header styling was changed so that the styling is applied to the whole line. When I migrated the code to these CSS styles, I did not realize the selecter I used, `.cm-header-*`, is only applied to small runs of characters within the header, not the entire line, so some characters within a header may not receive the styling. I noticed this because pilcrows and trailing space did not.

## Additional information

Linewrapping and header styling:

**CURRENT:**

https://github.com/user-attachments/assets/36d8e089-efcf-492e-9c75-b5f48a14c509

**PR:**

https://github.com/user-attachments/assets/952da956-e9b5-430d-8233-68738f655380

Tested on:  macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
+++ b/source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
@@ -42,8 +42,7 @@ class PilcrowWidget extends WidgetType {
     const span = document.createElement('span')
     span.className = 'cm-pilcrow'
     span.textContent = '¶'
-    span.style.userSelect = 'none'
-    span.style.pointerEvents = 'none'
+
     span.setAttribute('aria-hidden', 'true')
 
     return span
@@ -128,6 +127,10 @@ const pilcrowTheme = EditorView.baseTheme({
     // which otherwise causes the pilcrow to be offset into
     // the text.
     textIndent: 'initial',
+    // These help prevent issues related to
+    // pilcrows inserting literal text content
+    userSelect: 'none',
+    pointerEvents: 'none',
     // Styling
     color: '#aaa',
     opacity: '0.5',


source/win-main/MainEditor.vue
```diff
@@ -688,7 +688,7 @@ function maybeHighlightSearchResults (): void {
   }
 
   .cm-content {
-    overflow-x: hidden !important; // Necessary to hide the horizontal scrollbar
+    overflow-x: visible !important; // Necessary to hide the horizontal scrollbar
```
nathanlesage: What happens if you just remove this rule entirely…?

benniekiss: Oh, good idea. Briefly testing by opening developer tools and disabling it there doesn't seem to cause any issue.

benniekiss:
> I'm a bit wary with regard to the overflow, did you test this extensively? I am a bit afraid this could mess with CodeMirror's line wrapping calculation, or am I just over-cautious here?

I'll be testing this over the next couple days, so I'm fine letting it sit until I can report back. But have not had any issues so far. Text wrapping is normal, and from my understanding, `visible` and `hidden` are essentially the same, except for how the overflowing content is displayed

nathanlesage:
Okay, sounds good! First thing on my todo today is releasing 4.1.1 and then trying to convince Azure to use a new code cert, when that works I'll be less stressed as well 🙈

benniekiss:
Last two commits handle two issues I was seeing:

The line wrapping plugin was calculating in indent in narrow windows that shouldn't be there, and the issue was with the whitespace regex, which was matching on zero or more whitespace. I corrected it to require at least one whitespace character. Have not noticed any indent issues otherwise.

Wide tables were causing the content window to scroll. Adding back `overflow-x: hidden` fixed the problem, but I believe the scrolling was happening because the `.cm-content` element had `min-width: 100%;` set. Unsetting it to `min-width: 0` allows the content window to shrink enough to prevent scrolling.

Otherwise, haven't seen any other problems working the past two days

source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
```diff
@@ -42,6 +42,8 @@ class PilcrowWidget extends WidgetType {
     const span = document.createElement('span')
     span.className = 'cm-pilcrow'
     span.textContent = '¶'
+    span.style.userSelect = 'none'
```
nathanlesage: Shouldn't we move the style rules to the theme below?

source/common/vue/CodeEditor.vue
```diff
@@ -283,6 +283,11 @@ body {
       background-color: @base2;
       color: @base1;
     }
+
+    // Hide pilcrows
+    .cm-pilcrow {
+      opacity: 0;
+    }
```
nathanlesage, line 290: Why force-disable pilcrows in code editors…? Aren't there some who may prefer this…?

nathanlesage (approved):
Found two more minor things, but this thing looks mergeable!

source/common/vue/CodeEditor.vue
benniekiss: I think this makes it consistent with the codeblock rendering, but I'm fine putting them back in.

source/common/modules/markdown-editor/plugins/highlight-whitespace.ts
benniekiss: Good point! I'll go ahead and move them

source/common/vue/CodeEditor.vue
nathanlesage: Hm… I mean we can keep it in. I personally (!) don't use whitespace highlighting at all, but we can hope that others would complain, if we did that and they miss it, so I'm fine keeping it for now. Everything is better than stuffing yet another setting into the preferences dialog :D

benniekiss: Just to clarify -- keep pilcrows disabled in code editors?

nathanlesage:
Trés bien, and merging!

PR #6153, opened 2026-01-27
[Enhancement] Improve pdf viewer resize UX

benniekiss:
## Description

This PR changes how the PDF viewer determines whether to accept  `pointer-events` to improve UX when resizing the editor pane.

## Changes

Instead of watching for focus changes, `acceptClicks` is now toggled on `pointerenter` and `pointerleave`. This improves UX in two ways: the user no longer has to remember to unfocus the viewer before resizing, and the user no longer has to click to focus the viewer before using the UI.

The border around the active viewer has been removed because it no longer serves a purpose.

`overflow: hidden` was added to the iframe parent element to prevent scroll bars from appearing on the parent.

## Additional information

PR (still with the border present): 

https://github.com/user-attachments/assets/0f33063c-1a20-4759-95e2-05e8ec0228dd

Tested on:

### DIFF
--- a/source/win-main/file-viewers/PDFViewer.vue
+++ b/source/win-main/file-viewers/PDFViewer.vue
@@ -4,7 +4,8 @@
     class="pdf-viewer-container"
     role="region"
     v-bind:aria-label="`PDFViewer: Currently viewing file ${pathBasename(props.file.path)}`"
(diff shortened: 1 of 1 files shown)


nathanlesage:
Huh, smart! Thanks a bunch!

PR #6155, opened 2026-01-29
fix: Allow unicode emojis in tags

kyaso:
## Description
This PR fixes a recent issue that prevented having emojis in tags. If there was an emoji, the tag would not be parsed correctly.

## Changes
Fixed the tag regex to allow unicode emojis. Also refactored the regex to use unicode property escapes.

## Additional information

Tested on: WSL Ubuntu

Also closes: #6157

### DIFF
--- a/source/common/modules/markdown-editor/parser/zkn-tag-parser.ts
+++ b/source/common/modules/markdown-editor/parser/zkn-tag-parser.ts
@@ -17,7 +17,9 @@ import { type InlineParser } from '@lezer/markdown'
 // Any character allowed before a tag
 const allowedCharsBefore = /^[ \t\n\(\{\[]$/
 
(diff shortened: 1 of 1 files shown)


nathanlesage:
Does this by any Chance also fix #6157?

kyaso:
> Does this by any Chance also fix #6157?

I'm pretty sure it _should_, because `p{L}` is intended to match any letter from any language. But I will double check the offending example from #6157 later today!

kyaso:
> But I will double check the offending example from https://github.com/Zettlr/Zettlr/issues/6157 later today!

I just checked, and it indeed works again correctly with this fix.

nathanlesage:
Hah, that's what I suspected. Thank you!

PR #6164, opened 2026-02-06
Update file-extention-checks.ts

farvardin:
## Description
add gemtext support for gemini protocol (a markdown derivative)

## Changes
just add 3 new extensions names in ``source/common/util/file-extention-checks.ts``

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## GUI and Functionality
 
-(nothing here)
+- Add gemtext extension support for gemini protocol (a markdown derivative)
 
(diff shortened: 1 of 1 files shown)


nathanlesage:
I am not a fan of adding arbitrary extensions into the hardcoded core, especially niche ones that likely won't persist in the long term. It'd be more sustainable to allow custom extensions that users can define.

We might be able to address this together with the other request of having explicit plain-text files. We could use a token control for that, like the custom attachments extensions.

farvardin:
> We might be able to address this together with the other request of having explicit plain-text files.

sure, that would be great!

PR #6167, opened 2026-02-08
fix(linter): use correct node name for YAML frontmatter detection

kevinbader:
The YAML frontmatter linter was non-functional because
findYamlFrontmatterNode() searched for a node named 'FencedCode', but the
frontmatter parser creates nodes named 'YAMLFrontmatter'.

This mismatch meant the linter never found frontmatter content and
always returned an empty diagnostics array, causing all YAML syntax
errors (such as tabs used as indentation) to be silently ignored.

Change the node name check to 'YAMLFrontmatter' to match what
frontmatter-parser.ts actually creates, enabling proper YAML error
detection and reporting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 ## Under the Hood
 
+- Fixed YAML frontmatter linter not detecting any errors due to incorrect syntax
+  tree node name lookup.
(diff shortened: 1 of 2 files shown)


nathanlesage:
Very good catch! Thank you so much!

PR #6168, opened 2026-02-08
fix: respect indentation settings when pressing Tab

kevinbader:
Replace insertTab with indentMore in maybeIndentList to ensure the
Tab key respects the user's indentation settings (spaces vs tabs).

Previously, pressing Tab outside of Markdown lists would always insert
a literal tab character, regardless of the 'Indent using tabs instead
of spaces' setting.

The insertTab function from @codemirror/commands always inserts a
literal tab character, while indentMore respects the configured
indentUnit (spaces or tabs based on settings).

Additionally, add special handling for YAML frontmatter blocks. Since
YAML requires spaces for indentation (tabs break YAML parsing), the
editor now always inserts spaces when the cursor is inside a YAML
frontmatter block, regardless of the user's tab settings.

Changes:
- Replace insertTab with indentMore for consistent indentation behavior
- Add isInYamlFrontmatter() helper to detect YAML frontmatter context
- Force space-based indentation in YAML frontmatter blocks

Now Tab consistently uses the configured indentation throughout the
document, except in YAML frontmatter where spaces are always used to
maintain valid YAML syntax.

### DIFF
--- a/source/common/modules/markdown-editor/commands/lists.ts
+++ b/source/common/modules/markdown-editor/commands/lists.ts
@@ -17,10 +17,11 @@
 import { type ChangeSpec, type EditorState, type Transaction } from '@codemirror/state'
 import { type EditorView } from '@codemirror/view'
 import { syntaxTree } from '@codemirror/language'
-import { indentLess, indentMore, insertTab, moveLineDown, moveLineUp } from '@codemirror/commands'
+import { indentLess, indentMore, moveLineDown, moveLineUp } from '@codemirror/commands'
 import { type SyntaxNode } from '@lezer/common'
 import { markdownToAST } from '@common/modules/markdown-utils'
 import type { BulletList, OrderedList } from '@common/modules/markdown-utils/markdown-ast'
+import { configField } from '../util/configuration'
 
 /**
  * Tests if there is any list affected by the current editor selection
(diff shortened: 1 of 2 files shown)


source/common/modules/markdown-editor/commands/lists.ts
```diff
@@ -66,6 +67,44 @@ function isListTouchedBySelection (state: EditorState): boolean {
...
+ * we always use spaces in frontmatter regardless of user's tab settings.
+ *
+ * @param   {EditorState}  state  The state in question
+ *
+ * @return  {boolean}             Returns true if any selection is in YAML frontmatter
+ */
+function isInYamlFrontmatter (state: EditorState): boolean {
+  for (const range of state.selection.ranges) {
+    let inFrontmatter = false
+    syntaxTree(state).iterate({
```
benniekiss: I would recommend using the [`syntaxTree.resolve()`](https://lezer.codemirror.net/docs/ref/#common.Tree.resolve) method here to avoid iterating over the entire tree, and instead check whether the resolved node is one of `YAMLFrontmatter`, `YAMLFrontmatterStart`, or `YAMLFrontmatterEnd`

benniekiss:
```ts
const node = syntaxTree(state).resolve(range.from)

if (node.is('YAMLFrontmatter') || node.is('YAMLFrontmatterStart') || node.is('YAMLFrontmatterEnd') { 
  return true 
}
```

kevinbader: ah yes that's way better .. tbh I don't really grasp yet how this works, for example, the documentation of `resolve` says "Note that this will not enter overlays, and you often want `resolveInner` instead", but I've got no idea yet what overlays are and whether they're used in Zettlr. But I assume you do, so I'm happy to adopt your suggestion.

benniekiss: iirc, `resolveInner` could additionally return nodes parsed by a yaml code block parser, for example, on top of the main markdown parser, but it's been a bit since I've really investigated the difference. feel free to test both! I have used `resolve` in several places to detect the same nodes and found success.

source/common/modules/markdown-editor/commands/lists.ts
```diff
@@ -66,6 +67,32 @@ function isListTouchedBySelection (state: EditorState): boolean {
...
+ * @param   {EditorState}  state  The state in question
+ *
+ * @return  {boolean}             Returns true if any selection is in YAML frontmatter
+ */
+function isInYamlFrontmatter (state: EditorState): boolean {
+  const tree = syntaxTree(state)
+  for (const range of state.selection.ranges) {
+    // Use resolve() to directly find the node at the cursor position
+    // and walk up the tree to check for YAML frontmatter ancestry
+    let node = tree.resolve(range.from)
```
nathanlesage: Suggestion: Explicitly type this to `let node: SyntaxNode|null`. This will make the first linter error in the `while` go away. Then, you can also get rid of the `if`-condition, because that will automatically break out of the while loop in the next iteration.

nathanlesage (changes requested):
Looks good, except the linter has complained. See below for a fix.

nathanlesage:
Looks good! Thanks for fixing. I'll merge it now.

PR #6171, opened 2026-02-10
Add scroll-past-end extension to editor

kevinbader:
Enable scrolling past the end of documents so users can position the
last line in the middle of the viewport without adding empty lines.

This implements a custom ViewPlugin that adds bottom padding equal to
half the editor height. The padding updates dynamically when the editor
is resized. Unlike CodeMirror's built-in scrollPastEnd() which allows
scrolling to the top of the editor, this implementation centers the
last line in the viewport for a more comfortable writing experience.

---

Maybe it's just me, but I feel weird writing at the bottom of the screen - I want my focus to be in the center of the screen.

I'm wondering:

- Is this the right place or would `markdown-editor/plugins` be better?
- Is it okay to have this as the default behavior or should this be configurable? For me it's obviously fine like this, but who knows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## GUI and Functionality
 
+- **Feature**: The editor now allows scrolling past the end of documents for a
+  more comfortable writing experience.
 - **Feature**: New utilities for working with native Pandoc Divs and spans
   (#6032):
   - Added a new setting to switch between native highlights (`==mark==`) and
--- a/source/common/modules/markdown-editor/editor-extension-sets.ts
+++ b/source/common/modules/markdown-editor/editor-extension-sets.ts
@@ -29,6 +29,7 @@ import {
   EditorView,
   lineNumbers,
   dropCursor,
+  ViewPlugin,
   type ViewUpdate,
   type DOMEventHandlers
 } from '@codemirror/view'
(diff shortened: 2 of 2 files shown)


nathanlesage:
To first answer your questions:

1. The editor extension sets file should only provide sets of extensions, but not define them. Therefore, any extension should go into the appropriate folder. For such a plugin, the generic `plugins`-folder would be appropriate.
2. This absolutely should be configurable, because I personally (!) do not like this behavior, so I'd want to disable this. I'd fathom the perfect place for that would be the Editor -> Other settings fieldset, and the config option itself should be grouped into the `editor`-key of the config options.

Lastly, I'd like you not to implement this functionality using a custom plugin, because CodeMirror thankfully provides an extension for this: https://codemirror.net/docs/ref/#view.scrollPastEnd

Please use that ready-made plugin instead. To make it configurable, you probably want to use a settings compartment. To see how, search the codebase for `inputModeCompartment`. You will see that it essentially wraps the extension, and can be re-configured on config updates to include or exclude that extension, effectively turning it on and off.

kevinbader:
Hi Hendrik, thanks for taking a look. I wanted to use the scrollPastEnd extension, but it adds margin for the full height, so scrolling rapidly you would end up with a blank screen, which I personally don't like; and I couldn't find a way to configure it. Maybe there's a way after all, I could look into that again.

Obviously, you're fine with the current behavior, makes sense. I've started using Zettlr quite recently, and maybe it just takes some getting used to, looking at the lower part of the screen while writing. I'm also thinking about replicating something like [vim's scrolloff](https://vimhelp.org/options.txt.html#%27scrolloff%27)... I don't know, maybe I should just my monitor higher up haha

I'll change this to draft status for now and play around with this in my fork for a while. I don't want to suggest unnecessary features/options.

nathanlesage:
> you would end up with a blank screen

Does it not show at least one single line? That at least is similar behavior to a setting in VS Code.

Also, if you want the line to be in the center, I recommend you use the Typewriter mode. I use it often if I want to write but don't want to look at the bottom of the app all the time.

kevinbader:
> Also, if you want the line to be in the center, I recommend you use the Typewriter mode. I use it often if I want to write but don't want to look at the bottom of the app all the time.

That's pretty cool, I've missed typewriter mode so far :) One question though: when enabling "dark mode" through the view menu, it's displayed as enabled with a check mark next to it in the menu. For "typewriter mode" and "distraction-free mode" this is not the case. This surprised me - is this by design? If not, I can open an issue an attempt a fix.

Also, switching to a different file and back throws me out of typewriter mode again. Same question here, by design or could be improved?

nathanlesage:
The Former could be a bug. The latter means that we should store this information in the local editor cache together with the folded regions.

PR #6178, opened 2026-02-12
[fix] Add more pandoc div insertion checks

benniekiss:
## Description

This PR fixes some issues when inserting pandoc divs using the toolbar UI, and it adds some QOL improvements.

## Changes

After inserting a div using the toolbar, the editor is re-focused so the user does not have to click back into the editor to type.

Bounds checks were added when calculating the previous and next lines. This fixes an issue where inserting a div would fail on the first or last line.

Additional checks were added to improve UX when inserting empty divs.

Newlines are no longer inserted before or after div marks if the previous or next line is itself a div mark.

If the div is inserted on the first line, a newline is no longer inserted before it.

## Additional information

Tested on:  macOS 26


nathanlesage:
Looks good, thank you!

PR #6194, opened 2026-02-18
[Improvement] Improve filename extension handling

benniekiss:
## Description

This PR improves file extension handling when creating or renaming a file, as well as when loading and detecting supported documents.

## Changes

How file extensions are determined when creating or renaming a file was refactored. I encountered some annoying behavior that, when renaming a `.docx` file displayed in Zettlr, it would add an `.md` extension, and this was symptomatic of some deeper logic bugs. Now, files can be reliably named with any extension recognized by Zettlr, via those defined in `file-extention-checks.ts`.

When renaming a file, some additional checks are done: If the old and new file extensions do not match (i.e, they were changed), and the new name does not have a recognized extension, then the old file extension is used, or if that was not determinable, it defaults to `.md`. 

When creating a file using `file-new`, the `type` parameter was changed from a `string` to a `DocumentType`. This standardizes how supported documents are referenced in the app. Likewise, if a `DocumentType` was provided, but the actual  extension and the expected `DocumentType` extension do not match, or if the new extension is not recognized, the extension is set to the appropriate extension for the `DocumentType`, or otherwise `.md` if no type was provided.

Two helper functions were added to `file-extention-checks.ts`: `getExtensionForDocumentType` and `getDocumentTypeForExtension`. These utilities help convert between supported document types and file extensions. Likewise, the extensions which Zettlr supports were split into their own `_EXT` variables, and the `CODE_EXT` variable is now created by combining the various extension lists.

Some `any` type errors were corrected.

## Additional information

Tested on: 
macOS 26

### DIFF
--- a/source/app/service-providers/commands/file-new.ts
+++ b/source/app/service-providers/commands/file-new.ts
@@ -18,9 +18,9 @@ import path from 'path'
 import sanitize from 'sanitize-filename'
 import generateFilename from '@common/util/generate-filename'
 import { app } from 'electron'
-import { getExtensionForDocumentType, hasAnyRecognizedFileExtension } from '@common/util/file-extention-checks'
+import { getDocumentTypeForExtension, getExtensionForDocumentType, hasAnyRecognizedFileExtension } from '@common/util/file-extention-checks'
 import type { AppServiceContainer } from 'source/app/app-service-container'
-import type { DocumentType } from '@dts/common/documents'
+import { DocumentType } from '@dts/common/documents'
 
 export default class FileNew extends ZettlrCommand {
   constructor (app: AppServiceContainer) {
@@ -115,15 +115,30 @@ export default class FileNew extends ZettlrCommand {
         throw new Error('Could not create file: Filename was not valid')
       }
 
-      const ext = path.extname(filename)
-      const extType = arg.type !== undefined ? getExtensionForDocumentType(arg.type) : '.md'
-      const invalidExt = !hasAnyRecognizedFileExtension(filename, attachmentExtensions)
+      const docType = arg.type ?? DocumentType.Markdown
+      const docExtension = getExtensionForDocumentType(docType)
 
-      // If a type was provided, but the actual and expected extensions do not
-      // match, or if the new extension is not recognized, set the extension to
-      // the expected one, defaulting to `.md`
-      if ((arg.type !== undefined && ext !== extType) || invalidExt) {
-        filename += extType
+      // No DocumentType was provided, so this could be a generic file.
+      // Check if the file extension is recognized. If it is not, then
+      // append the default DocumentType extension to the filename.
+      if (arg.type === undefined && !hasAnyRecognizedFileExtension(filename, attachmentExtensions)) {
+        filename += docExtension
+      }
+
+      // A DocumentType was provided, so the DocumentType for the filename
+      // must match the provided DocumentType. If it doees not, then append
+      // the default DocumentType extension to the filename.
+      if (arg.type !== undefined) {
+        const ext = path.extname(filename)
+        const extDocType = getDocumentTypeForExtension(ext)
+
+        // If the document types do not match, then an invalid extension was provided.
+        // Append the default wanted extension to the filename. Do not try to
+        // strip and replace the existing extension, as this could clobber otherwise
+        // normal user input, such as dot-separated filename schemes.
+        if (docType !== extDocType) {
+          filename += docExtension
+        }
       }
 
       const absPath = path.join(dirpath, filename)
--- a/source/app/service-providers/commands/file-rename.ts
+++ b/source/app/service-providers/commands/file-rename.ts
@@ -44,43 +44,38 @@ export default class FileRename extends ZettlrCommand {
       return
     }
 
-    let oldExt = path.extname(arg.path)
-    let newExt = path.extname(newName)
+    let parsedOldPath = path.parse(arg.path)
+    let parsedNewPath = path.parse(newName)
 
-    const invalidExt = !hasAnyRecognizedFileExtension(newName, this._app.config.get().attachmentExtensions)
+    // If the old and new extensions do not match, we must check
+    // that the new extension is recognized by Zettlr, and if it
+    // is, confirm with the user to change the file extension. If
+    // it is not, simply append the old extension.
+    if (parsedOldPath.ext !== parsedNewPath.ext) {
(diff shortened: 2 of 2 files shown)


benniekiss:
The last few commits add a confirmation dialogue to change extensions, and they improve readability when renaming a file by changing the styling a little bit.

source/app/service-providers/commands/file-new.ts
```diff
@@ -114,22 +115,15 @@ export default class FileNew extends ZettlrCommand {
...
-            filename += '.tex'
-            break
-          case 'yaml':
-            filename += '.yml'
-            break
-          default:
-            filename += '.md'
-        }
+      const ext = path.extname(filename)
+      const extType = arg.type !== undefined ? getExtensionForDocumentType(arg.type) : '.md'
```
nathanlesage:
```suggestion
      const wantedExt = arg.type !== undefined ? getExtensionForDocumentType(arg.type) : '.md'
```

(So that we can better differentiate it from the `type`)

source/app/service-providers/commands/file-new.ts
```diff
@@ -114,22 +115,15 @@ export default class FileNew extends ZettlrCommand {
...
+      const invalidExt = !hasAnyRecognizedFileExtension(filename, attachmentExtensions)
+
+      // If a type was provided, but the actual and expected extensions do not
+      // match, or if the new extension is not recognized, set the extension to
+      // the expected one, defaulting to `.md`
+      if ((arg.type !== undefined && ext !== extType) || invalidExt) {
```
nathanlesage: Shouldn't we *replace* the extension instead of appending a valid one? There shouldn't be too many cases where there is an "extension" given that is really just part of the filename (such as with `bundle.app` folders on macOS). I'd fathom that users who want to add an explicit modifier in the file `such as file.backup.md` would make few mistakes and, if they did, be willing to rename the file after the fact.

I.e., something like:

```js
filename = path.basename(filename, path.extname(filename)) + wantedExt
```

source/app/service-providers/commands/file-rename.ts
```diff
@@ -44,9 +44,43 @@ export default class FileRename extends ZettlrCommand {
...
+    // If no new extension was provided, or the extension is otherwise invalid,
+    // assume it was a rename targeting the old extension. If the extensions
+    // are not the same, show a dialogue asking the user to confirm  if they
+    // would like to change the file extension.
+    //
+    // If they respond with `Keep`, the file is still renamed, but the new
+    // extension is replaced with the old one.
+    if (newExt !== '' && !invalidExt && oldExt !== newExt) {
+      const response = await dialog.showMessageBox({
+        title: trans('Confirm'),
```
nathanlesage:
```suggestion
        title: trans('Change file extension'),
```

source/app/service-providers/commands/file-rename.ts
```diff
@@ -44,9 +44,43 @@ export default class FileRename extends ZettlrCommand {
...
+        message: trans('Change file extension from %s to %s?', oldExt, newExt),
```
nathanlesage:
```suggestion
        message: trans('Do you want to change the file extension from %s to %s?', oldExt, newExt),
```

source/app/service-providers/commands/file-rename.ts
```diff
@@ -44,9 +44,43 @@ export default class FileRename extends ZettlrCommand {
...
+        ],
+        defaultId: 1
+      })
+
+      // If `Keep`, replace the new extension with the old one.
+      if (response.response === 1) {
+        let newExtPos = newName.lastIndexOf(newExt)
+
+        newName = newName.slice(0, newExtPos) + oldExt
+        newExt = oldExt
```
nathanlesage: This seems a bit convoluted, no?

Also, I'm wondering if we would gain clarity in the code if we treat filename and file extension as separate until the very end, so that we can create the "full" filename only at the end using something like `fs.rename(basename + ext)`

This might make the whole extension game a bit cleaner, I'd argue

source/win-main/file-manager/TreeItem.vue
```diff
@@ -477,6 +477,15 @@ const isSelected = computed(() => {
...
 watch(selectedDir, uncollapseIfApplicable)
 
+// We need to reset the scroll position when exiting editing mode
+// so that the filename is displayed within the element correctly.
+// It would otherwise be offset to the left edge of the container.
+watch(nameEditing, (isEditing) => {
+  if (!isEditing && displayText.value !== null) {
+    displayText.value.scrollLeft = 0
+  }
+})
```
nathanlesage, line 485: Can't we solve this in CSS…? Also, what I wanted to do was have the text input field maximally extend to the edge of the file manager. Currently, there's this bug where the text input field just extends indefinitely, which is super inconvenient. I don't exactly know when that went wrong…

benniekiss: I dont think this can be fixed with CSS. I did try, but it may be beyond my skills.

Does the text box extension happen on this branch? I didnt notice it in my tests, and I made sure to test rather long file names

nathanlesage: Mh, then let's leave it at that. I'll have to revisit that at some point anyhow!

source/app/service-providers/commands/file-new.ts
benniekiss: Initially I went with replacement, but I feel appending is less destructive to user input.

It also allows a user to not have to worry about an extension, where they can input a filename like `Test.Date.Part2`, and the ext is auto appended. If we went with replacement, this would end up `Test.Date.md`, which introduces more friction for the user

source/app/service-providers/commands/file-rename.ts
benniekiss: I can see about that.

This test was initially quite small, but it grew out of the various paths the user input can take, especially with ambiguous extensions that may not be extensions.

source/app/service-providers/commands/file-new.ts
nathanlesage: Good point! Okay, then let's keep it this way!

source/app/service-providers/commands/file-rename.ts
benniekiss: Please take a look at the most recent commit. I've cleaned up the handling to be much more straigtforward IMO

nathanlesage:
So, sorry for the delay, but I believe now this PR is ready for merging – thank you! :)

PR #6196, opened 2026-02-19
[fix] fix tab insertion

### LINKED ISSUE #6190: Pressing tab within a line indents the whole line
### Description

When trying to insert a tab somewhere in the middle of the line, the whole line is indented rather than the tab being inserted.

I tracked it down to this change: 4706bb086ad229ca725ef22d75e71bf515571468
and this line: https://github.com/Zettlr/Zettlr/blob/ff148283c9334d5503773f78977b112d95fa6d93/source/common/modules/markdown-editor/commands/lists.ts#L290

As `indentMore` applies the indent to the beginning of the line, we need to revert back to `insertTab`, or refactor the indent handling.

### Reproducing

Press tab in the middle of a paragraph

### Zettlr version

develop

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit

### Operating System Version

macOS 26

### Additional Information

_No response_

benniekiss:
## Description

This PR fixes tab insertion and improves YAML tab-to-space sanitation. 

Closes #6190

## Changes

The `codemirror` `insertTab` command was reimplemented. The native command always inserts tabs (when not calling `indentMore`), so we need a custom implementation that inserts spaces or tabs. The functionality was separated from the `list` helper function. As a result, a new keybinding was added. 

The `YAMLFrontmatter` sanitation was improved. The `isInYamlFrontmatter` function was moved to its own file and renamed to `yamlInSelection`, mirroring the `rangeInSelection` utility. In addition, if a selection is within a yaml node, the `indentWithTabs` config setting is overriden temporarily. This is necessary so that `indentMore` uses spaces when within `YAMLFrontmatter` nodes. Otherwise, we still get tab inserts.

Finally, the `isListTouchedBySelection` was refactored and renamed to `listInRange` to mirror the other similar functions. It was refactored to use `tree.resolve()` rather than iterate over the tree.

## Additional information

Tested on: 

macOS 26

### DIFF
--- a/source/win-preferences/schema/editor.ts
+++ b/source/win-preferences/schema/editor.ts
@@ -322,6 +322,7 @@ export function getEditorFields (config: ConfigOptions): PreferencesFieldset[] {
         {
           type: 'checkbox',
           label: trans('Always indent the current line when pressing Tab'),
+          info: trans('Zettlr always indents the line for list items. Turn this setting on to always indent any line.'),
           model: 'editor.alwaysIndentLineOnTab'
         },
         { type: 'separator' },


benniekiss:
The latest commit further changes `yamlInSelection` into a more generic function for checking if a list of nodes is within a selection. It is now called `nodeInSelection`, and an additional helper was added `posInSelection`, both of which consolidate logic used in a few places in the code

nathanlesage:
I appreciate this, especially ensuring that YAML frontmatters won't contain tab characters anymore.

However, I think we should spend at least a little bit on discussing the implementation, because — at least for me, personally — this PR would prove to be an unwanted change. To elaborate a bit:

`Tab` characters stem from a time when we didn't have tables and proper alignment, so they have lost their original meaning. Nowadays, they mostly are used to switch between program tabs, or to indent/unindent lists, but even this indentation is at least half of the time done with spaces. So, oftentimes, they are at the very least not necessary and thus a matter of preference.

If I understand you correctly, this would change the behavior so that, when one is somewhere within a line, this would no longer indent the line, but instead insert a literal tab character. This would be a regression, because I have grown accustomed to just clicking somewhere in a line to indent/unindent it.

However, I fully understand and support workflows that rely on tabs instead of spaces, so I am more than happy to implement this, but with some changes.

Thus, I have two refactoring requests:

1. Please separate the switching-off of tabs in the frontmatter from the indentation logic, if possible. (This might necessitate two separate commands both bound to Tab with a fall-through like behavior, I apologize for this in advance. But maybe this could be done within a single command…?)
2. Allow toggling of this behavior, with the default being the current behavior (i.e., always indent/unindent, never inserting a space). You can just amend the existing config objects and add this setting to the other section of the editor settings (unless you find a better place).

Thank you!

benniekiss:
Sure, I can make it configurable. I think it should default to allowing inserting tabs/spaces in the text, since it would otherwise be a big change from previous Zettlr behavior, which currently allows inserting tabs.

I don't think I should separate out the YAML logic, though. The YAML case would be an exact duplicate, except for the YAML check.

> If I understand you correctly, this would change the behavior so that, when one is somewhere within a line, this would no longer indent the line, but instead insert a literal tab character. This would be a regression, because I have grown accustomed to just clicking somewhere in a line to indent/unindent it.

Just to clarify, this would only be a regression from the changes in #6168. Before that PR, Zettlr used the codemirror `insertTab`, which either inserts a literal tab, or indents the line if the selection is non-empty.

benniekiss:
Another point -- there are three places in the code that could re-use the new `posInNode` function. I will save those refactors for a separate review, unless you think I should include them here.

- https://github.com/Zettlr/Zettlr/blob/a73f0dcdda60d57126104958e9d1a7174117bb8f/source/common/modules/markdown-editor/commands/autocorrect.ts#L25-L58

- https://github.com/Zettlr/Zettlr/blob/a73f0dcdda60d57126104958e9d1a7174117bb8f/source/common/modules/markdown-editor/plugins/default-context-menu.ts#L23-L46

- https://github.com/Zettlr/Zettlr/blob/a73f0dcdda60d57126104958e9d1a7174117bb8f/source/common/modules/markdown-editor/tooltips/hyperlinks.ts#L32-L45

nathanlesage:
> Sure, I can make it configurable. I think it should default to allowing inserting tabs/spaces in the text, since it would otherwise be a big change from previous Zettlr behavior, which currently allows inserting tabs.
> 
> I don't think I should separate out the YAML logic, though. The YAML case would be an exact duplicate, except for the YAML check.
> 
> > If I understand you correctly, this would change the behavior so that, when one is somewhere within a line, this would no longer indent the line, but instead insert a literal tab character. This would be a regression, because I have grown accustomed to just clicking somewhere in a line to indent/unindent it.
> 
> Just to clarify, this would only be a regression from the changes in #6168. Before that PR, Zettlr used the codemirror `insertTab`, which either inserts a literal tab, or indents the line if the selection is non-empty.

Oh god, I completely overlooked that, which might be due to me never wanting Tabs in my files, so I didn't realize this. So yes, your change is correct. Let me go through this again, because given the fact that it worked the last years without me complaining means it worked as intended. So no need to make this configurable.

nathanlesage:
Ah I see you made it configurable already. Then let's keep that, it can't hurt, and this way your effort wasn't wasted!

source/common/modules/markdown-editor/commands/lists.ts
```diff
@@ -269,26 +200,16 @@ function correctListMarkers (state: EditorState): Transaction {
  * @return  {boolean}             Whether the command has handled the keypress
  */
 export function maybeIndentList (target: EditorView): boolean {
-  const cmd = {
-    state: target.state,
-    dispatch: (transaction: Transaction) => target.dispatch(transaction)
+  const tree = syntaxTree(target.state)
+  if (nodeInSelection(target.state.selection, tree, [ 'OrderedList', 'BulletList' ], -1)) {
```
nathanlesage, line 204: Is `TaskList` a subtype of these nodes? If so, you can resolve this comment, but we want to make sure task lists count as lists!

source/app/service-providers/config/get-config-template.ts
```diff
@@ -132,6 +132,7 @@ export interface ConfigOptions {
     enableTableHelper: boolean
     indentUnit: number
     indentWithTabs: boolean
+    indentLine: boolean
```
nathanlesage:
```suggestion
    alwaysIndentLineOnTab: boolean
```

source/app/service-providers/config/get-config-template.ts
```diff
@@ -340,6 +341,7 @@ export function getConfigTemplate (): ConfigOptions {
       enableTableHelper: true, // Enable the table helper plugin
       indentUnit: 4, // The number of spaces to be added
       indentWithTabs: false,
+      indentLine: false, // Whether `Tab` always indents the current line
```
nathanlesage:
```suggestion
      alwaysIndentLineOnTab: false, // Whether `Tab` always indents the current line
```

source/common/modules/markdown-editor/util/configuration.ts
```diff
@@ -51,6 +51,7 @@ export interface EditorConfiguration {
   idGen: string
   indentUnit: number
   indentWithTabs: boolean
+  indentLine: boolean
```
nathanlesage:
```suggestion
  alwaysIndentLineOnTab: boolean
```

source/common/modules/markdown-editor/util/configuration.ts
```diff
@@ -114,6 +115,7 @@ export function getDefaultConfig (): EditorConfiguration {
     idGen: '',
     indentUnit: 4,
     indentWithTabs: false,
+    indentLine: false,
```
nathanlesage:
```suggestion
    alwaysIndentLineOnTab: false,
```

source/win-preferences/schema/editor.ts
```diff
@@ -319,6 +319,11 @@ export function getEditorFields (config: ConfigOptions): PreferencesFieldset[] {
           label: trans('Indent using tabs instead of spaces'),
           model: 'editor.indentWithTabs'
         },
+        {
+          type: 'checkbox',
+          label: trans('Always indent the current line when pressing Tab'),
+          model: 'editor.indentLine'
```
nathanlesage:
```suggestion
          model: 'editor.alwaysIndentLineOnTab'
```

source/win-main/MainEditor.vue
```diff
@@ -233,6 +233,7 @@ const editorConfiguration = computed<EditorConfigOptions>(() => {
   return {
     indentUnit: editor.indentUnit,
     indentWithTabs: editor.indentWithTabs,
+    indentLine: editor.indentLine,
```
nathanlesage:
```suggestion
    alwaysIndentLineOnTab: editor.alwaysIndentLineOnTab,
```

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -612,3 +614,45 @@ export function applyPandocDivOrSpan (target: EditorView, type: 'div'|'span', at
...
+ *
+ * This version inserts either the configured `indentUnit`, or, if the selection
+ * is within a YAMLFrontmatter node, it inserts `tabSize` number of spaces
+ *
+ * @param   {EditorView}  target      The target view
+ *
+ * @returns {boolean}
+ */
+export function insertTabOrSpace (target: EditorView): boolean {
+  const { indentWithTabs, indentLine } = target.state.field(configField)
```
nathanlesage:
```suggestion
  const { indentWithTabs, alwaysIndentLineOnTab } = target.state.field(configField)
```

source/common/modules/markdown-editor/commands/markdown.ts
```diff
@@ -612,3 +614,45 @@ export function applyPandocDivOrSpan (target: EditorView, type: 'div'|'span', at
...
+  if (indentWithTabs && nodeInSelection(target.state.selection, tree, [ 'YAMLFrontmatter', 'YAMLFrontmatterStart', 'YAMLFrontmatterEnd' ], -1)) {
+    // We need to temporarily override the `indentWithTabs` setting
+    // so that the `indentUnit` facet updates to insert spaces.
+    // This is necessary for `indentMore` to insert the correct indent.
+    target.dispatch({ effects: configUpdateEffect.of({ indentWithTabs: false }) })
+  }
+
+  let result = false
+  // If any range is not empty, indent the line
+  if (indentLine || target.state.selection.ranges.some(r => !r.empty)) {
```
nathanlesage:
```suggestion
  if (alwaysIndentLineOnTab || target.state.selection.ranges.some(r => !r.empty)) {
```

source/win-preferences/schema/editor.ts
```diff
@@ -319,6 +319,11 @@ export function getEditorFields (config: ConfigOptions): PreferencesFieldset[] {
           label: trans('Indent using tabs instead of spaces'),
           model: 'editor.indentWithTabs'
         },
+        {
+          type: 'checkbox',
+          label: trans('Always indent the current line when pressing Tab'),
```
nathanlesage, line 324:
```suggestion
          label: trans('Always indent the current line when pressing Tab'),
          info: trans('Zettlr always indents list items. Turn this setting on to always indent any line and never insert Tab characters.'),
```

benniekiss: So this description is not entirely accurate, as it doesn't control whether tab characters are inserted or not, that would be controlled by the setting above, `indentWithTab`. It only determines whether pressing tab always indents the line, or if it inserts an indent unit at the cursor.

source/common/modules/markdown-editor/commands/lists.ts
benniekiss: I believe `TaskList`s are indeed sub-nodes. Just tested with the alway-indent disabled, and task list lines still indent the line when pressing tab, so I assume this is fine.

source/win-preferences/schema/editor.ts
nathanlesage: Hmhm, I'll think about this and improve the wording later on. No need to worry about this detail at this point :)

nathanlesage:
Thank you!

PR #6209, opened 2026-02-24
[Feat] Add close all files/workspaces function

### LINKED ISSUE #6096: Add modifier key for simultaneously closing tab and its associated file.
### Description

Often, when I open a file that's not part of a workspace, it's something that I don't need to keep loaded in Zettlr long-term. Currently, when I want to remove that file from Zettlr, I have to complete two separate actions — _1)_ closing the tab in which the file is open and then _2)_ closing the file in the file manager.

I suggest streamlining this workflow by introducing a modifier key the user could press when closing a tab to simultaneously close the file in the file manager. For example, holding down `Shift` while clicking the `×` button on a tab or pressing `Ctrl-Shift-w` rather than simply `Ctrl-w` would complete both actions.

### Proposed Changes

See above.

### Caveats

The only tricky thing is that you wouldn't want this feature for files that were part of workspaces, since the idea of having workspace folders is to keep a list of all the files in a particular folder handy. If the user uses the modifier key on a tab containing a file from a workspace, my suggestion would be to simply close the tab and leave the workspace file listing alone. Possibly you could pop up a status message somewhere to explain that workspace files are not closed by the modifier. But I don't even think that's necessary.

No

### Zettlr Version

Nightly

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Operating System Version

Manjaro (Rolling)

### Additional Information

Thank you!

benniekiss:
## Description

This PR adds the ability to close all files or workspaces. In addition, it adds a collapse all directories button, a context menu item to the documents tab to close the file, and it improves handling of file closing.

## Changes

The main feature is that a button was added to the filetree that allows a user to close all open files or workspaces. The button is accessible by holding shift and hovering over the `Files` or `Workspaces` dropdown arrow, at which point, it will turn into an `X`. When the `X` is double-clicked, all files or workspaces are closed.

As a result, the `documents-provider` now exposes the `closeFileEverywhere` function via the `ipc` interface, and an additional method was added, `getOpenWorkspaceFiles`. This is needed when closing workspaces.

The `closeFileEverywhere` was refactored to re-use the logic of `forEachLeaf`.

Another button was added to `Workspace` section heading to collapse all open directories. If child directories are uncollapsed, then on first click, only children will be collapsed, leaving the roots open. If clicked and only root directories are uncollapsed, then they will be collapsed.

When a file or workspace is closed from the sidebar, all open tabs are now closed as well.

A context menu item was added to the documents tab bar to close the associated file.

the `askSaveChanges` dialogue now takes an additional `detail` parameter. This is used to include the filename of the document currently being asked for confirmation. Because multiple files can be opened from a workspace, when closing an entire workspace and some files have unsaved changes, this makes it easier to know which file is being prompted.

Collapsed directory handling UX was refactored in improved in some places. Mainly, this was to prevent uncollapsing the currently selected dir when uncollapsing a different one.

The hover time required to expand a directory when dragging was reduced from 2 seconds to 1 second.

## Additional information

This potentially closes #6096, although via a  different route.

Some recordings of the new behavior:

https://github.com/user-attachments/assets/6af0ff2a-c6ab-44d9-be3d-b2df8d8747b4

https://github.com/user-attachments/assets/829a4f74-3249-4343-9f0f-3031e4f49b60

https://github.com/user-attachments/assets/35f1ec46-d760-4363-bb09-295aa49d1d68

Tested on: 

macOS 26

### DIFF
--- a/source/win-main/file-manager/FileTree.vue
+++ b/source/win-main/file-manager/FileTree.vue
@@ -325,34 +325,39 @@ function closeAllWorkspaces (): void {
   }).catch(err => console.error(err))
 }
 
-// Collapse all folders.
-function collapseAll (): void {
-  const uncollapsed = [...windowStateStore.uncollapsedDirectories]
-  const roots = rootDescriptors.value.map(r => r.path)
-
-  // Only collapse child folders, not roots, if any children are uncollapsed.
-  let onlyRoots = true
-  for (const filePath of uncollapsed) {
-    if (!roots.includes(filePath)) {
-      let idx = windowStateStore.uncollapsedDirectories.indexOf(filePath)
-      if (idx > -1) {
-        windowStateStore.uncollapsedDirectories.splice(idx, 1)
-        onlyRoots = false
-      }
-    }
+// Collapse uncollapse folders. If `collapseRoots` is `true`, also collapse
+// root workspace directories.
+function collapseAll (collapseRoots: boolean): void {
+  // Collapse all folders and roots.
+  if (collapseRoots) {
+    windowStateStore.uncollapsedDirectories.splice(0)
+    return
   }
 
-  // If all of the children are collapsed, then collapse the roots.
-  if (onlyRoots) {
-    windowStateStore.uncollapsedDirectories.splice(0)
+  // Collapse only child folders, leaving roots uncollapsed
+  const roots = new Set(rootDescriptors.value.map(r => r.path))
+
+  const uncollapsed = windowStateStore.uncollapsedDirectories
+    .filter(path => !roots.has(path))
+
+  for (const filePath of uncollapsed) {
+    let idx = windowStateStore.uncollapsedDirectories.indexOf(filePath)
+    if (idx > -1) {
+      windowStateStore.uncollapsedDirectories.splice(idx, 1)
+    }
   }
 }
 
 // Context menu for the `Workspaces` header
 function workspaceRootContextMenu (event: MouseEvent): void {
+  const roots = new Set(rootDescriptors.value.map(r => r.path))
+
+  const onlyRoots = windowStateStore.uncollapsedDirectories
+    .every(path => roots.has(path))
+
   const template: AnyMenuItem[] = [
     {
-      label: trans('Collapse directories'),
+      label: trans('Collapse %s', onlyRoots ? 'workspaces' : 'folders'),
       id: 'collapse-all-workspaces',
       type: 'normal'
     },
@@ -370,7 +375,7 @@ function workspaceRootContextMenu (event: MouseEvent): void {
   showPopupMenu(point, template, (clickedID) => {
     switch (clickedID) {
       case 'collapse-all-workspaces':
-        collapseAll()
+        collapseAll(onlyRoots)
         break
       case 'close-all-workspaces':
         closeAllWorkspaces()


benniekiss:
Forgot to add that this also includes shift+clicking the empty sidebar to open a file instead of a workspace

source/app/service-providers/documents/index.ts
```diff
@@ -254,7 +256,7 @@ export default class DocumentManager extends ProviderContract {
 
       if (event === 'unlink') {
         // Close the file everywhere
-        this.closeFileEverywhere(filePath)
+        this.closeFileEverywhere(filePath).catch(err => console.error(err))
```
nathanlesage: In the main process, `console.error` is not captured by the logger, so we would lose this information. Please use the logger instance directly.

source/app/service-providers/documents/index.ts
```diff
@@ -1017,6 +1029,32 @@ current contents from the editor somewhere else, and restart the application.`
...
+   * For the provided root workspace directory at `filePath`,
+   * retrieve a list of filepaths representing every open file
+   * within the workspace.
+   *
+   * @param {string}      filePath  Path of the workspace directory
+   *
+   * @returns {string[]}            A list of file paths representing the
+   *                                open files within `filePath`.
+   */
+  public async getWorkspaceOpenFiles (filePath: string): Promise<string[]> {
```
nathanlesage:
```suggestion
  public async getOpenFilesForWorkspace (workspacePath: string): Promise<string[]> {
```

source/win-main/file-manager/util/item-composable.ts
```diff
@@ -25,6 +25,57 @@ import type { DocumentManagerIPCAPI } from 'source/app/service-providers/documen
 
 const ipcRenderer = window.ipc
 
+/**
+ * Close a file root path at `path`, including all open
+ * instances of the file.
+ *
+ * @param {string}  path    The filepath to close
+ */
+export function closeFile (path: string): void {
```
nathanlesage, line 34: You should, in here, check whether this path actually describes a root file, to avoid inexplicable errors. If you perform this check at other points, I suggest centralizing the logic here.

Hint: You can require the workspaces store (or wherever Zettlr stores the root descriptors) using the same invocation as in the Vue components (e.g., `const store = useStore()`)

source/win-main/file-manager/util/item-composable.ts
```diff
@@ -25,6 +25,57 @@ import type { DocumentManagerIPCAPI } from 'source/app/service-providers/documen
...
+    .catch(err => console.error(err))
+}
+
+/**
+ * Close a workspace root path at `path`, including all open files
+ * within the workspace
+ *
+ * @param {string}  path    The workspace path to close
+ */
+export function closeWorkspace (path: string): void {
```
nathanlesage, line 61: Same as with the file closer: Check that it's actually a workspace.

source/win-main/file-manager/util/item-composable.ts
```diff
@@ -100,6 +151,13 @@ export function useItemComposable (
           newTab: middleClick || (alt && type === 'file') // Force a new tab in this case.
         }
       } as DocumentManagerIPCAPI)
+        .then(() => {
+          configStore.setConfigValue('openDirectory', obj.value.dir)
+          // Finally, since it's a directory, uncollapse it.
+          if (!windowStateStore.uncollapsedDirectories.includes(obj.value.dir)) {
+            windowStateStore.uncollapsedDirectories.push(obj.value.dir)
+          }
+        })
```
nathanlesage, line 174: This is a bug fix, right?

source/win-main/file-manager/util/item-composable.ts
```diff
@@ -113,6 +171,13 @@ export function useItemComposable (
           // We leave leafId undefined
           payload: { path: obj.value.path, windowId }
         })
+          .then(() => {
+            configStore.setConfigValue('openDirectory', obj.value.dir)
+            // Finally, since it's a directory, uncollapse it.
+            if (!windowStateStore.uncollapsedDirectories.includes(obj.value.dir)) {
+              windowStateStore.uncollapsedDirectories.push(obj.value.dir)
+            }
+          })
```
nathanlesage, line 194: Same here, right?

nathanlesage: (I'm asking since I'll need to know for the changelog)

source/win-main/file-manager/TreeItem.vue
```diff
@@ -616,7 +593,7 @@ function enterDragging (_event: DragEvent): void {
   uncollapseTimeout.value = setTimeout(() => {
     windowStateStore.uncollapsedDirectories.push(props.item.path)
     uncollapseTimeout.value = undefined
-  }, 2000)
+  }, 1000)
```
nathanlesage, line 596: I'm not so sure we should decrease the timing here. But that is opinion. I'll ask around on Discord to see what others feel.

source/app/service-providers/commands/file-new.ts
```diff
@@ -161,13 +161,15 @@ export default class FileNew extends ZettlrCommand {
...
-    } catch (err: any) {
-      this._app.log.error(`Could not create file: ${err.message as string}`)
-      this._app.windows.prompt({
-        type: 'error',
-        title: trans('Could not create file'),
-        message: err.message
-      })
+    } catch (err: unknown) {
+      if (err instanceof Error) {
+        this._app.log.error(`Could not create file: ${err.message as string}`)
```
nathanlesage:
```suggestion
        this._app.log.error(`Could not create file: ${err.message}`)
```

benniekiss:
Personally I really do not like context menus, but I see how that approach fits in better with Zettlr here, so I'll get working on that.


> And one last thing: Could you elaborate a little bit more on how you changed the `uncollapseIfApplicable` behavior

I found that currently, if you have a directory selected (highlighted in the sidebar), and then collapse that directory, expanding a _separate_ directory would result in the highlighted directory also expanding. There was also some other strange collapse/uncollapse behavior, but I can't recall specifically. 

I think there was a lot of overlapping logic going on that I traced to the `uncollapseIfApplicable` function, and removing it cleared up the issue I was seeing.

nathanlesage:
Gotcha, thank you!

benniekiss:
This should be good to go on my end! I implemented the context menus and confirmation dialogue, and it's working in my tests.

nathanlesage:
Thank you for the changes! I've just tested it out locally.

1. When uncollapsing a bunch of directories, then right-clicking the workspaces section title, and selecting "Collapse directories", it only collapses directories up until the first level below the workspaces. Selecting the menu item a second time then finally collapses everything as expected.
2. The dialog boxes should also feature a question ("Do you really want to close all workspaces/files?"). Without, it looks empty and a bit weird, as if someone had forgotten something.
3. Now that I've observed it myself, I do believe that, adding an ellipsis button ("…") to the right of the section headers that also summons the context menu might be good UX so that users can discover the new functionality.

Other than that, it looks and works great!

benniekiss:
a button to summon the context menu sounds great imo! I'll add that.

The two-click collapse behavior is intended. This was inspired by VScode's behavior which I quite like

benniekiss:
Screenshots of the button, I'm open to styling changes.

The file button is `hovered`, which is why it is a different color

Screenshot 2026-03-08 at 11 30 30

nathanlesage:
Great! Maybe make the button a bit more squirkly? Like the toolbar buttons? Because afaik we don't have any entirely round borders in regular buttons. Other than that, looks good!

benniekiss:
Squarer corners and smaller border line:

Screenshot 2026-03-08 at 12 00 53

nathanlesage:
Maybe a bit more padding (= smaller icon) and give it a solid background instead of the border. I think the it's great!

Also the I think we can just see how it looks and adapt as we go :)

benniekiss:
I'd like to make the ellipsis a little bolder, but I'm not sure if that's possible with icon, otherwise I'm liking how this is so far.

Screenshot 2026-03-08 at 12 59 24

Screenshot 2026-03-08 at 12 59 30

nathanlesage:
Yes, this looks nice! Again, there's still time until the next update, so if you or I have some idea it'll make it in!

nathanlesage:
Ah, one final thing then: It is quite unusual having to click the context menu button twice. To make it more explicit, I'd like to simply exchange the label based on what it will do, e.g.:

* "Collapse folders"
* "Collapse workspaces"

We might even want to add a setting so that users can decide if they want all in one go or the two-stage design. I can see the appeal of having a two stage though.

nathanlesage:
Let me implement the setting, I will have to think about where to place it, so I'll take that onto my todo list

benniekiss:
I like that idea. I can ahead and push the conditional label, but I'll leave the setting configuration to you

nathanlesage:
One final comment on this: The `trans` function is less smart than you think. I have now fixed it in such a way that we can actually translate everything. In some instances, the strings `workspace` or `file` wasn't translated, and the `gettext` system can't decode conditionals in the translate function (it takes the string as a literal), so I fixed that. Just so you know!

benniekiss:
ahh, good too know

PR #6210, opened 2026-02-25
[Enhancement] Improve autocorrect undo history handling

benniekiss:
## Description

This PR refactors autocorrect replacement handling to improve the UX of the undo history. It makes it so that, after an autocorrect replacement occurs, undoing the transaction first reverts the replacement without removing the inserted space/newline.

## Changes

The replacement handling was refactored into two commands, `handleAutocorrectSpace` and `handleAutocorrectEnter`, so that the respective `space` or `enter` text insertions can occur **before** the autocorrect replacement happens. This improves UX for undoing autocorrect replacements.

Because the replacement now occurs **after** the space or enter, when a user tries to undo the autocorrect replacement, they are able to do so without also removing the newly entered space or newline. 

## Additional information

Currently, when a user tries to undo the replacement, it undoes the entire initial text and insertion.

However, it's not as simple as adding the `isolateHistory` annotation. Then, they must undo the history twice, once to remove the space, and again to remove the replacement. This then throws the user into a precarious loop where, on the next space or newline, the replacement will be triggered *again* because the cursor is placed adjacent to the autocorrect text, and so the user is forced to do some gymnastics to prevent the autocorrect. 

This PR makes it so that a user has to undo the history once, and they can continue typing.

Tested on: 

macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/commands/autocorrect.ts
+++ b/source/common/modules/markdown-editor/commands/autocorrect.ts
@@ -18,6 +18,7 @@ import { syntaxTree } from '@codemirror/language'
 import { EditorSelection, type ChangeSpec, type EditorState } from '@codemirror/state'
 import { type Command, type EditorView } from '@codemirror/view'
 import { configField } from '../util/configuration'
+import { insertNewlineAndIndent, isolateHistory } from '@codemirror/commands'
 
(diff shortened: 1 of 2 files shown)


nathanlesage:
So, sorry for the delay, but it looks good and can be merged!

PR #6225, opened 2026-03-07
Fix startup crash when restored files were deleted

### LINKED ISSUE #6223: Zettlr doesn't start when markdown document was moved in the meantime
### Description

Zettlr won't start again, when a document was moved or deleted.
Zettlr 4.2.0
Windows 10

### Reproducing

- Edit a markdown document i.e. test.md, save it an close Zettlr.
- Remove test.md
- Zettlr won't start again

### Zettlr version

4.2.0

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Windows 10

### Additional Information

_No response_

SergioChan:
## Description
Prevent session restore from aborting startup when `openFiles` includes a file path that no longer exists.

## Changes
- Wrap `getDescriptorFor(absPath)` in `FSAL.reindexFiles` with error handling.
- Skip stale file paths (moved/deleted while app was closed) and continue indexing remaining files/workspaces.
- Add an upcoming changelog entry for #6223.

## Additional information
This keeps the fix narrowly scoped to startup reindexing: missing paths are logged and ignored instead of throwing and terminating boot.

Tested on: Linux (WSL2) — ran `yarn lint:code source/app/service-providers/fsal/index.ts` successfully.

Fixes #6223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ verify any of your exports.
 - Fix indenting and tab insertion behavior (#6196).
 - Fix Zettlr not correctly parsing Pandoc extensions (#6212).
 - Improved error handling when opening PDF attachments.
-- Prevent startup crashes when session restoration references files that were moved or deleted while Zettlr was closed (#6223).
+- Prevent startup crashes when session restoration references files that were
+  moved or deleted while Zettlr was closed (#6223).
 
 ## Under the Hood
 
--- a/source/app/service-providers/fsal/index.ts
+++ b/source/app/service-providers/fsal/index.ts
@@ -287,7 +287,7 @@ export default class FSAL extends ProviderContract {
         // Zettlr was not running. Skip those stale entries so startup can
         // continue and the remaining files/workspaces still load.
         const message = err instanceof Error ? err.message : String(err)
-        this._logger.warning(`[FSAL] Skipping stale path during reindex: ${absPath} (${message})`)
+        this._logger.error(`[FSAL] Skipping stale path during reindex: ${absPath} (${message})`)
       }
     }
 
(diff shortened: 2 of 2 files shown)


benniekiss:
I'll test this out myself, but from a quick review this definitely seems like the right approach! Thank you for the PR

nathanlesage:
I have a hunch that this issue does not fix what the comments suggest. `readDirectoryRecursively` will always read the **current** directory contents, and not outdated cache information. So if `getDescriptorFor` throws an error, this indicates only that the file described by `absPath` has been deleted *after* it has been read in by `readDirectoryRecursively` but *before* the function call to `getDescriptorFor`.

If that fixes the error indicated in the issue you linked, there is something else afoul. I would appreciate it if we could get to the bottom of this.

I agree that this PR nonetheless includes a good fix in any case, since errors here should not completely grind Zettlr to a halt. Just a few notes:

1. Please make an error out of the logging call. The issue is that this really should not be considered a small hiccup, but a serious error due to how the logic operates (because this should not happen in regular instances).
2. Please adjust the Changelog (I have released a new update before I saw your PR, sorry!) Also, please respect the Changelog format (the line is too long).

Other than that, thank you, of course!

benniekiss:
The function also tries to get descriptors declared in `openFiles`, which I believe is where the error gets thrown, if one of those files no longer exists.

I believe there may need to be another check on `getAllLoadedDescriptors` as well, since that does a call to `getDescriptorFor` on the config paths

SergioChan:
Posting a clean follow-up (my previous comment got mangled by shell escaping):

Thanks for the detailed review — great call on openFiles/getAllLoadedDescriptors.

I pushed a follow-up commit (ee3f13463) that addresses all points:
- switched the reindex stale-path log from warning -> error
- added stale-path guards in getAllLoadedDescriptors() for both openFiles and workspace children, so missing files from session/config paths no longer abort descriptor collection
- updated the changelog entry formatting to match line-length style

Validation:
- attempted `npm run lint:types`
- blocked in this environment because `vue-tsc` is not executable (`sh: 1: vue-tsc: Permission denied`), so I could not complete the type-check locally here

nathanlesage:
It took quite a while to figure out what you were doing. You were definitely in the right spot, but after double-checking the source code you were working on, I realized what was going wrong.

Please see commit 2781c4436cf8fa39339ab588a90451342f81f56a which fixes #6223 correctly and implements the correct logic. I have ensured your efforts in this PR have been acknowledged, because without your initial research and help, it would've taken much longer to find and fix this bug!

The reason I implemented it myself is since I do want the `getDescriptorFor`-function to continue throwing errors. This forces upstream callers to properly check if a path is valid beforehand and thus handle the errors accordingly. Sometimes, it is fine if a path does not exist (then the caller can catch the errors), sometimes a file really must exist, and we want to bubble up the error to the original caller, which usually is a user action, meaning they see an error dialog.

PR #6229, opened 2026-03-07
[Enhancement] Refactor editor themes

benniekiss:
## Description

This PR refactors the codemirror theme code to be more centralized, structured, and customizable. 

It is not intended to change any styling beyond those needed for organization/simplicity

## Changes

While this PR is quite large, it is essentially reproducing a lot of boiler plate and reorganizing where things are located.

The main change is the implementation of CSS variables to apply various styles to editor components. By using variables, theme implementation becomes much simpler because themes only have to update the variable values rather than target individual elements. Elements styling also remains synchronized and consistent across themes. Finally, it simplifies custom themeing, as users can target the variables rather than craft unique selectors per element.

The geometry styling for the editor was moved to `editor.css` outside of the `main-override.ts` file. I did this so that `main-override` and the other themes only handle visual styling (color, font, size, etc).

The code styling was moved into the theme code from the `render-code` renderer. This integrates code styling into the main themes, again for centralization and organization.

The syntax tag highlighting code was refactored into `customTagHighlight` and `tagHighlight`. This deduplicates a lot of tags, and better organizes the code. 

## Additional information

My goal was to change the themes as little as possible, if at all. I believe there are one or two elements that had a color changed, but otherwise, I think I was successful in leaving everything as is! 

Tested on:


benniekiss:
I also plan to update the docs with the available variables and acceptable values once this has gotten the go-ahead on the implementation

source/common/modules/markdown-editor/editor.css
```diff
@@ -27,13 +27,96 @@
...
+   * 2.) It uses two identical animations for that.
+   * 3.) It uses a "step" animation to make the transition "harsh"
+   * 4.) For the animation source code, see:
+   *     https://github.com/codemirror/view/blob/main/src/theme.ts
+   *
+   * Our changes:
+   * * Remove the steps(1)-function to remove the harsh transitions.
+   * * Make the transition more smooth with a 15% opacity transition period.
+   */
+  @keyframes cm-blink {
```
nathanlesage, line 101: Something in these animations is off now. The cursor animation feels much slower than before. Can you take a look?

source/common/modules/markdown-editor/theme/editor.ts
```diff
@@ -13,126 +13,489 @@
...
  */
+
 import { EditorView } from '@codemirror/view'
 
-export const mainOverride = EditorView.baseTheme({
-  // General overrides
-  '&.cm-editor': {
-    height: '100%',
-    fontFamily: 'inherit',
+export interface ThemeVars {
```
nathanlesage, line 32: Can you provide a little bit of commentary on the general idea behind these variables? That is: How did you select which to make its own variable and how variables vs. hard-coded styling works here? We currently know little about how you intend this to be used, and that would be important. (Same with the Code variables below.)

source/common/modules/markdown-editor/theme/editor.ts
```diff
@@ -13,126 +13,489 @@
...
+  '.cm-hr':  {
+    font: 'var(--zettlr-editor-strong-style) var(--zettlr-editor-font-size) var(--zettlr-editor-font)',
+    color: 'var(--zettlr-editor-primary-color)',
+  },
+  '.cm-info-string': {
+    opacity: 'var(--zettlr-editor-opacity)',
+  },
+  '.cm-link': {
+    color: 'var(--zettlr-editor-primary-color)',
+    textDecoration: 'var(--zettlr-editor-line-decoration)',
```
nathanlesage, line 321: Are you certain we need non-color variables…? If so, it really would make sense to have some documentation (inside this file!) about the general logic.

source/common/modules/markdown-editor/theme/main-override.ts
```diff
@@ -13,126 +13,489 @@
...
+    font: 'var(--zettlr-editor-header-style) var(--zettlr-editor-header-5-size) var(--zettlr-editor-font)',
+    '--zettlr-editor-code-style': 'var(--zettlr-editor-header-style)',
+  },
+  '.cm-line:has(:not(.cm-quote).cm-header-6)': {
+    font: 'var(--zettlr-editor-header-style) var(--zettlr-editor-header-6-size) var(--zettlr-editor-font)',
+    '--zettlr-editor-code-style': 'var(--zettlr-editor-header-style)',
+  },
+})
+
+/* Code Theme
```
nathanlesage: Just a thought: Maybe separate the main theme and the code theme into their own respective files, and just merge them in here? That could help with keeping it a bit more easy to follow?

nathanlesage (changes requested):
Sooooo, sorry again for the delay! But now I've finally taken a look at it.

First, I really appreciate the cleanliness of the themes now, this looks great and more comprehensible. Also, I appreciate that the app now distinguishes between CodeMirror builtins and Zettlr-specific elements in the variables. Lastly, even though you've moved around a lot of code, it looks definitely more organized now. I tested this out locally, and it feels mostly similar, with some improvements and changes you made that feel like they are well-placed.

That said, I do have a few small comments. First, I am a little bit concerned by the `main-override.ts` file, specifically its size. This can also be due to the diff, but I wonder if it might make it easier if we split out the two theme variable definition blocks into their own utility files and just merge them in there? (Feel free to disagree if you've already tried that!) Lastly, a few more comments on the logic will help so that other contributors (read: primarily I) can adopt that new system correctly. For what it's worth, I think that your theme variable documentation that you plan can simply by a copy-and-paste of the interfaces you declared into the documentation, so this way we have the documentation exactly where we need it (inside the code), but also in the official documentation so that modders can quickly find it.

Oh, and, somehow the cursor blink animation feels slower now, that one you would need to take a look at. Other than that, I think it's a great effort!

source/common/modules/markdown-editor/theme/main-override.ts
benniekiss: That makes sense to me

source/common/modules/markdown-editor/theme/editor.ts
benniekiss: So my methodology was to combine all of the themes to find the elements which were styled, and to determine which elements shared styling. All the themes were grouped relatively the same. After finding the list of all groups, I broke down what was styled within each group. Any style that changed between themes was converted into a CSS variable. Anything that was the same across themes was kept hardcoded.

The variables are named loosely off previous variables (primary color), the general CSS attribute they apply to (font), or a vague descriptor on the color context (accent, error). This was intended more of a structural change, so I am not tied to any of the variables themselves or the internal theme structure. I think there is plenty of room for change in that area, including adding more variables or other styling, and changing how styling is applied.

~~I will add comments indicating which CSS attribute each variable applies to, as that will be the most helpful when updating them.~~ *EDIT: looks like I already did that* And I'll add a comment to the top of the interfaces elaborating on how to modify/add/remove variables and theme styles

source/common/modules/markdown-editor/theme/editor.ts
benniekiss: The original theme had these variables, so I kept them.

benniekiss:
Latest commit goes a little further with reorganizing the folder structure. I split `main-overrides` into an `editor` and `code` module. I renamed `main-override` to `index` to modularize the folder, and re-exported all of the themes from it.

I also moved the individual themes into a `themes` folder to reduce the parent clutter.

Added some comments to the top of `code.ts` and `editor.ts` briefly explaining maintenance (adding/renaming variables)

I still need to address the cursor animation, and I'll get to that today or tomorrow.

benniekiss:
Latest commit fixes the cursor animation.

@nathanlesage This is good to go on my end if you want to look it over

nathanlesage:
Hello hello! Looks good, I tested it and everything seems good to go! I'm just wondering because having a folder `theme/themes/berlin.ts` seems a bit silly. But you've done so much work already, I don't want to make you bother with it once more, this PR deserves to be merged. Thank you once again :)

benniekiss:
Feel free to change that as you see fit! I felt that they had to go into some sub folder, but I couldn't think of anything good enough. `styles` maybe? But that also felt redundant

nathanlesage:
That's why I didn't suggest any alternative, because styles also feels somewhat off. I mean, in the end it's just names, and we'll be fine to find the themes! I felt I shouldn't bother anyone with that nitpick :)

PR #6243, opened 2026-03-16
i18n: update zh-CN.po

hch12907:
## Description
Updates the Chinese translation of Zettlr, increasing the percentage of translated strings from 48% to 85-ish%. It is by no means complete, but it means Zettlr is now very usable for a normal Chinese user.

## Changes
I added a bunch of translations. :)

Tested on: Arch Linux (rolling)


krisfremen:
谢谢！

nathanlesage:
Thanks a lot!

PR #6245, opened 2026-03-17
Make clear that Ctrl-Shift-o opens a new workspace

rnestler:
To me it wasn't clear that I need to use Ctrl-Shift-o and was confused when I tried it with Ctrl-o. I didn't notice that the "O" was capitalized which indicates to use shift.

I realize that I would need to change this in all languages, which I can do if there is interest in the change.

## Description
Documentation change.

## Additional information

Tested on:


nathanlesage:
Good point, thank you.

PR #6260, opened 2026-03-24
Add Japanese tutorial

coolvitto:
## Description
Add Japanese tutorial.

## Additional information

Tested on: Windows 11 Build26200.8106


nathanlesage:
I love how you just kept the German Marx quotation!

(You can, if you so please, also change the examples to something possibly more culturally appropriate when you translate tutorials. It's more about showing users how to use the app, not perfect congruence between the data!)

coolvitto:
I see.
That's a very good suggestion.
I'd like to revise it next time if I find any good quotes.

nathanlesage:
Feel free to! But no rush, unless German is very weird for Japanese people I think it'll work just fine. No need to push it!

PR #6276, opened 2026-03-30
[Security] HTML insanity

nathanlesage:
Hi,

I'm growing tired of people complaining that there are holes open for potential XSS attacks. Instead of letting random people on the internet find one issue after the other in a very piecemeal fashion, I decided to give it a go and put a stop to the shenanigans.

Now, the app should be hopefully almost as tight as an airlock on the ISS. In short, what I've done:

* Replaced `sanitize-html` with `DOMPurify` everywhere
* Translations (which didn't really have HTML before) now actually aren't allowed to have HTML anymore (cf. the `v-html` rule change below).
* Everything – including stuff that *should* be safe is now punched through DOMPurify.
* Increased the CSP for the app. Unfortunately, trusted types can't be done at this point because Clarity Icons uses Lit, and that literally lights up like a Christmas tree if I turn trusted types on.
* Throw out all instances of `v-html`. Instead, re-arm the corresponding disallow-rule again, and remove all instances. There are a few exceptions, which are handled by mangling the stuff through DOMPurify before *explicitly* disabling the rule with a corresponding note on why that is.
* One notice, however: Where there's literally just a static string inserted into the HTML, I did not add DOMPurify (yet). That's deliberate.
* Moved all sanitization towards the edge (=next to the actual injection sinks), and not somewhere in the chain where we do not have context on whether the stuff is being edited afterwards again.

Things that reviewers should please take a look at:

* Did I overlook any injection sinks that security folk may take an issue with?
* Did I break anything?

I did a cursory look, and couldn't find anything.

Thanks a bunch!


source/common/modules/markdown-editor/tooltips/citations.ts
```diff
@@ -78,7 +79,7 @@ function citationTooltip (view: EditorView, pos: number, side: 1 | -1): Tooltip|
 
           content.innerHTML = [
             options.bibstart,
-            ...bibliography[1],
+            ...bibliography[1].map(item => DOMPurify.sanitize(item)),
             options.bibend
```
krisfremen: bibstart and bibend are sanitized in references, but not here in citations?

krisfremen (commented):
Only the citations stand out on first glance. I will test more soon.

source/common/modules/markdown-editor/tooltips/citations.ts
nathanlesage: Good catch!

nathanlesage:
Thank you! Also, don't get too worked out about this. This should just be a "significant" improvement, not a "complete". Worst case, we get a few more concerning emails by security folks. So don't be afraid to maybe miss an instance or so.

nathanlesage:
I'll merge this one now; if you spot anything else, just let me know and we can continue to patch any potential XSS vulnerabilities.

PR #6283, opened 2026-04-05
[Feat] Add active tab name to window title

RTnhN:
## Description
This adds the name of the active tab into the window title and the option to enable or disable it. 

## Changes
This first refactored the getTabText function to it's own file so that it could be used in the current place `DocuemntTabs.vue` file and in the `app.vue` where the page title is defined.

The main part of the change was changing the `app.vue` to change the title of the page to include the active tab name using the existing tab name function that was refactored. 

I was not sure if this is something that everyone would want, so I added an option in the settings to enable or disable this. 

## Additional information
A main driver for this is the [Activity Watch](https://github.com/activitywatch/activitywatch) project. Activity Watch is a free, open-source, and private activity automated time tracker. [Someone on the forum](https://forum.activitywatch.net/t/need-help-developing-a-watcher-for-zettlr/4880) asked if it was possible to have a watcher (a watcher is just a small program that tracks some sort of metric then sends that data to the user's local, private server) for Zettlr. I realized that Zettlr does not surface the tab name in the title like some other programs like VS Code. 

I think that for researchers interested in tracking time, it would be really helpful for them to know exactly how long they work on individual files rather than in the main program. They could just manually time track this, but this is error prone and not exact.   

Tested on: Windows 11 WSL Ubuntu 20.04.1 LTS


nathanlesage:
Funfact: This is something I always wanted to implement, because it's kind of a no-brainer, but never did. Why? Because I'm on macOS where there are no title bars. Hence you rarely see the window title.

That being said, in general I think your code looks perfectly fine. Just one request:

1. I strongly believe that nobody realistically will have anything against having the active filename's title in the window, so to keep the changes small I suggest removing the configurability. I'll make sure to announce this appropriately.

Once you've made this change, I'll take a closer look at the PR, but I couldn't find anything weird at all while quickly parsing it right now.

RTnhN:
Okay, great! I would be fine with that. I am usually a fan of giving people an option, but since it is a relatively minor thing, just making it default is a better option. 

By the way, this is more of a background change. On windows, it actually does not show the title either:

image

This title is really only available to programs looking at the processes. For example, the task manager on windows can see this:

image

Also, you can make programs that look up the title of the active window like a simple program that I made below:

image

I ran the linting, but all of the lint suggestions were for files that I did not touch with these changes.

tliebig:
Thoughtful and well developed PR. I second this (actually to be able to track my Zettlr usage more finegrained via ActivityWatch). Thanks @RTnhN 🙏🏻

source/win-main/util/get-tab-text.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { OpenDocument } from '@dts/common/documents'
+import { pathBasename } from '@common/util/renderer-path-polyfill'
+import type { ConfigOptions } from 'source/app/service-providers/config/get-config-template'
+import type { AnyDescriptor } from 'source/types/common/fsal'
+
+interface WorkspaceDescriptorLookup {
+  descriptorMap: Map<string, AnyDescriptor>
+}
+
+export default function getTabText (
```
nathanlesage: Please rename this function + file to `getDocumentTitle` to align with what it actually does.

source/win-main/util/get-tab-text.ts
```diff
@@ -0,0 +1,49 @@
...
+  doc: OpenDocument,
+  config: ConfigOptions,
+  workspaceStore: WorkspaceDescriptorLookup
```
nathanlesage: Since this function is only used in the renderer, my suggestion is to retrieve the necessary stores within this function. This has little overhead once they are instantiated, but it keeps the function very simple and the logic together. (i.e., use `const configStore = useConfigStore()`).

source/win-main/util/get-tab-text.ts
```diff
@@ -0,0 +1,49 @@
...
+): string {
+  const file = workspaceStore.descriptorMap.get(doc.path)
```
nathanlesage:
```suggestion
  const descriptor = workspaceStore.descriptorMap.get(doc.path)
```

nathanlesage (changes requested):
So, I was now able to review your changes. Technically, everything looks perfectly fine. I only have some requests for renaming and a bit of refactoring to align this PR with the overall code base.

Thanks for implementing this!

source/win-main/util/get-tab-text.ts
RTnhN: Sure, I can fix this. But I have one quick question. I don't have a problem naming the function to `getDocumentTitle`, but it seems that kebab-case is more common than camelCase for typescript file names like `tabs-context.ts` or `find-object.ts`. Would you still like for the file to also be called `getDocumentTitle.ts` or should it be `get-document-title.tx`?

source/win-main/util/get-tab-text.ts
RTnhN: Oh, I like that better. I can definitely do that. This also will remove a kind of awkward wrapper in the DocumentTabs.vue file below:

``` typescript
function getTabTextForDocument (doc: OpenDocument): string {
  return getTabText(doc, configStore.config, workspaceStore)
}
```

source/win-main/util/get-tab-text.ts
nathanlesage: Kebab-case for filenames, camelCase for functions!

source/win-main/util/get-tab-text.ts
nathanlesage: That was my intention.

RTnhN:
Ok, your three comments should be resolved with [745219a](https://github.com/Zettlr/Zettlr/pull/6283/commits/745219a29be7f81339a605d97da099301fb67a35). Let me know if you have any other comments or suggestions.

nathanlesage:
Looks good now, only one minor linter error. One that's fixed, we can go ahead and merge! Please see that I have just changed the Changelog. My suggestion would be to remove your change to the Changelog, then I would simply do this after your PR gets merged, that's the easiest.

Thank you!

RTnhN:
> Looks good now, only one minor linter error. One that's fixed, we can go ahead and merge! Please see that I have just changed the Changelog. My suggestion would be to remove your change to the Changelog, then I would simply do this after your PR gets merged, that's the easiest.
> 
> Thank you!

Great! I fixed the linting error and removed that change from CHANGELOG.

nathanlesage:
Looks good, thank you!

RTnhN:
You're welcome! Thanks for letting me contribute and for having the contribution process be so smooth.

PR #6302, opened 2026-04-23
Add HCL/Terraform support via CodeMirror in fenced code block

a-goodarzi:
## Description
This PR adds HCL/Terraform fenced code block support in the markdown parser and updates code block autocomplete to suggest both hcl and terraform.

## Changes
- Added CodeMirror HCL language mapping in `markdown-parser.ts` using [Codemirror-lang-hcl](https://github.com/haoqixu/codemirror-lang-hcl) and mapped both hcl and terraform selectors to the same HCL language
- Added autocomplete entries for both HCL and Terraform code fences

Breaking API changes: none.

## Additional information
Terraform syntax is based on HCL, so both fence names are highlighted through the same CodeMirror language support. 

Tested on: Cachyos Rolling and macOS Tahoe (26.04.1)


nathanlesage:
Looks good, thank you!

PR #6317, opened 2026-05-01
feat: Add curly quotes transformation command and context menu option

### LINKED ISSUE #6259: Straight quotes to curly quotes and vice versa
### Checklist

- [ ] Zettlr does not possess this feature (if it does, but the feature doesn't work out for you, propose an enhancement instead.)
- [ ] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)

### Feature Proposal

As a writer and journalist, I _always_ use curly quotes except when writing code. I noted in the documentation that there is a tool to convert curly quotes to straight, but no matching tool to convert straight to curly.

I can use keyboard shortcuts to insert the curly single and double quotes as needed. But, it would be nice when bringing in text from elsewhere to be able to fix all of them at once. Using find-and-replace 

### Example Workflow

Import a block of text from elsewhere that uses straight quotes, and use a tool to change them to curly quotes.

### Caveats

If this is a tool one can choose to use, rather than a setting, I think it would be better so people don’t invoke it accidentally.

No

### Platforms

- [x] Windows
- [x] macOS
- [x] Linux

### Additional Information

_No response_

wyl2003:
## Description

Add a "Curl quotes" text transform — the inverse of the existing "Straighten quotes" — that converts straight quotes (`"` and `'`) to curly (smart) quotes based on the user's configured Magic Quotes settings.

Closes #6259

## Changes

- **New file**: `source/common/modules/markdown-editor/commands/transforms/curl-quotes.ts` — the core transform function
- **Modified**: `source/common/modules/markdown-editor/context-menu/transform-items.ts` — added "Curl quotes" to the Transform context menu, placed right after "Straighten quotes"
- **New file**: `test/transform-selected-text/curl-quotes.spec.ts` — unit tests
- **Modified**: `CHANGELOG.md` — added entry under Upcoming

No breaking changes.

## Additional information

The algorithm reuses the same heuristic as the existing Magic Quotes autocorrect (`handleQuote` in `autocorrect.ts`): a quote preceded by a space, opening bracket, dash, or start-of-text becomes an opening quote; otherwise it becomes a closing quote. This ensures consistent behavior between typing and bulk-converting.

Tested on: Windows 11


source/common/modules/markdown-editor/commands/transforms/curl-quotes.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Convert straight quotes to curly (smart) quotes, using the user's configured
+ * Magic Quotes characters. This is the inverse of `straightenQuotes`.
+ *
+ * @param   {string}  primary    The primary (double) Magic Quotes config, e.g. `\u201c\u2026\u201d`
+ * @param   {string}  secondary  The secondary (single) Magic Quotes config, e.g. `\u2018\u2026\u2019`
+ *
+ * @return  {StateCommand}       A `StateCommand` that curls straight quotes.
+ */
+export function curlQuotes (primary: string, secondary: string): StateCommand {
```
nathanlesage: I'd suggest making the parameters accept only the actual quotes (such as: `primary: [string, string]`) instead of the config string that has to be split first.

The reason is that this way the function isn't bound to how the Config defines quotes and might be reused. In addition, you could keep the `split`-logic closer to the configuration that you retrieve (see below).

source/common/modules/markdown-editor/context-menu/transform-items.ts
```diff
@@ -69,6 +70,14 @@ export function getTransformSubmenu (view: EditorView): SubmenuItem {
         type: 'normal',
         action () { straightenQuotes(view) }
       },
+      {
+        label: trans('Curl quotes'),
+        type: 'normal',
+        action () {
+          const { magicQuotes } = view.state.field(configField).autocorrect
```
nathanlesage, line 77: I'd suggest adding the split-logic here. This will make it easier for us to understand in the future what we're doing (see my comment above).

source/common/modules/markdown-editor/context-menu/transform-items.ts
```diff
@@ -69,6 +70,14 @@ export function getTransformSubmenu (view: EditorView): SubmenuItem {
         type: 'normal',
         action () { straightenQuotes(view) }
       },
+      {
+        label: trans('Curl quotes'),
```
nathanlesage: "Curl quotes" sounds somewhat weird. Is there maybe a more natural sounding way to name this command? Could you take a look how other apps call such a feature?

test/transform-selected-text/curl-quotes.spec.ts
nathanlesage, line 1: I appreciate the dedication you put into writing up these tests. Would it be possible to make the file a bit more compact, so that it will be easier in the future to add new tests? (That is, factor out all the "Create Editor State"-logic into its own function here in the file, just to not repeat yourself so often)?

nathanlesage (changes requested):
Hi! Thank you so much for providing this PR. From what I've seen, this does what it says on the tin, which is great. However, there are few things here and there that I'd love you to address.

If you have any questions or concerns, just let me know!

source/common/modules/markdown-editor/context-menu/transform-items.ts
wyl2003: This name is actually meant as the opposite of the existing “Straighten quotes” feature. How about naming it “Quotes to smart quotes” (or something similar), since what it does is effectively applying smart quotes?

I also checked. In InDesign, there's a feature called "Use Typographer's Quotes".
image
In word, there are only autocorrect options as follows
image

source/common/modules/markdown-editor/commands/transforms/curl-quotes.ts
wyl2003: Thank you for your detailed and patient review!
I have updated `curlQuotes` to accept` [string, string] `tuples for both primary and secondary.

source/common/modules/markdown-editor/context-menu/transform-items.ts
wyl2003: Done — moved the `split('…')` into the action in `transform-items.ts`, right next to where `magicQuotes` is retrieved from config.

source/common/modules/markdown-editor/context-menu/transform-items.ts
nathanlesage: Thank you for looking this up! I think to remain with the verbiage Zettlr already uses, a suggestion would be to "turn straight quotes into Magic Quotes". A shorter alternative might be "Convert quotes to Magic Quotes".

I know that the correct wording is going to be typographic and we may change this in the future, but it's better to have everything referred to using the same terms!

wyl2003: Thanks for the suggestion! Updated to "Convert quotes to Magic Quotes".
image

test/transform-selected-text/curl-quotes.spec.ts
JinyiArthurChen: Thanks for the feedback! I have refactored the curl-quotes.spec.ts file by created a sunnyDayTestCases. This makes the file much cleaner and easier to extend.

wyl2003:
Hi,
Thank you again for all the helpful feedback! Just wanted to give you a quick ping that all the requested changes have been implemented.

The PR is ready for your review whenever you have a moment. Let me know if there's anything else needed!

nathanlesage:
Looks good now — once the tests pass we can merge. Thank you!

PR #6339, opened 2026-05-15
feat: Implement Search Provider

### LINKED ISSUE #6027: Full-Text Search Directory Filter Does Not Scroll
### Description

4.0.0-beta and 4.0.0-beta2

When executing a full-text search, the second input to restrict the search to a specific directory does not allow scrolling by mouse wheel or scroll bar.   Arrow keys do seem to send the selection down below the visible options but the user interface does not scroll down to track the selection.   This impedes search functionality for users with many directories or workspaces who would like to restrict a search to a specific directory.

### Reproducing

1.  Have many workspaces with a few directories each or a single workspace with many directories.
2.  Initiate a full-text search in the left sidebar.
3.  In the second input field, use the drop-down to open the list of directories.
4.  Attempt to scroll said list.

### Zettlr version

4.0.0-beta2

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Win11

### Additional Information

Zettlr is being used in the portable fashion.

nathanlesage:
## Description

This PR refactors the global search functionality and moves it mostly to the main process. This is done by creating a new provider for searches. The reason for this is that this allows us to implement stateful logic, including caching. This can allow us to provide better search experiences in the future which rely on pre-computed data (incl. vector search or BM25-variants).

## Changes

Two crucial functions of the full text search have been removed from the renderer and moved to the new Search Provider: The compile-search-terms function and the actual searching action.

In addition, the entire search has been overhauled:

* Clearly label it as a "boolean search" now
* Fixed an issue that would finish a search in a file when all search terms were matched immediately. This would foreclose later matches in the same file.
* Now searches are by default done only case-insensitive. Previously, the search would run twice, once in regular case, one in lower case. This cuts search times on average in half.
* Users can now specify whether they want to actually run a case sensitive search.
* Improved the terminology used throughout the search
* Removed a bunch of odd and messy choices
* Improve the search by combining results from the same file on the same line
* No longer just provide the entire line as a result; instead now a window around the search match(es) is returned of about 40 characters by default, which can include additional context information
* Moved some of the types of the search around to collocate them with the corresponding providers
* Modularize the search to allow for the inclusion of different types of search in the future

## Additional information

There currently still one iffy thing I'd like to address before merging: I found out while benchmarking that the search now is blazingly fast (good!) but this doesn't appear like such with the user, because the Vue reactivity adds so much overhead that a search that takes 0.5s visually takes ca. 4-5s until all elements are processed and rendered. There is lots of room for performance improvements here.

## AI Disclosure Statement

No AI has been used whatsoever in crafting this PR. (The search functionality was all over the place and is many years old at this point, I think an AI would've had a hard time improving that as fast as I did.)

Tested on: macOS Tahoe 26.5


nathanlesage:
@kyaso @benniekiss Anyone interested in giving this a quick sanity check? I've extensively tested this with my 10k notes library, and it works really, really well imo.

benniekiss:
i'll take a look today and tomorrow!

benniekiss:
I was able to look this over, and nothing stands out to me as being an issue. I should be able to do some actual testing today to see about any performance improvements

nathanlesage:
Wonderful! Let me know when you're done :)

source/win-main/GlobalSearch.vue
benniekiss: This needs to be a virtual scroller. Loading searches with thousands of results freezes the app.

source/win-main/GlobalSearch.vue
```diff
@@ -245,69 +279,72 @@ const directorySuggestions = computed<string[]>(() => {
...
+      if (result.file.path.toLowerCase().includes(lowercase) === true) {
+        return true
+      }
 
-    // No luck here.
-    return false
-  })
+      // No luck here.
+      return false
+    })
```
benniekiss, line 332: I have a hunch this block is the main contributor to the slow UI loading. Here, we have many many  iterations over the results lists

source/win-main/GlobalSearch.vue
nathanlesage: Ah, damn — I thought by throwing away everything not critically necessary I could keep it simple. I'll have to look into that since the virtual scroller strictly speaking requires same-height elements (or at least required it the last time I checked). I'll investigate!

source/win-main/GlobalSearch.vue
nathanlesage: Do you mean just the results filter…? If so, that should only engage once the search is over to actually filter the results, or is that specifically where you experience the lag?

benniekiss: Hm, I'll have to test again to make note of the specifics. But when I search for something like `the` in my large folder, the search takes quite a long time (>30sec). this appears along with UI lag, and a slow loading search progress bar.

What kind of lag were you seeing?

nathanlesage: Your hunch with the virtual scroller was pretty spot on. I have now done two things: (1) Do not render the search results while the search is still running (to avoid heavy updates); and (2) use a dynamic scroller to only render parts of the results.

The bad news is that this now means that the scroll bar is not totally accurate anymore (as the area dynamically shifts in size as new elements are rendered), but the good news is that even on my system, searching is much, much, much faster.

nathanlesage:
Did you have a chance to take another look? Or do you think that this is mergeable…?

benniekiss:
I can look over it now, actually! I'll do some profiling as well

source/app/service-providers/search/util/boolean-search.ts
```diff
@@ -0,0 +1,407 @@
...
+  const fileName = query.caseInsensitive ? descriptor.name.toLowerCase() : descriptor.name
+  const content = query.caseInsensitive ? fileContent.toLowerCase() : fileContent
+  const matches: SearchResult = []
+
+  // First, check for NOT-terms. Since NOT-operators immediately exclude a file,
+  // we can use this to reduce the amount of work.
+  const notOperators = query.terms.filter(elem => elem.operator === 'NOT')
+  const notWords = notOperators.reduce<string[]>((acc, curr) => acc.concat(curr.words), [])
+  for (const term of notWords) {
+    if (content.includes(term) || fileName.includes(term)) {
```
benniekiss: the `fileName` check should probably come first to short circuit on the faster path

source/app/service-providers/search/util/boolean-search.ts
```diff
@@ -0,0 +1,407 @@
...
+ * Determines the line of the provided index in the string.
+ *
+ * @param   {number}  index  The index
+ * @param   {string}  text   The full text
+ *
+ * @return  {number}         The line number
+ */
+function lineOf (index: number, text: string): number {
+  return [...text.slice(0, index).matchAll(/\n/g)].length + 1
+}
```
benniekiss, line 368: the `lineOf` calculation could be optimized here to avoid recalculating line numbers each iteration.

I would suggest saving the last found line number in `findAll` and only providing the slice between `startIndex` and `index` to `lineOf`, adding the result to the existing line number

Something like:

```ts
function findAll (searchString: string, text: string): FoundRange[] {
  const results: FoundRange[] = []

  let startIndex = 0
  let index = -1
  let lineNum = 1

  while ((index = text.indexOf(searchString, startIndex)) > -1) {
    lineNum +=  lineOf(startIndex, index, text)
    results.push({ from: index, to: index + searchString.length, line: lineNum })
    startIndex = index + 1
  }

  return results
}

/**
 * Determines the line of the provided index in the string.
 *
 * @param   {number}  start The start index
 * @param   {number}  end  The end index
 * @param   {string} text    The full text
 *
 * @return  {number}         The line number
 */
function lineOf (start: number, end: number, text: string): number {
  return [...text.slice(start, end).matchAll(/\n/g)].length
}
```

nathanlesage:
Good suggestions, thanks :) Anything else you could spot? Otherwise I'd say this is ready to merge! Any other improvements can also follow during the nightly testing…

benniekiss:
This lgtm! I agree about further improvements in nightly :)

nathanlesage:
Perfect, thanks a bunch! :)

Issue #6100, opened 2026-01-06
Triple click in a table cell does not select the whole line/cell, only the word


jdhoek:
### Description

Triple clicking on a word in a text editor generally leads to the whole line being selected (where just double click selects only the whole word). This is how text selection in Zettlr works too.

In the new table editor however, triple click does not extend beyond the word, but stays at the word level.

Expected: triple click on text in a table cell selects the whole line or all text in the cell, matching the behaviour of text elsewhere in Zettlr.

### Reproducing

* Create a table
* Add a sentence — a line of words separated by spaces — in one of the cells
* Triple click on any word of that line

### Zettlr version

4.0.0

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Ubuntu 24.04

benniekiss:
I can reproduce this. The issue is a conflict with the following code: https://github.com/Zettlr/Zettlr/blob/eeb2b1e4f4ef64d6c28b014f9ac03a513b4a7716/source/common/modules/markdown-editor/table-editor/subview.ts#L58-L80

The triple-click transaction is disallowed

nathanlesage:
Oh? My first instinct was to just say that the first click will be "swallowed" by the Table Editor anyhow because we need to mount an editor first, and thus you'd have to do a four-click. But if it's being registered, then great, because it makes our work easier!

In this case, I believe this issue is in the similar ballpark as the Home key issue. (Couldn't find it on the tracker, so it can be that this was only mentioned on Discord, but the Home key apparently doesn't move to the beginning of the table cell, but the End key does).

jdhoek:
> Oh? My first instinct was to just say that the first click will be "swallowed" by the Table Editor anyhow because we need to mount an editor first, and thus you'd have to do a four-click.

Current behaviour is that it appears to just cycle back. Four clicks does not select the sentence. Two clicks does correctly select the word, even if the table wasn't active (i.e., coming from a different section of the document), but also when the table is active (i.e., showing those `+` widgets).

nathanlesage:
Ah, here: https://github.com/Zettlr/Zettlr/issues/6083

For reference, the following are the CM commands bound to these keys:

```js
{ key: 'Home', run: cursorLineBoundaryBackward, shift: selectLineBoundaryBackward, preventDefault: true },
[...]

PR #6344, opened 2026-05-18
Fix: Allow triple-click selection in table cells by clamping to cell boundaries

wyl2003:
## Description
This PR fixes triple-click text selection inside table cells. Previously, a triple-click would attempt to select the entire line (which exceeds the cell's subview boundaries), causing the transaction to be outright disallowed — leaving the user unable to select all text in a cell with a triple-click.

The fix changes the boundary enforcement logic in `ensureBoundariesFilter `(in `subview.ts`) from disallowing out-of-bounds selection transactions to clamping them to the cell's boundaries instead. This allows triple-click select-all to work correctly within the cell.

## Changes
**`source/common/modules/markdown-editor/table-editor/subview.ts`**

- Replaced the blanket "disallow" behavior for out-of-bounds selections with clamping:
  if any selection range exceeds the cell boundaries, all ranges are clamped to the cell
  rather than the transaction being rejected.
- The existing `wasDeleted` guard is preserved — transactions that would delete the cell
  itself are still disallowed.

No breaking changes to public APIs or UX beyond the intended fix.

## Additional information
This fix was discussed in issue #6100. 

A "check-then-clamp" strategy was chosen over "always-clamp" after benchmarking. The guard avoids allocating new EditorSelection objects on every transaction during normal editing (the common case), where selections are already within bounds.
## AI Disclosure Statement

No AI was used in the development of this PR.

Tested on: Windows 11

### DIFF
--- a/source/common/modules/markdown-editor/table-editor/subview.ts
+++ b/source/common/modules/markdown-editor/table-editor/subview.ts
@@ -17,7 +17,7 @@
  */
 
 import { syntaxHighlighting, defaultHighlightStyle } from '@codemirror/language'
-import { EditorState, MapMode, Prec, StateField, type ChangeSpec, type Range } from '@codemirror/state'
+import { EditorState, MapMode, Prec, StateField, EditorSelection, type ChangeSpec, type Range } from '@codemirror/state'
 import { EditorView, drawSelection, type DecorationSet, Decoration, ViewPlugin, type ViewUpdate } from '@codemirror/view'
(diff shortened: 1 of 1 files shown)


wanzy147:
I performed the operation locally: triple-click text selection inside table cells. The result showed that the selection was successfully confined to the table cell boundaries. The original bug has been successfully resolved, and the`wasDeleted `protection mechanism remains effective.

Tested on: macOS

nathanlesage:
Looks good, thank you!

PR #6348, opened 2026-05-19
feat: Added around 300 PT-BR translations and fixed some translation issues

felipe-f-rocha:
Added around 300 PT-BR translations to pt-BR.po
Fixed some translation inconsistencies and wording issues


nathanlesage:
Impressive work — thank you very much!

PR #6349, opened 2026-05-20
Feature/open tabs at end

JinyiArthurChen:
## Description

Add a new "Open tabs at end" option for tab opening behavior. This allows users to choose whether new tabs open next to the current tab (default), replace the current tab, or always open at the end of the tab bar.

## Changes

1. Added a new enum value end_of_bar to the tab opening behavior setting.
2. Updated config migration logic to unify legacy booleans (avoidNewTabs, openTabsAtEnd) into the new tabOpeningBehavior enum.
3. Modified tab opening logic in the tab manager and document service to support opening new tabs at the end of the tab bar.
4. Updated Preferences UI to expose the new option under "Opening files behavior".
5. Updated config template types to include the new enum value.

No breaking changes.

## Additional information

1. The migration logic ensures backward compatibility with existing user configs.
2. The feature is accessible via General → Behavior → Opening files behavior.

Tested on: macOS 26.3.1 (a) (M4 Pro)

### DIFF
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -233,6 +233,7 @@ export interface ConfigOptions {
     deleteOnFail: boolean
     leaveAppRunning: boolean
     avoidNewTabs: boolean
+    openTabsAtEnd: boolean
     iframeWhitelist: string[]
     checkForUpdates: boolean
     zoomBehavior: 'gui'|'editor'
@@ -498,6 +499,7 @@ export function getConfigTemplate (): ConfigOptions {
       deleteOnFail: false, // Whether to delete files if trashing them fails
       leaveAppRunning: false, // Whether to leave app running in the notification area (tray)
       avoidNewTabs: false, // Whether to avoid opening new tabs for documents if possible
+      openTabsAtEnd: false, // Whether to always open new tabs at the end of the tab bar
       iframeWhitelist: [ 'www.youtube.com', 'player.vimeo.com' ], // Contains a list of whitelisted iFrame prerendering domains
       checkForUpdates: true,
       zoomBehavior: 'gui' // Used to determine what gets zoomed: The GUI or the editor
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -18,6 +18,7 @@ import { v4 as uuid4 } from 'uuid'
 import getLanguageFile from '@common/util/get-language-file'
 
 export type MarkdownTheme = 'berlin'|'frankfurt'|'bielefeld'|'karl-marx-stadt'|'bordeaux'
+export type TabOpeningBehavior = 'next_to_current'|'replace_current'|'end_of_bar'
 
 // This is a handy interface to add groups of file types to the settings in
 // order to allow users to display them in filemanager and/or sidebar, and open
@@ -232,8 +233,7 @@ export interface ConfigOptions {
   system: {
     deleteOnFail: boolean
     leaveAppRunning: boolean
-    avoidNewTabs: boolean
-    openTabsAtEnd: boolean
+    tabOpeningBehavior: TabOpeningBehavior
     iframeWhitelist: string[]
     checkForUpdates: boolean
     zoomBehavior: 'gui'|'editor'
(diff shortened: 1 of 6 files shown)


source/app/service-providers/config/get-config-template.ts
```diff
@@ -18,6 +18,7 @@ import { v4 as uuid4 } from 'uuid'
 import getLanguageFile from '@common/util/get-language-file'
 
 export type MarkdownTheme = 'berlin'|'frankfurt'|'bielefeld'|'karl-marx-stadt'|'bordeaux'
+export type TabOpeningBehavior = 'next_to_current'|'replace_current'|'end_of_bar'
```
nathanlesage:
```suggestion
export type TabOpeningBehavior = 'next-to-current'|'replace-current'|'end'
```

source/app/service-providers/config/get-config-template.ts
```diff
@@ -497,7 +498,7 @@ export function getConfigTemplate (): ConfigOptions {
     system: {
       deleteOnFail: false, // Whether to delete files if trashing them fails
       leaveAppRunning: false, // Whether to leave app running in the notification area (tray)
-      avoidNewTabs: false, // Whether to avoid opening new tabs for documents if possible
+      tabOpeningBehavior: 'next_to_current',
```
nathanlesage:
```suggestion
      tabOpeningBehavior: 'next-to-current',
```

source/app/service-providers/config/index.ts
```diff
@@ -286,6 +286,22 @@ export default class ConfigProvider extends ProviderContract {
         }
       }
     } // END: openPaths migration
+
+    // Migrate old tab opening booleans to the unified enum setting.
+    if ('system' in readConfig && typeof readConfig.system === 'object') {
+      const { avoidNewTabs, openTabsAtEnd } = readConfig.system
```
nathanlesage, line 292: Why is there `openTabsAtEnd` in the system config? That wasn't part of the existing config, right?

source/app/service-providers/config/index.ts
```diff
@@ -286,6 +286,22 @@ export default class ConfigProvider extends ProviderContract {
...
+      if (!('tabOpeningBehavior' in readConfig.system)) {
```
nathanlesage, line 293: I suggest moving this to the outer if statement. Removes one nested level

source/app/service-providers/config/index.ts
```diff
@@ -286,6 +286,22 @@ export default class ConfigProvider extends ProviderContract {
...
+        if (avoidNewTabs === true) {
+          readConfig.system.tabOpeningBehavior = 'replace_current'
+        } else if (openTabsAtEnd === true) {
+          readConfig.system.tabOpeningBehavior = 'end_of_bar'
+        } else {
+          readConfig.system.tabOpeningBehavior = 'next_to_current'
+        }
```
nathanlesage, line 300: Given that `openTabsAtEnd` seem to be a relic from your iterations, I suggest using a simple ternary if to assign this, with the defaults `replace` or `next` (since `end` is the new option that should not be on by default)

source/app/service-providers/config/index.ts
```diff
@@ -286,6 +286,22 @@ export default class ConfigProvider extends ProviderContract {
...
+      }
+      delete readConfig.system.avoidNewTabs
+      delete readConfig.system.openTabsAtEnd
```
nathanlesage, line 303: Is this necessary? Afaik the read config will be deleted anyhow once the system overrides its internal config state via `safeAssign`

source/common/modules/markdown-editor/tooltips/file-preview.ts
```diff
@@ -145,10 +145,9 @@ function getPreviewElement (metadata: FindFileAndReturnMetadataResult, linkConte
   openButton.addEventListener('click', openFunc)
   actions.appendChild(openButton)
 
-  // Only if preference "Avoid New Tabs" is set,
-  // offer an additional button on preview tooltip
-  // to open the file in a new tab
-  if (window.config.get('system.avoidNewTabs') === true) {
+  // In "replace current tab" mode, offer an extra button to force opening
+  // in a dedicated new tab.
+  if (window.config.get('system.tabOpeningBehavior') === 'replace_current') {
```
nathanlesage, line 150: I wonder if that isn't already handled in the main process? Could you please trace this call and see if it ends up in the document manager's `openFile` function? If so, we could even reduce the complexity of this part of the code.

source/win-preferences/schema/general.ts
```diff
@@ -4,7 +4,7 @@
  *
  * Contains:        General Preferences Schema
  * CVM-Role:        Model
- * Maintainer:      Hendrik Erz
+ * Maintainer:      Zilu Wang
```
nathanlesage, line 7: Do you want to take over maintenance of this entire file…?

nathanlesage (changes requested):
So far looks solid – a few questions remain, and there is one research task for you.

PR #6372, opened 2026-06-06
feat: apply list formatting > toggle list formatting

elvishcraftsman:
## Description

I added functionality to edit lists (ul, ol, task lists) better in the markdown editor, such as being able to remove list formatting, change between formats, format nested lists, and format lists within blockquotes.

## Changes

"Insert [some type of] list" becomes "toggle [some type of] list." Choosing this option with a list selected will remove the list formatting.
Or, if you choose a different type of list (e.g., the text is formatted with a bulleted list, and you choose a numbered list), it will change it to a numbered list.
Some list and blockquote creation code has been separated so that they don't interfere with each other.
Toggling a list within a blockquote now works as expected.
When inserting lists, indented lines are respected, becoming nested lists
Added keyboard shortcuts for ordered (Ctrl+Shift+O) and unordered (Ctrl+Shift+B) lists. I'm not tied to these options; feel free to change them to something else.
Added toolbar buttons, as well as settings for displaying or not displaying these buttons.
When a selection contains blank and non-blank lines, inserting a list will only apply to the non-blank lines.
Fixed an error that popped up when the selection went to the end of the document (the selection went out of bounds).

No breaking changes as far as I know.

## AI Disclosure Statement

I used Claude Sonnet 4.6 to write the new functionality. I reviewed the code and understand what it does. I've tested it carefully.

Tested on: 

Tested on Fedora 44 x86

### DIFF
--- a/source/app/service-providers/config/get-config-template.ts
+++ b/source/app/service-providers/config/get-config-template.ts
@@ -247,6 +247,8 @@ export interface ConfigOptions {
     showMarkdownLinkButton: boolean
     showMarkdownImageButton: boolean
     showMarkdownMakeTaskListButton: boolean
+    showMarkdownMakeOrderedListButton: boolean
+    showMarkdownMakeBulletListButton: boolean
     showInsertTableButton: boolean
     showInsertFootnoteButton: boolean
     showDocumentInfoText: boolean
@@ -513,6 +515,8 @@ export function getConfigTemplate (): ConfigOptions {
       showMarkdownLinkButton: true,
       showMarkdownImageButton: true,
       showMarkdownMakeTaskListButton: true,
+      showMarkdownMakeOrderedListButton: true,
+      showMarkdownMakeBulletListButton: true,
       showInsertTableButton: true,
       showInsertFootnoteButton: true,
       showDocumentInfoText: true,
--- a/source/common/modules/markdown-editor/commands/markdown.ts
+++ b/source/common/modules/markdown-editor/commands/markdown.ts
@@ -132,6 +132,24 @@ function insertPandocDiv (target: EditorView, attributes: string): void {
 }
 
 /**
+ * Helper function that removes only heading and blockquote markup from a line,
+ * leaving list markers intact.
+ *
+ * @param   {string}  line  The line text
+ *
+ * @return  {string}        The line text without heading/blockquote markup
+ */
+function removeHeadingAndBlockquoteMarkup (line: string): string {
+  let match
+  if ((match = /^#{1,6}\s+/.exec(line)) !== null) {
+    return line.substring(match[0].length)
+  } else if ((match = /^(\s{,3})>\s/.exec(line)) !== null) {
+    return match[1] + line.substring(match[0].length)
+  }
+  return line
+}
+
+/**
  * Helper function that removes block level markup from the provided line
  *
  * @param   {string}  line  The line text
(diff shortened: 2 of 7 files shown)


nathanlesage:
Hi! Thanks for this PR; I've been wanting to add this functionality, but didn't find any time for it, so thank you.

However, that being said, I have a possibly somewhat larger request. Namely, you've (correctly) oriented yourself at the existing commands for implementing this feature. However, I want to slowly rewrite all that logic to use the syntax tree from CodeMirror, rather than regular expressions. The RegExp logic stems from the older CodeMirror 5 setup that simply did not allow for such a fine-grained control over the document that we have now.

Using the syntax tree to find the correct list blocks rather than regular expressions has three major benefits:

1. It means that we can always treat the syntax tree as a source of truth. Whenever we have to change something in how the parser searches for lists (i.e., letter-based lists or `#.`-lists that we might want to add in the future), these commands could break if they are too strict with their regular expression matching, meaning what the commands see is not what the editor highlights.
2. Using the syntax tree makes it much simpler for you to find and replace parts of the syntax with confidence, knowing that you won't accidentally delete any text or other information, or end up with a malformed syntax.
3. It makes the entire code much easier to read and comprehend.

Would you be willing to implement this (somewhat larger) refactor? If you don't, then I could do an in-depth review and we implement this at a later stage, but if we could already start this refactoring work now it might make things easier for all of us.

What do you think? To see how much more transparent this approach could be, take a look at the [`parseNode` function](https://github.com/Zettlr/Zettlr/blob/88b31123cf802bec241d7851991dbbe000456c67/source/common/modules/markdown-utils/markdown-ast/index.ts#L541)

elvishcraftsman:
I don't fully understand the syntax tree and how it would work, but I'm willing to dig into it and see if I can figure it out. It would definitely make the best sense to future-proof the code while changing it. Could this patch be implemented in the interim until I can get to that?

nathanlesage:
> I don't fully understand the syntax tree and how it would work, but I'm willing to dig into it and see if I can figure it out. It would definitely make the best sense to future-proof the code while changing it. Could this patch be implemented in the interim until I can get to that?

Oh yes, sure! I'll do a review of this sometime this week or so. But I have a summer school that runs until next Friday, so I might be a little tight on time and/or forget this. CC @benniekiss if you have some time and want to have the honor before me!

benniekiss:
I'd rather not push a patch just to change it again.

The syntax node approach can probably be modeled after the `lang-markdown` [`deleteMarkupBackward`](https://code.haverbeke.berlin/codemirror/lang-markdown/src/commit/80e2aab74682bbe616501d2c98ce3c665ab07f93/src/commands.ts#L229-L276) command. You will likely have to refactor it to handle each line within a selection range, as the base command only handles markup directly before the cursor (empty selections).

also take a look at the [`insertTabOrSpace`](https://github.com/Zettlr/Zettlr/blob/9e85cb4fd80461158875b89ef7814f17f2c01071/source/common/modules/markdown-editor/commands/markdown.ts#L618-L658) command in the same file as this patch for how to use the syntax tree, and take note of the `nodeInSelection` helper.

The code in the [`renderers`](https://github.com/Zettlr/Zettlr/tree/9e85cb4fd80461158875b89ef7814f17f2c01071/source/common/modules/markdown-editor/renderers) folder is also a good place to look for use of the syntax tree.

PR #6373, opened 2026-06-06
feat: Add Portuguese tutorial files for Zettlr

felipe-f-rocha:
Created a Portuguese version of the Tutorial Files


nathanlesage:
Looks good, thank you! Did you also re-export the PDF file so that it's part of the tutorial files? Also, did you quickly check that Zettlr properly copies over that tutorial folder when starting in a Portuguese environment?

Other than that, it looks good and mergeable. The unit test error looks like a fluke and is unrelated to your PR.

felipe-f-rocha:
Yes. I checked the language resolution logic used by the tutorial loader.

The tutorial directories are discovered dynamically, and `findLangCandidates()` supports close language matches. Since the new tutorial directory is named `pt`, locales such as `pt-BR` and `pt-PT` should resolve to the Portuguese tutorial instead of falling back to English.

I have not yet tested this in a running Portuguese environment, but the code path appears to support it correctly.
And yes, I re-exported the PDF so it is included with the tutorial files.

nathanlesage:
I checked again, and I just missed the PDF file, apologies! I'm gonna trust you – worst case it doesn't find it and serves the English translation, so please make sure to test this out at some point so that people actually receive your translation!

Thank you again, I'm going to merge it now!

Issue #4596, opened 2023-09-02
No option to change text direction (RTL vs LTR) in version 3?


kavir1698:
### Checklist

- [X] Zettlr does not possess this feature (if it does, but the feature doesn't work out for you, propose an enhancement instead.)
- [X] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)
- [X] I have searched the documentation to make sure I did not overlook it.
- [X] I have searched both open and closed issues to see if someone has already proposed this.

### Feature Proposal

There was an option in the editor preferences to change the text direction. I just updated to version three and the feature is not there, nor does it automatically choose the direction. Could you please bring this option back?

### Example Workflow

A toggle to change the text direction.

### Caveats

none

### Do you wish to attempt implementing this yourself?

No

### Platforms

- [X] Windows
- [ ] macOS
- [ ] Linux

nathanlesage:
> nor does it automatically choose the direction

This is — allegedly — working out of the box, but I may have misread some part of CodeMirror's documentation. Feel free to play around with it to see what is needed to make the text direction actually change.

mfakih294:
I confirm this issue. In v3, Arabic text and English text are left aligned. In v2.3.0 Arabic text used to be displayed right to left automatically. A sample Arabic text: 
هذا السطر ينبغي أن يظهر من يمين الشاشة.

nathanlesage:
Okay, I just had another look at what CodeMirror does here and what I need to do. Turns out I just hadn't looked well enough. So here's what I have gathered:

1. CodeMirror relies as much as possible on CSS, and it turns out that setting `dir="rtl"` in the surrounding editor DIV makes the script go right-to-left (albeit it's an all-or-nothing thing).
2. There are a few methods that allow working with bidi, albeit I do not understand exactly where or how these are supposed to work.
3. Auto-detection does not appear to work (i.e. setting `dir="auto"`)
4. Fortunately, there is an example live on the website that wasn't there until a few weeks ago, so it seems we can make something properly work: https://codemirror.net/examples/bidi/
5. A related thread is [here](https://discuss.codemirror.net/t/overriding-bidi-spans-in-rtl-context/6695)

**So regarding a quick relief for RTL-writers here, I have a question**:

Would it be okay if I first provided only a very coarse direction-change setting to switch between RTL and LTR? That may mess up LTR-parts in the code, though, but would give us a first impression as to how that works and should at least improve the situation until we know more.

[...]

PR #6376, opened 2026-06-09
feat: Add text direction toggle to editor configuration

kavir1698:
## Description
This PR implements the text direction (writing direction) preference in the editor settings, replacing the placeholder that was previously in place, fixing #4596 . Users can now choose between Left-to-Right, Right-to-Left, or Auto-detect text direction, which is applied as the HTML `dir` attribute on the editor wrapper element.

## Changes
- Added `textDirection: 'ltr' | 'rtl' | 'auto'` to the `editor` config section in `ConfigOptions` (type + default)
- Added `textDirection` to the `EditorConfiguration` interface and its default config
- Replaced the "Writing direction" placeholder fieldset in the preferences schema with a working select field offering three options: Auto detect, LTR, RTL
- Mapped `editor.textDirection` through `MainEditor.vue` onto the editor wrapper div as a `dir` attribute
- Added validation rule for `editor.textDirection`
- No breaking changes — existing configs will receive the default `auto` via `safeAssign`

## Additional information
The `dir="auto"` value is natively supported by all browsers and automatically detects text direction based on the first strong directional character in the content, making this a zero-config sensible default.

## AI Disclosure Statement
This PR was implemented with assistance from Kilo code.

### DIFF
--- a/source/app/service-providers/config/config-validation.ts
+++ b/source/app/service-providers/config/config-validation.ts
@@ -35,6 +35,7 @@ const RULES = {
   'editor.boldFormatting': 'required|string|in:__,**|default:**',
   'editor.italicFormatting': 'required|string|in:_,*|default:_',
   'editor.readabilityAlgorithm': 'required|string|in:dale-chall,gunning-fog,coleman-liau,automated-readability|default:dale-chall',
+  'editor.textDirection': 'required|string|in:ltr,rtl,auto|default:auto',
   cslLibrary: 'optional|string|default:',
   'display.imageWidth': 'required|number|min:1|max:100|default:100',
   'display.imageHeight': 'required|number|min:1|max:100|default:100',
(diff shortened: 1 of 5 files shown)


nathanlesage:
Is that really all that it takes? 😳 Like, boy, we had so much frustration over the feature in the past that I was too afraid to implement anything that wasn't perfect, and since I do not know any RTL language I couldn't just try it out.

If that's it … looks all good to me 👀

benniekiss:
No, it's not this simple. IIRC, Zettlr has several places where is uses instances of `-left` and `-right` selectors/commands, when it should be using `-start` and `-end` to properly support RTL text. Even though this PR implements visual RTL, I do not believe the editing experience will be satisfactory.

This implementation is also not BIDI aware, which I believe should be the target for the editor, especially for things such as raw markdown formatting symbols.

The codemirror docs discuss this extensively: https://codemirror.net/examples/bidi/

nathanlesage:
Okay, I already thought that can't be it... we really need someone with extensive knowledge in an RTL language helping us get through this ...

PR #6388, opened 2026-06-14
chore: reuse `posInNode` function

benniekiss:
## Description

This PR reuses the `posInNode` function to centralize and deduplicate a lot of logic. It may have marginal performance improvements since calls to `syntaxTree` have been significantly reduced

## Changes

The changes touch URL tooltip rendering, link and citation context menus, and autocorrect replacement logic, if you are looking to test the effects.

## AI Disclosure Statement

None

Tested on: 
macOS 26

### DIFF
--- a/source/common/modules/markdown-editor/commands/autocorrect.ts
+++ b/source/common/modules/markdown-editor/commands/autocorrect.ts
@@ -68,7 +68,7 @@ export const handleReplacement: Command = (target: EditorView): boolean => {
     let pos = range.from - 1
 
     // Ignore those cursors that are inside protected nodes
-    if (posInNode(range.from, tree, PROTECTED_NODES, -1)) {
+    if (posInNode(pos, tree, PROTECTED_NODES, -1)) {
       continue
     }
 
@@ -85,7 +85,7 @@ export const handleReplacement: Command = (target: EditorView): boolean => {
 
     for (const { key, value } of replacements) {
       if (slice.endsWith(key)) {
-        const startOfReplacement = range.from - key.length
+        const startOfReplacement = pos - key.length
         if (posInNode(startOfReplacement, tree, PROTECTED_NODES, -1)) {
           break // `range.from` is not in a protected area, but start is.
         }
--- a/source/common/modules/markdown-editor/commands/autocorrect.ts
+++ b/source/common/modules/markdown-editor/commands/autocorrect.ts
@@ -15,49 +15,24 @@
 
 // The autocorrect plugin is basically just a keymap that listens to spaces and enters
 import { syntaxTree } from '@codemirror/language'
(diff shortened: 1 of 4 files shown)


source/common/modules/markdown-editor/commands/autocorrect.ts
```diff
@@ -109,8 +85,8 @@ export const handleReplacement: Command = (target: EditorView): boolean => {
 
     for (const { key, value } of replacements) {
       if (slice.endsWith(key)) {
-        const startOfReplacement = pos - key.length
-        if (posInProtectedNode(target.state, startOfReplacement)) {
+        const startOfReplacement = range.from - key.length
+        if (posInNode(startOfReplacement, tree, PROTECTED_NODES, -1)) {
```
nathanlesage, line 89: What does the change from `-1` to not reducing it by one do…?

source/common/modules/markdown-editor/util/node-in-selection.ts
```diff
@@ -51,7 +51,7 @@ export function nodeInSelection (
  * @param   {-1|0|1}            side              How nodes shoud be entered.
  *                                                Passed to `SyntaxTree.resolveInner`
  *
- * @return  {boolean}                             `true` if the position touches a node.
+ * @return  {SyntaxNode|null}                     The inner-most node that the position touches.
```
nathanlesage, line 54: Hah, I saw this, too, when double-checking your code.

This function is really lying by now. I wonder if we should rename it, maybe to `nodeAtPos` and call the node names `filter` or something.

But that should be another PR because that diff will probably be a nightmare.

nathanlesage (approved):
Looks good, but I have one question, and a follow-up PR. But I checked the code, should work fine.

source/common/modules/markdown-editor/commands/autocorrect.ts
benniekiss: oh good catch. I will double check. This may be lingering from before my previous autocorrect undo history patch.

source/common/modules/markdown-editor/util/node-in-selection.ts
benniekiss: I totally agree with both your renaming suggestions.

I think saving it for another PR would be fine, and make it easy to see if anything goes awry. I'll have that up once this is through, since intellisense should take care of most of it.

nathanlesage: Wonderful, thank you!

source/common/modules/markdown-editor/commands/autocorrect.ts
benniekiss: it was erroneous! Fixed up in latest commit.

nathanlesage:
Wonderful — and off we go.

PR #6389, opened 2026-06-14
feat: snippet fixes and improvements

benniekiss:
## Description

This PR fixes syntax highlighting for the snippet editor, adds an `abortSnippetRemoveContent` to cancel and remove a pending snippet, and improves the cursor association for snippet placeholders.

## Changes

Syntax highlighting was restored and improved for the snippet editor. Nested snippets should be highlighted correctly.

An `abortSnippetRemoveContent` command was added, bound to `Shift + Esc`, which removes all pending tabstops and associated default content for in-progress snippets.

Cursor association for tabstops with default content was improved. For such tabstops, text inserted *before* the tabstop remains separate and unassociated, i.e., the tabstop will not expand to encompass text inserted before it. Text inserted directly *after* a tabstop with default text will be associated with the tabstop, i.e., the tabstop will grow to encompass the inserted text. Text placed before or after a tabstop without default content(`$1`, `${2}`, etc), remains unassociated to keep the placeholder empty.

## AI Disclosure Statement

None

Tested on: 
macOS 26

### DIFF
--- a/source/common/modules/markdown-utils/snippets-syntax-extension.ts
+++ b/source/common/modules/markdown-utils/snippets-syntax-extension.ts
@@ -73,6 +73,15 @@ const tabstopDecorator = new MatchDecorator({
       return
     }
 
+    // Note: `from` is a document-relative position corresponding to the
+    // beginning of the matched text, while `start` and `end` are relative
+    // to whatever text was provided to the regex engine. In this case,
+    // codemirror matches per-line, so `start` and `end` are line-relative.
+    // i.e., a `start` of 0 corresponds to `view.state.doc.lineAt(from).from`.
+    //
+    // Furthermore, `from` cannot be used to get document-relative  positions
+    // for `start` and `end`, as `from` could refer to a position anywhere in the
+    // line. It would only work as an offset if `from` was at the start of a line.
     const line = view.state.doc.lineAt(from)
     const [ start, end ] = match.indices.groups.num
 
@@ -89,6 +98,15 @@ const variableDecorator = new MatchDecorator({
       return
     }
 
+    // Note: `from` is a document-relative position corresponding to the
+    // beginning of the matched text, while `start` and `end` are relative
+    // to whatever text was provided to the regex engine. In this case,
+    // codemirror matches per-line, so `start` and `end` are line-relative.
+    // i.e., a `start` of 0 corresponds to `view.state.doc.lineAt(from).from`.
+    //
+    // Furthermore, `from` cannot be used to get document-relative  positions
+    // for `start` and `end`, as `from` could refer to a position anywhere in the
+    // line. It would only work as an offset if `from` was at the start of a line.
     const line = view.state.doc.lineAt(from)
     const [ start, end ] = match.indices.groups.var
 
@@ -142,8 +160,14 @@ function getNestedRanges (state: EditorState, pos: number, text: string): Snippe
   return ranges
 }
 
-// Generates a `DecorationSet` for snippet placeholders with default values,
-// handling arbitrarily nested placeholders.
+/**
+ * Generates a `DecorationSet` for snippet placeholders with default values,
+ * handling arbitrarily nested placeholders.
+ *
+ * @param   {EditorView}    view  The editor view
+ *
+ * @returns {DecorationSet}       The generated decorations.
+ */
 function renderPlaceholders (view: EditorView): DecorationSet {
   const decos: Range<Decoration>[] = []
 
@@ -151,7 +175,7 @@ function renderPlaceholders (view: EditorView): DecorationSet {
     const text = view.state.sliceDoc(range.from, range.to)
     const nested = getNestedRanges(view.state, range.from, text)
 
-    for (const { id, from: start, to: end } of nested) {
+    for (const { id, from, to } of nested) {
       let idDeco = tabstopDeco
       let defaultDeco = placeholderDeco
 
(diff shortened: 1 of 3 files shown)


source/common/modules/markdown-utils/snippets-syntax-extension.ts
```diff
@@ -48,38 +65,122 @@ const SUPPORTED_VARIABLES = [
...
+  decorate: (add, from, _, match, view) => {
+    if (match.indices?.groups?.num === undefined) {
+      return
+    }
+
+    const line = view.state.doc.lineAt(from)
+    const [ start, end ] = match.indices.groups.num
+
+    add(from, from + 1, snippetMarkDeco)
+    add(line.from + start, line.from + end, tabstopDeco)
```
nathanlesage, line 89: This seems difficult to comprehend. Isn't `from, from + 1` equivalent to `start - 1, start`? If so, that would remove a dependency and make the code a bit clearer to understand.

source/common/modules/markdown-utils/snippets-syntax-extension.ts
```diff
@@ -48,38 +65,122 @@ const SUPPORTED_VARIABLES = [
...
+  decorate: (add, from, _, match, view) => {
+    if (match.indices?.groups?.var === undefined) {
+      return
+    }
+
+    const line = view.state.doc.lineAt(from)
+    const [ start, end ] = match.indices.groups.var
+
+    add(from, from + 1, snippetMarkDeco)
+    add(line.from + start, line.from + end, SUPPORTED_VARIABLES.includes(match.groups!.var) ? varDeco : invalidVarDeco)
```
nathanlesage, line 114: Same as above

source/common/modules/markdown-utils/snippets-syntax-extension.ts
```diff
@@ -48,38 +65,122 @@ const SUPPORTED_VARIABLES = [
...
+// Generates a `DecorationSet` for snippet placeholders with default values,
+// handling arbitrarily nested placeholders.
+function renderPlaceholders (view: EditorView): DecorationSet {
+  const decos: Range<Decoration>[] = []
+
+  for (const range of view.visibleRanges) {
+    const text = view.state.sliceDoc(range.from, range.to)
+    const nested = getNestedRanges(view.state, range.from, text)
+
+    for (const { id, from: start, to: end } of nested) {
```
nathanlesage: Would it be possible to simplify and not rename the properties here?

source/common/modules/markdown-utils/snippets-syntax-extension.ts
```diff
@@ -48,38 +65,122 @@ const SUPPORTED_VARIABLES = [
...
+
+  // Recurse on the interior text of the match. This is everything after the
+  // colon and before the final brace.
+  ranges.push(...getNestedRanges(state, pos + end, text.slice(end)))
+
+  return ranges
+}
+
+// Generates a `DecorationSet` for snippet placeholders with default values,
+// handling arbitrarily nested placeholders.
```
nathanlesage: Please use JS Docstrings for function documentation

nathanlesage (changes requested):
In principle looks good, just a few suggestions to make the code easier to understand.

source/common/modules/markdown-utils/snippets-syntax-extension.ts
benniekiss: I'll double check, but I dont think that'll work. IIRC, `start` is a line relative position, so to get a document position for a `start` of `0` would be  `line.pos + 0`. `from` is a document position, so it needs no adjustment.

Ill add a comment explaining that, though. I believe this behavior is mentioned in the nested snippet placeholder code, too

source/common/modules/markdown-utils/snippets-syntax-extension.ts
benniekiss: ~~personally, there are already a lot of `.from`/`.to` vars in this block, and using `start` and `end` was easier for me to parse, and it matches the terminology used in the other functions~~ I immediately disagreed with myself 🙃

I think removing the rename is fine!

source/common/modules/markdown-utils/snippets-syntax-extension.ts
benniekiss: I did recall correctly! I added a comment explaining the behavior in the latest commit

nathanlesage: Okay, so the snippetMark and the tabStop are not guaranteed to be adjancent here, correct?

benniekiss: no, they always should be, since the regex is `\$(?<num>\d+)`, so the `from` referes to the document-relative position of `\$`, and `start` and `end` are line-relative positions of the start and end positions of the capture group `(?<num>\d+)`.

nathanlesage: Okay, but so we could indeed use the same reference points for both markers! I'd rather have that to actually communicate that both elements will always be next to each other. Right now I know that we'll gonna have some issues figuring this out at a future point — just trying to think efficiently here.

benniekiss: I really don't think it makes more sense to derive the starting position from the position of the capture group, as they are two distinct tokens, and think it is more confusing to do so. 

The regex already communicates that the elements are next to eachother. Why duplicate that in an area where it doesn't need to be? Also, I think it is less efficient to derive positions when we are already provided the exact information needed.

All of these are more confusing to me, and still require knowing the distinction between line-position and document-position of `from` and `start`:

```ts
    const line = view.state.doc.lineAt(from)
    const [ start, end ] = match.indices.groups.num
    
    add(line.from + start - 1, line.from + start, snippetMarkDeco)
    add(line.from + start, line.from + end, tabstopDeco)
```

```ts
    const line = view.state.doc.lineAt(from)
    const [ _, end ] = match.indices.groups.num
    
    add(from, from + 1, snippetMarkDeco)
    add(from + 1, line.from + end, tabstopDeco)
```

I think adding a comment above each `.add()` with the tokens it is matching would be helpful, as I did this in other matchers below.

nathanlesage: Mh, yeah, I see your point. I just always try to be extra verbose. I mean there are now comments, so let's just keep it at that. But thank you for engaging with me here. Going to merge!

nathanlesage:
Sorry for being a bit nitpicky, but I think we can merge now.

PR #6393, opened 2026-06-17
feat: Reopen last closed tab(s)

### LINKED ISSUE #5741: Add shortcut `CTRL` + `SHIFT` + `T` to reopen last closed tabs
### Checklist

- [ ] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)
- [ ] I have searched the documentation to make sure I did not overlook it.
- [ ] I have searched both open and closed issues to see if someone has already proposed this.

### Feature Proposal

Common web browsers all share that same shortcut that is `CTRL`+`SHIFT`+`T` that reopens the latest closed tabs in order of closing time.

### Example Workflow

You have several tabs, you close one by mistake, `CTRL` +  `SHIFT` + `T` reopens that tab.

### Caveats

Browsers use that shortcut because they use `CTRL` + `T` to open a new tab. Which is not the case in Zettlr. 
But I think it makes sens because lot of people that are shortcut users have this one instinctively engraved in the fingers.

For example VSCodium the code editor does not have `CTRL` + `T` shortcut to open a new tab but uses `CTRL` + `SHIFT` + `T` to reopen latest closed tab too.

I don't use macOS but I guess `Cmd` + `SHIFT` + `T` does the same

No

### Platforms

- [x] Windows
- [x] macOS
- [x] Linux

### Additional Information

I'd love to try my hand at implementing it but I'm not sure I have the skills. Curious to read how it was done when it is though.

richdouglasevans:
Closes #5741

## Description
This adds a feature where the user can reopen the most recently closed tab (document), if any.

If the user has closed at least one tab (document) in this Zettlr session then invoking the _Reopen Tab_ action will reopen that document.

The most recently _closed_ tabs (documents) are maintained in a stack data structure so the user can close multiple documents and then repeatedly invoke the _Reopen Tab_ action to reopen each closed document.

If they have not closed any documents then the _Reopen Tab_ action silently does nothing. The UI elements exposing the _Reopen Tab_ action do not (currently) disable themselves if there are no recently closed documents.

The memory of "closed documents" is _per Zettlr session_; if a user closes a document and then closes all Zettlr windows/instances -- thus closing the session -- then when Zettlr is restarted there is not considered to be anything to reopen.

## Action Shots

Here's an action shot of the feature being used to reopen a recently-closed document from the application menu.

https://github.com/user-attachments/assets/e13c64fa-55da-43bf-868c-56e7e26831ed

Here's an action shot of the feature being used to reopen two recently-closed documents in succession from the application menu.

https://github.com/user-attachments/assets/6bfbb3c9-38e9-48a9-9eef-b56b7306d3e3

Here's an action shot of the feature being used to reopen a recently-closed document from the little menu activated by right-clicking on the tab bar.

https://github.com/user-attachments/assets/32d0ce70-eb16-43b6-9ca0-6b384da45b5b

## Changes
I'm aware that Hendrik is not personally keen on PRs that are self-annotated by the author; I have annotated this PR because the user who raised the ticket expressed an interest in grokking the implementation so a little self-annotation to guide them through the implementation feels okay in this instance.

I trust that Hendrik will forgive me this one time 🙇‍♂️ 

## Additional information

There is no keyboard shortcut / accelerator to invoke the _Reopen Tab_ action; this is a deliberate omission.

The most common accelerator for this action across apps is <kbd>CTRL+SHIFT+T</kbd>. This accelerator is already used by Zettlr's _Filter files_ action.

On the ticket Hendrik wrote this:

> Feel free to open a PR but as for now not yet bound to a shortcut, then we at least have that functionality and can do some reshuffling of the shortcuts in line with a custom shortcut implementation.

## AI Disclosure Statement

No AI was used.

Tested on: Windows 11 (x86)

### DIFF
--- a/source/app/app-service-container.ts
+++ b/source/app/app-service-container.ts
@@ -119,9 +119,8 @@ export class AppServiceContainer {
     this._targetProvider = new TargetProvider(this._logProvider, this._fsal)
     this._linkProvider = new LinkProvider(this._logProvider, this._configProvider, this._fsal)
     this._searchProvider = new SearchProvider(this._logProvider, this._fsal, this._configProvider)
-    
-    // The document provider accesses only the FSAL in its constructor
-    this._documentManager = new DocumentManager(this)
+
+    this._documentManager = new DocumentManager(this, this._recentDocsProvider)
     this._tagProvider = new TagProvider(this._logProvider, this._documentManager, this._configProvider, this._fsal)
     this._windowProvider = new WindowProvider(this._logProvider, this._configProvider, this._documentManager)
 
@@ -192,7 +191,7 @@ export class AppServiceContainer {
     await this._informativeBoot(this._tagProvider, 'TagProvider')
     await this._informativeBoot(this._trayProvider, 'TrayProvider')
     await this._informativeBoot(this._citeprocProvider, 'CiteprocProvider')
-    
+
     await this._informativeBoot(this._documentManager, 'DocumentManager')
     await this._informativeBoot(this._menuProvider, 'MenuProvider')
     await this._informativeBoot(this._updateProvider, 'UpdateProvider')
--- a/source/app/service-providers/commands/reopen-recently-closed-file.ts
+++ b/source/app/service-providers/commands/reopen-recently-closed-file.ts
@@ -1,0 +1,28 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        ReopenRecentlyClosedFile command
+ * CVM-Role:        <none>
+ * Maintainer:      Rich Douglas
+ * License:         GNU GPL v3
+ *
+ * Description:     Reopens the most recently closed file, if any.
+ *
+ * END HEADER
+ */
+
+import type { AppServiceContainer } from 'source/app/app-service-container'
+import ZettlrCommand from './zettlr-command'
+
+export default class ReopenRecentlyClosedFile extends ZettlrCommand {
+  constructor (app: AppServiceContainer) {
+    super(app, 'reopen-recently-closed-file')
+  }
+
+  run (_evt: string, _arg: string): Promise<boolean> {
+    const windowId = this._app.documents.windowKeys()[0]
+    const leafId = this._app.documents.leafIds(windowId)[0]
+    return this._app.documents.reopenRecentlyClosedFile(windowId, leafId)
+  }
+}
--- a/source/app/service-providers/documents/index.ts
+++ b/source/app/service-providers/documents/index.ts
@@ -37,6 +37,7 @@ import { trans } from '@common/i18n-main'
 import type FSALWatchdog from '@providers/fsal/fsal-watchdog'
 import { getDocumentTypeForExtension, hasImageExt, hasMdOrCodeExt, hasPDFExt } from 'source/common/util/file-extention-checks'
 import isDir from 'source/common/util/is-dir'
+import type RecentDocumentsProvider from '../recent-docs'
 
 type DocumentWindows = Record<string, DocumentTree>
 type DocumentWindowsJSON = Record<string, BranchNodeJSON|LeafNodeJSON>
@@ -141,6 +142,7 @@ export type DocumentManagerIPCAPI = IPCAPI<{
   'open-file': LeafLoc & { path: string, newTab: boolean }
   'close-file': LeafLoc & { path: string }
   'close-file-everywhere': { path: string }
+  'reopen-recently-closed-file': LeafLoc
   'get-open-workspace-files': { path: string }
   'sort-open-files': LeafLoc & { newOrder: string[] }
   'get-file-modification-status': unknown
@@ -225,7 +227,10 @@ export default class DocumentManager extends ProviderContract {
     leafId: string|undefined
   }
 
-  constructor (private readonly _app: AppServiceContainer) {
+  constructor (
+    private readonly _app: AppServiceContainer,
+    private readonly _recentDocs: RecentDocumentsProvider
+  ) {
     super()
 
     const containerPath = path.join(app.getPath('userData'), 'documents.yaml')
@@ -307,6 +312,10 @@ export default class DocumentManager extends ProviderContract {
         case 'close-file-everywhere': {
           const { path } = payload
           return this.closeFileEverywhere(path)
+        }
+        case 'reopen-recently-closed-file': {
+          const { windowId, leafId } = payload
+          return await this.reopenRecentlyClosedFile(windowId, leafId)
         }
         case 'get-open-workspace-files':
           const { path } = payload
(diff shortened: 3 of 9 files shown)


source/app/service-providers/commands/reopen-recently-closed-file.ts
```diff
@@ -0,0 +1,28 @@
...
+export default class ReopenRecentlyClosedFile extends ZettlrCommand {
+  constructor (app: AppServiceContainer) {
+    super(app, 'reopen-recently-closed-file')
+  }
+
+  run (_evt: string, _arg: string): Promise<boolean> {
+    const windowId = this._app.documents.windowKeys()[0]
+    const leafId = this._app.documents.leafIds(windowId)[0]
+    return this._app.documents.reopenRecentlyClosedFile(windowId, leafId)
+  }
```
richdouglasevans, line 27: We've added the functionality to track recently-closed files and to reopen them... but we haven't got a way for the user to access this functionality via the UI.

This Zettlr command is what _glues_ UI affordances -- such as menu items -- to the functionality we just added to the `DocumentManager` class. On line `26` we invoke our new `reopenRecentlyClosedFile` method.

The `windowId` and `leafId` arguments are derived from the same `DocumentManager`: the implementation is a bit "dumb" here in that it just grabs the first window (and leaf). This is _probably_ okay since most of the time I expect users will have just one Zettlr window open at a time.

However, if the user does have multiple Zettlr windows open then this introduces some potential surprise:

- the user may close a tab in window A
- then switch to window B
- then realise they want to reopen that tab so they click _Reopen Tab_ while in window B
- the tab reopens -- yay -- but it reopens in window A 'cos that's the first window and the code explicitly grabs the first window (`documents.windowKeys()[0]`)

I thought about enriching the stack of recently-closed documents to also track which window they were closed in... but you know, I thought I'll just raise this in the PR and let Hendrik (or others) guide me. It feels like a legit but niche case so I don't wanna go overboard: if Hendrik says that we should reopen tabs in the window they were closed in then I'll do that. (But then again, what if that window is no more?)

source/app/service-providers/menu/menu.darwin.ts
```diff
@@ -622,6 +622,14 @@ export default function getMenu (
...
           }
         },
+        {
+          id: 'menu.tab_reopen',
+          label: trans('Reopen Tab'),
+          click: function () {
+            commands.run('reopen-recently-closed-file', {})
+              .catch(e => logger.error(String(e.message), e))
+          }
+        },
```
richdouglasevans, line 632: Here's the first place where we wire up the UI to invoke our _Reopen Tab_ functionality.

All we do is run the command we defined earlier. On line `20` of that command we assigned an ID unique to that command `reopen-recently-closed-file`... and we supply that same ID here so that Zettlr knows which command to run.

source/app/app-service-container.ts
```diff
@@ -119,9 +119,8 @@ export class AppServiceContainer {
     this._targetProvider = new TargetProvider(this._logProvider, this._fsal)
     this._linkProvider = new LinkProvider(this._logProvider, this._configProvider, this._fsal)
     this._searchProvider = new SearchProvider(this._logProvider, this._fsal, this._configProvider)
-    
-    // The document provider accesses only the FSAL in its constructor
-    this._documentManager = new DocumentManager(this)
+
+    this._documentManager = new DocumentManager(this, this._recentDocsProvider)
```
nathanlesage, line 123: Unnecessary, since you can already get the recent docs from the service container. That's why the document manager gets a reference to the app service container, because it needs a lot of those other providers, too.

source/app/service-providers/documents/index.ts
```diff
@@ -225,7 +227,10 @@ export default class DocumentManager extends ProviderContract {
     leafId: string|undefined
   }
 
-  constructor (private readonly _app: AppServiceContainer) {
+  constructor (
+    private readonly _app: AppServiceContainer,
+    private readonly _recentDocs: RecentDocumentsProvider
+  ) {
```
nathanlesage, line 233: See other comment -- unnecessary

source/app/service-providers/menu/menu.darwin.ts
nathanlesage: I do not understand why you use the `commands` var here. If you look above, you can see that, typically, we'll send a "shortcut" command to the renderer, which then sends a message to main. Why this odd double-message? Because this way we are able to intercept and request more information from the user if necessary.

This then solves also your problem above: Because you can actually retrieve the correct window based on which renderer sent the request.

source/app/service-providers/recent-docs/index.ts
```diff
@@ -94,6 +94,41 @@ export default class RecentDocumentsProvider extends ProviderContract {
...
+   *
+   * @returns the filepath to the most recently-closed document or `undefined`
+   */
+  getMostRecentlyClosedDocument (currentlyOpenDocs: string[]): string | undefined {
+    for (const recentDoc of this._recentDocs) {
+      const isCurrentlyOpen = currentlyOpenDocs.indexOf(recentDoc) >= 0
+      if (!isCurrentlyOpen) {
+        return recentDoc
+      }
+    }
```
nathanlesage, line 127: My strong suggestion here is to use sets, and create the difference between recent docs and open docs, and return the newest one.

Also, this here might only work on Linux. Iirc `_recentDocs` only stores recent docs for Linux, for the other platforms, we'll hook into an Electron API.

Finally, the comment of this function is oddly long, so please see whether you might be able to cut it a bit.

source/app/service-providers/menu/menu.darwin.ts
nathanlesage: I have just seen that you did indeed implement such a function, well below in the context menu of the document tabs. You can just reuse that IPC hook in the documents manager.

source/app/service-providers/documents/index.ts
```diff
@@ -1023,6 +1032,22 @@ current contents from the editor somewhere else, and restart the application.`
...
 
+  /**
+   * Reopens the most recently closed file, if possible.
+   *
+   * If there are no such closed files then this is a no-op.
+   *
+   * @returns {boolean}            `true` iff the most recently closed file was reopened.
+   */
+  public async reopenRecentlyClosedFile (windowId: string, leafId: string): Promise<boolean> {
+    const filePath = this._recentDocs.getMostRecentlyClosedDocument(this.documents.map((doc) => doc.filePath))
```
nathanlesage, line 1043: I wonder if you could rather implement the check here, and make the recent docs provider just return a list of recent documents?

nathanlesage: Yeah please do that, the recent docs provider has nothing to do with a list of open files

source/app/service-providers/documents/index.ts
```diff
@@ -308,6 +313,10 @@ export default class DocumentManager extends ProviderContract {
           const { path } = payload
           return this.closeFileEverywhere(path)
         }
+        case 'reopen-recently-closed-file': {
+          const { windowId, leafId } = payload
+          return await this.reopenRecentlyClosedFile(windowId, leafId)
+        }
```
nathanlesage, line 319: See how it just becomes possible to quickly access windowId/leafId here?

source/app/service-providers/commands/reopen-recently-closed-file.ts
nathanlesage: This entire command is unnecessary, so please remove this file. See all my other comments for refactor suggestions.

nathanlesage (changes requested):
I see where this is going and it has potential. However, I'd like to request a bit of refactoring. I commented most of the changes directly where they're necessary.

PR #6399, opened 2026-06-19
chore: rename posInNode to align with functionality

benniekiss:
## Description

This docstring and function have drifted quite a bit from when I originally wrote this. This just syncs them up again.

As discussed: https://github.com/Zettlr/Zettlr/pull/6388#discussion_r3443778635

## AI Disclosure Statement

Tested on:


nathanlesage (approved):
HAH, that was what I was expecting: now it is clear that often we use this function to check "Is there a node at that position which is in our small filter list?" And then we do something with that information. We do quite a lot of implicit null checks (if(node)), so now one final thing: let's make those null checks explicit.

That was partly what was bugging me, that I knew some of those Boolean checks were actually not-null checks. Now it's much more transparent what happens when we run this function to check whether the cursor is at a specific node. Would you mind adding those null checks to the if statements/booleans? Then I think we're golden.

Thank you!

benniekiss:
done!

nathanlesage:
Wonderful, thank you so much!

PR #6401, opened 2026-06-20
fix(search): Prevent log for unsupported files

kyaso:
## Description
While testing the new search provider, I have noticed that verbose logs are also printed for images (e.g. png). After looking at the code, it turns out that those files are _not_ actually searched, but only a log message is printed (for any file). Unfortunately, this pollutes the logs, if there are many images.

## Changes
Move the log call to the `searchFileBoolean` method of the `SearchProvider` class, so that it only gets called if the file to be searched is of a valid descriptor type.

## Additional information
If the current behavior, that is, to log _any_ file, is indeed intended, then please forget my PR :)

## AI Disclosure Statement
Only used for code completion.

### DIFF
--- a/source/app/service-providers/search/index.ts
+++ b/source/app/service-providers/search/index.ts
@@ -137,8 +137,6 @@ export class SearchProvider implements ProviderContract {
       return
     }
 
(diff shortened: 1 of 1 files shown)


nathanlesage:
Oh, makes sense, thanks!

Issue #6022, opened 2025-11-19
Keyboard Shortcut for Preview/Raw mode


rayachristian:
### Description

I suggest adding a keyboard shortcut (e.g., Ctrl+Alt+S) to switch between raw mode and preview mode.

### Proposed Changes

Just add a shortcut on the toggle button Raw/Preview.
I suggest Ctrl+Alt+S because it's already the shortcut of MarkText.

### Caveats

I think there's no problem adding this feature.

### Do you Wish to Attempt Implementing this Yourself?

Yes

### Zettlr Version

None

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

gertsylvest:
I've implemented this — adding a keyboard shortcut `Cmd+Alt+M` (macOS) / `Ctrl+Alt+M` (Win/Linux) to toggle rendering mode via the View menu. PR incoming.

Issue #5912, opened 2025-09-21
Add keyboard shortcut to toggle a task on and off.


go-ubx:
### Checklist

- [x] Zettlr does not possess this feature (if it does, but the feature doesn't work out for you, propose an enhancement instead.)
- [ ] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)
- [x] I have searched the documentation to make sure I did not overlook it.
- [x] I have searched both open and closed issues to see if someone has already proposed this.

### Feature Proposal

I recently switched to Zettlr because of it's simplicity but a major turn off is that __it doesn't have a dedicated keyboard shortcut to toggle a task/checkbox__.
Having to use the mouse after typing a task is not that fast. And not everyone wants to use the mouse while writing their notes.
Adding a shortcut key for toggling a task will be very helpful.

The combinations can be:

> Ctrl+D
> Ctrl+Shift+T
> F12

Or whatever the devs like.

### Example Workflow

- For those who don't want to use the mouse much while writing, this will be a great feature to manage their to-do list.
- It will be just faster to not have to use the mouse.
- More focused workflow.

### Caveats

I don't think there are any Caveats of this.

### Do you wish to attempt implementing this yourself?

No

### Platforms

- [ ] Windows
- [ ] macOS
- [x] Linux

PR #6411, opened 2026-06-25
feat: Implement Custom Editor Shortcuts

nathanlesage:
## Description

This PR adds the ability to customize the editor shortcuts to the app. It makes many commands accessible via keyboard shortcuts and allows the user to set and/or override them in the preferences.

## Changes

* The main keymap is now simplified. It only contains `keys` and does no longer differentiate between different platforms.
* The main keymap now listens to configuration updates to programmatically pull in changed keyboard shortcuts
* Simplified the table editor keymap somewhat to be a bit more efficient
* Implemented a new Shortcut Recording Control and a Shortcut Display component to facilitate displaying and changing shortcuts
* Implemented customizable shortcuts in the preferences window
* The shortcut for the remove line breaks transform that had been added in an ad-hoc fashion is gone again. Users can now set it if needed.

## Additional information

For now, this only allows customizing editor shortcuts. More shortcuts will come in the future. Also, it is important to recognize that I have now just committed to a few decisions, since input is always extremely hard. There may be certain combinations of wanted keyboard shortcuts and layouts that won't work as expected.

## Open TODOs

- [x] Resetting shortcuts
- [x] Implement CodeMirror->Electron shortcut transform
- [x] Add remaining shortcuts to the preferences
- [x] Validate shortcuts and determine if a few more are needed
- [x] Cleanup and style fixes

## AI Disclosure Statement

No AI has been used in drafting this PR whatsoever.

Tested on: macOS 26.5.1


nathanlesage:
@kyaso @benniekiss If you're bored, this was a lot of chores but only few technical changes.

nathanlesage:
I've now sherlocked an older PR and implemented menu shortcut customization. Also, I've cleaned everything up a bit.

benniekiss:
Going through issues an noticed this request: https://github.com/Zettlr/Zettlr/issues/6022

Another shortcut request: #5912

Another issue that will be closed by this: #6411

source/common/modules/markdown-editor/editor-extension-sets.ts
```diff
@@ -170,7 +170,8 @@ function getCoreExtensions (options: CoreExtensionOptions): Extension[] {
     // keymap.
     inputModeCompartment.of(inputMode),
     // Then, include the default keymap
-    defaultKeymap(),
+    zettlrKeymap(options.initialConfig.shortcuts),
+    darkMode({ darkMode: options.initialConfig.darkMode, ...themes[options.initialConfig.theme] }),
```
benniekiss: this should be removed (duplicate line)

source/common/vue/form/elements/ShortcutCaptureControl.vue
```diff
@@ -46,34 +143,99 @@ function handleKeydown (event: KeyboardEvent): void {
...
+    padding: 2px;
+    display: flex;
+    flex-wrap: wrap;
+    gap: 6px;
+    align-items: center;
+    cursor: text;
+  
+    &.is-set {
+      color: rgb(30, 30, 30);
+    }
```
benniekiss: The styling here could definitely be improved.

In darkmode, it is almost impossible to see the captured input keys, since they are black, and in light mode, the default key values are too pale:

Image
Image

benniekiss:
The text transform shortcuts do not seem to actually apply the shortcut from my testing. I haven't pinpointed why yet.

Also, I think having a conflict indicator would be very helpful, both with each custom shortcut, and those defined by the app and codemirror. 

Currently, it seems that the custom shortcuts have a lower precedence than default/non-configurable shortcuts, but this is kind of unexpected, and I think it could lead users to performing unintended actions. The custom shortcuts should probably be set to the highest precedence.

benniekiss:
> The text transform shortcuts do not seem to actually apply the shortcut from my testing. I haven't pinpointed why yet.

I think this was actually because of a conflict with another shortcut, not with the code itself. So my point regarding a conflict indicator is really the only thing I think this PR is missing

nathanlesage:
> he text transform shortcuts do not seem to actually apply the shortcut from my testing. I haven't pinpointed why yet.

Reminds me of what @wideaperture has reported, too… I'll investigate during the week when I have some time.

> Also, I think having a conflict indicator would be very helpful, both with each custom shortcut, and those defined by the app and codemirror.

I agree, BUT what has stopped me from implementing that is that I didn't have a way to figure out potential conflicts. One way would be to move literally every shortcut into the shortcut map, and just not expose most of them via the config; then I could implement a duplicate checker — at least for the editor shortcuts, which would be something. But I think, until then people hopefully can recognize potential issues…?

> Currently, it seems that the custom shortcuts have a lower precedence than default/non-configurable shortcuts, but this is kind of unexpected, and I think it could lead users to performing unintended actions. The custom shortcuts should probably be set to the highest precedence.

Good point. I didn't want to touch the ordering right now because there is a lot of overloading going on, and some commands need to be tried before others. But I can take a look how to optimize the overloads in such a way that anything customizable can run first.

benniekiss:
> One way would be to move literally every shortcut into the shortcut map, and just not expose most of them via the config; then I could implement a duplicate checker

Is the `keymap` facet iterable/extractable in UI context https://codemirror.net/docs/ref/#view.keymap? Then instead of moving all shortcuts to be explicitly defined, it can just search in the current keymap set with a simple `filter()`

>  But I think, until then people hopefully can recognize potential issues…?

It was not initially evident to me why shortcuts were not working, and it ends up being a rather annoying guess-and-check to find one that is open. I think since there are already several hard-coded shortcuts, showing a conflict is rather necessary

nathanlesage:
> Is the keymap facet iterable/extractable in UI context https://codemirror.net/docs/ref/#view.keymap? Then instead of moving all shortcuts to be explicitly defined, it can just search in the current keymap set with a simple filter()

Maybe, but that would not solve the conflict issue with the broader UI. Menu shortcuts will override editor shortcuts, etc ... so having a custom, central location would be better than extracting just from the editor context.

nathanlesage:
Sooo, a bunch of improvements:

1. I have now modified the type in the configuration so that we only have to specify/store those shortcuts we actually want to, while we can add whichever shortcuts we want to the default keymap.
2. There is now conflict detection built in (see screenshot) that warns the user if one of the shortcuts in the editor defaults list conflicts with another.
3. I have fully re-ordered the key bindings to ensure that the overloads work in the correct way. I did even find a few duplications and other oddities that I've now fixed in this case.

Unrelated to that: I have also decided to move all the logic to the actual curlQuotes command, because having half of a command out in the open bothered me. This now broke the unit tests, but the fix should be easy. I'll do so asap.

Please, feel free to take another look!

***

image

nathanlesage:
So, unit tests are fixed as well. Please take a look!

nathanlesage:
I've now done a bit of additional refactoring, moved the conflict checker into the preferences pane (where it belongs); made sure that conflicts are now detected across both maps; and added a new one (filter files), partially to draft a small instruction for how to do that.

@benniekiss Wanna give it a new spin? I'd like to include this in the next release because I know people will come screaming at us for which shortcuts to customize only after they have the feature :D

source/common/vue/form/elements/ShortcutCaptureControl.vue
benniekiss: I swear, I do not know how I continue to struggle using this review interface.

2 weeks later...

benniekiss:
I am looking this over again now, but I like the new conflict picker! Superficially looks good to me. I'll do some testing today and report back. 

The only thing I have ATM is the styling noted in the above comment.

source/common/vue/form/elements/ShortcutCaptureControl.vue
nathanlesage: Haha no worries. Regarding your changes: I'll see to fixing this CSS error.

Regarding your suggestion:

I wanted to mimic the "placeholder" design for default keys, that's why I gave them the placeholder gray color. This way, you can better distinguish them from your custom shortcuts. The thought for this was that default shortcuts can all be found in the documentation, so there's no need to make them scream. Thoughts?

benniekiss: I think the different color for defaults is perfect! I just think the contrast could be increased a step or two

nathanlesage:
Now I think I got it. Decided to make it a bit more IBM Model M like, with a dose of beige:

image

And in dark mode:

image

Did you find anything else, or is this now ready to merge?

benniekiss:
I think this lgtm!

nathanlesage:
Perfect, thank youu (I have muted the colors a bit after thinking about it, but other than that everything's in place.)

PR #6417, opened 2026-06-27
feat: add text transforms and additional shortcuts to link/image shortcut menu

benniekiss:
## Description

This PR adds several additional context menu items to for links and images.

## Changes

Context menu items were added for bold,italic, cut, copy, and paste, as well as the text transforms context menu.

Link and image title text can contain markdown formatting, so having these items available for links and images makes sense to me. I also use link formatting for some internal things, so I'm pasting regular text within brackets, and I really enjoy having these transforms available.

## AI Disclosure Statement

N/A

Tested on:

### DIFF
--- a/source/common/modules/markdown-editor/context-menu/default-menu.ts
+++ b/source/common/modules/markdown-editor/context-menu/default-menu.ts
@@ -18,10 +18,9 @@ import { trans } from '@common/i18n-renderer'
 import showPopupMenu, { type AnyMenuItem } from '@common/modules/window-register/application-menu-helper'
 import { type SyntaxNode } from '@lezer/common'
 import { forEachDiagnostic, type Diagnostic, forceLinting, setDiagnostics } from '@codemirror/lint'
-import { applyBold, applyItalic, insertLink, applyBlockquote, applyOrderedList, applyBulletList, applyTaskList } from '../commands/markdown'
-import { cut, copyAsPlain, copyAsHTML, paste, pasteAsPlain } from '../util/copy-paste-cut'
-import { getTransformSubmenu } from './transform-items'
 import { extractLTSpellcheckSuggestionsFrom, isLanguageToolMisspelling } from '../linters/language-tool'
+import { getFormatMenuItems } from './format-items'
+import { getInsertSubmenu } from './insert-items'
 
 const ipcRenderer = window.ipc
 const suggestionCache = new Map<string, string[]>()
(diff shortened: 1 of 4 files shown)


nathanlesage:
I like it! I just wonder if this could be a chance to just make the context menu modular and define a "formatting" template similar to the transformations, and then just compose context menus accordingly. This seems a little bit repetitive and I'm afraid that we might forget one when we have to change something.

Also, I see that you return the popup menu for the link image menu when there is no url found -- how about then just falling back to showing the regular default menu in that case ...?

benniekiss:
> I just wonder if this could be a chance to just make the context menu modular and define a "formatting" template similar to the transformations,

That makes perfect sense to me and I'm not sure why I didn't do that initially, but it seems simple enough...

> how about then just falling back to showing the regular default menu in that case

I did consider this, but the default menu include several `insert {formatting}` commands that I wasn't sure were appropriate for the context. Happy to add them though, just thought it was worth considering.

benniekiss:
Latest commit refactors the format menu items into its own file, and it moves the insert element items into a separate file, as well.

I changed the insert element commands to be in a submenu, since there were quite a few of them. I think the new context menu is much more readable, now. With this refactor, I think it makes sense to add the insert menus into the image/link context menu, since they are fairly out of the way now.

Screenshot 2026-06-27 at 17 02 56
Screenshot 2026-06-27 at 17 03 00

nathanlesage:
> I did consider this, but the default menu include several insert {formatting} commands that I wasn't sure were appropriate for the context.

Mh, because some are block based, right? We might want to do some additional work in the future to better take context into account when displaying a default context menu, but I think that's out of scope for this PR. Also I think that if statement will never execute and is basically only to satisfy typescript.

In general, I like the restructuring! Can you show an example of how it looks like in an image/link context?

benniekiss:
> Also I think that if statement will never execute and is basically only to satisfy typescript.

Which statement is this?

> Can you show an example of how it looks like in an image/link context?

Screenshot 2026-06-28 at 08 09 46

nathanlesage:
> Which statement is this?

The one in your screenshot -- because url shouldn't be null ever once we reach that point, unless Marijn seriously messes up the parser, which I doubt is anywhere likely!

benniekiss:
ah, the url can be null for bare links like `[no url]`

benniekiss:
@nathanlesage would you take a look at this when you get a chance? I think its ready to go!

nathanlesage:
Sure! Next week is quite busy for me, but maybe I'll be able to give it a Review in between!

PR #6418, opened 2026-06-27
chore(css): misc toolbar styling improvements

benniekiss:
## Description

This PR cleans up the log window toolbar on macos and increases the draggable space between the insert footnote and word count widgets.

## Changes

The log window toolbar on macos no longer permanently shows the scroll overflow buttons, and the toolbar was resized to fill the space. A small tolerance had to be added to `checkOverflow`, since the `clientWidth` would occassionaly be reported as fractions of a pixel (which is what was causing the buttons to show in the log viewer).

The `TextControl` component was changed to use `width: fit-content` instead of the fixed-width of `width: 150px`. This increases the draggable area on the toolbar, as outlined in the pic below. An small spacer was added between the pomodoro popover and word count popover to account for the potentially smaller width.

## Additional information

Yellow is the current draggable area:

Screenshot 2026-06-27 at 15 54 03

Screenshot 2026-06-27 at 16 05 53

## AI Disclosure Statement

N/A

Tested on:

### DIFF
--- a/source/common/vue/window/WindowToolbar.vue
+++ b/source/common/vue/window/WindowToolbar.vue
@@ -140,11 +140,15 @@ const hasRTLTrafficLights = ref<boolean>(false)
 
 const scrollArea = ref<HTMLDivElement|null>(null)
 const canScroll = ref(false)
+// To account for really small differences between `scrollWidth` and `clientWidth`
+// (sometimes only fractions of a pixel), add a small additional tolerance to
+// `clientWidth` to prevent erroneously displaying of the scroll buttons.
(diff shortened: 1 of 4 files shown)


nathanlesage:
Looks good, thanks! The log window needs an overhaul anyhow soon…

btw, I see you're using stats, too. What's the 8W number indicating? (I'm asking since I wanted to also get a quick indicator of how fast my laptop charges with various chargers)

benniekiss:
its current power draw, I believe, but clicking the widget brings up a whole host of power info, including charge rate

Issue #6093, opened 2026-01-04
Word and character counts are inflated after the parsing change for identical Markdown content.


RippleTester:
#### Expected Behavior
- Document statistics (word and character count) should be consistent for the same input across builds.
- The link parsing refactor should not change basic statistics computation unless explicitly intended.

#### Actual Behavior
- Before the change, stats show ~2 words / 14 characters (matching the visible text “Wikipedia test”).
- After the change, stats show ~4 words / 26 characters, even though the preview still displays only “Wikipedia test”.

#### Notes
This is introduced by Pull Request https://github.com/Zettlr/Zettlr/pull/5982 , namely Commit https://github.com/Zettlr/Zettlr/commit/a99afc6773bf57a1979e78fefb04de106fb475bb

#### Screenshot Before

<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/47a90c75-0e72-4290-a13f-ddf250bbc768" />

#### Screenshot After

<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/4bca05e1-ad17-4a5f-89ba-02abd5565c2b" />

#### Steps to Reproduce

1. Create a new Markdown file.

2. Enter: `[Wikipedia test](https://en.wikipedia.org/wiki/Function_(mathematics) "Math article")`

3. Observe the word/character counters (top-right header stats and the bottom status bar for the active document).

### Zettlr version

4.0.0-beta.1

### Installation Method

From the Website or GitHub

### Your Platform

- [ ] Windows
- [x] macOS
- [x] Linux

### Architecture

- [ ] Intel 64bit
- [x] ARM 64bit (includes Apple Silicon)

### Operating System Version

ubuntu:22.04, macOS 15.3.2 (24D81)

### Additional Information

This also happend when input is `[![Updated](https://img.shields.io/badge/Updated-2025--10--13-darkgreen)](https://pandoc.org/)`

#### Screenshot before
<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/de486679-6026-4e96-a659-e113cfbddbcf" />

#### Screenshot after

<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/92d5bce5-3722-47d5-aaae-a9f8094d0a5a" />

benniekiss:
First issue i think is intended behavior. You have a link title, so its included in the word count.

[...]

PR #6425, opened 2026-06-29
fix: exclude link/image title from word & character counts (#6093)

CedricConday:
## What

A link's (or image's) **title** — the quoted tooltip after the URL — is incorrectly counted toward the document's **word and character statistics**.

`extractTextnodes` pushes both the visible text (`alt`) **and** the `title` as text nodes. Since the counter (and the spellchecker) consume those text nodes, the title inflates the stats even though it isn't visible document text.

### Reproduction (from #6093)

```
[Wikipedia test](https://en.wikipedia.org/wiki/Function_(mathematics) "Math article")
```

- Expected: **2 words / 14 chars** ("Wikipedia test")
- Actual: **4 words / 26 chars** ("Wikipedia test" + "Math article")

## Fix

In `extractTextnodes`, push only the visible text (`alt`) for `Link`/`Image` nodes; drop the `title`. The title is metadata, so it should neither be counted nor spellchecked.

## Test

Adds a regression case to `test/counter.spec.ts`. Fails on the previous code, passes with this change. Full unit suite: **377 passing**.

---

Addresses #6093 (the link-title case). Per @benniekiss's review, the **image-within-a-link** case (an image's URL counted inside link text) is a separate remaining problem — happy to follow up on that in a dedicated PR, so leaving the issue open rather than auto-closing.

### DIFF
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -152,10 +152,10 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
 
     case 'Image':
     case 'Link': {
+      // Only the visible link text (alt) is document text. The title — the
+      // quoted tooltip after the URL — is metadata and must not be counted
+      // (nor spellchecked), otherwise it inflates word/char statistics. See #6093.
       textNodes.push(ast.alt)
-      if (ast.title !== undefined) {
-        textNodes.push(ast.title)
-      }
       break
     }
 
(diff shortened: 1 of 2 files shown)


benniekiss:
Thank you @CedricConday ! I was incorrect in thinking the title was part of the document text.

This doesn't actually completely close the linked issue, as there is still a problem with images within links having the image URL included in word counts, but it's a step!

LGTM

benniekiss (approved):

PR #6426, opened 2026-06-29
fix: count image alt, not URL, when an image is a link's text (#6093)

CedricConday:
## Follow-up to #6425

In #6425 @benniekiss noted that PR still left one case open:

> This doesn't actually completely close the linked issue, as there is still a problem with **images within links having the image URL included in word counts**

This PR fixes that case.

## Bug

When a link's visible text is an image — a clickable banner, `[alt text](https://example.com)` — the parser stored the **raw image Markdown** (including the image URL) in the link's `alt` range, because `alt` is taken as the raw substring between the link marks. The counter then counted that raw text:

| Markdown | Counted before | Correct |
|---|---|---|
| `alt text` (bare image) | 2 words / 8 chars | 2 / 8 |
| `[alt text](…)` (image as link text) | **5 words / 40 chars** | 2 / 8 |

## Fix

In `parseNode` (`markdown-ast`), when a `Link`'s text is exactly an `Image`, build the link's `alt` from the inner image's alt text instead of the raw substring. Guarded so it only triggers when the image fills the link text — plain links, bare images, links with titles, and links with text *surrounding* an image are all unchanged.

## Test

Adds a regression case to `test/counter.spec.ts` (`[alt text](…)` → 2 words / 8 chars). Full suite green: **377 passing** (375 → 377; the new case adds the word + char assertions).

### DIFF
--- a/source/common/modules/markdown-utils/markdown-ast/index.ts
+++ b/source/common/modules/markdown-utils/markdown-ast/index.ts
@@ -571,6 +571,21 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
         }
       }
 
+      // A link whose visible text is itself an image — e.g. a clickable banner,
+      // `[![alt](img.png)](url)` — stores the raw image Markdown, including the
+      // image URL, between its link marks. Counting that raw text inflates the
+      // word/character statistics, so when the link text is exactly an image,
+      // count that image's alt text instead. See #6093.
+      const innerImage = node.name === 'Link' ? node.getChild('Image') : null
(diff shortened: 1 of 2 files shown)


benniekiss:
I'll need to refamiliarize myself with the AST parser, but my initial feeling is that a better approach would be to refactor the link AST to properly parse the interior nodes, since links can include any inline formatting. This would mean that the link, `[some alt text, *an image*, alt text, and another image alt text](https://example.com)`, would parse the correct alt-text. However, I haven't looked at it in a bit, so this may not currently be possible without a significant change.

If it ends up being too big a change, I think this patch as-is is a good compromise for the time being.

PR #6427, opened 2026-06-29
fix: parse link interior nodes so counts use visible text only (#6093)

CedricConday:
## The structural fix from #6426

In #6426, @benniekiss suggested the proper approach:

> refactor the link AST to properly parse the interior nodes, since links can include any inline formatting

This PR does that. A link node now carries its **parsed inline children** (text, emphasis, images, …) instead of a single flat `alt` string, and `extractTextnodes` recurses into them.

## Why

The flat `alt` was the raw text between the brackets, so the counter (and spellchecker) mis-counted anything that wasn't plain text inside a link:
- a link **title** was counted (#6093),
- an **image used as a link's text** counted the image URL (#6426),
- **mixed inline content** counted every URL.

Parsing the interior fixes all three from one place. Each piece contributes only its **visible** text — an image contributes its alt, never its URL — and the title (metadata) is excluded.

| Markdown | Counted |
|---|---|
| `[alt text](…)` | **2 words / 8 chars** (the image alt) |
| `[Wikipedia test](… "Math article")` | **2 / 14** (title excluded) |
| `[some alt text, *an image*, alt text, and another image alt text](…)` | **12 / 61** (visible text, no URLs) |

## Implementation

- `parseLinkInterior()` parses only the nodes whose range lies between the link's `[` `]` marks, which naturally excludes the marks, URL, and title (the part @benniekiss flagged as the tricky bit).
- `LinkOrImage` gains an optional `children` field; `alt`/`url`/`title` are untouched, so `markdown-to-html` and other consumers are unaffected.

## Tests

Adds counter regression cases for image-in-link, link-title, and the mixed example above. **Full suite green** (381 passing).

## Relationship to the other PRs

This **supersedes the single-image special-case in #6426** and also covers the title case from #6425 — it's a complete fix for #6093. Happy to let #6426 land as the interim and merge this when you're ready, or close #6426 in favour of this — whichever you prefer, @benniekiss.

### DIFF
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -152,12 +152,8 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
 
     case 'Image':
     case 'Link': {
-      // Count only the visible text. Both links and images (per the CommonMark
-      // image-description rules) count the visible text of their bracketed
-      // description: recurse into the parsed inline children so each contributes
-      // its own visible text — nested emphasis by its text, a nested image by
-      // its alt, never the raw Markdown, URL, or title. Fall back to `alt` for
-      // the degenerate case of no parsed children. See #6093.
+      // Links and images count the visible text of their bracketed description
+      // via their parsed children; fall back to `alt` when there are none.
       if (ast.children.length > 0) {
         for (const child of ast.children) {
           textNodes = textNodes.concat(extractTextnodes(child, filter))
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -152,12 +152,13 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
 
     case 'Image':
     case 'Link': {
-      // Count only the visible text. For a link with parsed inline children
-      // (text, emphasis, nested images) recurse so each contributes its own
-      // visible text — an image contributes its alt, never its URL. The title,
-      // the quoted tooltip after the URL, is metadata and must not be counted.
-      // See #6093.
-      if (ast.type === 'Link' && ast.children.length > 0) {
+      // Count only the visible text. Both links and images (per the CommonMark
+      // image-description rules) count the visible text of their bracketed
+      // description: recurse into the parsed inline children so each contributes
+      // its own visible text — nested emphasis by its text, a nested image by
+      // its alt, never the raw Markdown, URL, or title. Fall back to `alt` for
+      // the degenerate case of no parsed children. See #6093.
+      if (ast.children.length > 0) {
         for (const child of ast.children) {
           textNodes = textNodes.concat(extractTextnodes(child, filter))
         }
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -157,7 +157,7 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
       // visible text — an image contributes its alt, never its URL. The title,
       // the quoted tooltip after the URL, is metadata and must not be counted.
       // See #6093.
-      if (ast.type === 'Link' && ast.children !== undefined && ast.children.length > 0) {
+      if (ast.type === 'Link' && ast.children.length > 0) {
         for (const child of ast.children) {
           textNodes = textNodes.concat(extractTextnodes(child, filter))
         }
--- a/source/common/modules/markdown-utils/index.ts
+++ b/source/common/modules/markdown-utils/index.ts
@@ -152,9 +152,17 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
 
     case 'Image':
     case 'Link': {
-      textNodes.push(ast.alt)
-      if (ast.title !== undefined) {
-        textNodes.push(ast.title)
+      // Count only the visible text. For a link with parsed inline children
+      // (text, emphasis, nested images) recurse so each contributes its own
+      // visible text — an image contributes its alt, never its URL. The title,
+      // the quoted tooltip after the URL, is metadata and must not be counted.
+      // See #6093.
+      if (ast.type === 'Link' && ast.children !== undefined && ast.children.length > 0) {
+        for (const child of ast.children) {
+          textNodes = textNodes.concat(extractTextnodes(child, filter))
+        }
+      } else {
+        textNodes.push(ast.alt)
       }
       break
     }
--- a/source/common/modules/markdown-utils/markdown-ast/index.ts
+++ b/source/common/modules/markdown-utils/markdown-ast/index.ts
@@ -147,6 +147,12 @@ export interface LinkOrImage extends MDNode {
    * Optional title text (i.e. what can be added after the URL in quotes)
    */
   title?: TextNode
+  /**
+   * For links, the parsed inline contents between the brackets (text, emphasis,
+   * images, etc.), so consumers can recurse into the real inline structure
+   * instead of treating `alt` as a flat string. See #6093.
+   */
+  children?: ASTNode[]
 }
 
 /**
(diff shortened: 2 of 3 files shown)


source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -538,6 +544,33 @@ export type ASTNodeType = ASTNode['type']
...
+      index = child.to
+    }
+    child = child.nextSibling
+  }
+  if (index < to) {
+    children.push(genericTextNode(index, to, markdown.substring(index, to)))
+  }
+  return children
+}
+
```
benniekiss: please see the `parseChildren` function in the `parse-children.ts` file. It should do what you need here. This function should be removed.

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -585,6 +618,13 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
           : genericTextNode(url.from, url.to, markdown.substring(url.from, url.to))
       }
 
+      // For links, parse the bracketed interior into child nodes so nested
+      // inline content (emphasis, images, ...) is counted/spellchecked by its
+      // visible text only — not the raw image Markdown or URLs. See #6093.
+      if (node.name === 'Link' && marks.length >= 2) {
+        astNode.children = parseLinkInterior(node, marks[0].to, marks[1].from, markdown)
```
benniekiss: Update this callsite, relying on `parseChildren`

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -147,6 +147,12 @@ export interface LinkOrImage extends MDNode {
    * Optional title text (i.e. what can be added after the URL in quotes)
    */
   title?: TextNode
+  /**
+   * For links, the parsed inline contents between the brackets (text, emphasis,
+   * images, etc.), so consumers can recurse into the real inline structure
+   * instead of treating `alt` as a flat string. See #6093.
+   */
+  children?: ASTNode[]
```
benniekiss: this attribute should not be optional:

```suggestion
  children: ASTNode[]
```

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -147,6 +147,12 @@ export interface LinkOrImage extends MDNode {
    * Optional title text (i.e. what can be added after the URL in quotes)
    */
   title?: TextNode
+  /**
+   * For links, the parsed inline contents between the brackets (text, emphasis,
+   * images, etc.), so consumers can recurse into the real inline structure
+   * instead of treating `alt` as a flat string. See #6093.
```
benniekiss: make the comment less verbose:

```suggestion
   * For links, the parsed inline contents between the brackets (text, emphasis,
   * images, etc.)
```

source/common/modules/markdown-utils/index.ts
```diff
@@ -152,9 +152,17 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
...
     case 'Link': {
-      textNodes.push(ast.alt)
-      if (ast.title !== undefined) {
-        textNodes.push(ast.title)
+      // Count only the visible text. For a link with parsed inline children
+      // (text, emphasis, nested images) recurse so each contributes its own
+      // visible text — an image contributes its alt, never its URL. The title,
+      // the quoted tooltip after the URL, is metadata and must not be counted.
+      // See #6093.
+      if (ast.type === 'Link' && ast.children !== undefined && ast.children.length > 0) {
```
benniekiss:
```suggestion
      if (ast.type === 'Link' && ast.children.length > 0) {
```

benniekiss:
> One thing worth a look: `parseChildren` walks **all** of the node's Lezer children — for a `Link` that includes the `LinkMark`s, the `URL`, and the `LinkTitle` (none are in `EMPTY_NODES`), so passing them straight through would count the URL/title text. I kept only the children whose range falls inside the brackets (`[marks[0].to, marks[1].from)`),

Please add those nodes to the empty nodes list, and remove the postprocess filter. That is was the list is for.

However, please check for any unexpected behavior from adding those nodes to the list, especially in the print preview window.

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -582,7 +587,18 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
...
+      }
+
+      // Parse the bracketed interior into child nodes so nested inline content
+      // (emphasis, images, ...) is counted/spellchecked by its visible text
+      // only, never the raw image Markdown or URL. parseChildren walks all of
+      // the node's children (link marks, URL, title included), so keep only
+      // those that lie within the brackets. See #6093.
+      if (marks.length >= 2) {
+        astNode.children = parseChildren({ ...astNode }, node, markdown).children
+          .filter(child => child.from >= marks[0].to && child.to <= marks[1].from)
```
benniekiss, line 600: Remove this filter, and add the nodes you are trying to exclude to the EMPTY_NODES list

source/common/modules/markdown-utils/index.ts
```diff
@@ -152,9 +152,17 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
...
     case 'Image':
     case 'Link': {
-      textNodes.push(ast.alt)
-      if (ast.title !== undefined) {
-        textNodes.push(ast.title)
+      // Count only the visible text. For a link with parsed inline children
+      // (text, emphasis, nested images) recurse so each contributes its own
+      // visible text — an image contributes its alt, never its URL. The title,
+      // the quoted tooltip after the URL, is metadata and must not be counted.
+      // See #6093.
```
benniekiss: This comment is unnecessary

```suggestion
```

benniekiss:
I was also just reviewing the commonmark spec, and this patch should be applied to images as well:

https://spec.commonmark.org/0.31.2/#image-description

source/common/modules/markdown-utils/markdown-ast/index.ts
benniekiss, line 575: this should also use `parseChildren`. 

You can probably move the child parsing above this block

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -582,7 +587,18 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
...
+          : genericTextNode(url.from, url.to, markdown.substring(url.from, url.to)),
+        children: []
+      }
+
+      // Parse the bracketed interior into child nodes so nested inline content
+      // (emphasis, images, ...) is counted/spellchecked by its visible text
+      // only, never the raw image Markdown or URL. parseChildren walks all of
+      // the node's children (link marks, URL, title included), so keep only
+      // those that lie within the brackets. See #6093.
+      if (marks.length >= 2) {
```
benniekiss, line 598: I believe you can remove this conditional

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -582,7 +587,18 @@ export function parseNode (node: SyntaxNode, markdown: string): ASTNode {
...
-          : genericTextNode(url.from, url.to, markdown.substring(url.from, url.to))
+          : genericTextNode(url.from, url.to, markdown.substring(url.from, url.to)),
+        children: []
+      }
+
+      // Parse the bracketed interior into child nodes so nested inline content
+      // (emphasis, images, ...) is counted/spellchecked by its visible text
+      // only, never the raw image Markdown or URL. parseChildren walks all of
+      // the node's children (link marks, URL, title included), so keep only
+      // those that lie within the brackets. See #6093.
```
benniekiss: This comment is unnecessary. It is the commonmark spec.

```suggestion
```

source/common/modules/markdown-utils/index.ts
```diff
@@ -152,9 +152,17 @@ export function extractTextnodes (ast: ASTNode, filter?: (node: ASTNode) => bool
...
+      if (ast.type === 'Link' && ast.children.length > 0) {
```
benniekiss:
```suggestion
      if (ast.children.length > 0) {
```

benniekiss:
lmao.

Please review this projects AI policy: https://github.com/Zettlr/Zettlr/blob/develop/CONTRIBUTING.md#ai-usage-policy

If I suspect any further bot communication, I am going to close this PR.

For posterity, the original comment I was responding to:

````md
Hi Carina, thanks for the careful review.

On the AI disclosure — I've added it to the PR description per the policy: I'm an AI engineer and use Claude Code in my workflow; I find, fix, and test, then review and verify everything before opening it under my name. Happy to walk through any part of the reasoning.

On the approach — I agree, silently shrinking `noverlap` didn't sit right with me either. I've pushed an update that instead **drops** good-data spans shorter than `n_per_seg` with a warning, and raises a clear `ValueError` if *every* span is too short. Glad that matches where you landed.

Happy to sync with @drammock before this is finalized. On your open question — I don't want to guess at why `noverlap < n_per_seg` didn't raise in older MNE, so I'll dig into the history and report back here, so we're deciding on the real prior behavior rather than assumptions.

Thanks!

````

Issue #5357, opened 2024-09-02
Strange warning message in links to footnotes with only a single word


rruusu:
### Description

When a footnote only contains a single word, an error message is generated on the footnote link:

> Unexpected shortcut reference link (\`\[text\]\`), expected collapsed reference (\`\[text\]\[\]\`)
remark-lint (no-shortcut-reference-link)

Adding a second word to the text of the footnote removes this error.

### Reproducing

Create a new empty document. Put into it the following text:

~~~
This is just a test.[^1]

[^1]: Foo
~~~

The warning message in the description appears as a triangle in the margin and an underscoring of the link. Hovering over the link shows both the "Edit" button and the warning message.

![image](https://github.com/user-attachments/assets/c3f15406-0f7d-44ff-b2de-c3c8f6ec9d37)

Add a word "bar" after the "Foo" in the footnote:

~~~
This is just a test.[^1]

[^1]: Foo bar
~~~

The warning message is removed. Only the "Edit" button is shown when hovering over the link.

### Zettlr version

3.2.0

### Installation Method

From the Website or GitHub

### Your Platform

- [X] Windows
- [ ] macOS
- [ ] Linux

### Architecture

- [X] Intel 64bit
- [ ] ARM 64bit (includes Apple Silicon)

### Operating System Version

Windows 11 Pro 23H2, build 22631.4037

nathanlesage:
This seems like an easy fix: We'll need to tell the Markdown linter that footnote syntax (with the `^`) is still valid even if it just contains a single word. It's benign, so nothing to worry about. Thanks for reporting!

mike-a:
Fixed yet?

nathanlesage:
> Fixed yet?

If it were, the issue would be closed.

benniekiss:
This is actually caused by this lint: https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-shortcut-reference-link

I do not think there is a workaround other than disabling it in the code, which I think would be fine.

Another approach would be to implement a settings pane to customize which markdown lints are active.

Fightxing:
> This is actually caused by this lint: https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-shortcut-reference-link
> 
> I do not think there is a workaround other than disabling it in the code, which I think would be fine.
> 
> Another approach would be to implement a settings pane to customize which markdown lints are active.

Hi @benniekiss ,

[...]

PR #6434, opened 2026-07-01
feat: add remark-gfm to support GFM

Fightxing:
## Description
Fixed Issue #5357 

## Changes
-  Added `remark-gfm` as a dependency.
- Enabled `remark-gfm` and `remark-math` in `md-lint.ts` before lint rules are applied.
- Enabled `remark-gfm` in `html-to-md.ts`.

## Additional information
See Issue #5357 

## AI Disclosure Statement

DeepSeek v4.1-Pro was used to analyse the problem and codebase. The actual code implementation was done by myself.

Tested on: Arch Linux

### DIFF
--- a/source/common/util/html-to-md.ts
+++ b/source/common/util/html-to-md.ts
@@ -14,6 +14,7 @@
 
 import rehypeParse from 'rehype-parse'
 import rehypeRemark from 'rehype-remark'
+import remarkGfm from 'remark-gfm'
 import remarkStringify, { type Options } from 'remark-stringify'
 import rehypeRemoveComments from 'rehype-remove-comments'
 import { unified } from 'unified'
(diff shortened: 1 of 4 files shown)


benniekiss:
I forgot the CI won't run if there are changes to the lock file :(

source/common/util/html-to-md.ts
```diff
@@ -32,12 +33,14 @@ export default async function html2md (html: string, stripComments = false, stri
...
     .use(rehypeParse)
     .use(rehypeRemark)
+    .use(remarkGfm)
     .use(remarkStringify, stringifyOptions)
 
   const procRemoveComments = unified()
     .use(rehypeParse)
     .use(rehypeRemoveComments, { removeConditional: true })
     .use(rehypeRemark)
+    .use(remarkGfm)
```
benniekiss, line 43: I am not certain if this does anything here. The plugin docs give me the impression that this is not necessary, but they are not clear. Have you tested the output with/without the plugin enabled?

benniekiss: see https://github.com/remarkjs/remark-gfm#what-is-this

Fightxing: That's because `remarkGfm` is not a linter rule; in fact it fundamentally changes how the parser constructs and interprets the AST. Without it, the praser won't (or shouldn't) create correct GFM Markdown syntax (like tables and to-do lists), it should only create CommonMark syntax.

But Zettlr uses GFM as its baseline Markdown syntax, so we need this to ensure that any HTML converted back into Markdown is correctly structured as GFM-compliant AST nodes from the start. That's why I included it here.

nathanlesage (approved):
Good find, looks good to me

benniekiss:
Oop, I forgot to mark this as good earlier!

lgtm

nathanlesage:
hahaha :D

PR #6456, opened 2026-07-13
feat: Look Up feature in contextual menu on macOS

### LINKED ISSUE #6455: "Look Up" system feature in macOS contextual menu
### Checklist

- [x] Zettlr's current features cannot be used to approximate this behaviour (otherwise, propose an enhancement instead.)

### Feature Proposal

I am constantly reaching for (and missing) the macOS system "Look Up"  highlighted-word contextual menu when working in Zettlr.

Could it be incorporated?

Image

### Example Workflow

- Right-click over a word, or on highlighted text
- Select 'Look up "[word]"'
- Pop-up of macOS system dictionary lookup of [word] is displayed

### Caveats

None

Yes

### Platforms

- [ ] Windows
- [x] macOS
- [ ] Linux

### Additional Information

_No response_

fiveangle:
## Description
Implement "Look Up" feature request on macOS platform (Closes #6455)

## Changes
Implemented the feature as a helper in similar fashion to existing project standards and applied to the relevant contextual menu sites throughout the app, limiting the feature in link prose to only non-empty highlighted text. Purposefully not applied to text input sites, as they appear limited mostly to custom user-specific preference items (regex patterns, etc)

## Additional information

macOS UI standards dictate the "Look Up" item be included as the first item in the relevant contextual menu (which I concur with, for consistency sake) but since Zettlr currently does not adhere to the guidance, the implementation would be a marked change for all users accustomed to the current behavior. At maintainer's discretion, feel free to accept either the feature as-is which renders the Look Up feature at the top of the contextual menu, or the refactor commit addition which renders the feature below the current Zettlr structure, prior to the rest of the system menu additions.

- hand-tested functionality on all targeted sites via project test build frame work (very nice setup ! thanks for making it trivial for us 😃 )
- yarn lint: Passed with exit code 0. It reported 147 existing repository warnings and zero errors; no change from current `develop` head.
- yarn test: Passed—all 375 tests.
- Tested on: macOS Tahoe 26.5.2 (M5)
- Support for other platforms investigated, but none have a similar base-OS-provided feature without additional apps/plugins.

## AI Disclosure Statement

OpenAI model `OpenAI Codex GPT-5.6 Sol` was used in the making of this PR with complete human inspection of all final code.

### DIFF
--- a/source/app/service-providers/windows/index.ts
+++ b/source/app/service-providers/windows/index.ts
@@ -78,6 +78,7 @@ export type WindowControlsIPCAPI = IPCAPI<{
   'copy': unknown
   'paste': unknown
   'selectAll': unknown
+  'show-definition-for-selection': undefined
   'inspect-element': { x: number, y: number }
   'drag-start': { filePath: string }
   'show-item-in-folder': { itemPath: string }
@@ -191,6 +192,12 @@ export default class WindowProvider extends ProviderContract {
           break
         case 'selectAll':
           event.sender.selectAll()
+          break
+        // This command is only available on macOS
+        case 'show-definition-for-selection':
+          if (process.platform === 'darwin') {
+            event.sender.showDefinitionForSelection()
+          }
           break
         case 'inspect-element':
           event.sender.inspectElement(Math.round(payload.x), Math.round(payload.y))
--- a/source/common/modules/markdown-editor/context-menu/default-menu.ts
+++ b/source/common/modules/markdown-editor/context-menu/default-menu.ts
@@ -283,7 +283,7 @@ export async function defaultMenu (view: EditorView, node: SyntaxNode, coords: {
 
   const lookupItem = getMacOSLookupItem(view)
   if (lookupItem !== undefined) {
-    tpl.unshift(lookupItem, { type: 'separator' })
+    tpl.push({ type: 'separator' }, lookupItem)
   }
 
   // If we found a diagnostic earlier and a word, add the suggestion items
--- a/source/common/modules/markdown-editor/context-menu/default-menu.ts
+++ b/source/common/modules/markdown-editor/context-menu/default-menu.ts
@@ -22,6 +22,7 @@ import { applyBold, applyItalic, insertLink, applyBlockquote, applyOrderedList, 
 import { cut, copyAsPlain, copyAsHTML, paste, pasteAsPlain } from '../util/copy-paste-cut'
 import { getTransformSubmenu } from './transform-items'
 import { extractLTSpellcheckSuggestionsFrom, isLanguageToolMisspelling } from '../linters/language-tool'
+import { getMacOSLookupItem } from './macos-lookup-item'
 
 const ipcRenderer = window.ipc
 const suggestionCache = new Map<string, string[]>()
@@ -279,6 +280,11 @@ export async function defaultMenu (view: EditorView, node: SyntaxNode, coords: {
     },
     getTransformSubmenu(view)
   ]
+
+  const lookupItem = getMacOSLookupItem(view)
+  if (lookupItem !== undefined) {
+    tpl.unshift(lookupItem, { type: 'separator' })
+  }
 
   // If we found a diagnostic earlier and a word, add the suggestion items
   if (diagnostic !== undefined && misspelledWord !== undefined) {
--- a/source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
+++ b/source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
@@ -1,0 +1,45 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        macOS lookup menu item
+ * CVM-Role:        Utility function
+ * Maintainer:      Hendrik Erz
+ * License:         GNU GPL v3
+ *
+ * Description:     Creates a context menu item for looking up the current
+ *                  editor selection in the macOS dictionary.
+ *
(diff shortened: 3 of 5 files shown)


source/app/service-providers/windows/index.ts
```diff
@@ -78,6 +78,7 @@ export type WindowControlsIPCAPI = IPCAPI<{
   'copy': unknown
   'paste': unknown
   'selectAll': unknown
+  'show-definition-for-selection': undefined
```
nathanlesage, line 81:
```suggestion
  'show-definition-for-selection': unknown
```

source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * @ignore
+ * BEGIN HEADER
+ *
+ * Contains:        macOS lookup menu item
+ * CVM-Role:        Utility function
+ * Maintainer:      Hendrik Erz
```
nathanlesage, line 7: Please add your name here, since the file comes from you

source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Description:     Creates a context menu item for looking up the current
+ *                  editor selection in the macOS dictionary.
+ *
+ * END HEADER
+ */
+
+import { type EditorView } from '@codemirror/view'
+import { trans } from '@common/i18n-renderer'
+import { type AnyMenuItem } from '@common/modules/window-register/application-menu-helper'
+import { type WindowControlsIPCAPI } from '@providers/windows'
```
nathanlesage, line 19: Please review all of these imports and only keep the ones actually used in this file. In addition, all imports in the renderer from the main process are required to have the `type` annotation outside the brackets because we can not import anything from main, except the types.

source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const ipcRenderer = window.ipc
+
+export function getMacOSLookupItem (view: EditorView): AnyMenuItem|undefined {
```
nathanlesage, line 23: Please add a JSDoc string to this method

source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
```diff
@@ -0,0 +1,45 @@
...
+  // Only macOS provides a native dictionary popover for the current selection
```
nathanlesage, line 24: This comment does not seem factual…?

source/common/modules/markdown-editor/context-menu/default-menu.ts
```diff
@@ -280,6 +281,11 @@ export async function defaultMenu (view: EditorView, node: SyntaxNode, coords: {
     getTransformSubmenu(view)
   ]
 
+  const lookupItem = getMacOSLookupItem(view)
+  if (lookupItem !== undefined) {
+    tpl.push({ type: 'separator' }, lookupItem)
+  }
```
nathanlesage, line 287: From what I can see on my machine, the "Look up" item is located at the top of the context menu, not the bottom. I suggest unshifting that instead in reverse order.

source/common/modules/markdown-editor/table-editor/context-menu.ts
```diff
@@ -214,6 +215,11 @@ export function displayTableContextMenu (event: MouseEvent, mainView: EditorView
     }
   ]
 
+  const lookupItem = getMacOSLookupItem(subviewOrView)
+  if (lookupItem !== undefined) {
+    template.push({ type: 'separator' }, lookupItem)
+  }
```
nathanlesage, line 221: Same as above: The lookup item should be at the top of the context menu.

nathanlesage (changes requested):
So far looks fine and the linters and tests haven't complained, but there are a few things I'd like you to improve before we can merge this. All of which should be things you can quickly do yourself.

fiveangle:
Sorry for late reply. Just got back from Burningman and will review/update PR at earliest convenience.

source/common/modules/markdown-editor/context-menu/macos-lookup-item.ts
fiveangle: i did not find a similar feature in Win 11 (APIs for spellcheck naturally exist, but none for definitions that i could find... only that MS Edge provides such a feature within it only). For Electron apps (e.g. Zettlr for macOS), their documentation states showDefinitionForSelection() as [macOS-only](https://www.electronjs.org/docs/latest/api/web-contents#contentsshowdefinitionforselection-macos). Linux of course has alternatives such as GNOME Dictionary and KDE’s dictionary runner, but integrating those would require separate handling and availability checks as they are not guaranteed to be present, unlike the feature on macOS. I’ll narrow the comment to the Electron API limitation, or omit it entirely if desired.

source/common/modules/markdown-editor/context-menu/default-menu.ts
fiveangle: yes, that is my preference as well. I mentioned in the PR above that the separate "bottom-loaded" commit was in case you didn't want the "Look Up" feature to be injected into the top, where existing users would not expect it, if they were used to the current behavior. I'll remove the optional bottom-loaded commit and change code to match the render order.

PR #6458, opened 2026-07-18
Allow snippets inside YAML frontmatter

### LINKED ISSUE #6240: Snippet menu can't be invoked when cursor is in YAML front matter.
### Description

I have a couple snippets I recently created for regular use. One is to create the YAML front matter for a document and the other is to add a `header-includes` section to an existing YAML front matter section that I use in only some documents. However, I'm unable to use the second snippet because when I place my cursor inside an existing YAML front matter section, the `:` character does not work to invoke the snippet menu. It continues to work just fine _outside_ of the YAML front matter section. Here's a quick video:

Image

Note, the GIF got sped up or cropped a bit. So, for posterity, I'll note that it doesn't matter how long you wait after pressing the `:` character—the snippet menu doesn't pop up.

### Reproducing

See above. Also, a few things I've considered and tried: It does not matter whether the cursor placement follows a comment or whether it occurs on the first or last line of the front matter.

### Zettlr version

4.2.1+nightly

### Installation Method

From the Website or GitHub

AppImage

### Your Platform

- [ ] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Manjaro (Rolling)

### Additional Information

Thanks, as always!

deeferentleeg:
## Summary

Fixes #6240 — the snippet menu (triggered by `:`) does not appear when the cursor is inside a YAML frontmatter block, even though snippets (e.g. one that inserts a `header-includes` section) are genuinely useful there.

This implements the fix @nathanlesage described in the issue: move the "forbidden tokens" logic away from the central autocomplete source and into the individual autocompletes, so each plugin decides for itself where it should be suppressed.

## The problem

`autocompleteSource` in `source/common/modules/markdown-editor/autocomplete/index.ts` applied a blanket guard for *all* autocompletes:

```ts
const forbiddenTokens = ['YAMLFrontmatter', 'YAMLFrontmatterStart', 'YAMLFrontmatterEnd', 'MathEquation']
if (ctx.tokenBefore(forbiddenTokens) !== null) {
  return null
}
```

This runs before any plugin's `applies()` is consulted, so every autocomplete — including snippets — is suppressed inside YAML frontmatter and math blocks. That makes sense for tags, citations, file links, headings, and code-block infostrings, but snippets are user-defined text expansions that are valuable inside frontmatter.

## The fix

Two small, targeted changes:

**`autocomplete/index.ts`** — added an optional `forbiddenTokens?: string[]` field to the `AutocompletePlugin` interface. The central guard was removed; instead, each plugin's own `forbiddenTokens` (defaulting to the previous YAML + math set) is checked inside the dispatch loop, right before that plugin's `applies()` runs. All existing plugins keep the default, so their behavior is unchanged.

**`autocomplete/snippets.ts`** — the snippets plugin sets `forbiddenTokens: ['MathEquation']`, explicitly opting back into YAML frontmatter while still suppressing snippets inside math blocks (where a `:` is part of the equation, not a trigger).

## Why this approach

- **Minimal & behavior-preserving:** every plugin except snippets behaves exactly as before. Only snippets-in-YAML is newly allowed, which is the reported bug.
- **Matches maintainer intent:** this is the refactor @nathanlesage proposed — "moving that logic away from the central logic to the individual autocompletes."
- **Extensible:** if another plugin later needs to opt into a previously-forbidden context (e.g. tags inside YAML for `#tags:` frontmatter), it just sets its own `forbiddenTokens` — no further changes to the central source needed.

## Scope touched

- `source/common/modules/markdown-editor/autocomplete/index.ts` — interface field added, central guard moved into the dispatch loop
- `source/common/modules/markdown-editor/autocomplete/snippets.ts` — one field added to the plugin export

No changes to `applies()`, `entries()`, or any snippet parsing/tabstop logic — the existing `test/snippets.spec.ts` exercises `template2snippet` and is unaffected.

## Testing

I reviewed the existing `test/snippets.spec.ts`, which covers `template2snippet`/placeholder parsing and is untouched by this change (the forbidden-token behavior lives in the central source, not in the parsing function). I did not add a new spec because the snippet-menu trigger depends on a full CodeMirror editor + syntax-tree + config-facet setup that the existing unit tests don't scaffold; happy to add an integration-style test if you'd prefer one.

## AI usage disclosure

Closes #6240

### DIFF
--- a/source/common/modules/markdown-editor/autocomplete/index.ts
+++ b/source/common/modules/markdown-editor/autocomplete/index.ts
@@ -54,9 +54,16 @@ export interface AutocompletePlugin {
    */
   entries: (ctx: CompletionContext, query: string) => Completion[]
   fields?: Array<StateField<any>>
+  /**
+   * The list of CodeMirror syntax-tree tokens inside which this autocomplete
+   * should NOT trigger. By default, every plugin is suppressed inside YAML
+   * frontmatter and math blocks. A plugin can override this list (e.g. set it
+   * to an empty array) to opt back into those contexts.
+   */
+  forbiddenTokens?: string[]
 }
 
-const forbiddenTokens = [
+const defaultForbiddenTokens = [
   'YAMLFrontmatter',
   'YAMLFrontmatterStart',
   'YAMLFrontmatterEnd',
@@ -67,17 +74,20 @@ const autocompleteSource: CompletionSource = function (ctx): CompletionResult|nu
   // This function is called for every keystroke and shall determine whether to
   // actually start the autocomplete.
 
-  // With this function we check whether we are currently within "forbidden"
-  // tokens (i.e. codeblocks, YAML stuff, etc.)
-  if (ctx.tokenBefore(forbiddenTokens) !== null) {
-    return null
-  }
-
   let plugin: AutocompletePlugin|undefined
   let startpos = ctx.pos
 
   // NOTE: Headings has to be checked before tags
   for (const p of [ codeBlocks, citations, files, headings, tags, snippets ]) {
+    // Each plugin decides for itself in which contexts it should be suppressed
+    // (i.e. codeblocks, YAML stuff, etc.). This allows individual plugins to opt
+    // back into contexts that were previously forbidden for everyone — such as
+    // snippets inside YAML frontmatter (see issue #6240).
+    const forbidden = p.forbiddenTokens ?? defaultForbiddenTokens
+    if (forbidden.length > 0 && ctx.tokenBefore(forbidden) !== null) {
+      continue
+    }
+
     const res = p.applies(ctx)
     if (res !== false) {
       plugin = p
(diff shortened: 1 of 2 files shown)


nathanlesage:
Hi, thanks for opening a bunch of PRs that seem to fix a good amount of issues. Before I go into reviews, I have noticed that you did not use the PR description template, and the descriptions you provided look like they have been generated with an LLM. Could you specify why you did not use the PR description template, and whether you drafted the descriptions for these PRs yourself?

To quote the AI Usage Policy from the Contributing guide:

> You agree to follow ethical principles in whatever you do. This especially holds true for "Respect for persons": You are interacting with humans here, and we expect you to respect them. This means that nobody wishes to talk to a GPT-model. Communicate on your own, regardless of how much AI you use to help you contribute.

Thank you for clarifying!

nathanlesage:
Closing due to lack of contributor response.

PR #6459, opened 2026-07-18
fix(autocomplete): strip citeproc HTML from citation picker preview

### LINKED ISSUE #6409: HTML <span />s as plain-text in citation picker preview pane
### Description

When adding a citation in-text, I noticed that html-tags were written out as plain text in the citation preview picker, making the title difficult to read.

Image

Note that this is not the case for the preview of the citation once it has been added:
Image

It seems like the bib-syntax of double curly brackets to maintain the Case is translated into <span class="nocase" />s, and this html code is not parsed as-is but first encoded before being displayed in the preview pane.

### Reproducing

On: Zettlr 4.6.0 (Mac 26.5.1)
File format: Markdown
Bib reference:
```bib
@article{stirlingOpeningClosingPower2008,
  title = {"{{Opening Up}}" and "{{Closing Down}}": {{Power}}, {{Participation}}, and {{Pluralism}} in the {{Social Appraisal}} of {{Technology}}},
  shorttitle = {“{{Opening Up}}” and “{{Closing Down}}”},
  author = {Stirling, Andy},
  date = {2008-03-01},
  journaltitle = {Science, Technology, \& Human Values},
  volume = {33},
  number = {2},
  pages = {262--294},
  publisher = {SAGE Publications Inc},
  issn = {0162-2439},
  doi = {10.1177/0162243907311265},
  url = {https://doi.org/10.1177/0162243907311265},
  urldate = {2026-05-27},
  langid = {english}
}
```

Start typing @ and move selection to the bib reference, triggering the preview pane.

### Zettlr version

4.6.0

### Installation Method

Homebrew

_No response_

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit

### Operating System Version

26.5.1 (25F80

### Additional Information

_No response_

deeferentleeg:
## Summary

Fixes #6409 — the citation autocomplete picker's info pane shows raw `<span class="nocase">…</span>` (and other citeproc HTML) as plain text instead of the title.

## Root cause

In `source/common/modules/markdown-editor/autocomplete/citations.ts`, each citation completion is built with `info: entry.displayText`. That `displayText` comes straight from the citeproc database, and bibTeX's case-preserving double-curly-brace syntax (`{{Opening Up}}`) is serialised by citeproc as `<span class="nocase">Opening Up</span>`. Other markup like `<i>`/`<b>` can appear in titles too.

CodeMirror renders a **string** `info` as escaped plain text in the autocomplete info pane, so the tags show up literally — e.g. the reporter's example displayed the raw `<span class="nocase">` tags around the title words instead of the title itself.

## The fix

Add a small `stripCitationHtml` helper that removes HTML tags from the display text before it's exposed as the completion `info`, leaving just the title text:

```ts
function stripCitationHtml (displayText: string): string {
  return displayText
    .replace(/<[^>]+>/g, '')
    .replace(/\s+/g, ' ')
    .trim()
}
```

…then `info: stripCitationHtml(entry.displayText)` in the `citekeyUpdateField` mapper.

This matches @nathanlesage's read on the issue — *"Ah, that should be easy to fix. In fact, I think I can use this to improve the citation preview."*

## Why strip rather than render

We only ever display the text content of the title in the info pane, so dropping the tags is both safe and correct, and it avoids any HTML-rendering / sanitisation surface. Stripping also keeps the `entries()` query filter honest — it matches against the same cleaned text the user actually sees.

## Scope

- `source/common/modules/markdown-editor/autocomplete/citations.ts` — one helper + one call-site change.

No changes to the citation `applies()`/`apply()` logic, the parser, or the database loader. The `nocase` styling isn't used anywhere in Zettlr's CSS (it's a citeproc/biblatex convention), so removing the wrapper tags has no visual regression elsewhere.

## Verification

- Checked `test/` — no existing autocomplete-citations test references `citekeyUpdate`, so nothing to break.
- The exact data shape is confirmed by `test/database-loader.spec.ts`, which has a title `Social <span class="nocase">Revolutions</span> in the <span class="nocase">Modern World</span>` → after stripping: `Social Revolutions in the Modern World`.

## AI usage

Fixes #6409

### DIFF
--- a/source/common/modules/markdown-editor/autocomplete/citations.ts
+++ b/source/common/modules/markdown-editor/autocomplete/citations.ts
@@ -20,6 +20,30 @@ import { configField } from '../util/configuration'
 import { extractCitationNodes, nodeToCiteItem } from '../parser/citation-parser'
 
 /**
+ * Strips inline HTML that citeproc/biblatex emits in citation display text so
+ * that it can be shown verbatim in the autocomplete info pane.
+ *
+ * BibTeX's double-curly-brace case-preserving syntax (e.g. `{{Opening Up}}`)
+ * is serialised by citeproc as `<span class="nocase">…</span>`, and other
+ * markup such as `<i>`/`<b>` can appear in titles too. CodeMirror renders a
+ * string `info` as escaped plain text, so without stripping, the raw tags show
+ * up literally in the citation picker preview (see issue #6409). We only ever
+ * display the text content, so dropping the tags is both safe and correct.
+ *
+ * @param   {string}  displayText  The raw display text, possibly containing HTML.
+ *
+ * @return  {string}               The display text with any HTML tags removed.
+ */
+function stripCitationHtml (displayText: string): string {
+  // Remove anything that looks like an HTML tag, then collapse the whitespace
+  // that collapsing span boundaries can leave behind.
+  return displayText
+    .replace(/<[^>]+>/g, '')
+    .replace(/\s+/g, ' ')
+    .trim()
+}
+
+/**
  * Use this effect to provide the editor state with a set of new citekeys
  */
 export const citekeyUpdate = StateEffect.define<Array<{ citekey: string, displayText: string }>>()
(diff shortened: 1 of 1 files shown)


nathanlesage:
Closing due to rule breaking.

PR #6473, opened 2026-07-24
fix(snippets): correctly update snippet statefield

### LINKED ISSUE #6467: typewriter mode interaction with snippets misbehaving
### Description

Certain interactions when inserting snippets while in typewriter mode cause unexpected behaviour, including text that can no longer be modified and a "$1" label that does not go away. These do not fix themselves unless the document is closed and reopened, sometimes not until the app is closed and reopened. See reproduction for details.

Edit: Possibly related to https://github.com/Zettlr/Zettlr/issues/6058?

### Reproducing

1. Turn on typewriter mode.
2. Insert snippet. I am using the following `[${1:text}]{custom-style="gr"}`
3. The block `[text]{custom-style="gr"}$1` appears. This is usually well-behaved if you type something to overwrite "text" immediately and press tab, but _sometimes_ moving the cursor or switching to another window and back will cause the `$1` to move erratically. In the following example, I've clicked away from "text" to copy and paste in "57":

Image

4. Sometimes pressing tab will make `$1` go away and for everything to resume normally, but sometimes the `$1` persists on screen, jumping locations and lines with further input. I've also seen this bug causing the window to just reject all new typed text (you type but nothing appears), or for the placeholder "text" to reappear after you type something to overwrite it and click away. This misbehaviour is unpredictable and I haven't found a reliable way to reproduce it, but it requires restarting the app to fix. Closing and reopening the document sometimes fixes it, but can introduce other issues like the dropdown box for `:` no longer showing snippets (but still showing emoji).

### Zettlr version

Zettlr Version: 4.6.0 Build number: fc1408c (Sunday, June 14, 2026)

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Architecture

- [ ] Intel 64bit

### Operating System Version

macOS Tahoe Version 26.3.1 (a)

### Additional Information

_No response_

benniekiss:
## Description

Statefields should be immutable, but the snippet field was updating `activeSelections` on the field itself which causes problems with the `transactionExtender` and `transactionFilter` functions. As such, the statefield will stop updating, and for snippets, this means the placeholder decorations start to drift down the page. 

This seems to be explicitly caused by the `scrollIntoView` transaction, as this forum post reports an identical issue: https://discuss.codemirror.net/t/using-editorview-scrollintoview-in-transactionextender-breaks-statefield-updates/7476

Closes #6467

## AI Disclosure Statement

None

## Declarations

- [ x ] I hereby confirm that I am solely responsible for the code provided in
  issues with my code.
- [x  ] I hereby confirm that I wrote this PR description myself and that I did
- [ x ] I have specified any open issues that this PR fixes/closes accordingly in

### DIFF
--- a/source/common/modules/markdown-editor/autocomplete/snippets.ts
+++ b/source/common/modules/markdown-editor/autocomplete/snippets.ts
@@ -147,7 +147,7 @@ export const snippetsUpdateField = StateField.define<SnippetStateField>({
   update (val, transaction) {
     for (const effect of transaction.effects) {
       if (effect.is(snippetsUpdate)) {
-        val.availableSnippets = effect.value.map(entry => {
+        let availableSnippets = effect.value.map(entry => {
           return {
             label: entry.name,
             info: entry.content,
@@ -155,16 +155,16 @@ export const snippetsUpdateField = StateField.define<SnippetStateField>({
           }
         })
 
-        return { ...val }
+        return { ...val, availableSnippets }
       } else if (effect.is(snippetTabsEffect)) {
-        val.activeSelections = effect.value
+        let activeSelections = effect.value
 
         // Calculate the association when the effects come in
         // because we need access to the current tab stop
         // range, which is the `transaction.selection` value.
-        val.association = getAssociation(val.activeSelections, transaction.selection?.main.from)
+        let association = getAssociation(val.activeSelections, transaction.selection?.main.from)
 
-        return { ...val }
+        return { ...val, activeSelections, association }
       } else if (effect.is(shiftNextTabEffect)) {
         // NOTE: We cannot shift the range in the nextTab() command, as this
         // change is not transparent to the library (hence it would render a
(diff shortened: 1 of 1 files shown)


benniekiss:
I'll have a follow up PR auditing and cleaning up the other statefields, as well.

nathanlesage:
> Statefields should be immutable

I Never thought about that, but this totally explains some other issues I had in the past with state fields 🫣

I wonder if this could be made clearer by declaring `val` in the update method as readonly ... anyway, this looks perfectly good, thank you for taking care of this! I think we can already include this in 4.7 to help fix the typewriter mismatch bug!

benniekiss:
> I Never thought about that, but this totally explains some other issues I had in the past with state fields 🫣

Maybe I'll ask to add something to the codemirror docs to make it a little more explicit

> I wonder if this could be made clearer by declaring val in the update method as readonly 

I will definitely be checking this route

PR #6496, opened 2026-08-06
feat: Make tags in frontmatters clickable again

nathanlesage:
## Description

This PR addresses the clickability of tags in the frontmatter. The editor now has the capability to properly detect keywords in both the "keywords" and "tags" properties and emit the appropriate search command.

## Changes

To do so, I had to do a few silly things. First, I had to insert a new dummy node into the frontmatter that can be replaced with the YAML syntax tree. The reason is that this way the styling of the CodeText node isn't harmed. Next, the YAML frontmatter is properly mounted (no longer as an overlay), making the syntax tree available to the editor, thereby replacing said dummy node. Finally, the click handlers have been amended with an odd construction that tests whether a clicked node is a keyword which is part of the correct tree structure to be a valid keyword, and, if so, a tag event is emitted that contains that value. Quoted strings are unquoted in the process.

## Tested on

macOS Tahoe 26.6

## Additional information

I have received a few range errors while testing, I'm not sure if those are emanating from the search or from the new implementation. I almost can't imagine that they're due to this PR, but I'd like someone to sanity check this first.

## AI Disclosure Statement

I doubt any LLM would have been able to come up with this weird contraption I did there. So none used for this.

## Declarations

  issues with my code.

### DIFF
--- a/scripts/test-gui/test-files/Miscellaneous/YAML Frontmatter.md
+++ b/scripts/test-gui/test-files/Miscellaneous/YAML Frontmatter.md
@@ -1,0 +1,16 @@
+---
+title: "YAML Frontmatter Test File"
+# tags is an official way to denote keywords in the frontmatter
+tags: [a, list, of, tags]
+# keywords is the same, but specifically for Pandoc
+keywords:
+  - Keywords with spaces
+  - More
+  - 12345 # A number which may produce a different node type
+---
+
+# YAML Frontmatter Test File
+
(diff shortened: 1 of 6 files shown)


nathanlesage:
If any of you two have some time, I'd appreciate a review. Thanks! @benniekiss @kyaso

nathanlesage:
Range errors have been fixed, now everything should work well. I've also added a test file for future reference. Ready to review!

benniekiss:
I'll be able to give this some attention in the next couple of days!

PR #6497, opened 2026-08-07
feat: Add admonitions

### LINKED ISSUE #532: [FEATURE] Add support of admonitions
## Description
Admonitions are a simple way to have shout-out boxes. They were originally introduced by `mkdocs` (I think) and are usually written in markdown as
```
!!! note
    You should note that the title will be automatically capitalized.
```

It would be nice if Zettlr would support them by properly rendering such notes.

render example

## Use cases
- Highlight important information
- It would be fantastic to obtain a global list of all admonition (by type). This would make it possible to use it to collected related things that are specific to one page (e.g. one does a review of some published paper and got an idea for a further project from this; add an adminition "idea" to the page that contains notes about this paper and the overview page can give a nice summary of all ideas one got from various papers).

## Additional Information
Description:
https://squidfunk.github.io/mkdocs-material/extensions/admonition/
https://github.com/qjebbs/vscode-markdown-extended#admonition
https://python-markdown.github.io/extensions/admonition/

Implementation:
https://github.com/vsch/flexmark-java/wiki/Admonition-Extension

Related:
https://asciidoctor.org/docs/asciidoc-writers-guide/#admonitions

nathanlesage:
## Description

> [!important]
> This PR enables admonition parsing and rendering in the main editor. This addresses the six-year-old, and highly in demand issue #532.

Admonition support follows both GitHub, with some additions:

* Define an admonition like a blockquote, but with an additional keyword: `> [!keyword]`
* Supported keywords (case-insensitive) are "note", "tip", "important", "warning", "caution"
* Unsupported keywords make the parser fall back to regular blockquote styling
* Not supported by GitHub, but by Obsidian and VuePress, VitePress, and others: Optional titles: `> [!keyword] optional title`
* Also not supported by GitHub, but by the rest: Nested admonitions

> [!tip]
> Using an unsupported keyword will cause a fallback to regular block styling.

> [!warning]
> Despite Zettlr now pre-rendering admonitions, you will need a system that understands admonitions. If you work on, say, Markdown files intended for GitHub or Obsidian, it will work flawlessly. But if you want to export via Pandoc, additional support (e.g., via Lua filters) is necessary.

> [!note]
> Yeah, I'm being silly.

## Changes

1. Added a parser that mimicks blockquotes, but with the change that it only affects blockquotes that fulfill the (GitHub style) admonition syntax `> [!keyword]`
2. Support five keywords: `note`, `tip`, `important`, `warning`, `caution`
3. Support optional titles (which GitHub does not support, but Obsidian and others do)
4. Add basic syntax highlighting, ensuring that even without the preview renderer, admonitions are discernible from both blockquotes and surrounding text
5. Add a renderer that pre-renders these admonitions with colored backgrounds, leaving it with the standard note = gray, tip = green, important = purple, warning = yellow, and caution = red scheme.
6. Added corresponding theme variables in line with work that @benniekiss has already started to allow customization of the admonition colors
7. Add the corresponding settings to allow enabling/disabling of the renderer appropriately.

## Tested on

macOS 26.6

## Additional information

> [!note]
> Closes #532.

#### Light + Rendering

image

#### Dark + Rendering

image

#### Light + Raw

image

#### Dark + Raw

image

## AI Disclosure Statement

As always, as soon as I realized the chores I had to do it was too late to engage a coding agent. All thinking and choring done by my warm, non-dead hands.

## Declarations

  issues with my code.


nathanlesage:
@benniekiss I think I'd prefer if you could review this PR, since it is closely aligned with work you have already done. Then we can leave #6496 to either @kyaso or maybe @krisfremen if you want.

Also, apparently this change wrecked the table parser (??) of all things. I'll have to investigate.

nathanlesage:
Found the table bug, that's fixed. One thing I just remembered that I still want to add is auto-continue markup thing, because that works with block quotes but not with admonitions at the moment.

nathanlesage:
Ah, also that needs to be added to the AST builder, should be simple too.

benniekiss:
I'll be able to go into this in more detail later in the week, but this looks good overall!

Some initial thoughts. I think that the various `Admonition{keyword}` are too redundant and clutter the parse trree. The nodes should be made into a generic `AdmonitionKeyword` node, and the styling/specific keyword handling should just be moved into the renderer. The renderer can simply add a class to the decoration with the name of the keyword so users can target it for styling.

To keep the fallback  to regular blockquotes, a lookup table in the parser should be added with a list of support keywords, and if the provided one isn't available, parsing moves on. This would make it way simpler to add more keywords in the future, if they ever come up.

I also think the styling variables could be made more generic. Instead of specific admonition variables, I think these should be made into generic `-warning`, `-info`, `-error` variables. that way info/warning/error styling is synchronized across elements. There already is an `--zettlr-editor-error-color` variable, so this approach would align with that.

source/common/modules/markdown-editor/parser/admonition-parser.ts
```diff
@@ -0,0 +1,138 @@
...
+    // AdmonitionHeader to make parsing the node easier.
+    const linestart = ctx.lineStart + line.pos
+    ctx.addElement(ctx.elt('QuoteMark', linestart, linestart + 1))
+
+    const skippedSpace = ctx.lineStart + line.skipSpace(line.pos + 1)
+    const [ kwStart, kwEnd ] = match.indices.groups.keyword
+    const keyword = match.groups.keyword.toLowerCase()
+    // The admonition marker encompasses both the keyword and the two code
+    // marks, because this way it can be styled in its entirety.
+    const headerChildren: Element[] = [ctx.elt('CodeMark', skippedSpace, skippedSpace + 2)]
```
benniekiss: I wouldn't use `CodeMark` here. I would rather have a dedicated `AdmonitionMark` node for clarity.

source/common/modules/markdown-editor/parser/admonition-parser.ts
```diff
@@ -0,0 +1,138 @@
...
+      headerChildren.push(ctx.elt('AdmonitionNote', linestart + kwStart, linestart + kwEnd))
+    } else if (keyword === 'tip') {
+      headerChildren.push(ctx.elt('AdmonitionTip', linestart + kwStart, linestart + kwEnd))
+    } else if (keyword === 'important') {
+      headerChildren.push(ctx.elt('AdmonitionImportant', linestart + kwStart, linestart + kwEnd))
+    } else if (keyword === 'warning') {
+      headerChildren.push(ctx.elt('AdmonitionWarning', linestart + kwStart, linestart + kwEnd))
+    } else if (keyword === 'caution') {
+      headerChildren.push(ctx.elt('AdmonitionCaution', linestart + kwStart, linestart + kwEnd))
+    }
```
benniekiss: this should be condensed into a single `AdmonitionKeyword` node, and the validation should just use a `VALID_KEYWORDS.includes(keyword)` check.

source/common/modules/markdown-editor/renderers/render-admonitions.ts
```diff
@@ -0,0 +1,203 @@
...
+          } else if (admonitionHeader.getChild('AdmonitionImportant')) {
+            inlineRanges.push(getTitleDeco(admonitionTitle ?? trans('Important')).range(from, to))
+            classes.push('important')
+          } else if (admonitionHeader.getChild('AdmonitionWarning')) {
+            inlineRanges.push(getTitleDeco(admonitionTitle ?? trans('Warning')).range(from, to))
+            classes.push('warning')
+          } else if (admonitionHeader.getChild('AdmonitionCaution')) {
+            inlineRanges.push(getTitleDeco(admonitionTitle ?? trans('Caution')).range(from, to))
+            classes.push('caution')
+          }
```
benniekiss: If changed to a generic `AdmonitionKeyword` node, this should be changed to a single `admonitionHeader.getChild('AdmonitionKeyword')` call and just push the text content, which should already be validated by the parser.

source/common/modules/markdown-editor/renderers/render-admonitions.ts
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        if (titleMarker !== null) {
+          inlineRanges.push(hiddenDeco.range(titleMarker.from, titleMarker.to))
+        }
+
+        const quoteMarkers = node.node.getChildren('QuoteMark')
+        for (const q of quoteMarkers) {
+          inlineRanges.push(hiddenDeco.range(q.from, q.to))
+        }
```
benniekiss: I think this should be moved out of this renderer into the `renderEmphasis` plugin. That's where all other hiddenDeco handling is done.

source/common/modules/markdown-editor/theme/editor.ts
```diff
@@ -73,6 +73,26 @@ export interface ThemeVars {
...
+  /** CSS `<color>` value */
+  '--zettlr-editor-admonition-note-bg': string
+  /** CSS `<color>` value */
+  '--zettlr-editor-admonition-tip-bg': string
+  /** CSS `<color>` value */
+  '--zettlr-editor-admonition-important-bg': string
+  /** CSS `<color>` value */
+  '--zettlr-editor-admonition-warning-bg': string
+  /** CSS `<color>` value */
+  '--zettlr-editor-admonition-caution-bg': string
```
benniekiss, line 95: these should all be made generic, a la `--zettlr-editor-error-color` below. Essentially removing the `-admonition-` from the name.

source/common/modules/markdown-editor/util/custom-tags.ts
```diff
@@ -22,6 +22,15 @@ export const customTags = {
   YAMLFrontmatter: Tag.define(tags.monospace),
   YAMLFrontmatterStart: Tag.define(tags.contentSeparator),
   YAMLFrontmatterEnd: Tag.define(tags.contentSeparator),
+  // Admonitions
```
benniekiss, line 25: this should include a little example of the syntax

source/common/modules/markdown-utils/markdown-ast/index.ts
```diff
@@ -502,6 +502,22 @@ export interface TextNode extends MDNode {
   value: string
 }
 
+export interface Admonition extends MDNode {
+  type: 'Admonition',
+  /**
+   * The type of this admonition.
+   */
+  admonitionType: 'note'|'tip'|'important'|'warning'|'caution',
+  /**
```
benniekiss: I think this should be renamed `keyword` and hold a string value

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
```diff
@@ -35,6 +42,18 @@ const EMPTY_NODES = new Set([
...
+  // we parse the "Admonition" itself, all of these nodes are parsed into the
+  // Admonition AST node, but we need to parse the admonition node's children,
+  // which are siblings to the admonition header.
+  'AdmonitionHeader',
+  'AdmonitionTitle',
+  'AdmonitionNote',
+  'AdmonitionTip',
+  'AdmonitionImportant',
+  'AdmonitionWarning',
+  'AdmonitionCaution',
```
benniekiss: I'll revisit this, but the comment makes me think that the header node isn't really all that necessary while adding a little too much complexity.

source/common/modules/markdown-utils/markdown-to-html.ts
```diff
@@ -200,7 +200,10 @@ function addAttribute (node: ASTNode, attributeName: string, ...values: string[]
  * @return  {string}                The HTML string
  */
 export function nodeToHTML (node: ASTNode|ASTNode[], options: MD2HTMLOptions, indent: number = 0): string {
-  const HIDDEN_GENERIC_NODES = ['Document']
+  const HIDDEN_GENERIC_NODES = [
+    'Document',
+    'AdmonitionHeader'
```
benniekiss: just another indication the header node isn't needed.

nathanlesage:
Thank you so much for your initial thoughts — I'll let it rest a bit to pause and ponder on some of your concerns. My initial reactions:

I'll be able to go into this in more detail later in the week, but this sounds like great feedback.

> I think that the various Admonition{keyword} are too redundant and clutter the parse trree. The nodes should be made into a generic AdmonitionKeyword node, and the styling/specific keyword handling should just be moved into the renderer. The renderer can simply add a class to the decoration with the name of the keyword so users can target it for styling.

That would make it impossible to distinguish any of those admonitions when the preview is turned off, because to the theme, each admonition would look the same. Each parse target should be style-able without any preview renderers turned on as shown in the screenshots I provided.

I will give you that my current construction is definitely a bit convoluted and needs improvement. What I was thinking about was instead creating a “factory”[^1] for admonition parsers that yields five separate block parsers which parse one type of admonition. That would then make adding particular nodes for the keywords redundant, as you say.

> To keep the fallback to regular blockquotes, a lookup table in the parser should be added with a list of support keywords, and if the provided one isn't available, parsing moves on. This would make it way simpler to add more keywords in the future, if they ever come up.

That's a good point, and definitely easier to maintain than what I contracted here. Thank you!

Would the following parse tree be better, or do I overlook a potential optimization here?

```
    >        [!        keyword          ]      Optional title
QuoteMark|CodeMark|AdmonitionKeyword|CodeMark|AdmonitionTitle
|                  Admonition{keyword}                      |
```

Again, I strongly believe we need to mount a corresponding node for the keyword into the parse tree, because otherwise it will be impossible for the raw-parser to allow any type of customization. I believe that having colors denoting those keywords are important to visually distinct these from blockquotes if rendering is off.

> I also think the styling variables could be made more generic. Instead of specific admonition variables, I think these should be made into generic -warning, -info, -error variables. that way info/warning/error styling is synchronized across elements. There already is an --zettlr-editor-error-color variable, so this approach would align with that.

I had the exact same thought, but I have come to a different conclusion after thinking a bit. What people will want to style are *semantics*, namely the type of admonition. Is “caution” really always “error”-coded? Or couldn't it also be a warning? Also, I've seen “important” notes both with purple and blue. “Notes” sometimes also have a color, and even “tips” don't necessarily have to be green. By making the variables tied to the semantics of *what* they're styling, we keep it separate from the info/warning/error semantics that do not fully overlap with what admonitions do. (Also, I've already gotten the first request for additional keywords, so keeping `admonition-note` allows us to see those things as “groups” that users can re-style as they see fit.) At least that's my current thought process behind those variables.

Again, I'll take a look at your comments and ponder some of these things over the week (I have a conference going on, so I'll be busy anyhow), and then go back to the drawing board. I think there's more to think about before this is actually ready to merge. Thanks already for these initial thoughts.

[^1]: A.k.a. literally a function that generates parsers. The reason is if you `startComposite` providing the name `AdmonitionNote` and then you have the composite function, that will make CodeMirror throw an error because suddenly you have several nodes being started by the same parser which it doesn't like. That's the primary reason I resorted to custom keyword nodes for now, but I'm not too happy with that.

benniekiss:
> That would make it impossible to distinguish any of those admonitions when the preview is turned off, because to the theme, each admonition would look the same. Each parse target should be style-able without any preview renderers turned on as shown in the screenshots I provided.

I think this could be addressed by a separate renderer which applies the inline decorations and the keyword as a class name to the node. This would always be enabled so the classes can be styled, even in raw mode. The pretty-renderer would then only apply a blockwrapper when enabled, rather than managing inline and blockwrapper decorations.

My main concern is really just maintaining a clear separation between styling/rendering and document structure, and the individual keywords don't seem distinct enough for me to see a reason to have separate nodes from a structural standpoint.

> I had the exact same thought, but I have come to a different conclusion after thinking a bit. What people will want to style are semantics, namely the type of admonition. Is “caution” really always “error”-coded? Or couldn't it also be a warning? Also, I've seen “important” notes both with purple and blue. “Notes” sometimes also have a color, and even “tips” don't necessarily have to be green. By making the variables tied to the semantics of what they're styling, we keep it separate from the info/warning/error semantics that do not fully overlap with what admonitions do. (Also, I've already gotten the first request for additional keywords, so keeping admonition-note allows us to see those things as “groups” that users can re-style as they see fit.) At least that's my current thought process behind those variables.

I think you've misunderstood me. I think keeping the semantics is fine! A -caution, -tips, -notes, etc variable exactly. But they should not be strictly named for admonitions. Like, other elements in the UI that may indicate `Caution` should probably all use the same caution styling, etc.

benniekiss:
Note: Pandoc supports this syntax with the alerts extension: https://pandoc.org/MANUAL.html#extension-alerts

Should the extension be added to the export profiles?

nathanlesage:
> > That would make it impossible to distinguish any of those admonitions when the preview is turned off, because to the theme, each admonition would look the same. Each parse target should be style-able without any preview renderers turned on as shown in the screenshots I provided.
> My main concern is really just maintaining a clear separation between styling/rendering and document structure, and the individual keywords don't seem distinct enough for me to see a reason to have separate nodes from a structural standpoint.

I didn't think of adding yet another renderer but you're right, and I mean this is what we do already in some places. I think that is the smartest solution, and I will implement that

> I think you've misunderstood me. I think keeping the semantics is fine! A -caution, -tips, -notes, etc variable exactly. But they should not be strictly named for admonitions. Like, other elements in the UI that may indicate `Caution` should probably all use the same caution styling, etc.

Ah, now I see! Yes, sure! I'm thinking if that shouldn't even become a global style, too, so that we can reuse the colors in the rest of the UI (including ZtrAdmonition)

nathanlesage:
> Note: Pandoc supports this syntax with the alerts extension: https://pandoc.org/MANUAL.html#extension-alerts
> 
> 
> 
> Should the extension be added to the export profiles?

God, yes, please -- sorry that I didn't check the Pandoc extension list for updates, but that's precisely what we need. Maybe sans title and nesting support, I'll have to check this.

benniekiss:
> Ah, now I see! Yes, sure! I'm thinking if that shouldn't even become a global style, too, so that we can reuse the colors in the rest of the UI (including ZtrAdmonition)

exactly my thoughts as well!

source/common/modules/markdown-editor/renderers/render-admonitions.ts
nathanlesage: I think that wouldn't be wise. If someone, for whatever reason, disables the emphasis renderer, but keeps the admonition renderer, they would see quote marks in the rendered admonitions.

source/common/modules/markdown-editor/parser/admonition-parser.ts
nathanlesage: Do you think that is necessary? After I figured out that we can just reuse existing nodes that this might make it less complex, and since these are indeed codemarks, it feels simpler this way.

nathanlesage:
Sooo, I have now simplified the parser a lot. I have opted to create dedicated composite parsers for each class of admonition just so that it's possible to style them in raw mode, and I have implemented already quite some of your suggestions. There's still a bit more missing, but it feels more ordered already.

source/common/modules/markdown-editor/parser/admonition-parser.ts
benniekiss: Yes, for a couple of reasons. All the other parsers add nodes for their various marks, except for the math parser, which re-uses the codemark. Re-using nodes can also have unintended side effects for things that look for the nodes in the partial tree during parsing. Using a separate node ensures that does not happen. this is why the pandoc span parser adds a separate PandocSpanMark, for instance.

I also think codemark would be the wrong node to reuse, anyway.   I don't think they are filling the role of code marks, which are used to denote a block containing some discrete syntax. `[!` and `]` are also very much link-like, or more generally, reference-like syntax, which I believe fits in with the keyword semantics. But I don't think any of the existing link-like nodes are a good match or should be reused, so the best option would be to add a discrete `AdmonitionMark` node, or more verbose, `AdmonitionKeywordMark`.

nathanlesage:
So, did some refactoring to the renderer. I'll still need to work on the technical changes as requested.

However, I managed to test out what the alerts extension does, and here are my observations:

* It exclusively accepts uppercase keywords, lowercase does not work
* Nesting is supported
* Custom titles is not supported
* There is no default styling of the HTML export (this one I can address with a small PR to Pandoc itself to add basic default styling)

nathanlesage:
Most things are now in order. The only remaining change is our discussed semantic change of the color variables, which I'll do this evening. @benniekiss If you'd like you can do another review already.

Also, an update: I have now opened a PR that fixes the uppercase/lowercase discrepancy between Pandoc's alert extension and GitHub's (and other) ability to consume lowercase keywords: https://github.com/jgm/pandoc/pull/11836

Once that change gets in, Pandoc and Zettlr should be in alignment again as to what they produce.

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
```diff
@@ -25,6 +25,13 @@ import { genericTextNode } from './generic-text-node'
  * marks) who can be reconstructed without the verbatim value, as well as larger
  * container nodes (whose contents is represented via their children).
  *
+ * Essentially, what this means is that these nodes will be ignored by the
+ * `parseChildren` function, meaning they don't end up in the AST. This makes
+ * the AST more abstract. But this also means that all of these nodes should be
+ * treated in the `parseNode` function. Also, this only applies to nodes where
+ * `parseNode` calls `parseChildren`, because unless you do, `parseNode` treats
+ * the entire node.
+ *
```
benniekiss, line 35: This comment could be clarified a bit, specifically about what you mean by "treated in the `parseNode` function"

nathanlesage:
The comment you mentioned is now a bit clearer. Also, I believe I found a bug in the AST parser in that those empty nodes aren't fully ignored. At least, they are ignored, but their children are not, because the `parseNode` function is still being called on them. But that's OOS for this PR, I'll have to revisit this logic later.

nathanlesage:
@benniekiss Are there any other oustanding todos for this PR, or is this good to go?

benniekiss:
> @benniekiss Are there any other oustanding todos for this PR, or is this good to go?

Let me set aside time tomorrow (Friday) to give this one last look over.

nathanlesage:
Thank you!

source/common/modules/markdown-editor/renderers/render-admonitions.ts
```diff
@@ -0,0 +1,216 @@
...
+
+const hiddenDeco = Decoration.replace({})
+
+function renderAdmonitionWrappers (view: EditorView): { wrappers: RangeSet<BlockWrapper>, inlines: RangeSet<Decoration> } {
+  const wrapperRanges: Range<BlockWrapper>[] = []
+  const inlineRanges: Range<Decoration>[] = []
+  const includeAdjacent = view.state.field(configField, false)?.previewModeShowSyntaxWhenCursorIsAdjacent ?? true
+
+  for (const { from, to } of view.visibleRanges) {
+    syntaxTree(view.state).iterate({
```
benniekiss, line 50: this syntaxTree should be moved out of the loop:

```suggestion
  const tree = syntaxTree(view.state)
  for (const { from, to } of view.visibleRanges) {
    tree.iterate({
```

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
```diff
@@ -22,8 +22,16 @@ import { genericTextNode } from './generic-text-node'
...
+ * marks) which can be reconstructed without the verbatim value, as well as
+ * larger container nodes (whose contents is represented via their children).
+ *
+ * Essentially, what this means is that these nodes will be ignored by the
+ * `parseChildren` function, meaning they don't end up in the AST. This makes
+ * the AST a bit cleaner, but this also means that you should treat these nodes
+ * in the `parseNode` function. For example, we ignore all child nodes of the
+ * admonitions here, which means that `parseChildren` will simply ignore them
+ * (and any of their children), but in turn this implies that you should have
+ * already treated them in the `parseNode` function.
```
benniekiss, line 34: This is still not correct, and I think this whole second paragraph should just be deleted, except for the first sentence: "Essentially, what this means is that these nodes will be ignored by the`parseChildren` function, meaning they don't end up in the AST.`"

The whole part about needing to be handled by the `parseNode` function is wrong -- several of the nodes listed are not handled at all by `parseNode`.

benniekiss:
I still think it is the wrong move to have distinct node types for each admonition. Currently, it just adds an unnecessary layer of indirection. Why not just remove the unique nodes and get the string value of the `AdmonitionKeyword`? That already is essentially what is done, except the keywords are now hardcoded several different places and have to now all be kept in sync.

I would remove the parser factory, only pushing a generic `Admonition` container. In the parser, I would keep a list of valid keywords:

```ts
export const admonitionKeywords = [
  'note',
  'tip',
  'important',
  'warning',
  'caution'
] as const
```

And then just change this:

```ts
      if (!admonitionTypes[admonitionType].includes(match.groups.keyword.toLowerCase())) {
        return false // Wrong admonition type
      }
```

to this:

```ts
      if (!admonitionKeywords.includes(match.groups.keyword.toLowerCase())) {
        return false // invalid keyword
      }
```

Then, this section in the renderer:

```ts
        const classes = ['admonition-wrapper']
        let genericKeywordTranslated = trans('Note')
        if (node.type.name === 'AdmonitionNote') {
          classes.push('note')
        } else if (node.type.name === 'AdmonitionTip') {
          classes.push('tip')
          genericKeywordTranslated = trans('Tip')
        } else if (node.type.name === 'AdmonitionImportant') {
          classes.push('important')
          genericKeywordTranslated = trans('Important')
        } else if (node.type.name === 'AdmonitionWarning') {
          classes.push('warning')
          genericKeywordTranslated = trans('Warning')
        } else if (node.type.name === 'AdmonitionCaution') {
          classes.push('caution')
          genericKeywordTranslated = trans('Caution')
        }
```

Could be simplified to:

```ts
    const kw = view.state.sliceDoc(keywordMarker.from, keywordMarker.to).toLowerCase()
    classes.push(keyword)
    genericKeywordTranslated = trans(keyword)
```

The parser already has determined what is valid, so the test in the renderer is unneeded.

To keep styling when in raw view, I really think a separate renderer is a better approach than adding all of these nodes to the ast. 

It would only be:

```ts
function renderAdmonitionKeywords (view: EditorView): RangeSet<Decoration> {
  const inlineRanges: Range<Decoration>[] = []
  const includeAdjacent = view.state.field(configField, false)?.previewModeShowSyntaxWhenCursorIsAdjacent ?? true
  const tree = syntaxTree(view.state)
  for (const { from, to } of view.visibleRanges) {
        tree.iterate({

        ....

        const keywordMarker = node.node.getChild('AdmonitionKeyword')
        if (keywordMarker === null) {
          return
        }

        deco = Decoration.mark({ class: view.state.sliceDoc(keywordMarker.from, keywordMarker.to) })
        inlineRanges.push(deco.range(keywordMarker.from, keywordMarker.to))
    }) 
  } 
  
  return Decoration.set(inlineRanges, true)
}
```

This would then be enabled as an always-on renderer, like the code renderer. The styling then could be kept in this same file.

I don't like the current factory approach, as I find it cluttered and confusing. There are, as I count, eight places where a unique node name is used, so adding or removing a node now requires synchronizing these eight places. I think with my approach, there would be three: one in the parser allow list, and two in the renderer style definitions. In fact, you could probably consolidate the blockwrapper and keyword styling into one selector.

nathanlesage:
> I still think it is the wrong move to have distinct node types for each admonition. Currently, it just adds an unnecessary layer of indirection. Why not just remove the unique nodes and get the string value of the `AdmonitionKeyword`? That already is essentially what is done, except the keywords are now hardcoded several different places and have to now all be kept in sync.
> 
> 
> 
> I would remove the parser factory, only pushing a generic `Admonition` container. In the parser, I would keep a list of valid keywords:
> 
> 
> 
> ```ts
> 
> export const admonitionKeywords = [
> 
>   'note',
> 
>   'tip',
> 
>   'important',
> 
>   'warning',
> 
>   'caution'
> 
> ] as const
> 
> ```
> 
> 
> 
> And then just change this:
> 
> 
> 
> ```ts
> 
>       if (!admonitionTypes[admonitionType].includes(match.groups.keyword.toLowerCase())) {
> 
>         return false // Wrong admonition type
> 
>       }
> 
> ```
> 
> 
> 
> to this:
> 
> 
> 
> ```ts
> 
>       if (!admonitionKeywords.includes(match.groups.keyword.toLowerCase())) {
> 
>         return false // invalid keyword
> 
>       }
> 
> ```
> 
> 
> 
> Then, this section in the renderer:
> 
> 
> 
> ```ts
> 
>         const classes = ['admonition-wrapper']
> 
>         let genericKeywordTranslated = trans('Note')
> 
>         if (node.type.name === 'AdmonitionNote') {
> 
>           classes.push('note')
> 
>         } else if (node.type.name === 'AdmonitionTip') {
> 
>           classes.push('tip')
> 
>           genericKeywordTranslated = trans('Tip')
> 
>         } else if (node.type.name === 'AdmonitionImportant') {
> 
>           classes.push('important')
> 
>           genericKeywordTranslated = trans('Important')
> 
>         } else if (node.type.name === 'AdmonitionWarning') {
> 
>           classes.push('warning')
> 
>           genericKeywordTranslated = trans('Warning')
> 
>         } else if (node.type.name === 'AdmonitionCaution') {
> 
>           classes.push('caution')
> 
>           genericKeywordTranslated = trans('Caution')
> 
>         }
> 
> ```
> 
> 
> 
> Could be simplified to:
> 
> 
> 
> ```ts
> 
>     const kw = view.state.sliceDoc(keywordMarker.from, keywordMarker.to).toLowerCase()
> 
>     classes.push(keyword)
> 
>     genericKeywordTranslated = trans(keyword)
> 
> ```
> 
> 
> 
> The parser already has determined what is valid, so the test in the renderer is unneeded.
> 
> 
> 
> To keep styling when in raw view, I really think a separate renderer is a better approach than adding all of these nodes to the ast. 
> 
> 
> 
> It would only be:
> 
> 
> 
> ```ts
> 
> function renderAdmonitionKeywords (view: EditorView): RangeSet<Decoration> {
> 
>   const inlineRanges: Range<Decoration>[] = []
> 
>   const includeAdjacent = view.state.field(configField, false)?.previewModeShowSyntaxWhenCursorIsAdjacent ?? true
> 
>   const tree = syntaxTree(view.state)
> 
>   for (const { from, to } of view.visibleRanges) {
> 
>         tree.iterate({
> 
> 
> 
>         ....
> 
> 
> 
>         const keywordMarker = node.node.getChild('AdmonitionKeyword')
> 
>         if (keywordMarker === null) {
> 
>           return
> 
>         }
> 
> 
> 
>         deco = Decoration.mark({ class: view.state.sliceDoc(keywordMarker.from, keywordMarker.to) })
> 
>         inlineRanges.push(deco.range(keywordMarker.from, keywordMarker.to))
> 
>     }) 
> 
>   } 
> 
>   
> 
>   return Decoration.set(inlineRanges, true)
> 
> }
> 
> ```
> 
> 
> 
> This would then be enabled as an always-on renderer, like the code renderer. The styling then could be kept in this same file.
> 
> 
> 
> I don't like the current factory approach, as I find it cluttered and confusing. There are, as I count, eight places where a unique node name is used, so adding or removing a node now requires synchronizing these eight places. I think with my approach, there would be three: one in the parser allow list, and two in the renderer style definitions. In fact, you could probably consolidate the blockwrapper and keyword styling into one selector.
> 
> 
> 
> 

That was my initial approach. However, that seemed to render it impossible to have a block level renderer style and this seems, at least to me, much more clean. But I will take another look, maybe I have overlooked something. I don't want to dismiss your concerns. I just went that route already (since it is, as you agree, the more logical choice), and figured it was less ideal than using a factory.

But again, I'll see if I can simplify it again. Because now that I'm trying I can't remember what it was that prompted me towards the factory approach.

nathanlesage:
Just took another look. The crux really is the theming in code view.

> To keep styling when in raw view, I really think a separate renderer is a better approach than adding all of these nodes to the ast.

I see where you are coming from, but after having thought about this, I believe I will need to disagree. That renderer would have to be a bit more convoluted. What you're proposing merely changes the keyword itself, but I would like to style at the very least also the code marks being inserted. That would involve a lot of extra nodes. While relatively benign in terms of code (as we can simply iterate through the code marks and add classes to them), that makes the resulting DOM structure much more complex, and the DOM structure we already produce is already quite hard. Having several top-level nodes makes styling all of them much easier. Yes, it has the downside of having more places where we'd have to maintain nodes (but I will do one final sweep and centralize all available nodes in the admonition parser itself, to reduce the maintenance cost).

The only other way I could think of would be to make use of node props to attach the type of admonition to the generic Admonition node, but that unfortunately isn't supported by the Markdown Parser Extension API.

So unless you have some compelling argument I didn't consider yet, I think I will have to disagree with you (while at the same time agreeing, because I want to be absolutely clear that I can fully see where your critique is coming from).

benniekiss:
> but I would like to style at the very least also the code marks being inserted.

~~Do you mean the `[!` and `]`? Since those are already added as syntax nodes, can't they just use the regular tag-based node styling? I also don't see how that is achieved with the current styling. The same goes for the blockquote marks~~ Nevermind, I see how the structure works.

I see what you mean about the raw-renderer needing to be more involved, but I still don't think it would be much. It would just wrap a decoration mark around the entire admonition node, not just the keyword, so one more node lookup. Styling would then just have to target `.class .cm-admonition-{keyword,mark,etc}`.

I still think it is worth it to encapsulate this in one renderer rather than throughout the ast structure. I don't think additional DOM nodes are an issue, but I understand the concern.

EDIT:

I was curious and tested out a separate raw renderer, and this works in highlighting keywords and formatting marks, to the same level as the current implementation, but it does not rely on the unique admonition nodes. I'll leave this here for consideration, but I get wanting to keep the current approach:

<details><summary>Details</summary>
<p>

```ts
function renderAdmonitionRaw (view: EditorView): RangeSet<Decoration> {
  const inlineRanges: Range<Decoration>[] = []

  const tree = syntaxTree(view.state)
  for (const { from, to } of view.visibleRanges) {
    tree.iterate({
      from,
      to,
      enter: (node) => {
        if (!node.type.is('Admonition')) {
          return
        }

        const keywordNode = node.node.getChild('AdmonitionKeyword')
        if (keywordNode === null) {
          return false
        }

        const deco = Decoration.mark({
          class: view.state.sliceDoc(keywordNode.from, keywordNode.to).toLowerCase(),
        })
        inlineRanges.push(deco.range(node.from, node.to))
      },
    })
  }

  return Decoration.set(inlineRanges, true)
}

const admonitionRawPlugin = ViewPlugin.fromClass(
  class {
    inlines: RangeSet<Decoration>

    constructor (view: EditorView) {
      this.inlines = renderAdmonitionRaw(view)
    }

    update (update: ViewUpdate) {
      if (update.docChanged || update.viewportChanged || update.selectionSet) {
        this.inlines = renderAdmonitionRaw(update.view)
      }
    }
  },
  {
    provide: (plugin) => {
      return [
        EditorView.decorations.of(
          (view) => view.plugin(plugin)?.inlines ?? Decoration.none,
        ),
      ]
    },
  },
)

export const renderRawAdmonitions = [
  admonitionRawPlugin,
  EditorView.baseTheme({
  // Admonitions
    '.note > :is(.cm-admonition-mark, .cm-admonition-keyword, .cm-code-mark)': { color: 'var(--zettlr-note-color)' },
    '.tip > :is(.cm-admonition-mark, .cm-admonition-keyword, .cm-code-mark)': { color: 'var(--zettlr-tip-color)' },
    '.important > :is(.cm-admonition-mark, .cm-admonition-keyword, .cm-code-mark)': { color: 'var(--zettlr-important-color)' },
    '.warning > :is(.cm-admonition-mark, .cm-admonition-keyword, .cm-code-mark)': { color: 'var(--zettlr-warning-color)' },
    '.caution > :is(.cm-admonition-mark, .cm-admonition-keyword, .cm-code-mark)': { color: 'var(--zettlr-caution-color)' },
  }),
]

```

</p>
</details>

nathanlesage:
HAH! That's actually the missing puzzle piece. What I was afraid of was the prospect at having to quite literally litter the AST with a bunch of decorations. I was not aware that they are now finally able to span entire blocks with newlines in them. I'll still have to see if that works properly with having two renderers flip-flop, but that's definitely a start.

nathanlesage:
@benniekiss Now it should all be in order. Thanks for writing out the entire raw admonition parser, I have literally just copy-pasted it (with attribution) and it worked flawlessly.

Let me know if there are any additional things, or if this is finally ready for the production line!

source/common/modules/markdown-editor/renderers/render-admonitions.ts
benniekiss: This should be addressed before merging

source/common/modules/markdown-editor/util/custom-tags.ts
benniekiss: one last nit

source/common/modules/markdown-utils/markdown-ast/parse-children.ts
benniekiss: Another nit

benniekiss:
Just three lingering nits before this is ready to go!

nathanlesage:
> Just three lingering nits before this is ready to go!

I can't see any new comments ... is this a bug on GH Mobile or did you forget to submit the review ...?

nathanlesage:
I think we got it all now. That was a long PR process — but I strongly believe that this was worth it and has made the feature more robust as previous ones where I didn't have a second pair of eyes looking over them. Thank you for your time in making this happen!

If you spot anything later on, just let me know; I'll let this all simmer a bit now and see if the nightlies unearth a few new bugs, and then release the next update.

PR #6513, opened 2026-08-16
Fix add-row button doing nothing in the last table row

### LINKED ISSUE #6512: Add row button on visual table editor intermittent functionality
### Description

When working on a table, the `+` icon to add a new row to the end of the table frequently does nothing when the cursor is in the last row. If you click on a different row, then click back to the bottom row it will function normally. Adding a row in the middle of the table works as intended.

Tested with 4.7.0, though this has been an issue for a few versions.

### Reproducing

1. Create a table with a few rows and populate them
2. With the cursor in the bottom row, click the `+` icon to add a row to the end
3. Button will intermittently not function.

### Zettlr version

4.7.0

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [ ] Windows
- [x] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Debian 13, MacOS Tahoe

### Additional Information

_No response_

atirna:
## Description

Fixes the "+" add-row button in the visual table editor intermittently doing nothing when the cursor is in the last row.

The command computed the new cursor position by adding the current line's length to the document offset. Since the appended row is shorter than the current one (the cell contents get cleared), a cursor near the end of a populated last row overshot the new row's end, and the resulting transaction threw `Selection points outside of document`. That exception was swallowed inside the button's mousedown handler, so the click appeared to do nothing. Re-selecting the row moved the cursor to a position where the math happened to work, which is why it felt intermittent.

Fixes #6512

## Changes

- `addRowAfter` now places the cursor at the same column in the appended row, clamped to the new row's length, instead of adding the line length to the document offset.
- Added unit tests for the row command (append after last row, cursor placement, cursor at end of document, cursor in header row).

## Tested on

- macOS (arm64), via `yarn test` and `vue-tsc --noEmit`.

## Additional information

Before the fix, a cursor in most columns of a wide last row triggered the exception; after the fix the same sweep succeeds for every column. The new tests reproduce the failure on the old code (both the exception and a wrong-column cursor placement) and pass on the new one.

## AI Disclosure Statement

AI was used to help investigate this issue, implement the fix, and verify it; I reviewed and own the final code and PR.

## Declarations

  issues with my code.


nathanlesage:
Hi, sorry for the delay, but thanks for the PR! I'll see to giving you a review on the weekend.

test/table-editor-rows.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+import { doesNotThrow, ok, strictEqual } from 'assert'
+import { EditorState } from '@codemirror/state'
+import { EditorView } from '@codemirror/view'
+import markdownParser from 'source/common/modules/markdown-editor/parser/markdown-parser'
+import { addRowAfter } from 'source/common/modules/markdown-editor/table-editor/commands/rows'
+
+// The jsdom environment (see test/setup.js) does not provide requestAnimationFrame,
+// which the CodeMirror EditorView requires upon construction.
+window.requestAnimationFrame = (cb: FrameRequestCallback) => setTimeout(cb, 0) as unknown as number
+window.cancelAnimationFrame = (id: number) => clearTimeout(id)
```
nathanlesage: …then this code should go into the setup code, I fathom?

test/table-editor-rows.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @return  {EditorView}        The editor view
+ */
+function createView (doc: string, head: number): EditorView {
+  const state = EditorState.create({ doc, extensions: [markdownParser()] })
+  const view = new EditorView({ state, parent: document.body })
+  view.dispatch({ selection: { anchor: head } })
+  return view
+}
+
+const TABLE = '| A | B |\n| --- | --- |\n| 1 | 2 |\n| 3 | 4 |'
```
nathanlesage, line 22: Please move consts and any data to the top of the file (below the imports), then utility functions, finally the tests.

test/table-editor-rows.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+describe('TableEditor#addRowAfter()', function () {
+  it('should append an empty row after the last row', function () {
+    const view = createView(TABLE, 36) // Cursor inside the last row
+    const changed = addRowAfter(view)
```
nathanlesage: When I saw that you were creating a view, I thought you wanted to programmatically click some buttons, but since you are only calling the commands (which I absolutely prefer!), I would like you to not create a view and instead only work with a state. This keeps this test more slim, and aligned with other tests. I believe the transformation tests can show you how to merely instantiate a state, run a command on it, and check the result. It may be necessary to move some of the utility functions from that suite out to make them available to other commands. We didn't do that until now because no tests except theirs needed the state creation utilities.

test/table-editor-rows.spec.ts
nathanlesage: Nevermind, see my comments below.

test/table-editor-rows.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  })
+
+  it('should place the cursor at the same column in the new row', function () {
+    // The last row "| 3 | 4 |" starts at offset 34; the cursor at offset 36
+    // sits in the first cell ("3"), two characters in.
+    const view = createView(TABLE, 36)
+    addRowAfter(view)
+
+    // The new row starts at offset 44 (after the inserted newline), so the
+    // same column is offset 46.
```
nathanlesage:
```suggestion
    // same column should be at offset 46.
```

nathanlesage (changes requested):
Hi, so sorry for the delay, but I had a look. I think everything looks fine in terms of your changes, but I believe we can make the unit tests a bit more efficient. Right now you're instantiating entire views, which is not necessary to test out state-editing commands. I recommend you take a look into the transform command tests, which contain a bunch of utility functions for making testing commands painless.

If you have any questions, just let me know.

atirna:
thx for the review!, i am fixing the bloat tests
tightened the tests while waiting on review. two of the four were only asserting the document text, and addRowAfter inserts the same row with or without this fix, so those two passed on the unfixed tree. rest is folded into one table driven case list like build-pipe-table.spec.ts does, and each case now pins the resulting cursor line and column instead of an absolute offset.

all three fail on develop and the full test suite is green.

atirna:
Reworked the test around EditorState and dispatch, so it no longer constructs an EditorView. `yarn test test/table-editor-rows.spec.ts` passes.

nathanlesage:
Wonderful, now it looks much cleaner and easier to comprehend. Since the linters now pass, I think we can merge this. Thank you for the contribution.

PR #6523, opened 2026-08-25
fix(file-manager): Remove IPC listeners on unmount

### LINKED ISSUE #6522: Tab/file clicks stop responding after a RangeError: Maximum call stack size exceeded triggered by external file change + open race
**Environment**
- Zettlr version: 4.7.0
- OS: macOS 26.6.2 (Apple M3, MacBook Air)
- Install source: Homebrew Cask (`brew install --cask zettlr`)

**Description**

After several hours of use with ~13 tabs open, the app UI becomes unresponsive to clicks — tabs don't switch when clicked — while the process itself stays alive (not "Not Responding" in Activity Monitor, no high CPU usage while idle).

The renderer log shows a two-stage failure pattern, observed identically on two separate days:

1. An external process modifies a file that is currently open in a Zettlr tab (detected via the file watchdog), and a file-open/tab-switch action happens in Zettlr at nearly the same time.
2. Shortly after, the renderer throws repeated `Uncaught RangeError: Maximum call stack size exceeded` errors.
3. From that point on, every subsequent action (opening files, tab clicks) triggers `TypeError: Cannot read properties of null (reading 'Symbol(_vei)')` — Vue's internal event-listener tracking appears to be corrupted — and the UI stops responding to tab clicks.

**Log excerpt (paths anonymized)**

```
[10:39:07] [Info] [WATCHDOG] Emitting event: change:<file-A>.md
[10:39:07] [Info] [DocumentManager] Processing change for <file-A>.md
[10:39:20] [Info] [Application] Opening file <file-B>.md.
[10:39:42] [Info] [Application] Opening file <file-C>.md.
[10:39:44] [Error] [R] [Main Window] Uncaught RangeError: Maximum call stack size exceeded (index.js:1)
  (repeated ~19 times within the same second)
[10:39:45] [Error] [R] [Main Window] RangeError: Maximum call stack size exceeded (index.js:2)
...
[11:21:16] [Error] [R] [Main Window] TypeError: Cannot read properties of null (reading 'Symbol(_vei)') (index.js:2)
  (recurs on every subsequent file-open/tab action; ~130+ occurrences over following hours)
```

The same pattern (WATCHDOG change event immediately followed by a file-open action, then the stack overflow, then repeated `_vei` errors) appeared the previous day as well, at a different file/time.

**Reproduction (best guess, not confirmed deterministic)**
1. Open several tabs (10+) in one window
2. Have an external process (e.g., a script, another editor, a sync tool) modify a file that is currently open in a tab
3. At roughly the same moment, click to open/switch to a different tab
4. Stack overflow errors appear in the log; afterward tab clicks stop working

**Workaround**
Quitting and relaunching Zettlr restores normal behavior (tabs/documents are restored from `documents.yaml`).

**Additional context**
- `autoSave` was set to `"off"` in config.json
- Update check at the time failed with `ETIMEDOUT` (network issue, unrelated) — unable to confirm whether a newer version already fixes this

tosh13:
## Description

`FileItem` and `TreeItem` in the file manager registered IPC listeners on mount but never removed them, leaking one `fsal-event` listener per unmounted tree item. This fixes the leak by unsubscribing in `onUnmounted`.

## Changes

Both components registered an `fsal-event` listener (and, for directories, a `shortcut` listener) inside `onMounted` without ever removing it. Tree items are unmounted and re-created whenever the tree re-renders — collapsing/expanding directories, typing in the filter box, switching workspaces — so listeners accumulate over the lifetime of a window. On my machine this reliably produced

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
101 fsal-event listeners added.
```

in the renderer log, typically within 2–20 minutes of starting the app, depending on how much I used the file manager.

Both components now keep the unsubscribe callback returned by `ipcRenderer.on()` and call it in `onUnmounted`, matching how other components in this repository handle IPC listeners (`FilterTab.vue`, `DefaultsTab.vue`, `CustomCSS.vue`, `SnippetsTab.vue`, `GlobalSearch.vue`, `win-update/App.vue`).

The event handler bodies are unchanged — they were only extracted into named functions so the same reference can be passed to `on()` and used with its returned unsubscribe function. No behavioural change is intended.

## Tested on

macOS 26.6.2 (Apple Silicon), built from `develop`.

- `yarn lint` — 0 errors (147 pre-existing warnings, none in the changed files)
- `yarn test` — 375 passing
- `yarn start` — app boots, file manager works, clean shutdown, no new warnings or errors in the log

## Additional information

Two caveats I want to be upfront about:

1. I have not empirically verified that listener growth actually stops. My `yarn start` session was short and did not exercise the file manager heavily enough for the warning to appear even on unpatched builds, so "no warning appeared" is not evidence by itself. The reasoning for the fix is from the code: the listener is registered per component instance and never removed.

2. I suspect, but cannot prove, that this leak is related to #6522, where tab clicks stop responding after a `RangeError: Maximum call stack size exceeded` and every subsequent interaction throws `TypeError: Cannot read properties of null (reading 'Symbol(_vei)')`. Both the listener warnings and those errors appeared in the same sessions on my machine. I am not claiming this PR fixes #6522.

Transparency about AI use, per the checklist: I investigated this issue and prepared the patch with the assistance of an AI coding agent (Claude Code). I reviewed the diff, ran the checks above myself, and take full responsibility for the code. No AI is listed as an author or co-author on the commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@
 ## Under the Hood
 
 - Update Electron to `v43.4.0`.
+- Fixed a memory leak in the file manager: `FileItem` and `TreeItem` registered
+  `fsal-event` (and, for directories, `shortcut`) IPC listeners on mount without
+  ever removing them, so listeners accumulated as tree items were unmounted and
+  re-created. This surfaced as recurring
+  `MaxListenersExceededWarning: 101 fsal-event listeners added` warnings in the
+  renderer log.
 
 # 4.7.0
 
--- a/source/win-main/file-manager/FileItem.vue
+++ b/source/win-main/file-manager/FileItem.vue
@@ -159,7 +159,7 @@ import formatSize from '@common/util/format-size'
 import PopoverDirProps from './util/PopoverDirProps.vue'
 import PopoverFileProps from './util/PopoverFileProps.vue'
 
-import { ref, computed, toRef, watch, onMounted } from 'vue'
+import { ref, computed, toRef, watch, onMounted, onUnmounted } from 'vue'
 import { type AnyDescriptor, type MDFileDescriptor } from '@dts/common/fsal'
 import { useConfigStore, useTagsStore, useWindowStateStore } from 'source/pinia'
 import { useItemComposable } from './util/item-composable'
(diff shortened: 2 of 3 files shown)


benniekiss:
> I have not empirically verified that listener growth actually stops. My yarn start session was short and did not exercise the file manager heavily enough for the warning to appear even on unpatched builds, so "no warning appeared" is not evidence by itself. The reasoning for the fix is from the code: the listener is registered per component instance and never removed.

Please try to verify with some confidence that your patch does what its intended to do.

nathanlesage:
+1 to what @benniekiss has mentioned. Also, please use the `beforeUnmount` hook instead of the unmounted, to ensure that all of this gets stripped before the DOM element is removed, which is safer than using the after unmount hook.

benniekiss:
Closing due to lack of engagement from OP.

Feel free to comment with either confirmation that this achieves what it claims, or new information, and we can consider reopening.

PR #6543, opened 2026-09-01
fix: correct scroll behavior

### LINKED ISSUE #6531: Scroll bar jumps up when file is modified externally
### Description

My markdown files are about 2000 words long. I often use Claude CLI to make edits, but every time it modifies the file, the scroll bar in Zettlr jumps up and I lose my position. It's more noticeable towards the end of the file.

### Reproducing

Open a markdown file of about 2000 words. Scroll to bottom of file. Open the same file from any other app (eg. Notepad), and add something to the end of the file from Notepad, click save. When Zettlr detects the save it will jump up a few scroll positions and you will lose your place.

### Zettlr version

4.7.0 on Windows and Linux

### Installation Method

From the Website or GitHub

_No response_

### Your Platform

- [x] Windows
- [ ] macOS
- [x] Linux

### Architecture

- [x] Intel 64bit

### Operating System Version

Linux OpenSuse, Windows 11

### Additional Information

_No response_

felipe-f-rocha:
## Description

Fixes the scroll behavior described in issue #6531 by applying the solution proposed by @zelvault.

## Changes

* Implemented the scroll fix proposed by @zelvault.
* Adjusted the affected scroll handling logic to prevent the incorrect scrolling behavior.
* No breaking changes to the UI/UX or internal APIs.

## Tested on

* Windows
* Chrome
* Local development environment

## Additional information

Closes #6531.

The solution and implementation approach were proposed by @zelvault.

## AI Disclosure Statement

AI (GPT-5.6 Luna) was used as a development assistance tool to help understand the existing code and review the proposed changes. The implementation was based on the solution proposed by @zelvault.

## Declarations

* [x] I hereby confirm that I am solely responsible for the code provided in
  issues with my code.
* [x] I hereby confirm that I wrote this PR description myself and that I did
* [x] I have specified any open issues that this PR fixes/closes accordingly in

### DIFF
--- a/source/common/modules/markdown-editor/index.ts
+++ b/source/common/modules/markdown-editor/index.ts
@@ -29,7 +29,7 @@ import EventEmitter from 'events'
 // CodeMirror imports
 import { Decoration, type DecorationSet, EditorView } from '@codemirror/view'
 import {
-  EditorSelection,
+  type EditorSelection,
   EditorState,
   Text,
   type StateEffect,
@@ -382,39 +382,16 @@ export default class MarkdownEditor extends EventEmitter {
       // Now that the correct document has been loaded, there will be content
       // and we can restore the persisted information.
       const { scrollSnapshot, selection, foldedRanges } = persistentState
-      const docLength = this._instance.state.doc.length
 
       const effects: StateEffect<any>[] = [scrollSnapshot]
 
       const cursor = foldedRanges.iter()
       while (cursor.value) {
-        const from = Math.min(cursor.from, docLength)
-        const to = Math.min(cursor.to, docLength)
-        if (to > from) {
-          effects.push(foldEffect.of({ from, to }))
-        }
+        effects.push(foldEffect.of({ from: cursor.from, to: cursor.to }))
         cursor.next()
-      } 
-
-      let clampedSelection = selection
-      if (selection.ranges.some(r => r.anchor > docLength || r.head > docLength)) {
-        try {
-          clampedSelection = EditorSelection.create(
-            selection.ranges.map(r => EditorSelection.range(
-              Math.min(r.anchor, docLength),
-              Math.min(r.head, docLength)
-            )),
-            selection.mainIndex
-          )
-        } catch (err: any) {
-        // Clamping can make multiple ranges overlap; fall back to a single
-        // cursor at the end of the (shrunk) document.
-          console.warn(`Could not restore selection after reload: ${String(err.message)}`)
-          clampedSelection = EditorSelection.single(docLength)
-        }
       }
 
-      this._instance.dispatch({ selection: clampedSelection, effects })
+      this._instance.dispatch({ selection, effects })
     }
 
     // Ensure the theme switcher picks the state change up; this somehow doesn't
@@ -460,7 +437,7 @@ export default class MarkdownEditor extends EventEmitter {
    * a setting has changed that requires extensions to be fully reloaded.
    */
   async reload (): Promise<void> {
-    await this.loadDocument(this.persistentState)
+    await this.loadDocument()
   }
 
   /**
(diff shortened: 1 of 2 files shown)


felipe-f-rocha:
The translation file isn't related to this PR. It'll be removed as soon as possible, so the PR can be properly done.

source/common/modules/markdown-editor/index.ts
```diff
@@ -382,16 +382,39 @@ export default class MarkdownEditor extends EventEmitter {
...
       const effects: StateEffect<any>[] = [scrollSnapshot]
 
       const cursor = foldedRanges.iter()
       while (cursor.value) {
-        effects.push(foldEffect.of({ from: cursor.from, to: cursor.to }))
+        const from = Math.min(cursor.from, docLength)
+        const to = Math.min(cursor.to, docLength)
+        if (to > from) {
+          effects.push(foldEffect.of({ from, to }))
+        }
```
nathanlesage: What exactly is this check doing? It appears that we could reduce this to merely checking `if from > docLength || to > docLength`, the `Math.min` calls appear unnecessary. Could you check this?

source/common/modules/markdown-editor/index.ts
```diff
@@ -382,16 +382,39 @@ export default class MarkdownEditor extends EventEmitter {
...
+      let clampedSelection = selection
+      if (selection.ranges.some(r => r.anchor > docLength || r.head > docLength)) {
+        try {
+          clampedSelection = EditorSelection.create(
+            selection.ranges.map(r => EditorSelection.range(
+              Math.min(r.anchor, docLength),
+              Math.min(r.head, docLength)
+            )),
+            selection.mainIndex
+          )
```
nathanlesage: This right now produces overlapping selections because it will just squash all excess selections onto the exact same range, which might break CodeMirror. Rather, we can simplify this, and clamping should be done by something like the following:

```ts
const newRanges = selection.ranges.filter(r => r.anchor <= docLength && r.head <= docLength)
const clampedSelection = EditorSelection.create(newRanges)
// Plus some treating of the `mainIndex` which may have gotten invalid at this point due to filtering out selections
```

nathanlesage (changes requested):
Thanks for this! So essentially the fix seems to be to just check the persistent state that its saved information is still valid after re-loading the information. The PR seems to exclusively ensure that folded ranges and selection ranges do not exceed the new (possibly shorter) document length after reloading.

I think in both cases there can be some improvements made to ensure we do the least amount of work and don't end up with potentially unwanted states.

It would be great if you could take a look at the comments and improve the code accordingly. If you need any pointers, feel free to let me know.

felipe-f-rocha:
Thanks for the feedback! I went through both comments and adjusted the persistent state restoration logic accordingly.

For folded ranges, I removed the `Math.min` clamping. Instead of modifying a persisted range when it exceeds the new document length, the range is now only restored if both its `from` and `to` positions are still valid in the newly loaded document. This avoids creating a different or potentially unintended fold range from the persisted state.

For selections, I also removed the clamping approach. Previously, multiple selections extending beyond the new document length could all be clamped to `docLength`, potentially causing overlapping ranges. The new implementation filters out invalid ranges entirely and only restores ranges whose `anchor` and `head` are still within the new document.

I also accounted for `mainIndex`, since filtering ranges can change the index of the main selection. While iterating over the original ranges, I track the new position of the original main range so that the restored selection keeps the correct main selection when possible.

Finally, if all persisted selection ranges are invalid after the document reload, the implementation falls back to a single cursor at the end of the newly loaded document. This prevents attempting to create an empty `EditorSelection`.

Overall, the restoration now validates the persisted state against the newly loaded document instead of modifying invalid ranges, which should avoid unnecessary work and prevent potentially unwanted CodeMirror states.

PR #6576, opened 2026-09-20
fix: never show 1000 or more of a smaller size unit

PloneMraz:
Found by reading `source/common/util/format-size.ts`, not from an issue report.

## The bug

The function picks the unit with **1024-based** thresholds but divides by **1000**. Sizes that fall in the gap between the two bases therefore render as a number that should already have rolled over into the next unit:

| input | current output | expected |
| --- | --- | --- |
| 999 999 B | `1000 KB` | `1 MB` |
| 1 023 999 B | `1024 KB` | `1 MB` |
| 999 999 999 B | `1000 MB` | `1 GB` |
| 1 023 999 999 B | `1024 MB` | `1 GB` |

`1024 KB` is not a size anyone displays, and it shows up in the file manager for any file in those bands.

## The change

The unit is stepped up whenever the rounded value reaches 1000, so the displayed number always stays below the next unit whatever the input is.

I kept the `< 1024` boundary for plain Bytes deliberately: changing it to 1000 would relabel every file between 1000 and 1023 bytes, and that is a cosmetic preference rather than a defect. As it stands, **every size that formatted correctly before formats identically now** — only the two bands in the table change. If you would rather the whole function went decimal (or binary, with KiB/MiB labels), say so and I will follow that instead; it is a one-line change either way.

Not touched: `Math.round` on the largest unit means 2.5 GB still shows as `3 GB`. That is a precision choice rather than a correctness one, so it felt like a separate decision.

## Tests

`test/format-size.spec.ts` is new — the module had no spec. 19 cases: both broken bands, the unit boundaries either side, the largest unit having nothing to roll over into, and the long labels.

- `mocha --spec test/format-size.spec.ts`: 19 passing
- Full suite with this change: **397 passing, 11 failing**

The same 11 failures either side are pre-existing on Windows (`makeValidUri` builds `safe-file:////…` with a fourth slash, plus one BibTeX attachment case) and have nothing to do with this change. The difference is exactly the 19 new tests.

- Restoring the old implementation under the new spec: **17 failing** instead of 11 — the 6 cases covering the two bands — so the spec is pinned to the fix.
- `eslint` on both files: clean.


nathanlesage:

PR #6577, opened 2026-09-20
fix: match file extensions case-insensitively

PloneMraz:
Found by reading `source/common/util/file-extention-checks.ts`, not from an issue report.

## The bug

```ts
export function hasExt (filePath: string, extensions: string[]): boolean {
  return extensions.some(ext => filePath.endsWith(ext))
}
```

No case folding. So:

| file | recognised today |
| --- | --- |
| `IMG_1234.JPG` | ✗ not an image |
| `README.MD` | ✗ not Markdown |
| `Scan.PDF` | ✗ not a PDF |
| `Thesis.TEX` | ✗ not code |

Cameras and scanners produce uppercase extensions by default, and on Windows and macOS — both case-insensitive filesystems — `README.MD` and `readme.md` are the same file type by every other measure.

Everything built on `hasExt` inherits it: `documents/index.ts` refusing to open the file, the workspace filter in `config/index.ts`, drag-and-drop in `md-paste-drop-handlers.ts`, image rendering, and the validation in `file-new` / `file-rename`.

## Why this is a fix rather than a preference

The rest of the codebase already folds case before comparing an extension:

- `is-attachment.ts` — `ext.includes(extname(p).toLowerCase())`
- `citeproc/util/database-loader.ts` — `path.extname(databasePath).toLowerCase()`
- `commands/save-image-from-clipboard.ts` — twice
- `commands/importer/check-import-integrity.ts`

So `file-extention-checks.ts` is the outlier, and this brings it in line rather than introducing a new rule.

The extension **list** is lowercased too, not just the path: `hasAnyRecognizedFileExtension` is called with `attachmentExtensions` straight from the user's config, where an entry can perfectly well be typed as `.JPG`. Today such an entry only matches uppercase files, which is the same bug from the other side.

## Tests

`test/file-extention-checks.spec.ts` is new — the module had no spec. 14 cases: uppercase path, uppercase extension list, mixed case, the negative cases that must keep failing (`.md` must not match `note.mdx` or a file called `readme`), an empty list, and one custom-attachment case.

- Full suite with this change: **392 passing, 11 failing**
- Restoring the case-sensitive comparison under the new spec: **383 passing, 20 failing** — the 9 uppercase cases — so the spec is pinned to the fix

The 11 failures either side are pre-existing on Windows (`makeValidUri` emitting a fourth slash, plus one BibTeX case) and unrelated to this change; the difference is exactly the 14 new tests.

`eslint` on the source file: clean. (The `test/` folder is outside the eslint config, so the spec reports "File ignored" there, same as the existing specs.)


nathanlesage:

PR #6578, opened 2026-09-20
fix: compare extended language subtags by value, not by identity

PloneMraz:
Found by reading `source/common/util/find-lang-candidates.ts`, not from an issue report.

## The bug

```ts
if (parsedLang.extendedLanguageSubtags.length > 0) {
  if (parsedLang.extendedLanguageSubtags !== cand.extendedLanguageSubtags) {
    continue
  }
}
```

`bcp47.parse()` returns `extendedLanguageSubtags` as an **array**, and `!==` on two arrays compares identities, which are never equal for two separately parsed tags. So the moment the requested language carries an extended subtag, every candidate is skipped — including one whose tag is character-for-character the query.

Measured against the current `develop`:

```
query zh-cmn-Hans-CN   candidates ["zh-cmn-Hans-CN","zh-Hans-CN"]  ->  exact=-  close=-
query en-US            candidates ["en-US","en-GB"]                ->  exact=en-US
query en               candidates ["en-US","en-GB"]                ->  close=en-US
```

The first line is the bug: an identical tag, and nothing comes back.

## The change

Compares the subtag lists element by element. Tags without an extended subtag never entered that branch, so the ordinary cases (`en-US`, `de-DE`, `sr-Latn`) are untouched.

## Something else I found but did not change

One line further down:

```ts
if (parsedLang.script !== '' && parsedLang.script !== cand.script) {
```

`bcp-47@2` reports an absent script as `null`, not `''`, so the "only check the script if one was requested" guard never actually short-circuits. The effect is that a script-less query skips every candidate that has one:

```
query sr   candidates ["sr-Latn","sr-Cyrl"]  ->  exact=-  close=-
```

which reads against the comment right above it ("if there are no variants requested, then we assume the user does not care which variant s/he gets").

I left it alone because fixing it is not self-contained: the exact-vs-close test below compares region, variants and extensions but **not** script, so simply making the guard nullish-aware would promote `sr-Latn` to an *exact* match for the query `sr`. That needs a decision about what "exact" should mean, which is yours rather than mine. Happy to follow up with whichever shape you prefer.

## Tests

`test/find-lang-candidates.spec.ts` is new — the module had no spec. 7 cases: exact match, close match, a different language, the extended-subtag case above, a candidate whose extended subtag genuinely differs (must still be rejected), a requested script still matching exactly, and an empty candidate list.

- Full suite with this change: **385 passing, 11 failing**
- New spec against the unfixed function: **384 passing, 12 failing** — the extended-subtag case, so the spec is pinned to the fix

The 11 failures either side are pre-existing on Windows (`makeValidUri` emitting a fourth slash, plus one BibTeX case) and unrelated; the difference is exactly the 7 new tests.

`eslint` on the source file: clean.


nathanlesage:

